PR #42, opened 2015-01-01
Weed "minimal" docker image

OGrape:
I think the weed docker image which build from the given Dockerfile is too large, so I made this minimal version and it's only about 19.6 MB.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,21 @@
-FROM cydev/go
-RUN go get github.com/chrislusf/weed-fs/go/weed
+FROM progrium/busybox
+
+WORKDIR /opt/weed
+
+RUN opkg-install curl
+RUN echo insecure >> ~/.curlrc
+
+RUN \
+  curl -Lks https://bintray.com$(curl -Lk http://bintray.com/chrislusf/Weed-FS/seaweed/_latestVersion | grep linux_amd64.tar.gz | sed -n "/href/ s/.*href=['\"]\([^'\"]*\)['\"].*/\1/gp") | gunzip | tar -xf - -C /opt/weed/ && \
+  mv weed_* bin && \
+  chmod +x ./bin/weed
+
 EXPOSE 8080
 EXPOSE 9333
+
 VOLUME /data
+
+ENV WEED_HOME /opt/weed
+ENV PATH ${PATH}:${WEED_HOME}/bin
+
 ENTRYPOINT ["weed"]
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,21 @@
...
-RUN go get github.com/chrislusf/weed-fs/go/weed
+FROM progrium/busybox
+
+WORKDIR /opt/weed
+
+RUN opkg-install curl
+RUN echo insecure >> ~/.curlrc
+
+RUN \
+  curl -Lks https://bintray.com$(curl -Lk http://bintray.com/chrislusf/Weed-FS/seaweed/_latestVersion | grep linux_amd64.tar.gz | sed -n "/href/ s/.*href=['\"]\([^'\"]*\)['\"].*/\1/gp") | gunzip | tar -xf - -C /opt/weed/ && \
```
samuelngs, line 9: The dockerfile will download the latest pre-compiled binary from Bintray automatically

Dockerfile
```diff
@@ -1,6 +1,21 @@
...
+  mv weed_* bin && \
```
OGrape: rename folder to bin

chrislusf:
Is it possible to have 2 socket files?

OGrape:
Yep actually we can. Maybe something like this?

~~~
`-- project
     |-- minimal
     |    |-- Dockerfile
     `-- Dockerfile
~~~

or minimal Dockerfile in the root? Which one do you prefer?

chrislusf:
I like the minimal docker file as the default. The original one can be called something like "Dockerfile.go_build".

OGrape:
I just pushed an update :)
thanks

chrislusf:
Thanks. One more thing, can you please also adjust the documentation page "doc/gettingstarted.rst", in order to make it consistent with the change?

OGrape:
No problem! I'm gonna have a look on that.

OGrape:
I did some changes on the gettingstarted.rst

chrislusf:
Thanks!

I think the "--rm -t 'cydev/weed' " can be removed from this?
  docker build --no-cache --rm -t 'cydev/weed' .

Also, the "docker run" in "In production" section needs change.

OGrape:
Yeah --rm is not necessary but -t, it's for the repository name.
I named it "cydev/weed" so that people can start up a container with same command line (Doesn't matter Local or Docker Hub image)

~~~
docker run --name weed -d -p 9333:9333 -p 8080:8080 \
	-v /opt/weedfs/data:/data cydev/weed server -dir="/data"
~~~

Sorry but could you tell me what to change in the "In production" section, as I see there's nothing needs to be changed.

chrislusf:
thanks a lot!

PR #79, opened 2015-02-04
use sync.RWMutex when masterNode changes

aszxqw:
It needs lock.

https://groups.google.com/forum/?hl=zh-CN#!topic/golang-nuts/zyQnord8hyc

### DIFF
--- a/go/weed/weed_server/volume_server.go
+++ b/go/weed/weed_server/volume_server.go
@@ -3,6 +3,7 @@ package weed_server
 import (
 	"math/rand"
 	"net/http"
+	"sync"
 	"time"
 
 	"github.com/chrislusf/weed-fs/go/glog"
@@ -12,6 +13,7 @@ import (
 
 type VolumeServer struct {
 	masterNode   string
+	mnLock       sync.RWMutex
 	pulseSeconds int
 	dataCenter   string
 	rack         string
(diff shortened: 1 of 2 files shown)


chrislusf:
Even the race happens, it would not impact the system since it's in a transition period.

However, I still do not see it's possible to have any data race related error related to this. It's just changing a string pointer, which is one cpu operation. But I do not know much about Go deep internals. Having a reproducible test case would be more convincing.

aszxqw:
1.
`Even the race happens, it would not impact the system since it's in a transition period.`
When the race happens, your process will be dead.  

2.  `It's just changing a string pointer, which is one cpu operation. ` You mean a pointer change is atomic operation ? Are you sure ?

I just hope this project to be more stable, avoid potential error happening.
But now that you really think it is no need to do it, so let's forget it and move on.

chrislusf:
Looks like you are winning on the mailing list. Reopening this. But it still has compilation error. Please check again.

aszxqw:
Now it's ok.

chrislusf:
Try to use atomic.Value to simplify this if possible.

aszxqw:
Nice suggestion, but I am not familiar with atomic.Value, maybe someone else can do it.

Issue #94, opened 2015-03-04
What is the name of this project?


stuartpb:
I'm looking to create a Docker image for this project, and maybe some OS-level packages as well. The trouble is, the naming around this project, with regards to its binary, its repo, and its documentation, are contradictory: the documentation refers to it as "Seaweed-FS", its name on sites like bintray is "seaweed", its repo name on GitHub is "weed-fs", and the name of its binary is "weed".

I want to call this project "seaweed", while its daemon name is "weed", in a similar fashion to how the "MongoDB" project's binaries are called "mongos" and "mongod". With this, the name for Docker images for this would be called "seaweed", and the package names to install this would be "seaweed".

Should there be a different name for external packages? If so, what should this name be?

chrislusf:
The names on bintray or github are due to historical reason.

How about use the name "SeaweedFS", similar to "MongoDB", and leave the binary as "weed"?

stuartpb:
I can handle that, though it'd make more sense if the repo were named "SeaweedFS" to reflect that. (GitHub handles repo redirects now, so it wouldn't be so disruptive.)

ArtworkAD:
The repo is still "weed-fs". @chrislusf Are you going to rename it?

chrislusf:
I tried, but did not find a clean way just to rename it.  So I gave up. Anyway, the repo name and the project name do not need to match.

ArtworkAD:
@chrislusf agree but I just started exploring seaweed-fs and it confused me that there are two names for it. If the name is weed-fs then keep it. If it is seaweed-fs then it is worth renaming it. I would create a new rename branch, replace weed-fs with seaweed-fs and give it a try.

chrislusf:
There are also many url redirects, source code changes. Not trivial.

stuartpb:
If you go to https://github.com/chrislusf/weed-fs/settings#rename_field, enter "seaweedfs" as the new name, and click "Rename", GitHub will set up the URL redirects for you.

As for the source code changes, I'll submit a pull request that's ready to go as soon as the repo is moved.

stuartpb:
*Excellent*.

chrislusf:
Big thanks to @stuartpb !

PR #118, opened 2015-04-16
Change all chrislusf/weed-fs links to point to chrislu/seaweedfs

stuartpb:
To be merged after renaming the repo (#94)


stuartpb:
(The Travis failure is to be expected - this change will only work after the repo is moved.)

PR #126, opened 2015-04-19
FIXED: When RaftServer cannot find a leader, Return a more readable erro...

yanyiwu:
...r.

Before:
curl -F "file=1234" "http://127.0.0.1:9333/submit"
{"error":"Post http:///dir/assign: http: no Host in request URL"}
After:
curl -F "file=1234" "http://127.0.0.1:9333/submit"
{"error":"Raft Server not initialized!"}


yanyiwu:
A more readable error message can be used to help developers find what is the real problem.

Issue #134, opened 2015-04-30
[in-progress] Export data changed since X


fedotxxl:
Hello. Can you add param `-changed-since` which will export all files changed since some value? We need it for implementing incremental backups - exporting whole data every day is too expensive 

Technically we can `diff` two tar files but even exporting single volume will generate too many wasteful disk load

chrislusf:
Here is how to backup data efficiently.

https://github.com/chrislusf/seaweedfs/wiki/Data-Backup

fedotxxl:
@dyeldandi added -newer param. We make 1 full backup per week + incremental backup every day by `-newer` param

PR #145, opened 2015-05-20
Using LastModified in export to export only new files and fill up ctime and mtime in tar archive

dyeldandi:
Added -newer argument, which takes a date/time in RFC3339 format and exports only files newer than this time
Using LastModified in TarHeader ctime and mtime


dyeldandi:
Hello!

I think it covers issue #134

PR #174, opened 2015-08-13
try to delete inconsistent file

seifer:
I think we need to delete file in master volume (and replicas to) if there was a failed replication. In order thar vaccum can reclaim it in future. What do you think?

### DIFF
--- a/go/topology/store_replicate.go
+++ b/go/topology/store_replicate.go
@@ -41,6 +41,8 @@ func ReplicatedWrite(masterNode string, s *storage.Store,
 				return err == nil
 			}) {
 				ret = 0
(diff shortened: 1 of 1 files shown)


seifer:
And where specifically there is a race? I cant find this code (( And what to do with trash which will collect as a result of bad replication?

If you explain me where race occurs maybe i can fix it ))

seifer:
Propably this 

for i := 0; i < length; i++ {
			ret = ret && <-results
		}

?

chrislusf:
If the replicated upload is re-tried with the same key, the previous partially successful uploaded entry will be re-used.

PR #187, opened 2015-09-05
Allow whitelisting by CIDR range, not just literally.

skx:
This allows you to write something like this:

         /usr/local/bin/weed master -mdir /srv/weed/master -whiteList=192.168.0.0/24,127.0.0.1

This will whitelist all the 192.168.0.XX hosts, as well as localhost.


chrislusf:
Nice fix!

PR #210, opened 2015-10-19
[bug fix]weed daemon exit due to hangup signal when terminal close

maggot611:
I use cmd below to run weedfs in server.
``` shell
nohup ./weed -log_dir="./log/" volume -ip=10.144.152.250 -mserver="10.164.30.155:9333" -port=8887  -dir="./data0" -max=10000 &
```
when I close the terminal(i use Xshell), weedfs exit, too. I add the log and find weedfs receive signal "hangup", so I ignore syscall.SIGHUP


chrislusf:
good to learn this!

PR #220, opened 2015-11-15
Go1.4.2 signal handling fix and Makefile.

euforia:
Since `signal.Ignore` doesn't seem to be present in 1.4.2 I've created an ignore channel to handle the SIGHUP.

Added a simple makefile as well.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,11 @@
+
+.clean:
+	go clean -i -v ./go/weed/
+
+.deps:
+	go get -d ./go/weed/
+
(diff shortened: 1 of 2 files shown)


chrislusf:
Nice fix! Just wondering why not just move to go 1.5

euforia:
Yeah I realized that after I did the PR, which is by I cancelled. :).  I can still add the makefile if you'd like?  Let me know.

chrislusf:
Make file is nice.

PR #236, opened 2016-01-04
collection name could be specified in url

hxiaodon:
uploading file with filer service could specify the collection name in get/post field, if no collection specified, the default collection setting against the filer service will be used


chrislusf:
Thanks for the pull request! How is it tested?

PR #239, opened 2016-01-13
dereplicate filer post request & support filer get req with raw fileId

hxiaodon:
The filer service uploading/updating operation does not have a dereplication
if someone run the following 2 commands orderly:
curl -F "filename=@server.crt" "http://localhost:8888/path/to/b/"
curl -F "filename=@server.crt" "http://localhost:8888/path/to/b/"
2 file ID will be assigned finally and the 1st one will not be deleted and the corresponding data would occupy the storage service for ever

### DIFF
--- a/go/weed/weed_server/filer_server_handlers.go
+++ b/go/weed/weed_server/filer_server_handlers.go
@@ -6,7 +6,6 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
-	"math/rand"
 	"net/http"
 	"net/url"
 	"strconv"
@@ -31,15 +30,6 @@ func (fs *FilerServer) filerHandler(w http.ResponseWriter, r *http.Request) {
 		fs.PostHandler(w, r)
 	case "POST":
 		fs.PostHandler(w, r)
-	}
-}
-
-func (fs *FilerServer) publicReadOnlyfilerHandler(w http.ResponseWriter, r *http.Request) {
-	switch r.Method {
-	case "GET":
-		fs.GetOrHeadHandler(w, r, true)
-	case "HEAD":
-		fs.GetOrHeadHandler(w, r, false)
 	}
 }
 
@@ -70,6 +60,7 @@ func (fs *FilerServer) listDirectoryHandler(w http.ResponseWriter, r *http.Reque
 	m["Files"], _ = fs.filer.ListFiles(r.URL.Path, lastFileName, limit)
 	writeJsonQuiet(w, r, http.StatusOK, m)
 }
+
 func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request, isGetMethod bool) {
 	if strings.HasSuffix(r.URL.Path, "/") {
 		if fs.disableDirListing {
@@ -79,40 +70,21 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request, 
 		fs.listDirectoryHandler(w, r)
 		return
 	}
-	var fileId string
-	uri_elements := strings.Split(r.URL.Path, "/")
-	//glog.V(1).Infof("debug uri_elements %#v", uri_elements)
-	if len(uri_elements) == 2 {
-		// /3,01e587647e will be split into []string{"","3,01e587647e"}
-		fileId = uri_elements[1]
-	} else {
-		var err error
-		fileId, err = fs.filer.FindFile(r.URL.Path)
-		if err == leveldb.ErrNotFound {
-			glog.V(3).Infoln("Not found in db", r.URL.Path)
-			w.WriteHeader(http.StatusNotFound)
-			return
-		}
-	}
-	parts := strings.Split(fileId, ",")
-	if len(parts) != 2 {
-		glog.V(1).Infoln("Invalid fileId", fileId)
+
+	fileId, err := fs.filer.FindFile(r.URL.Path)
+	if err == leveldb.ErrNotFound {
+		glog.V(3).Infoln("Not found in db", r.URL.Path)
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
-	lookup, lookupError := operation.Lookup(fs.master, parts[0])
-	if lookupError != nil {
-		glog.V(1).Infoln("Invalid lookup", lookupError.Error())
+
+	urlLocation, err := operation.LookupFileId(fs.master, fileId)
+	if err != nil {
+		glog.V(1).Infoln("operation LookupFileId %s failed, err is %s", fileId, err.Error())
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
-	if len(lookup.Locations) == 0 {
-		glog.V(1).Infoln("Can not find location for volume", parts[0])
-		w.WriteHeader(http.StatusNotFound)
-		return
-	}
-	urlLocation := lookup.Locations[rand.Intn(len(lookup.Locations))].Url
-	urlString := "http://" + urlLocation + "/" + fileId
+	urlString := urlLocation
 	if fs.redirectOnRead {
 		http.Redirect(w, r, urlString, http.StatusFound)
 		return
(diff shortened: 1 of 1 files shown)


go/weed/weed_server/filer_server_handlers.go
```diff
@@ -68,11 +79,20 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request,
...
 	}
-	fileId, err := fs.filer.FindFile(r.URL.Path)
-	if err == leveldb.ErrNotFound {
-		glog.V(3).Infoln("Not found in db", r.URL.Path)
-		w.WriteHeader(http.StatusNotFound)
-		return
+	var fileId string
+	uri_elements := strings.Split(r.URL.Path, "/")
+	//glog.V(1).Infof("debug uri_elements %#v", uri_elements)
+	if len(uri_elements) == 2 {
```
chrislusf: you are supporting /file_id url. This format will prevent files under root.

go/weed/weed_server/filer_server_handlers.go
```diff
@@ -32,6 +34,15 @@ func (fs *FilerServer) filerHandler(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
+func (fs *FilerServer) publicReadOnlyfilerHandler(w http.ResponseWriter, r *http.Request) {
```
chrislusf: seems it is your own feature accidentally included? Not bad but it is not complete. Put in a separate pull request please.

go/weed/weed_server/filer_server_handlers.go
```diff
@@ -134,14 +161,63 @@ func (fs *FilerServer) PostHandler(w http.ResponseWriter, r *http.Request) {
...
+		writeJsonError(w, r, http.StatusInternalServerError, err)
 		return
+	} else if fileId != "" {
+		parts := strings.Split(fileId, ",")
+		if len(parts) != 2 {
+			glog.V(1).Infoln("Invalid fileId", fileId)
+			w.WriteHeader(http.StatusNotFound)
+			return
+		}
+		lookup, lookupError := operation.Lookup(fs.master, parts[0])
```
chrislusf: operations.Lookup can handle "3,01e587647e". No need for additional split code above.

go/weed/weed_server/filer_server_handlers.go
```diff
@@ -134,14 +161,63 @@ func (fs *FilerServer) PostHandler(w http.ResponseWriter, r *http.Request) {
...
+	buf, _ := ioutil.ReadAll(r.Body)
+	r.Body = analogueReader{bytes.NewBuffer(buf)}
+	fileName, _, _, _, _, _, _, pe := storage.ParseUpload(r)
+	if pe != nil {
+		glog.V(0).Infoln("failing to parse post body", pe.Error())
+		writeJsonError(w, r, http.StatusInternalServerError, pe)
+		return
+	}
+	//reconstruct http request body for following new request to volume server
+	r.Body = analogueReader{bytes.NewBuffer(buf)}
```
chrislusf: This would require the request be held in memory in filer during the process. Is it possible to avoid this or only do this if it is an update?

hxiaodon: Chris, so far I could not think of a better way to avoid this additional http body copy,  could you give me some suggestion?
if we restrict above code only be implemented when end user trigger an update operation(HTTP PUT instead of HTTP POST), it will be easier and follow the restful API's rules.
Please feel free to give me your opinion and point out my mistake~ Thanks~

go/weed/weed_server/filer_server_handlers.go
hxiaodon: Could I add a regex verification for the url Path , if the root file path does not match the following expression
   ^[0-9]+,([0-9a-fA-F]{2}){5,12}$
then I will treat it as a filePath but not file Id, only the files under root need this calculation

go/weed/weed_server/filer_server_handlers.go
hxiaodon: OK, thanks!

go/weed/weed_server/filer_server_handlers.go
hxiaodon: I think what you mean is the operation.LookupFileId API, I will replace the redudant code(post and get handler) with the following:
        var le error
        urlLocation, le = operation.LookupFileId(fs.master, fileId)
        if le != nil {
            glog.V(1).Infoln("operation LookupFileId %s failed, err is %s", fileId, le.Error())
            return
        }

go/weed/weed_server/filer_server_handlers.go
chrislusf: Separating to PUT and POST is a nice idea!

go/weed/weed_server/filer_server_handlers.go
chrislusf: yes!

go/weed/weed_server/filer_server_handlers.go
chrislusf: It's better not to have this feature. This should be taken care by the volume server or master server redirecting. Mix it to file server is not ideal.

craveica:
@chrislusf  @hxiaodon  
The fix doesn't seam to work when using filer with cassandra.

hxiaodon:
Chris, I think we should wrap and hide the error at  cassandra_store.go:65 , as long as the error is equal to "Not found" but not  "service unavailable/service error"

hxiaodon:
@craveica,  when you try to reproduce this problem with PUT request, you got an 5xx error, is it?

craveica:
No, I don't get any errors. It's just dereplication doesn't work.

```
weed filer -cassandra.server 10.8.202.1  -master 10.8.181.1:9333
curl -F "filename=@server.crt" "http://localhost:8888/path/to/b/"
curl -F "filename=@server.crt" "http://localhost:8888/path/to/b/"
2 file ID will be assigned finally and the 1st one will not be deleted and the corresponding data would occupy the storage service for ever 
```

hxiaodon:
@craveica 
Try using the following command:
curl -X PUT -F "filename=@server.crt" "http://localhost:8888/path/to/b/"
Per Chris's opinion,POST in restful API is a creating operation while PUT means updating
If you get HTTP internal error(5xx), it means that /path/to/b/server.crt is not found in cassandra, then I think we could modify the relevant code(cassandra_store.go:65) to solve this problem ultimately.

craveica:
@hxiaodon 
You're right.
 
```
curl -v -X PUT -F "filename=@img/flags/us.png" "http://localhost:8888/test/img/"

HTTP/1.1 500 Internal Server Error
{"error":"No file id found for /test/img/us.png"}


I0119 05:37:16  6312 filer_server_handlers.go:160] failing to find path in filer store /test/img/us.png No file id found for /test/img/us.png
```

craveica:
One more thing. curl -X DELETE only removes entry from cassandra, but the img is still on volume.
```
SELECT fids FROM seaweed.seaweed_files WHERE path='/test/img/us.png'
 fids
------------------
 ['8,1a8fbc2a25']

curl -v -X DELETE "http://localhost:8888/test/img/us.png"

SELECT fids FROM seaweed.seaweed_files WHERE path='/test/img/us.png';

 path | fids
------+------

curl -I http://10.8.181.12:8080/8,1a8fbc2a25
HTTP/1.1 200 OK
Accept-Ranges: bytes
Content-Disposition: filename="us.png"
Content-Length: 940
Content-Type: image/png
Etag: "fb80808b"
Last-Modified: Tue, 19 Jan 2016 03:45:58 GMT
Date: Tue, 19 Jan 2016 03:54:19 GMT

```

hxiaodon:
Yeah,   this is caused by the following code:
image
I will ask Chris's suggestion to confirm we can change this code or not, if permitted, we can get the fid firstly then delete the record in above code
Then the filer handler could delete the file in volume server
Any way, I will create another issue for tracking

chrislusf:
This line should take care of actual deletion from volume servers.

https://github.com/chrislusf/seaweedfs/blob/master/go/weed/weed_server/filer_server_handlers.go#L256

PR #247, opened 2016-01-22
fix deleting inconsistent problems

rookie-xy:
1. Start master and volume
./weed master -port=9333 -mdir=m/1 -defaultReplication=001
./weed volume -dir=v/1 -port=7080 -mserver="localhost:9333" -dataCenter=dc1 -rack=A -max=2
./weed volume -dir=v/2 -port=7081 -mserver="localhost:9333" -dataCenter=dc1 -rack=A -max=2

2. Access using RESTful interface
curl -F file=@seaweedfs.png http://localhost:9333/submit
{"fid":"1,017c58a505","fileName":"seaweedfs.png","fileUrl":"127.0.0.1:7081/1,017c58a505","size":71155}

curl -X DELETE http://127.0.0.1:7081/1,017c58a505
{"size":71204}

OK! No problem, but bug will appear in the following situations.

3. Bug
If volume server 7080 crashes, then delete a file on the server 7081, you will be prompted to {" error: "deletion Failed."}, but file has been deleted on volume server 7081, and file is not deleted on volume 7080, because it has already been crashing, Is not consistent with 7081 of the data when the volume server 7080 is restarted.

### DIFF
--- a/go/topology/store_replicate.go
+++ b/go/topology/store_replicate.go
@@ -56,13 +56,14 @@ func ReplicatedDelete(masterNode string, store *storage.Store,
 	//check JWT
 	jwt := security.GetJwt(r)
 
-	ret, err := store.Delete(volumeId, n)
-	if err != nil {
-		glog.V(0).Infoln("delete error:", err)
-		return
-	}
-
-	needToReplicate := !store.HasVolume(volumeId)
+	// fix in 0.70 beta
+	// Contributions: Meng Shi (rookie-xy)
+	// date:          2016-01-22 16:02:55
+	// reason: 
+	//        When using the replication technique, found deleting
+	//        inconsistent (or called sync deletion) problems.
+	//
+	needToReplicate := store.HasVolume(volumeId)
 	if !needToReplicate && ret > 0 {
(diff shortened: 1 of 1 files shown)


chrislusf:
The inconsistent state will still happen if there are 3 servers and one of them crash?

rookie-xy:
./weed master -port=9333 -mdir=m/1 -default=001
./weed volume -dir=v/1 -port=7080 -mserver="localhost:9333" -dataCenter=dc1 -rack=A
./weed volume -dir=v/2 -port=7081 -mserver="localhost:9333" -dataCenter=dc1 -rack=A
./weed volume -dir=v/3 -port=7082 -mserver="localhost:9333" -dataCenter=dc1 -rack=A
./weed volume -dir=v/4 -port=7083 -mserver="localhost:9333" -dataCenter=dc1 -rack=A

My test result is that the replication will occupy only two servers and not three or four. Is it where I do wrong?

chrislusf:
The number of volumes will not grow unless manually configured or existing volumes are full. I guess you already have some volumes and they are not full.

PR #254, opened 2016-02-19
Redis filer metastore

zhaozhi406:
I followed the embedded filer, and extend redis_store to support directory related manipulations, like list sub-directories, files, or moving them.

I tried my best to modify fewest source files, and keep the source code clean.

The data structures in redis are as follow:
directories hash, d:/path/to/dir -> {dirname1:dirId1, dirname2, dirId2,...}
files hash, df:/path/to/dir -> {fname1:fid1, fname2:fid2, ...}
max dir id, swfs:dir-max-id->nnn

BUG report:

1. the listDirectoriesHandler(...) in filer_server_handlers.go makes it impossible to list sub-directories and files simultaneously.

2. because now use lastFileName for files list pagination, it's  impossible to show the first file of the first page.

### DIFF
--- a/go/filer/flat_namespace/flat_namespace_store.go
+++ b/go/filer/flat_namespace/flat_namespace_store.go
@@ -1,9 +1,16 @@
 package flat_namespace
 
-import ()
+import "github.com/chrislusf/seaweedfs/go/filer"
 
 type FlatNamespaceStore interface {
 	Put(fullFileName string, fid string) (err error)
 	Get(fullFileName string) (fid string, err error)
 	Delete(fullFileName string) (fid string, err error)
+
+	//redis store support these functions
+	FindDirectory(dirPath string) (dirId filer.DirectoryId, err error)
+	ListDirectories(dirPath string) (dirs []filer.DirectoryEntry, err error)
+	ListFiles(dirPath string, lastFileName string, limit int) (files []filer.FileEntry, err error)
+	DeleteDirectory(dirPath string, recursive bool) (err error)
+	Move(fromPath string, toPath string) error
 }
--- a/go/filer/redis_store/directory_manager_test.go
+++ b/go/filer/redis_store/directory_manager_test.go
@@ -1,0 +1,186 @@
+package redis_store
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"testing"
+
+	redis "gopkg.in/redis.v2"
+)
+
+var dm *DirectoryManager = nil
+
+func TestDirectory(t *testing.T) {
+	//root dir
+	did, err := dm.MakeDirectory("/")
+	if err != nil {
+		t.Errorf("make root dir error:%v\n", err)
+	}
+	if did != 1 {
+		t.Errorf("root dir's id is not 1!")
+	}
+	//make some dirs
+	dir := "/a/b/c"
+	did, err = dm.MakeDirectory(dir)
+	if err != nil {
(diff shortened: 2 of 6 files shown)


go/filer/flat_namespace/flat_namespace_store.go
```diff
@@ -1,9 +1,16 @@
...
 
-import ()
+import "github.com/chrislusf/seaweedfs/go/filer"
 
 type FlatNamespaceStore interface {
 	Put(fullFileName string, fid string) (err error)
 	Get(fullFileName string) (fid string, err error)
 	Delete(fullFileName string) (fid string, err error)
+
+	//redis store support these functions
```
chrislusf: FlagNamespaceStore means just fullpath-to-fid mapping.

Please create a new interface for these operations.

go/filer/redis_store/directory_manager_test.go
```diff
@@ -0,0 +1,264 @@
...
+	n, err := client.Del(dirMaxIdKey).Result()
+	if err != nil {
+		fmt.Printf("del dirMaxIdKey:%s, error:%v\n", dirMaxIdKey, err)
+	} else {
+		fmt.Println("del", n, " keys.")
+	}
+	return err
+}
+func TestMain(m *testing.M) {
+	redisClient := redis.NewTCPClient(&redis.Options{
```
chrislusf, line 273: Any way to make the test run without these special setup?

chrislusf:
Thanks a lot for your effort! 

The continuous integration test is not passing, can you please look into it also?

about bug 2: using LastFileName="" would help to see the first file?

go/filer/flat_namespace/flat_namespace_store.go
zhaozhi406: ok! I will create another interface

go/filer/redis_store/directory_manager_test.go
zhaozhi406, line 273: These setup is necessary to connect to a redis server, when removed I think it's impossible to run the tests. Can you give me some suggestion?

chrislusf, line 273: A redis mock would be nice.

zhaozhi406, line 273: I found it's hard to make a mock for gopkg.in/redis.v2, could you give me an example or I just expose the redis settings from TestMain() to command line arguments? I need help.

chrislusf, line 273: Having a command line for the test would also help. With it, we can just manually run the test pointing to any redis server, instead of requiring users to install redis just to run the test.

But the "go test" would not cover this if switching to manually run the test. 

The test would also need to be in a "main" package if need to run manually.

If you have problem with mock, usually you can think about how to refactor current functions. Usually you can take the redis specific operations out to a different file, and create mock for the redis specific operations.

zhaozhi406:
I have move the redis configuration arguments out to the command line, but the continuous-integration still failed, maybe it can not figure out how to setup the arguments, count you please handle it manually? thank you.

chrislusf:
The command line options will have to run manually with a redis setup. It would not happen regularly.

I would hope any redis specific operation can be separated out and mocked, so that we can test the non-redis part of code regularly. This pull request is fairly large. We should have regular tests to not break it.

chrislusf:
try to use this?

https://github.com/rafaeljusto/redigomock

PR #281, opened 2016-04-06
replace util/bytes.go with binary.BigEndian

thinxer:
The built-in package `binary` provides the same functionality, possibly faster since it has no for-loops.

PR #284, opened 2016-04-09
replace util/bytes.go with binary.BigEndian (again)

thinxer:
Another attempt to fix #281.

There are two places in the code where calling `Uint64` on variant length bytes.

1. ParseKeyHash: should use `strconv` instead.
2. Reading LastModifiedBytes: It's 5 bytes. Replaced by `bytes[0] << 32 | Uint32(bytes[1:5])`.

Now the benchmark runs correctly.

Please review and see if there are any remaining problems.


thinxer:
Besides, here's the benchmark for `ParseKeyHash`

```
before:
BenchmarkParseKeyHash-4	10000000	       241 ns/op	      16 B/op	       1 allocs/op

after:
BenchmarkParseKeyHash-4	20000000	       111 ns/op	       0 B/op	       0 allocs/op
```

chrislusf:
Already merged. But I am thinking why not just change the implementation of util.BytesToUint64(), etc, to binary.BigEndian.Uint64()? BigEndian is just one way to encode. Having a util wrapper to hide this detail makes the code easier to read.

yanyiwu:
yes, agree with `Having a util wrapper to hide this detail makes the code easier to read.`

thinxer:
The fewer indirection, the better.

Why is `binary.BigEndian.Uint64` harder to read than `util.BytesToUint64`?  By using `BigEndian`, it is also clearer that the encoding is BigEndian.  Otherwise, you will have to document the `BytesToUint64` function and describe it as BigEndian.

I agree that, in the `LastModifiedBytes` case, it is harder to reason about.  However, I don't think it is a good idea to use 5 bytes to represent a `uint64` anyway.

chrislusf:
It's easier to read Bytes to Uint64, than BigEndian to Uint64. BigEndian is implementation detail.

The indirection can be optimized out when Go complier inline those functions.

thinxer:
Actually BigEndian is not an implementation detail.  It must not be changed to LittleEndian or things go wrong.  Some usages depend on the converter being BigEndian, like the old implementation of `ParseKeyHash`, which converts hex string to big-endian bytes first.

On the readability part, I am assuming everyone is (or should be ) familiar with Go's built-in packages.  Readers should understand that the `binary` package is for value and bytes conversion.

chrislusf:
"implementation detail" does not mean it can be changed anyway we want. It just mean it is a lower level details. When I was coding it at the beginning, I just think it is a []byte to uint64 conversion. Nothing else. And randomly chose to use BigEndian. But most other developers does not really need to know this.

chrislusf:
I reverted this change due to the above reasons. However, I do like your implementation for parseKeyHash() and the test cases. You are welcome to PR for them.

And do not be discouraged for this. Your changes are welcome.

thinxer:
In the future you may have more comments before merging.  Merging and reverting results in cluttered git history and may cause confusion for future contributors.

I'll have another PR for ParseKeyHash implementation, though it may not be very important.

In the meanwhile, I've found several places where the built-in `binary` package could actually simplify the implementation.  Again, these may not meet your requirement for _readability_.  I'll see if they can be merged, or a fork may be the only solution.

PR #285, opened 2016-04-10
storage: do not copy sync.Mutex by value

thinxer:
`baseNeedleMapper` contains `sync.Mutex` which should not be copied by value.


chrislusf:
Thanks for this PR! Is this just by reading the source code or noticed during some tests?

thinxer:
`go vet` did the job.

PR #302, opened 2016-05-12
Fix Content-Disposition header

alexanderk23:
Fixes **Content-Disposition** header as defined by RFC6266
Adds **dl=(0|1)** URI param to select inline (default) or attachment content disposition


chrislusf:
Super!!!

PR #313, opened 2016-06-08
made a little change for Amazon S3 API's compatibility

hxiaodon:
Chris
We have a scenario that the file's uploading would not follow http multipart way. [Amazon S3 PUT API](http://docs.aws.amazon.com/AmazonS3/latest/API/RESTObjectPUT.html)

I made a simple adapter for above case, please help to review the code change
Thanks!

### DIFF
--- a/weed/server/filer_server_handlers_write.go
+++ b/weed/server/filer_server_handlers_write.go
@@ -2,10 +2,16 @@ package weed_server
 
 import (
 	"bytes"
+	"crypto/md5"
+	"encoding/base64"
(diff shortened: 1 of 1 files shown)


chrislusf:
How about refactor it a bit? The function is getting longer and longer. It's a good time to put the logic into smaller functions with your changes.

hxiaodon:
Good suggestion, I will refactor and submit again

hxiaodon:
Will create another new merge request after refactoring work

PR #316, opened 2016-06-10
retry delete fileId

030io:


030io:
I was wrong. Not the reason.

PR #327, opened 2016-06-23
add options params for the Assign func

hxiaodon:
This fix is to solve the problem that explicitly specifying the main [dataCenter | rack | dataNode] now is disabled

### DIFF
--- a/weed/command/benchmark.go
+++ b/weed/command/benchmark.go
@@ -205,7 +205,11 @@ func writeFiles(idChan chan int, fileIdLineChan chan string, s *stat) {
 		start := time.Now()
 		fileSize := int64(*b.fileSize + rand.Intn(64))
 		fp := &operation.FilePart{Reader: &FakeReader{id: uint64(id), size: fileSize}, FileSize: fileSize}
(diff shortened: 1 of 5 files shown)


chrislusf:
Thanks for the nice hack! However, this may lead to future API usage confusions. Explicit declaration is better.

Could you please refactor a bit more, so that the API becomes

`
type VolumeAssignRequest stuct{
  count uint64
  replication string
  collection string
  ttl string
  dataCenter string
  rack string
  server string
}
func Assign(server string, r VolumeAssignRequest) (*AssignResult, error) 
`

hxiaodon:
Sure!  We should not leave the technical debt for other ones and bring in hidden trouble for future
And pointer type should be better(avoid additional copy and ref count for each struct field), while the object is still on stack space(escape analysis) 
And the usage should be:

`
volumeAssignReq := &VolumeAssignRequest{
count : 1, 
replication : "001",
}

func Assign(server string, r \*VolumeAssignRequest) (*AssignResult, error)
`

I will submit later (by the end of this weekend), I'm working on seaweedfs's online ops work at prod environment with a lot of trivial matter...

chrislusf:
Let's add the field one by one to be clear and fast.

PR #333, opened 2016-06-29
fix volume_layout bug, add new feature for volume etag write,filer pa…

hxiaodon:


weed/server/volume_server_handlers_write.go
```diff
@@ -42,6 +42,8 @@ func (vs *VolumeServer) PostHandler(w http.ResponseWriter, r *http.Request) {
 		ret.Name = string(needle.Name)
 	}
 	ret.Size = size
+	etag := needle.Etag()
+	w.Header().Set("Etag", etag)
```
chrislusf, line 46: This way to use Etag seems not specified in http://tools.ietf.org/html/rfc7232#section-2.4

chrislusf:
Later try to keep them in separated commits at least.

PR #345, opened 2016-07-21
recover filer read handler logic

hxiaodon:
@cschiano @chrislusf 

I just merge the new list directory function and existed read file logic together. Please take a review, any concern, please let me know

Thanks!


chrislusf:
Thanks for fixing this.

PR #352, opened 2016-08-05
Adding AutoChunk/MaxMB Support to Filer API

mtolman:
This is related to the following issue I added to chrislusf/seaweedfs:
https://github.com/chrislusf/seaweedfs/issues/342

I'm VERY new to golang (and github), so let me know if there is anything I should be doing differently.

Thanks!
Mike


chrislusf:
Thanks!

Your code looks professional. :) That's what I like about Go. It helps to focus on the actual logic, rather than syntax.

PR #354, opened 2016-08-09
if replicated volume has one copy in readonly mode at one node,it sho…

hxiaodon:
…uld be removed from writable list

Disk fault or power-down may lead to the volume file in readonly mode at a specific machine. In this scenario, if the volume is a replicated one,  the sending heartbeat from different volume nodes to master node will switch between adding  it into and removing it from the master's writable list 

So I add a fix for above case

@chrislusf Chris, please help to review the code.


chrislusf:
Thanks!

PR #361, opened 2016-08-31
add mysql filer support

hxiaodon:
Chris,
Please help to take a review. 
I add a simple hash sharding for mysql storage since billions of records at one table will lead to the performance bottleneck, and involving the mysql proxy(middleware solution) for the weed 's filer store is overkill


weed/filer/mysql_store/mysql_store.go
```diff
@@ -0,0 +1,224 @@
...
+
+func (s *MySqlStore) Get(fullFilePath string) (fid string, err error) {
+	instance_offset, tableFullName, err := s.parseFilerMappingInfo(fullFilePath)
+	if err != nil {
+		return "", err
+	}
+	fid, err = s.query(fullFilePath, s.dbs[instance_offset], tableFullName)
+	if err == sql.ErrNoRows {
+		//Could not found
+		err = nil
```
chrislusf: Should return some error here if not found.

weed/filer/mysql_store/mysql_store.go
```diff
@@ -0,0 +1,224 @@
...
+	instance_offset, table_postfix := s.hash(path)
+	instanceId = instance_offset
+	tableFullName = fmt.Sprintf("%s_%04d", tableName, table_postfix)
+	return
+}
+
+func (s *MySqlStore) Get(fullFilePath string) (fid string, err error) {
+	instance_offset, tableFullName, err := s.parseFilerMappingInfo(fullFilePath)
+	if err != nil {
+		return "", err
```
chrislusf: Please use:
fmt.Errorf("MySqlStore Get operation can not parse file path %s: %v", fullFilePath, err)

This way, it'll be clear when the error happens. Same changes are needed for everywhere that's returning error as is.

weed/filer/mysql_store/filer_mapping.sql
```diff
@@ -0,0 +1,11 @@
+CREATE TABLE IF NOT EXISTS `filer_mapping` (
```
chrislusf: This file is not needed since you are auto creating the table.
You can add a README.txt file and add the SQL there.

Also, add example json config files in the README.

Hashing into tables should be an option. Or the user can specify number of tables to 1 in the json config file.

weed/filer/mysql_store/mysql_store.go
```diff
@@ -0,0 +1,224 @@
...
+}
+
+func (s *MySqlStore) Close() {
+	for _, db := range s.dbs {
+		db.Close()
+	}
+}
+
+var createTable = `
+CREATE TABLE IF NOT EXISTS %s_%04d (
```
chrislusf: If the user specify one shard table, it's better to remove the underscore in the table name.

weed/filer/mysql_store/mysql_store.go
```diff
@@ -0,0 +1,224 @@
...
+  id bigint(20) NOT NULL AUTO_INCREMENT,
+  uriPath char(256) NOT NULL DEFAULT "" COMMENT 'http uriPath',
+  fid char(36) NOT NULL DEFAULT "" COMMENT 'seaweedfs fid',
+  createTime int(10) NOT NULL DEFAULT 0 COMMENT 'createdTime in unix timestamp',
+  updateTime int(10) NOT NULL DEFAULT 0 COMMENT 'updatedTime in unix timestamp',
+  remark varchar(20) NOT NULL DEFAULT "" COMMENT 'reserverd field',
+  status tinyint(2) DEFAULT '1' COMMENT 'resource status',
+  PRIMARY KEY (id),
+  UNIQUE KEY index_uriPath (uriPath)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```
chrislusf: InnoDB is required? Maybe just these options empty.

weed/filer/mysql_store/mysql_store.go
hxiaodon: Could we make a convention which will not throw NotFound error in each filer mapping store package? If the <urlPath,fid> mapping entry is not found, fid="" while err=nil
Then in filer API's handler function, we do not need write the following logic:
`If err!=levelDB.NotFound && err!=mysql.NotFound`
Instead, the code should be:

```
if err!=nil{
...
} else{
    //not found
    if fid==""{
    }
}
```

Let me know if my understanding has something wrong. Thanks!

weed/filer/mysql_store/mysql_store.go
hxiaodon: Sure, it does make sense

weed/filer/mysql_store/mysql_store.go
hxiaodon: Good suggestion. I will add flag field in the json conf file

weed/filer/mysql_store/mysql_store.go
chrislusf: To follow Golang's convention, any error should not be simply passed to the upper level.

For mysql specific error, it should be wrapped to another error meaningful to current logic. Just return the error will work, but not recommended.

if err == mysql.NotFound {
    return fmt.Errorf("Failed to find key: %v", err)
}

weed/filer/mysql_store/README.md
```diff
@@ -0,0 +1,66 @@
...
+        {
+            "User": "root",
+            "Password": "root",
+            "HostName": "127.0.0.2",
+            "Port": 3306,
+            "DataBase": "seaweedfs"
+        }
+    ],
+    "IsSharding":true,
+    "ShardingNum":1024
```
chrislusf: "ShardingNum" can be changed to shardCount?

weed/filer/mysql_store/mysql_store.go
chrislusf: For existence check, I think the common practice is to return one additional boolean indicating existence.

hxiaodon: If we wrap the sql.ErrNoRows(Not Found) error in mysql_store.go, and return it to the filer_servers_handler_read.go, the fs.filer.FindFile 's logic will be:

```
func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request, isGetMethod bool) {
    if strings.HasSuffix(r.URL.Path, "/") {
        if fs.disableDirListing {
            w.WriteHeader(http.StatusMethodNotAllowed)
            return
        }
        fs.listDirectoryHandler(w, r)
        return
    }

    fileId, err := fs.filer.FindFile(r.URL.Path)
        if err == leveldb.ErrNotFound || strings.Contains(err.Error(),"Failed to find key") {
        glog.V(3).Infoln("Not found in db", r.URL.Path)
        return
    }

```

Does above code make sense?  Or we have some other better way to solve this problem?(wrapped error VS [predefined|hidden] error) Sometimes we need to know the accurate meaning against wrapped error(e.g It's a "not found" error but not a "sql execution" error) , because the following logic will send back 404 or 5xx http status code depending on the meaning

weed/filer/mysql_store/README.md
hxiaodon: Sure, I will make the change~

weed/filer/mysql_store/mysql_store.go
chrislusf: you can also refactor existing leveldb redis code to return a weedfs.FileNotFound error.

hxiaodon: OK, I will do that. Not found error in each kind of filer store will be wrapped and unified to a const fileNotFound error

chrislusf:
Big thanks!

PR #363, opened 2016-09-05
support filer with mysql

biliucheng:


chrislusf:
this is the same as https://github.com/chrislusf/seaweedfs/pull/361
It's better to consolidate the effort.

your code is simpler. but I would like to avoid using the command line options to configure databases.

PR #366, opened 2016-09-08
Changed Dockerfile so it uses a special entrypoint script.

jze:
All parameters are passed through to weed. Depending on the command the entrypoint.sh script adds parameters to link containers.


chrislusf:
Thanks! Does the README also need any changes?

PR #375, opened 2016-09-29
supplemental data between compacting and commit compacting

hxiaodon:
Chris:

This is the draft code for issue [374](https://github.com/chrislusf/seaweedfs/issues/374), please help to review and give me feedback.
Thanks!

### DIFF
--- a/weed/storage/volume.go
+++ b/weed/storage/volume.go
@@ -10,11 +10,6 @@ import (
 	"github.com/chrislusf/seaweedfs/weed/glog"
 )
 
-type keyField struct {
-	offset uint32
-	size   uint32
-}
-
 type Volume struct {
 	Id            VolumeId
 	dir           string
@@ -29,8 +24,8 @@ type Volume struct {
 	dataFileAccessLock sync.Mutex
 	lastModifiedTime   uint64 //unix time in seconds
 
-	lastCompactingIndexOffset       uint64
-	incrementedHasUpdatedIndexEntry map[uint64]keyField
+	lastCompactIndexOffset uint64
+	lastCompactRevision    uint16
 }
 
 func NewVolume(dirname string, collection string, id VolumeId, needleMapKind NeedleMapType, replicaPlacement *ReplicaPlacement, ttl *TTL) (v *Volume, e error) {
--- a/weed/storage/volume.go
+++ b/weed/storage/volume.go
@@ -10,6 +10,11 @@ import (
 	"github.com/chrislusf/seaweedfs/weed/glog"
 )
 
+type keyField struct {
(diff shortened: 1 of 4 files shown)


weed/storage/volume.go
```diff
@@ -23,6 +28,9 @@ type Volume struct {
 
 	dataFileAccessLock sync.Mutex
 	lastModifiedTime   uint64 //unix time in seconds
+
+	lastCompactingIndexOffset       uint64
+	incrementedHasUpdatedIndexEntry map[uint64]keyField
```
chrislusf: incrementedHasUpdatedIndexEntry does not need to be in this struct.

hxiaodon: Yeah, I know current garbage collection is ordered(one by one volume). But I have the concern that when concurrent compacting is involved in weed some day, these 2 fields should become volume specific.

hxiaodon:
the old index offset directly writing into the new index  has something, the new dat file's offset should be used. Anyway, I will modify the code finally after consolidate all you comments

weed/storage/volume.go
chrislusf: I am only talking about incrementedHasUpdatedIndexEntry. The map only used locally in makeupDiff()

weed/storage/volume_vacuum.go
```diff
@@ -55,7 +71,85 @@ func (v *Volume) commitCompact() error {
 	return nil
 }
 
-func makeupDiff(newDatFile, newIdxFile, oldDatFile, oldIdxFile string) (err error) {
+func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldIdxFileName string) (err error) {
```
chrislusf, line 90: How about adding a test case for this function?

hxiaodon: Is there some good ways to do it? I 'm thinking about creating 4 test volume/index files in advance(new creating, updating and deleting appended data should be in them), then write a test case  with these volume files as input

weed/storage/volume.go
hxiaodon: Got it, my local code has been modified

weed/storage/volume_vacuum.go
chrislusf: yes. this would work.

hxiaodon:
Chris,
Please take a review of my UT case. Since the UT input files differ one from each other, so I add the comments for it and comment the UT case
And I did run the ut case at my local, fix some bugs(such as fakedDelNeedle.) Please help to check my fix and give your feedback
Thanks!

PR #395, opened 2016-11-11
concurrent loading volume when starting volume server

hxiaodon:
Chris:
Please help to review, thanks!

### DIFF
--- a/weed/storage/disk_location.go
+++ b/weed/storage/disk_location.go
@@ -22,11 +22,43 @@ func NewDiskLocation(dir string, maxVolumeCount int) *DiskLocation {
 	return location
 }
 
-func (l *DiskLocation) loadExistingVolumes(needleMapKind NeedleMapType) {
-	l.Lock()
-	defer l.Unlock()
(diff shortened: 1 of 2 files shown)


weed/storage/disk_location.go
```diff
@@ -25,29 +26,53 @@ func (l *DiskLocation) loadExistingVolumes(needleMapKind NeedleMapType) {
...
+	const concurrency int = 10
+	var wg sync.WaitGroup
+	var mutex sync.RWMutex
+	for workerNum := 0; workerNum < concurrency; workerNum++ {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for dir := range task_queue {
+				name := dir.Name()
+				if !dir.IsDir() && strings.HasSuffix(name, ".dat") {
```
chrislusf: The logic is clear. But it will be better to move the actual loading code into its own function. The concurrent loading code can also be another separate function.

chrislusf:
Thanks!!!

PR #411, opened 2016-12-08
Adding PostgreSQL Filer Support

mtolman:
Hello - I've been working on a PostgreSQL filer store. It is based on the existing mysql implementation.
To use it, you use the same filer.confFile configuration file, and add a "postgres" section instead of a "mysql" section.

My golang experience is still weak, but it seems to be working fine.


chrislusf:
Thanks!

PR #427, opened 2016-12-31
add another filer authentication support

hxiaodon:
Chris:

Please help to review about this new authentication support.
It use AWS Restful API's V2 authentication way. [link](http://docs.aws.amazon.com/AmazonS3/latest/dev/RESTAuthentication.html)

Hope for your suggestion~

### DIFF
--- a/weed/security/guard.go
+++ b/weed/security/guard.go
@@ -1,10 +1,16 @@
 package security
 
 import (
+	"bytes"
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"net"
 	"net/http"
+	"net/url"
+	"sort"
 	"strings"
 
 	"github.com/chrislusf/seaweedfs/weed/glog"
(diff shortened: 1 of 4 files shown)


chrislusf:
Please write up the design first. What's the expected work flow. What it can do and what it can not do.

hxiaodon:
Hi,Chris:
Please take a review, anything I said is not clear, please let me know
Thanks!

chrislusf:
I originally was thinking to use something similar to amazon's json web token for file access on volume servers. Is this related to that or just orthogonal to it?

hxiaodon:
My understanding is that AWS  [API_AccessKey](http://docs.aws.amazon.com/IAM/latest/APIReference/API_AccessKey.html) and [JWT](https://aws.amazon.com/cn/blogs/compute/tag/jwt/) are 2 authentication ways but somewhat similar(both have hash signature at C/S side). JWT is more flexible since it could include more meta info in raw json.

Considering API_AccessKey has been widely used, merging it into seaweedfs should probably be reasonable for the scenarios that we do not want to bring in trouble/risk when changing existed end user's client sdk from accessKey supported one to jwt supported one. 

The AppKeyId and AppKeySecret is generated at server side. Take AWS as an example, please refer to [link](http://docs.aws.amazon.com/general/latest/gr/aws-sec-cred-types.html#access-keys-and-secret-access-keys). In my local online environment, I wrote a simple tool to generate the access-key and secret-access-key, OPs config them in filer.json and restart the service, tell the specific end user about the key pair, they will update their client sdk(such as [sdk source code](https://github.com/aliyun/aliyun-oss-go-sdk/blob/master/sample/new_bucket.go) line 13)

The bucket is indeed confusing, I think it is related to multi-tenancy technology. 
If the cloud storage service provide a service with a domain name, you could access the resource by the following 2 ways:
1 . http://$bucketName.domainName/objectName  (subdomain)
2. http://domainName/bucketName/objectName (hybrid deployment at single logical cluster?)

I feel that Seaweedfs's collection feature is similar as the 2nd way, although seaweedfs get the collection name from the url param instead of 1st level uri path

Because at my local environment, different collection(bucket) has different accessKey,accessKeySecret, so in my committing file,  I add a new argument(AppConf) for the NewGuard function which seems like a rough solution

JWT just need a secret key while my new Authentication code need a json conf. Do you have good suggestions for it?  Feel free to cancel this pull request if seaweed could benifit from it

Thanks~

PR #430, opened 2017-01-05
Update Dockerfile

vancepym:
Replace with Alpine Linux based image


chrislusf:
I do not really use docker. What's the difference?

chrislusf:
ok. Seen this. https://livewyer.io/blog/2015/02/24/slimming-down-your-docker-containers-alpine-linux/

Thanks!

PR #432, opened 2017-01-08
support additional header name-value pairs

sparklxb:
The name of pairs must begin with 'Seaweed-'. Common use cases have been tested, other than some API I'm unfamiliar with, such as 'Filer server'.

### DIFF
--- a/weed/operation/upload_content.go
+++ b/weed/operation/upload_content.go
@@ -36,13 +36,13 @@ func init() {
 
 var fileNameEscaper = strings.NewReplacer("\\", "\\\\", "\"", "\\\"")
 
-func Upload(uploadUrl string, filename string, reader io.Reader, isGzipped bool, mtype string, pairs []byte, jwt security.EncodedJwt) (*UploadResult, error) {
+func Upload(uploadUrl string, filename string, reader io.Reader, isGzipped bool, mtype string, pairMap map[string]string, jwt security.EncodedJwt) (*UploadResult, error) {
 	return upload_content(uploadUrl, func(w io.Writer) (err error) {
 		_, err = io.Copy(w, reader)
 		return
-	}, filename, isGzipped, mtype, pairs, jwt)
+	}, filename, isGzipped, mtype, pairMap, jwt)
 }
-func upload_content(uploadUrl string, fillBufferFunction func(w io.Writer) error, filename string, isGzipped bool, mtype string, pairs []byte, jwt security.EncodedJwt) (*UploadResult, error) {
+func upload_content(uploadUrl string, fillBufferFunction func(w io.Writer) error, filename string, isGzipped bool, mtype string, pairMap map[string]string, jwt security.EncodedJwt) (*UploadResult, error) {
 	body_buf := bytes.NewBufferString("")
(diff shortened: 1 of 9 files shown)


weed/storage/needle.go
```diff
@@ -55,8 +60,17 @@ func (n *Needle) String() (str string) {
 }
 
 func ParseUpload(r *http.Request) (
-	fileName string, data []byte, mimeType string, isGzipped bool,
+	fileName string, data []byte, mimeType string, pairs []byte, isGzipped bool,
 	modifiedTime uint64, ttl *TTL, isChunkedFile bool, e error) {
+	pairMap := make(map[string]string)
+	for k, v := range r.Header {
+		if len(v) > 0 && strings.HasPrefix(k, PairNamePrefix) {
+			pairMap[k] = v[0]
```
chrislusf, line 67: If the key is "Seaweed-xxxx" with value "yyyy", the map should store "xxxx"~"yyyy"

weed/storage/needle.go
```diff
@@ -55,8 +60,17 @@ func (n *Needle) String() (str string) {
 }
 
 func ParseUpload(r *http.Request) (
-	fileName string, data []byte, mimeType string, isGzipped bool,
+	fileName string, data []byte, mimeType string, pairs []byte, isGzipped bool,
```
chrislusf: return pairs as map[string]string

weed/operation/upload_content.go
```diff
@@ -36,13 +36,13 @@ func init() {
 
 var fileNameEscaper = strings.NewReplacer("\\", "\\\\", "\"", "\\\"")
 
-func Upload(uploadUrl string, filename string, reader io.Reader, isGzipped bool, mtype string, jwt security.EncodedJwt) (*UploadResult, error) {
+func Upload(uploadUrl string, filename string, reader io.Reader, isGzipped bool, mtype string, pairs []byte, jwt security.EncodedJwt) (*UploadResult, error) {
```
chrislusf: It will be better if pairs is map[string]string

weed/storage/needle.go
sparklxb: @chrislusf Considering the situation when uploading file directly with addition pairs, keys are only trimmed off prefix when needle is stored (in function NewNeedle).

weed/storage/needle.go
```diff
@@ -109,19 +120,19 @@ func ParseUpload(r *http.Request) (
...
+		ext = strings.ToLower(fileName[dotIndex:])
+		mtype = mime.TypeByExtension(ext)
+	}
+	contentType := part.Header.Get("Content-Type")
+	if contentType != "" && mtype != contentType {
+		mimeType = contentType //only return mime type if not deductable
+		mtype = contentType
+	}
+
 	if !isChunkedFile {
```
chrislusf, line 135: This section is moved out of the !isChunkedFile clause. Any reason for this?

chrislusf: I will merge this PR and move this section back.

sparklxb: It's a mistake. Now I figure out there is no need to get mimeType where uploading chunked files. It's alright to move back.

PR #434, opened 2017-01-09
fix the bug that we can't get filename when download file.

wangjie:
fix the bug that we can't get filename when download file.


chrislusf:
Thanks!

chrislusf:
fix https://github.com/chrislusf/seaweedfs/issues/433

PR #442, opened 2017-01-18
Fix counting volume size caused by changed tombstone size

brstgt:
Please review this patch. This bug made our cluster mostly unwriteable.


chrislusf:
thanks for the PR! I fixed it following this PR in commit 59022b6

PR #443, opened 2017-01-20
Online mount, unmount + deletion of single volumes

brstgt:
- Logs volume id if actual volume size differs dataFileSize
- Add routes to delete, mount and unmount volumes online without restarting server

### DIFF
--- a/weed/storage/volume_checking.go
+++ b/weed/storage/volume_checking.go
@@ -12,28 +12,28 @@ func getActualSize(size uint32) int64 {
 	return NeedleHeaderSize + int64(size) + NeedleChecksumSize + int64(padding)
 }
 
-func CheckVolumeDataIntegrity(v *Volume, indexFile *os.File) (int64, error) {
+func CheckVolumeDataIntegrity(v *Volume, indexFile *os.File) (error) {
 	var indexSize int64
 	var e error
 	if indexSize, e = verifyIndexFileIntegrity(indexFile); e != nil {
-		return 0, fmt.Errorf("verifyIndexFileIntegrity %s failed: %v", indexFile.Name(), e)
+		return fmt.Errorf("verifyIndexFileIntegrity %s failed: %v", indexFile.Name(), e)
 	}
 	if indexSize == 0 {
-		return int64(SuperBlockSize), nil
+		return nil
 	}
 	var lastIdxEntry []byte
 	if lastIdxEntry, e = readIndexEntryAtOffset(indexFile, indexSize-NeedleIndexSize); e != nil {
-		return 0, fmt.Errorf("readLastIndexEntry %s failed: %v", indexFile.Name(), e)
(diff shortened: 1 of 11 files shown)


ijunaid8088:
is this related to that issue you mentioned a few days ago? in latest version?

brstgt:
Yes, somehow.

1. I have to know which volumes are not integer
2. As I have to repair terabytes of data I want to do it as parallel as possible without taking whole nodes down if I only want to operate on a single volume. So I can unmount some files working on them in parallel (on many nodes in parallel) without risking that these volumes become completely unavailable.

weed/storage/volume_checking.go
```diff
@@ -12,28 +12,28 @@ func getActualSize(size uint32) int64 {
 	return NeedleHeaderSize + int64(size) + NeedleChecksumSize + int64(padding)
 }
 
-func CheckVolumeDataIntegrity(v *Volume, indexFile *os.File) (int64, error) {
+func CheckVolumeDataIntegrity(v *Volume, indexFile *os.File) (error) {
```
chrislusf, line 15: The offset is needed to check volume dat file integrity. Any reason to remove this?

weed/storage/volume_read_write.go
```diff
@@ -139,18 +131,15 @@ func (v *Volume) deleteNeedle(n *Needle) (uint32, error) {
...
 	if ok && nv.Size != TombstoneFileSize {
 		size := nv.Size
-		// println("adding tombstone", n.Id, "at offset", v.dataFileSize)
-		if err := v.nm.Delete(n.Id, uint32(v.dataFileSize/NeedlePaddingSize)); err != nil {
+		offset, err := v.dataFile.Seek(0, 2)
+		if err != nil {
 			return size, err
 		}
-		if offset, err := v.dataFile.Seek(0, 2); err != nil {
+		if err := v.nm.Delete(n.Id, uint32(offset/NeedlePaddingSize)); err != nil {
```
chrislusf, line 138: Good fix!

weed/storage/volume_checking.go
brstgt: I only removed the return value which was used to initialize Volume.dataFileSize. I did not change the logic of the integrity check. The problem was that the returned value was 0 if the last needle was a tombstone, so the dataFileSize was then set to 0 which was definitively wrong. But because I removed the whole dataFileSize property the return value was then unused anyway and can be completely removed.

chrislusf: Actually you are right.

chrislusf:
Thanks!

brstgt:
Thanks, too :)

chrislusf:
btw: I added locking to some of functions you added..

brstgt:
Oh thanks! Right, I also had this in my mind today, but then forgot.

2017-01-20 19:20 GMT+01:00 Chris Lu <notifications@github.com>:

PR #469, opened 2017-03-10
complement weed mount

sparklxb:
File read is not supported yet. It needs more modification of filer backend.

### DIFF
--- a/weed/command/mount_std.go
+++ b/weed/command/mount_std.go
@@ -107,8 +107,6 @@ func (dir *Dir) ReadDirAll(ctx context.Context) ([]fuse.Dirent, error) {
 	return ret, nil
 }
 
-const greeting = "hello, world\n"
-
 type File struct {
 	Id uint64
 	// FileId filer.FileId
@@ -117,11 +115,6 @@ type File struct {
 
 func (file *File) Attr(context context.Context, attr *fuse.Attr) error {
 	attr.Inode = file.Id
-	attr.Mode = 0444
-	attr.Size = uint64(len(greeting))
+	attr.Mode = 0000
 	return nil
-}
-
-func (file *File) ReadAll(ctx context.Context) ([]byte, error) {
-	return []byte(greeting), nil
 }
--- a/weed/command/mount_std.go
+++ b/weed/command/mount_std.go
@@ -7,6 +7,8 @@ import (
 	"os"
 	"runtime"
 
+	"path"
+
 	"bazil.org/fuse"
 	"bazil.org/fuse/fs"
 	"github.com/chrislusf/seaweedfs/weed/filer"
@@ -23,7 +25,7 @@ func runMount(cmd *Command, args []string) bool {
 		return false
 	}
 
-	c, err := fuse.Mount(*mountOptions.dir)
+	c, err := fuse.Mount(*mountOptions.dir, fuse.LocalVolume())
 	if err != nil {
 		glog.Fatal(err)
 		return false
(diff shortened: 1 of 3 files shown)


weed/command/mount_std.go
```diff
@@ -51,61 +53,75 @@ func runMount(cmd *Command, args []string) bool {
...
+	// FileId filer.FileId
+	Name string
 }
 
-func (File) Attr(context context.Context, attr *fuse.Attr) error {
-	attr.Inode = 2
+func (file *File) Attr(context context.Context, attr *fuse.Attr) error {
+	attr.Inode = file.Id
 	attr.Mode = 0444
+	attr.Size = uint64(len(greeting))
```
chrislusf: better skip the size, since it will need to call volume server to get the size info.

weed/server/filer_server_handlers_api.go
```diff
@@ -0,0 +1,40 @@
...
+	case "listDirectories":
+		res := filer.ListDirectoriesResult{}
+		res.Directories, err = fs.filer.ListDirectories(apiRequest.Directory)
+		if err != nil {
+			res.Error = err.Error()
+		}
+		writeJsonQuiet(w, r, http.StatusOK, res)
+	case "listFiles":
+		res := filer.ListFilesResult{}
+		limit := 1
```
chrislusf: limit can always be 100. The apiRequest.FileName is more like "offset" in a file list query "limit 100 offset x".

weed/command/mount_std.go
```diff
@@ -51,61 +53,75 @@ func runMount(cmd *Command, args []string) bool {
...
+		if dirent.Type == fuse.DT_File {
+			return &File{dirent.Inode, dirent.Name}, nil
+		}
+		return &Dir{
+			Id:   dirent.Inode,
+			Path: path.Join(dir.Path, dirent.Name),
+		}, nil
 	}
-	return nil, fmt.Errorf("File Not Found for %s", name)
+	return nil, fuse.ENOENT
```
chrislusf, line 81: better not to return  direct fuse package error if possible.

weed/command/mount_std.go
sparklxb: If skip the size, the size will be 0, so when reading this file, we get nothing. Do I need to make the file mode unreadable? If there is a need to to support file read, I'll add it latter.

chrislusf: do what you think is right here. I know no more than you here.

weed/command/mount_std.go
sparklxb: This error will received by fuse package and not be exported. So just follow the rule of this package.

weed/command/mount_std.go
sparklxb: In order to get the size, call volume server to get the size info is not a good idea, which will make many request. I suggest file size be saved in filer server.

chrislusf: Storing file size on filer is a much bigger change. Unlikely we will go that direction.

Use some constant value, such as 0, or 100, for size.

sparklxb: If we use constant value, there is a dilemma. Supposing 5 is used and we read a file with content 'Hello world', we can only get 'Hello', and the remain can not be read. I suggest use 0 and make file unreadable as my commit do.

chrislusf: the file content will use the fid to retrieve.

what will truncate the file content by this file size?

sparklxb: It's the OS. Package bazil.org/fuse don't truncate content, but OS only read bytes of file size. We can figure it out by stracing the system calls when `cat` the file content.

PR #608, opened 2017-12-21
extend export command to show tombstone + change output format to CSV

brstgt:
What do you think about that?
Exporting modification times + tombstone information may be pretty helpful when building external tools like sync/repair scripts
If you don't want to break BC, we could introduce an option to output CSV. I think for bulk operations, CSV is better suited as it is more dense and doesn't repeat the key names in every row.

### DIFF
--- a/weed/command/export.go
+++ b/weed/command/export.go
@@ -49,10 +49,11 @@ func init() {
 }
 
 var (
-	output = cmdExport.Flag.String("o", "", "output tar file name, must ends with .tar, or just a \"-\" for stdout")
-	format = cmdExport.Flag.String("fileNameFormat", defaultFnFormat, "filename formatted with {{.Mime}} {{.Id}} {{.Name}} {{.Ext}}")
-	newer  = cmdExport.Flag.String("newer", "", "export only files newer than this time, default is all files. Must be specified in RFC3339 without timezone, e.g. 2006-01-02T15:04:05")
+	output      = cmdExport.Flag.String("o", "", "output tar file name, must ends with .tar, or just a \"-\" for stdout")
+	format      = cmdExport.Flag.String("fileNameFormat", defaultFnFormat, "filename formatted with {{.Mime}} {{.Id}} {{.Name}} {{.Ext}}")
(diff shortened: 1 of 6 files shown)


weed/command/export.go
```diff
@@ -125,6 +144,10 @@ func runExport(cmd *Command, args []string) bool {
 
 	var version storage.Version
 
+	if tarOutputFile == nil {
+		fmt.Printf("key;name;size;gzip;mime;modified;ttl;deleted\n")
```
chrislusf: use comma here also

weed/command/export.go
```diff
@@ -62,6 +63,24 @@ var (
 	localLocation, _             = time.LoadLocation("Local")
 )
 
+func printNeedle(vid storage.VolumeId, n *storage.Needle, version storage.Version, deleted bool) {
+	key := storage.NewFileIdFromNeedle(vid, n).String()
+	size := n.DataSize
+	if version == storage.Version1 {
+		size = n.Size
+	}
+	fmt.Printf("%s;%s;%d;%t;%s;%d;%s;%t\n",
```
chrislusf: use comma, and maybe escape on name since it may also contain comma.

Or we can just use tsv instead of csv.

brstgt:
- I properly quoted the string fields in the CSV export
- Outputting timestamps as ISO8601 helps humans to read the output
- Added a "volumeInfo" option to output volume meta data
- Save timestamps with tombstones in "lastModified" field. This helps to sync volumes if replication is/was broken. This is more a preparational step towards a proper repair script

Any thoughts on all that? To be honest, not having any tests for volume manipulation doesn't feel very good. Maybe I have a bit more time next year to work on Seaweed, so having a couple of tests for the most critical parts would be great!

chrislusf:
do not bundle multiple changes together. this diff is too big and need to split.

brstgt:
Ok, no problem but don't you have any other comments like "I like the idea" or "It's ok, but ..."

chrislusf:
I think any new requirement/feature coming out of actual usage is nice.

PR #609, opened 2017-12-22
Store and replicate timestamp of file deletion in tombstone marker

brstgt:
I pulled out that of the recent PR. It's necessary for a proper repair / sync tool

### DIFF
--- a/weed/server/volume_server_handlers_write.go
+++ b/weed/server/volume_server_handlers_write.go
@@ -9,6 +9,8 @@ import (
 	"github.com/chrislusf/seaweedfs/weed/operation"
 	"github.com/chrislusf/seaweedfs/weed/storage"
 	"github.com/chrislusf/seaweedfs/weed/topology"
+	"time"
+	"strconv"
 )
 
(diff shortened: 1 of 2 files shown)


weed/server/volume_server_handlers_write.go
```diff
@@ -87,6 +89,13 @@ func (vs *VolumeServer) DeleteHandler(w http.ResponseWriter, r *http.Request) {
 		count = chunkManifest.Size
 	}
 
+	n.LastModified = uint64(time.Now().Unix())
+	if len(r.FormValue("time")) > 0 {
```
chrislusf, line 93: "time" is too generic. we can use "LastModified" here to be clear about the meaning and the unit.

brstgt:
Last modified is semantically not really correct when we talk about
deletion. Then let's call it "deletionTime"?

Am 22.12.2017 20:42 schrieb "Chris Lu" <notifications@github.com>:

chrislusf:
LastModified is defined in http header and should be clear.

deletion is just one kind of modification.

brstgt:
Agreed

Am 22.12.2017 20:49 schrieb "Chris Lu" <notifications@github.com>:

Issue #562, opened 2017-09-17
Cannot grow volume group！


Dr13am:
when i  use weedfs， i often meet  “Cannot grow volume group”， i find  volume -max  （dufalut “7”） ，I need big storage ，so I modify the upper limit，But there are still problems ，what should  i do ？

chrislusf:
increase the upper limit or post more details here.

Dr13am:
I recently found such a mistake:

 runtime: out of memory: cannot allocate 1310720-byte block (3484090368 in use)
fatal error: out of memory,

 how can I solve it?

chrislusf:
Add memory?

chrislusf:
the pull request may have resolved this issue.

PR #615, opened 2018-01-17
fix reserve volume on node

tengattack:
This will fix replication on difference data center complains `{"error":"Cannot grow volume group! No free volume slot found!"}`.

This problem may related to Issue #562.

Thanks!


chrislusf:
how to verify this fix?

tengattack:
Simply we can do:

```sh
curl -X POST http://localhost:9333/dir/assign?replication=100
```

If no errors, it fixed.

chrislusf:
Thanks! I am curious why this obvious bug did not surface before.

tengattack:
Yah, I made the same kind of mistake before. It's not easy to find out during the coding.

tengattack:
BTW, I'm not very clear about the different between dc replication and rack replication. Could you make some brief introduction?

PR #623, opened 2018-01-28
new: add support for memcached metadata store

NoneGG:
## change
Add support for metadata store in memcached

## about test
I test the code of memcached_store package in my test env and it runs well.
I am not sure if it is convenient for you to add memcached server in your CI(test) env so i didn't add them to this pr, they will be added if you allow.

### DIFF
--- a/weed/filer/memcached_store/memcached_store.go
+++ b/weed/filer/memcached_store/memcached_store.go
@@ -1,0 +1,56 @@
+package memcached_store
+
+import (
+	memcache "github.com/bradfitz/gomemcache/memcache"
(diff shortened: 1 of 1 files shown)


chrislusf:
1. Is this the only file to change? I suppose we need to invoke this store.
2. memcache is transient. What's your use case?

NoneGG:
### about invoking
I read the wiki and think that only store part is needed and invoking part will get added if merged. 

### about persistence
I didn't consider the problem so carefully, persistence is really needed and memcached does not fit this 
scene indeed. This pr will be closed.

PR #630, opened 2018-02-09
Close file before return

AceDarkkinght:


chrislusf:
no logic difference here.

PR #729, opened 2018-09-11
Add filer to docker-compose and entrypoint and build from local Dockerfile

hans-strudle:
I was playing around with seaweedfs in docker, and added a filer option to the `entrypoint.sh` as well as in the `docker-compose.yml`, to easily bring up a filer alongside the master & volume. This includes adding in the `filer.toml` which I have set to use the memory store.

I also would make the argument that this docker-compose should be building from the local Dockerfile, hence the change to using `build: .` instead of `image: chrislusf/seaweedfs` in the `docker-compose.yml`.

Also sorry for the dirty commit log with the odd usernames :)


docker/filer.toml
```diff
@@ -0,0 +1,3 @@
+[memory]
+enabled = true
```
chrislusf: how about default to leveldb? 

If just in memory, the directories will be gone while the files are still stored on volume servers.

docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '2'
 
 services:
   master:
-    image: chrislusf/seaweedfs
+    build: .
```
chrislusf: maybe put a comment here to show the different options, either build local or use the image. same for other 2 places.

chrislusf:
Thanks for the PR!

Added a few comments. Do not worry about Travis CI build error.

docker/filer.toml
hans-strudle: Great point, I assumed leveldb required some sort of extra setup, but just tested and it does preserve the information on a container restart. Although state will be lost if the container is rebuilt without the use of a volume of course.

docker/docker-compose.yml
hans-strudle: Added comments :)

PR #754, opened 2018-10-23
use released version of azblob

h8liu:


h8liu:
build break found by [gopkgs.io](https://gopkgs.io). example build log: 
https://gopkgs.io/builds/20181023-084847-c31e59e1

root cause is that the lib removed the `2016-05-31` directory:
https://github.com/Azure/azure-storage-blob-go/pull/82/files

not sure how to update the glide thing though..

chrislusf:
Thanks for this PR! There are other API changes. I fixed it in commit https://github.com/chrislusf/seaweedfs/commit/b8618999195674fe03710a1e2bd5687826dbcb14

h8liu:
thanks for the fix.

PR #821, opened 2019-01-09
Set redis key ttl with entry.TTLSec

yunbj:
When a file is deleted by ttl, filer store entry needs to be deleted also. (if the filerstore supports ttl)


chrislusf:
thanks!

btw: added ttl for Cassandra also.

PR #827, opened 2019-01-10
change log_dir to logdir for the "flag redefined:log_dir" error when use another package with "github.com/golang/glog"

suilongfei:
change log_dir to logdir for the "flag redefined:log_dir" error when use another package with "github.com/golang/glog"


chrislusf:
Thanks! I did not know this small change can resolve this issue.

PR #829, opened 2019-01-10
Fix https://github.com/chrislusf/seaweedfs/issues/823

PapaYofen:
- Due to be not familar with filer,   **only master & volume non-streaming rpc** timeout is set
-  http timeout is set

timeout is set to **_5s both rpc and http_**.


weed/topology/topology_vacuum.go
```diff
@@ -81,7 +85,10 @@ func batchVacuumVolumeCommit(vl *VolumeLayout, vid storage.VolumeId, locationlis
 	for _, dn := range locationlist.list {
 		glog.V(0).Infoln("Start Commiting vacuum", vid, "on", dn.Url())
 		err := operation.WithVolumeServerClient(dn.Url(), func(volumeServerClient volume_server_pb.VolumeServerClient) error {
-			_, err := volumeServerClient.VacuumVolumeCommit(context.Background(), &volume_server_pb.VacuumVolumeCommitRequest{
```
chrislusf, line 84: Commit may be longer than 5s. Leave it as is should be good

PapaYofen:
Thanks!

PR #848, opened 2019-01-24
refactor code 

bingoohuang:
1. make volumeSizeLimit argument more clear
2. use humanize.MiByte instead of 1024 * 1024
3. constantized NodeTypes together


chrislusf:
could you please avoid changing to "volumeSizeLimit"?

Just using numbers are easier. I do not want to look up for the syntax of volumeSizeLimit.

PR #865, opened 2019-02-26
Fix https://github.com/chrislusf/seaweedfs/issues/864

PapaYofen:


PapaYofen:
if there is no different opinion, merge this PR please

PR #922, opened 2019-04-09
Add replication-job Mode For The Docker Container

zicklag:
The container job is working, but for some reason I get an error when trying to use `weed shell` with the Docker Compose setup ( and therefore the replication-job mode with Compose ):

```
error: rpc error: code = Unimplemented desc = unknown service master_pb.Seaweed
```

I did successfully test the replication job with a local cluster on Docker swarm so I know that the job is running correctly.

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -44,6 +44,14 @@ case "$1" in
   	exec /usr/bin/weed $@ $ARGS
 	;;
 
+  'cronjob')
+	MASTER=${WEED_MASTER-localhost:9333}
+	CRON_SCHEDULE=${CRON_SCHEDULE-*/5 * * * * *}
+	echo "$CRON_SCHEDULE" 'echo "volume.fix.replication" | weed shell -master='$MASTER > /crontab
+	echo "Running Crontab:"
+	cat /crontab
+	exec supercronic /crontab
+	;;
   *)
   	exec /usr/bin/weed $@
 	;;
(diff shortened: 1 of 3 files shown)


chrislusf:
Thanks! How about renaming "replication-job" to "cronjob" or "admin-jobs", etc?

In the future, there can be more administrative jobs running as cron jobs.

docker/entrypoint.sh
```diff
@@ -44,6 +44,14 @@ case "$1" in
   	exec /usr/bin/weed $@ $ARGS
 	;;
 
+  'replication-job')
+	MASTER=${WEED_MASTER-localhost:9333}
```
chrislusf, line 48: Is this correct? Not sure.

master:9333-localhost:9333

zicklag: That just sets the default master to `localhost:9333` so that it is the same as the default for the `weed shell` command. You can override the setting by setting the WEED_MASTER environment variable, like in the compose file.

zicklag:
> How about renaming "replication-job" to "cronjob" or "admin-jobs", etc?

There you go. :slightly_smiling_face: Renamed it to "cronjob".

docker/seaweedfs-compose.yml
```diff
@@ -26,6 +26,16 @@ services:
     depends_on:
     - master
     - volume
+  replication-job:
```
chrislusf: rename it to "cronjob" also?

zicklag:
Fixed it.

PR #997, opened 2019-07-01
Adjust the variable name to increase readability

iliul:
- adjust the variable name to increase readability (size unit not GB)
- remove unreachable code reported by `go vet`

Signed-off-by: Lei Liu <liul.stone@gmail.com>

### DIFF
--- a/weed/util/constants_4bytes.go
+++ b/weed/util/constants_4bytes.go
@@ -4,5 +4,5 @@ package util
 
 const (
 	sizeLimit         = "30GB"
-	VolumeSizeLimitGB = 30
+	VolumeSizeLimitMB = 30
 )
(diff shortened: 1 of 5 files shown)


weed/util/constants_4bytes.go
```diff
@@ -4,5 +4,5 @@ package util
 
 const (
 	sizeLimit         = "30GB"
-	VolumeSizeLimitGB = 30
+	VolumeSizeLimitMB = 30
```
chrislusf, line 7: This is not expected.

iliul:
```
if *masterOptions.volumeSizeLimitMB > util.VolumeSizeLimitGB*1000 {
		glog.Fatalf("masterVolumeSizeLimitMB should be less than 30000")
	}
```

这里很容易让刚看代码的以为是要限制大小为 30*1000 GB，其实是要限制在 30 * 1000 MB

iliul: 也许应该改为 `VolumeSizeLimit = 30 * 1000` 更好一些

PR #1021, opened 2019-07-25
use go mod to manager dependencies

divinerapier:
It's surprising the dependencies are out of control.


chrislusf:
Tried go mod several times and could not get it working properly so I still prefer just to using the latest version.

I will merge it and let's see how it goes.

divinerapier:
Next time I will add a wiki page to introduce how to work with go mod.

fengye87:
Well, this patch is problematic.

In the travis config, this project is building agains go 1.10.x, 1.11.x and 1.12.x. For go 1.10.x, there's no go mod, while for the latter two go versions, go mod is disabled when the project is in GOPATH. To really enable go mod in this project, we should also change the travis setup too, to make it work with go mod.

Since the CI is still working in the old non-go-mod way, there's no surprise that currently, this project wouldn't build successfully using go mod. If you go build this project outside of GOPATH using go 1.12.x, in which case go mod is enabled, it would fail with "weed/s3api/filer_multipart.go:26:14: assignment mismatch: 2 variables but uuid.NewV4 returns 1 values".

The cause of this error is that, in go.mod file, package "github.com/satori/go.uuid" is locked to v1.2.0, and in this version of go.uuid, `uuid.NewV4()` returns no error. While if go mod is disabled, which is the case of CI setup, `go get -d ./weed/` would fetch master branch of package "github.com/satori/go.uuid", whose `uuid.NewV4()` returns extra error.

To sum up, we should probably need to:
1. drop 1.10.x, so that we can totally move towards go mod
2. update travis pipeline accordingly to NOT build in GOPATH (and without `go get ...`)
3. fix weed/s3api/filer_multipart.go:26 line if we are going to use v1.2.0 of "github.com/satori/go.uuid", or override the dependency in go.mod file

PR #1024, opened 2019-07-30
Move travis-ci pipeline to go mod

fengye87:
This should enable go mod in CI, and the build error introduced in #1021 . If the master branch of go.uuid is still preferred, I could also make it work, just let me know.


weed/s3api/filer_multipart.go
```diff
@@ -23,7 +23,7 @@ type InitiateMultipartUploadResult struct {
 }
 
 func (s3a *S3ApiServer) createMultipartUpload(ctx context.Context, input *s3.CreateMultipartUploadInput) (output *InitiateMultipartUploadResult, code ErrorCode) {
-	uploadId, _ := uuid.NewV4()
+	uploadId := uuid.NewV4()
```
chrislusf: Can we upgrade the uuid package version instead?

fengye87: Sure, updated the patch

PR #1049, opened 2019-08-19
use std encoding lib

divinerapier:
use stdlib makes code clear


Kimbsen:
this commit causes a panic here: https://github.com/chrislusf/seaweedfs/blob/master/weed/storage/needle/needle_read_write.go#L253

as it gets passed 5 bytes while binary.BigEndian.Uint64 requires 8 bytes.

chrislusf:
@Kimbsen thanks for reporting. Reverted this commit earlier today.

PR #1053, opened 2019-08-23
volume: add fsync threshold

xushuxun:
Volume server count how many object has been written, 
if the number of objects is bigger than threshold, sync the kernel cache.

===
给volume server加了启动参数fsyncThreshold，volume每写fsyncThreshold个对象，执行一次fsync。

### DIFF
--- a/weed/command/volume.go
+++ b/weed/command/volume.go
@@ -14,7 +14,7 @@ import (
 
 	"github.com/chrislusf/seaweedfs/weed/glog"
 	"github.com/chrislusf/seaweedfs/weed/pb/volume_server_pb"
-	"github.com/chrislusf/seaweedfs/weed/server"
+	weed_server "github.com/chrislusf/seaweedfs/weed/server"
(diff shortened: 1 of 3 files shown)


chrislusf:
this fsync is for the store, but since there are many volume files, you may need a counter for each volume file.

My original thinking is to improve API and allow the write request to do a fsync. I am still debating which approach is better.

chrislusf:
how about adding a flush command to volume servers for a particular volume?

xushuxun:
fsync would flush all kernel cache, so it might make no sense to flush a particular volume.

chrislusf:
> fsync would flush all kernel cache, so it might make no sense to flush a particular volume.

This commit used `v.dataFile.Sync()` and is volume specific.
https://golang.org/pkg/os/#File.Sync

chrislusf:
Added https://github.com/chrislusf/seaweedfs/commit/e4af63a721f62c280596f626302b64729baeaee9 to accept fsync=true requests

PR #1060, opened 2019-09-06
fix: non-thread-safe rand will panic

divinerapier:
The rand package in standard library isn't thread safe. It will panic cross multiple threads.

[Please see the doc of `rand.NewSource`](https://github.com/golang/go/blob/master//src/math/rand/rand.go#L41)

Here is a repo I copied from the stdlib, and added a new function `NewLocked` which is thread safe. For more information, please see the [repo](https://github.com/divinerapier/rand).

Thanks.

### DIFF
--- a/weed/wdclient/vid_map.go
+++ b/weed/wdclient/vid_map.go
@@ -3,11 +3,11 @@ package wdclient
 import (
 	"errors"
 	"fmt"
-	"math/rand"
+	"math"
 	"strconv"
 	"strings"
 	"sync"
-	"time"
+	"sync/atomic"
 
 	"github.com/chrislusf/seaweedfs/weed/glog"
 )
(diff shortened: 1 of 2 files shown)


chrislusf:
There are only one usage for this rand, and it is already locked. Do you see panic related to this?

divinerapier:
Yes, it panics in out production environment. And it is unsafe even if not panics.
Please look at the code in `weed/wdclient/wid_map.go`
```
func (vc *vidMap) GetRandomLocation(vid uint32) (serverUrl string, err error) {
	vc.RLock()
	defer vc.RUnlock()

	locations := vc.vid2Locations[vid]
	if len(locations) == 0 {
		return "", fmt.Errorf("volume %d not found", vid)
	}

	return locations[vc.r.Intn(len(locations))].Url, nil
}
```
This function is protected by the read and write. But function `(*Rand) Intn()` will modify the inner fields, it is unsafe.

chrislusf:
> This fucntion is protected by a RwLock. But the function (* Rand) Intn() modifies the internal fields, which is not safe.

I do not quite understand why there could panic if this is already protected.

divinerapier:
This is the function that generates a random number which is in package `math/rand/rng.go`.
The version of `go` on my pc is
``` bash
$ go version
go version go1.13 darwin/amd64
```
``` go
// Uint64 returns a non-negative pseudo-random 64-bit integer as an uint64.
func (rng *rngSource) Uint64() uint64 {
	rng.tap--                            // pay attention to
	if rng.tap < 0 {
		rng.tap += rngLen    // pay attention to
	}

	rng.feed--                         // pay attention to
	if rng.feed < 0 {
		rng.feed += rngLen  // pay attention to
	}

	x := rng.vec[rng.feed] + rng.vec[rng.tap]
	rng.vec[rng.feed] = x       // pay attention to
	return uint64(x)
}
```

chrislusf:
RLock still can allow multiple goroutines to access the rand object. Changing it to a normal Lock would be enough.

divinerapier:
It dose work, but I don't think it is a good idea. 

1. Even if vid is invalid, it needs to wait for a lock.
2. There may be some reasons why `math/rand` did not do this.
3. I thought about it again. I think `atomic.AddUint64(&cursor, 1) % len(locations)` is better than `math/rand.Intn(len(locations))`. The cursor is a field of vidMap and returns 0 when it overflows.

chrislusf:
The atomic operation is faster and more reasonable.

divinerapier:
It's my pleasure.

divinerapier:
@chrislusf PTAL

PR #1096, opened 2019-10-27
Update README.md

BruceWangNo1:
"export GO111MODULE=on" specified


chrislusf:
Why this is needed? This is default to on for Go later versions.

BruceWangNo1:
Yes, you are right. But some developers may have changed it to off at some point and later have trouble figuring out why go get does not work properly.

chrislusf:
Prefer not to add more lines here. The README is not to prevent all possible human errors. It's just to show the simplest way to get it up and running.

PR #1099, opened 2019-10-31
enhanced master api logic

iliul:
1，Layouts consistent with other fields, capitalized
2，/vol/vacuum、/col/delete return the appropriate http status code


weed/server/master_server_handlers_admin.go
```diff
@@ -35,7 +36,10 @@ func (ms *MasterServer) collectionDeleteHandler(w http.ResponseWriter, r *http.R
 			return
 		}
 	}
-	ms.Topo.DeleteCollection(r.FormValue("collection"))
+	ms.Topo.DeleteCollection(collectionName)
+
+	w.WriteHeader(http.StatusNoContent)
```
chrislusf, line 41: should a good deletion status be NoContent?

iliul: > should a good deletion status be NoContent?

HTTP 204 is more reasonable for some delete operations because there is no need to return the body content when the request is successful, refer to some ceph rgw delete relate interfaces

PR #1132, opened 2019-11-22
AWS S3 Select compatible feature underway

BruceWangNo1:


chrislusf:
this seems an error.

BruceWangNo1:
@chrislusf Yes, there is. It's only partial. Work is underway.

PR #1135, opened 2019-11-23
AWS S3 Select feature now available

BruceWangNo1:


weed/s3api/s3api_object_handlers.go
```diff
@@ -224,3 +227,98 @@ func getObject(vars map[string]string) string {
...
+		}
+		if err = s3Select.Open(getObject); err != nil {
+			if serr, ok := err.(s3select.SelectError); ok {
+				encodedErrorResponse := encodeResponse(cmd.APIErrorResponse{
+					Code:       serr.ErrorCode(),
+					Message:    serr.ErrorMessage(),
+					BucketName: bucket,
+					Key:        object,
+					Resource:   r.URL.Path,
+					RequestID:  w.Header().Get(xhttp.AmzRequestID),
```
chrislusf: try to avoid this package. seems you are only using one constant.

weed/s3api/s3api_object_handlers.go
```diff
@@ -224,3 +227,98 @@ func getObject(vars map[string]string) string {
...
+			return
+		}
+
+		s3Select.Evaluate(w)
+		s3Select.Close()
+	}
+	s3a.proxyToFilerS3Select(w, r, destUrl, passThroughResponseSelectObjectContent)
+}
+
+func (s3a *S3ApiServer) proxyToFilerS3Select(w http.ResponseWriter, r *http.Request, destUrl string,
```
iliul, line 291: Why did you write a new function instead of enhance the existing proxyToFiler func. It seems that only the http METHOD is different.

BruceWangNo1: I guess I did it because it's more straight forward. I'll do what you suggested in a future commit. Is that OK? @iliul 
And could you open a new branch "s3-select" to which I can make pull request?

PR #1142, opened 2019-11-27
fuse: check mount point available before do runmount

iliul:
- Check mount point available before do runmount
- Use more readable mount point information

eg:

**$ df -Th**

```
Filesystem              Type            Size  Used Avail Use% Mounted on
localhost:8888:/        fuse.seaweedfs  206G   512  206G   1% /mnt/weedfs
```

**$ mount | grep weedfs**

```
localhost:8888:/ on /mnt/weedfs type fuse.seaweedfs (rw,nodev,relatime,user_id=0,group_id=0,default_permissions,allow_other)
```

refer to cephfs and glusterfs

```
// glusterfs
datanode011.gluster-node.cc:/training-data fuse.glusterfs  161T   94T   68T  59% /mnt/glusterfs

// cephfs
10.10.100.10:6789:/     ceph      326T  295T   32T  91% /mnt/cephfs
```


iliul:
It's unclear how to judge the target mountpoint on darwin and freebsd

chrislusf:
Could this switch to use mount_linux.go, instead of hard coded if statement?

iliul:
> Could this switch to use mount_linux.go, instead of hard coded if statement?

updated. How about this version ?

chrislusf:
the util/mount.go logic is Linux specific. How about merging it into the
command/mount_linux.go file?

iliul:
> the util/mount.go logic is Linux specific. How about merging it into the command/mount_linux.go file?

Done.  

Please take a look again.

PR #1149, opened 2019-12-08
improve data consistency when replication write

song-zhang:
If a volume with 2 replications assigned to node1 and node2,  after node2 down, a file delete operation send to node1 will lead to file deleted successfully on replication in node1,  but file in node2 deleted failed. If node2 restart, the file may sometimes readable, but sometimes not readable. It also applies to write operation.

This patch can improve the data consistency by determining if all the replications are writable before write/delete operation.

### DIFF
--- a/weed/topology/store_replicate.go
+++ b/weed/topology/store_replicate.go
@@ -25,58 +25,61 @@ func ReplicatedWrite(masterNode string, s *storage.Store,
 	//check JWT
 	jwt := security.GetJwt(r)
 
+	var remoteLocations []operation.Location
+	if r.FormValue("type") != "replicate" {
+		remoteLocations, err = getWritableRemoteReplications(s, volumeId, masterNode)
+		if err != nil {
+			glog.V(0).Infoln(err)
+			return
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


chrislusf:
This pessimistic additional checking should be optional.

song-zhang:
a delete operation can always cause this problem if a replication is unavailable, because it use a exist fid.  this patch can reduce the possibility of data inconsistencies. the vid cache can also cause the problem, but this is another issue.

chrislusf:
please read my previous comment

song-zhang:
do you mean add a optional parameter, 
such as http://host:port/1,12121212?consistency_check=true  ?

chrislusf:
Right.

>

song-zhang:
update. thanks for your suggestion.

chrislusf:
Sorry. I think you previous change without the "consistency_check=true" is correct. Is a simple re-ordering the operations, although it needs to wait for all lookup to complete.

Please help to revert the commit for consistency_check=true.

song-zhang:
reset and force update

PR #1150, opened 2019-12-08
add risk warning to compact command

song-zhang:


chrislusf:
It should support compacting a running volume. If it causes problem, it is a bug.

song-zhang:
I got the problem when use weed compact command:

I1209 01:24:16 23452 file_util.go:19] Folder /var/lib/weed20 Permission: -rwxr-xr-x
I1209 01:24:16 23452 volume_loading.go:89] volumeDataIntegrityChecking failed verifyNeedleIntegrity /var/lib/weed20/2.idx failed: entry not found: offset 11552026320 found id 0 size 0, expected size 10312
I1209 01:24:16 23452 volume_loading.go:93] loading index /var/lib/weed20/2.idx to memory readonly true


The compact command is conflict with compact operation initiated by master. This is the same issue: https://github.com/chrislusf/seaweedfs/issues/1011

Since compact command and volume server are different processes, it's difficult to ensure only one exclusive vacuum process. Change the compacted file name(such as .dat2/.idx2) when compacting in command mode is a workaround, but it can't ensure the compacted volume file has the latest file which is write during compacting.

So, I think give a risk warning is reasonable. Otherwise, we should add a single volume compact API to master server.

PR #1153, opened 2019-12-11
Filer: Enable delete entries without flag volume

LazyDBA247-Anyvision:
same as happens on the s3 API on delete bucket,
to reduce the IO requests on the system, 
you can just clear recursively folders without flagging them in volume server.
this will remove the user's ability to retrieve data under these folders without io spike or delays related to updating the volume servers (will not release disk space! , files/folders will be removed on delete bucket/drop collection)


KobiAnyVision:
Up Vote this is a critical issue!

chrislusf:
I will tweak the parameter names a bit.

chrislusf:
modified via https://github.com/chrislusf/seaweedfs/commit/f06b65a6fd5636d6d385cdf027eeebb42e4ec62a

PR #1165, opened 2019-12-20
Add readOnly and routeByLatency options to Redis Cluster Store

dqsully:
I am running SeaweedFS in a multi-zone AWS environment where latency between the Filer and the Redis cluster is very important. However, by default the Go Redis client communicates with only the master node, which may be in a different AWS zone. This change will add configuration options to allow the Filer's Redis client to read from the closest Redis node, while still writing to the master Redis.

### DIFF
--- a/weed/filer2/redis/redis_cluster_store.go
+++ b/weed/filer2/redis/redis_cluster_store.go
@@ -22,13 +22,17 @@ func (store *RedisClusterStore) Initialize(configuration util.Configuration) (er
 	return store.initialize(
 		configuration.GetStringSlice("addresses"),
 		configuration.GetString("password"),
+		configuration.GetBool("connection_use_read_only"),
+		configuration.GetBool("connection_route_by_latency"),
 	)
 }
 
-func (store *RedisClusterStore) initialize(addresses []string, password string) (err error) {
+func (store *RedisClusterStore) initialize(addresses []string, password string, readOnly, routeByLatency bool) (err error) {
 	store.Client = redis.NewClusterClient(&redis.ClusterOptions{
-		Addrs:    addresses,
-		Password: password,
+		Addrs:          addresses,
+		Password:       password,
+		ReadOnly:       readOnly,
+		RouteByLatency: routeByLatency,
 	})
 	return
 }
--- a/weed/filer2/redis/redis_cluster_store.go
+++ b/weed/filer2/redis/redis_cluster_store.go
@@ -22,8 +22,8 @@ func (store *RedisClusterStore) Initialize(configuration util.Configuration) (er
 	return store.initialize(
 		configuration.GetStringSlice("addresses"),
 		configuration.GetString("password"),
-		configuration.GetBool("connection_use_read_only"),
-		configuration.GetBool("connection_route_by_latency"),
+		configuration.GetBool("useReadOnly"),
+		configuration.GetBool("routeByLatency"),
 	)
 }
 
--- a/weed/command/scaffold.go
+++ b/weed/command/scaffold.go
@@ -136,6 +136,8 @@ addresses = [
     "localhost:30006",
 ]
 password = ""
+readOnly = true
+routeByLatency = true
 
 [etcd]
 enabled = false


weed/filer2/redis/redis_cluster_store.go
```diff
@@ -22,13 +22,17 @@ func (store *RedisClusterStore) Initialize(configuration util.Configuration) (er
 	return store.initialize(
 		configuration.GetStringSlice("addresses"),
 		configuration.GetString("password"),
+		configuration.GetBool("connection_use_read_only"),
```
chrislusf: use readOnly

weed/filer2/redis/redis_cluster_store.go
```diff
@@ -22,13 +22,17 @@ func (store *RedisClusterStore) Initialize(configuration util.Configuration) (er
...
+		configuration.GetBool("connection_route_by_latency"),
```
chrislusf: use routeByLatency

chrislusf:
add it to scaffold.go also

chrislusf:
A side question: why do you prefer SeaweedFS over s3?

dqsully:
In regards to your question about why SeaweedFS over S3:

We needed a shared filesystem with all of these features:
* Highly Available
* Scalable
* High Performance
* Mounts to a Linux filesystem
* If self-hosted, can run in Kubernetes without being tied to a specific machine
* Can efficiently calculate disk usage

We considered pretty much every option out there:
* Amazon S3 (doesn't support Unix filesystem properties very well)
* Amazon EFS (only easy to calculate usage of total drive, and is limited to 100 drives for a 4-zone VPC, we would have needed >900)
* Rook CephFS (tied to specific machines, hard to manage)
* Rook EdgeFS (couldn't get it working, hard to manage)
* GlusterFS (also tied to specific machines)
* NFS (could not figure out high availability)
* Portworx (also tied to specific machines)

And I think there were a couple others that I can't remember.

We didn't originally consider SeaweedFS even though I had come across it in my searches because there wasn't any Helm charts for it, but in the end it proved to be the best solution to our needs:
* Highly Available (using StatefulSets, headless Services, and some Helm templating)
* Scalable (^ same things for different reasons)
* High Performance (especially after adding AWS zone affinity for both Mounter <-> Filer and Filer <-> Redis)
* Mounts to a Linux filesystem (using a DaemonSet to manage privileged containers for mounting)
* Is not tied to specific machines (Volume, Master, and Filer servers can run as normal pods backed by EBS volumes, which Kubernetes moves and manages)
* Can efficiently calculate disk usage (since we are using Redis for the Filer backend, we can write code to scan through the raw file metadata and quickly tally up usage numbers)

dqsully:
I also have a question about the values in `scaffold.go`. Do you want these options to be enabled or disabled by default? I committed them as enabled by default since it should only help performance, but the Go Redis library has them disabled by default, so I'll leave it up to you.

Also, I don't think I explained this very well, but by default the Redis client will only read and write to the master server for a given shard. The `readOnly` option allows the Redis client to read from a shard's slave servers as well as the master, but all writes will still go to the master. `routeByLatency` enables `readOnly` when true, and allows the Redis client to query and automatically use the closest Redis server for reads instead of picking a random one.

In our case, one of our processes needs to get file information as fast as possible in a synchronous manner, so even sub-millisecond latencies add up. Enabling `routeByLatency` decreased our overall process times by at least 10% in zones that didn't have the Redis master, and probably decreased our time interacting with SeaweedFS by 30% or more.

chrislusf:
Thanks for the detailed answer! It is eye-opening to get this detailed feedback! Thanks!

I address the default value concern with 62b1a346ef33e0d4aa84a568e32a53550b75b590

dqsully:
I don't think that commit is quite what I meant. The Go Redis client already has `readOnly` and `routeByLatency` set to false by default. My question was if you thought the scaffold code I committed should show them set to true or not.

chrislusf:
How about fcbd04c4253927da2c16821ec1a6565b29de6a6f ?

dqsully:
Yeah, that is the code I was talking about. I'm not sure if you want those set to true or false by default. The author of the Go Redis client has them false by default within his library, but I think setting them to true should only increase performance.

chrislusf:
Right. The default values should be true without any downside. Adjusted them in a9b1ee18afdaee0806d91f253fbb35e5ffd49c76

dqsully:
Cool, looks good! :+1:

PR #1198, opened 2020-02-13
when Store.MountVolume(), the volume server will be core

stlpmo-jn:
in the function of Store.MountVolume(), when loadExistingVolume() failed, the return value of funtion location.LoadVolume() will also be true,
the next funtion of s.findVolume() will return nil volume, the next opertion based on nill volume will make segment fault.

```
func (s *Store) MountVolume(i needle.VolumeId) error {
	for _, location := range s.Locations {
		if found := location.LoadVolume(i, s.NeedleMapType); found == true {
			glog.V(0).Infof("mount volume %d", i)
			v := s.findVolume(i)
			s.NewVolumesChan <- master_pb.VolumeShortInformationMessage{
				Id:               uint32(v.Id),
				Collection:       v.Collection,
				ReplicaPlacement: uint32(v.ReplicaPlacement.Byte()),
				Version:          uint32(v.Version()),
				Ttl:              v.Ttl.ToUint32(),
			}
			return nil
		}
	}

	return fmt.Errorf("volume %d not found on disk", i)
}
```

so we make loadExistingVolume() return bool, the problem will be resolved.

### DIFF
--- a/weed/storage/disk_location.go
+++ b/weed/storage/disk_location.go
@@ -50,29 +50,39 @@ func parseCollectionVolumeId(base string) (collection string, vid needle.VolumeI
 	return collection, vol, err
 }
 
-func (l *DiskLocation) loadExistingVolume(fileInfo os.FileInfo, needleMapKind NeedleMapType) {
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks for the explanation and fix!

PR #1206, opened 2020-02-23
DiskStats: adding Total & Percent Usage

LazyDBA247-Anyvision:
making relevant data visible and readable.
image


LazyDBA247-Anyvision:
@chrislusf maybe you can use the function bytesToHumanReadble on the other tables also :)

PR #1214, opened 2020-03-03
FILER/S3: filer Dir Buckets

LazyDBA247-Anyvision:
updated config, to the new 1.58


LazyDBA247-Anyvision:
@chrislusf  redeployed the updated chart :)
found these two issues.   copy paste & white labeling cleanups

PR #1219, opened 2020-03-06
schedule new volume by free volume number of nodes

song-zhang:
Now weed master pick node randomly when assign a new volume, but it has some problems in scenarios where node capacities are not equal,  here is an example to illustrate our problem: 

We have 5 nodes with different free volumes, 50, 50, 30, 30, 30, and we use 3 replications. after uploading some files, the free volume list changes to 20, 20, 0, 0, 0.  Even there are 40 available volumes, we can't assign a new volume satisfy 3 copy.

This patch provide a new strategy to schedule volume by node weights instead of picking node randomly. Here is a test:

First, start seaweedfs cluster with 6 volume servers, each has different max volume number.
# ./weed master -ip 10.128.34.48 -port=6333 -mdir=/var/lib/weed1/master -defaultReplication=100
# ./weed volume -ip 10.128.34.48 -port=8081 -dir=/var/lib/weed1/volume -max 2000 -mserver=10.128.34.48:6333 -dataCenter=dc1
# ./weed volume -ip 10.128.34.48 -port=8082 -dir=/var/lib/weed2/volume -max 2000 -mserver=10.128.34.48:6333 -dataCenter=dc2
# ./weed volume -ip 10.128.34.48 -port=8083 -dir=/var/lib/weed3/volume -max 1000 -mserver=10.128.34.48:6333 -dataCenter=dc3
# ./weed volume -ip 10.128.34.48 -port=8084 -dir=/var/lib/weed4/volume -max 1000 -mserver=10.128.34.48:6333 -dataCenter=dc4
# ./weed volume -ip 10.128.34.48 -port=8085 -dir=/var/lib/weed5/volume -max 500 -mserver=10.128.34.48:6333 -dataCenter=dc5
# ./weed volume -ip 10.128.34.48 -port=8086 -dir=/var/lib/weed6/volume -max 500 -mserver=10.128.34.48:6333 -dataCenter=dc6
 
Assign 1000 volumes use default "100" replication strategy:
# curl "http://10.128.34.48:6333/vol/grow?count=1000&collection=benchmark"

now we can see actual volume distribution,  agree with weight distribution on the whole:
# for((i=1;i<7;i++));do ls -l /var/lib/weed${i}/volume/benchmark_*.dat | wc -l;done
526
537
327
306
156
148

### DIFF
--- a/weed/topology/node.go
+++ b/weed/topology/node.go
@@ -62,56 +62,64 @@ type NodeImpl struct {
 }
 
 // the first node must satisfy filterFirstNodeFn(), the rest nodes must have one free slot
-func (n *NodeImpl) RandomlyPickNodes(numberOfNodes int, filterFirstNodeFn func(dn Node) error) (firstNode Node, restNodes []Node, err error) {
-	candidates := make([]Node, 0, len(n.children))
+func (n *NodeImpl) PickNodesByWeight(numberOfNodes int, filterFirstNodeFn func(dn Node) error) (firstNode Node, restNodes []Node, err error) {
+	var totalWeights int64
 	var errs []string
 	n.RLock()
+	candidates := make([]Node, 0, len(n.children))
+	candidatesWeights := make([]int64, 0, len(n.children))
+	//pick nodes which has enough free volumes as candidates, and use free volumes number as node weight.
 	for _, node := range n.children {
+		if node.FreeSpace() <= 0 {
+			continue
+		}
+		totalWeights += node.FreeSpace()
+		candidates = append(candidates, node)
+		candidatesWeights = append(candidatesWeights, node.FreeSpace())
+	}
+	n.RUnlock()
+	if len(candidates) < numberOfNodes {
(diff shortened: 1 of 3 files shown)


weed/topology/node.go
```diff
@@ -62,56 +63,65 @@ type NodeImpl struct {
...
-	for _, node := range n.children {
-		if node.Id() == firstNode.Id() {
-			continue
-		}
-		if node.FreeSpace() <= 0 {
-			continue
+	//pick nodes randomly by weights, the node picked earlier has higher final weights
+	sortedCandidates := make([]Node, 0, len(candidates))
+	for i:=0; i<len(candidates); i++  {
+		rand.Seed(time.Now().UnixNano())
```
chrislusf: Seed() here seems not so useful.

chrislusf:
Thanks! This is a very good improvement! Please help to add tests for this. If possible, keep the original random algo and test both.

weed/topology/node.go
song-zhang: yes, removed.

Issue #1289, opened 2020-04-27
os.Exit directly after all signal handler done running


wyxloading:
**Describe the bug**
Some "bug" in code. I was caught in the bug which make my program fail to graceful shutdown. BUT with a 0 exit code.
https://github.com/chrislusf/seaweedfs/blob/508ec01ab0668ab7ec7ba8a8484e043ac6c7b237/weed/util/signal_handling.go#L28-L35

<s>I think it's ok to remove the line `os.Exit(0)` directly. </s>

**System Setup**
- master

**Expected behavior**
We should not call `os.Exit()` in any signal handler.
Which might make the normal graceful shutdown fail at any moment.

EDIT: Remove the like `os.Exit(0)` directly is not ok.

wyxloading:
I have required seaweedfs to my project, as i need the shell package (which depend on the util package), make my graceful shutdown a mess.

chrislusf:
Not sure how to help you.

what you need from the shell package?

PR #1293, opened 2020-04-28
refactor: move signal handling and pprof to grace package

wyxloading:
relate to #1289 

I move `signal_handling.go` and `pprof.go` into package `grace`.
Then the usage of `weed/shell` package won't initialize the grace package, which can avoid the graceful shutdown mess.

I use the `weed/shell` package to implement something like:
```
// to get all collections with: collection.list
// for each collection
//   check collection name or something to determine whether it should do `tier.upload` or `delete` or nothing
//   volume.tier.upload or collection.delete
// end
```

Seems like i can't do this kind of logic in the `scirpt` block in `[master.maintenance]`. Also the `weed shell` do not support passing a single command, then get the structure data it returned.
So i try to use the `weed/shell` package directly.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/Azure/azure-pipeline-go v0.2.2 // indirect
 	github.com/Azure/azure-storage-blob-go v0.8.0
 	github.com/DataDog/zstd v1.4.1 // indirect
+	github.com/OneOfOne/xxhash v1.2.2
 	github.com/Shopify/sarama v1.23.1
 	github.com/aws/aws-sdk-go v1.23.13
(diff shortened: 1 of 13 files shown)


chrislusf:
cool. you are using volume.tier.upload already?

wyxloading:
Yes, we migrate the old data to the next tier storage.

chrislusf:
tiered to AWS S3?

wyxloading:
Exactly.

chrislusf:
Could you please share a bit more about your use case? Or you can contact me directly via slack/wechat.

wyxloading:
Of course. May be after the May Day Holiday. slack/wechat both sound good to me.

PR #1337, opened 2020-05-29
add Volume Ids column only for max 100 volumes for convenience in the master ui.

bingoohuang:
check volume ids directly in the master ui in demo likely (max 100 volumes) environment.


weed/topology/data_node.go
```diff
@@ -190,3 +192,19 @@ func (dn *DataNode) ToDataNodeInfo() *master_pb.DataNodeInfo {
 	}
 	return m
 }
+
+// GetVolumeIds returns the human readable volume ids limited to count of max 100.
+func (dn *DataNode) GetVolumeIds() string {
+	volumesLen := len(dn.volumes)
+	if volumesLen > 100 {
+		return "..."
```
chrislusf: how about showing the first 99 volume ids and one `...`?

bingoohuang: good idea

PR #1338, opened 2020-06-01
add Weed-FileIds to the response of `GET /123.png?verbose=true` for d…

bingoohuang:
…ebugging like internal FileIds checking


chrislusf:
I prefer not to duplicate the functionality as `fs.meta.cat` in `weed shell`

bingoohuang:
> I prefer not to duplicate the functionality as `fs.meta.cat` in `weed shell`

I think they serves different scenarios.

`fs.meta.cat` in `weed shell` is for command line on the host **server**.
`Weed-FileIds` in the response is for the HTTP rest, which is available in the **client** browser DEV tools or curl scripts output.

PR #1351, opened 2020-06-05
Set volumes ReadOnly if low free disk space

ekozlov-search:
Periodically check free disk space on volumes disks and mark volumes as readonly fi free space is low. 
ReadOnly volumes exclude from writable volumes and not assign for new files.


chrislusf:
Naming is the most difficult. How about changing `FreeDiskSpaceWatermark` to `minFreeSpacePercent` or `minFreeDiskSpacePercent`?

Ruakij:
I kinda found a problem with this.
The volumes are set to readOnly, but if currVolumes < maxVolumes, the master will send write requests to the volume which then fail. (and the mounter really doesnt stop trying when writes fail e.g. disk full, no volumes..)

I wrote a temporary patch for myself to report max volumes as 0 when disk-space is full, but i dont think that should be the solution.

PR #1358, opened 2020-06-11
Return NotSuchKey error code for bucket S3 DeleteObject method

DXist:
Fixes for S3 DeleteObject

* return 404 status code instead of 204 No content
* return xml encoded error with code `NoSuchKey`

Sample s3 response:
```
curl -v -X DELETE localhost:8333/not/exist
*   Trying ::1:8333...
* TCP_NODELAY set
* Connected to localhost (::1) port 8333 (#0)
> DELETE /not/exist HTTP/1.1
> Host: localhost:8333
> User-Agent: curl/7.68.0
> Accept: */*
>
* Mark bundle as not supporting multiuse
< HTTP/1.1 404 Not Found
< Accept-Ranges: bytes
< Content-Length: 201
< Content-Type: application/xml
< Date: Thu, 11 Jun 2020 14:56:48 GMT
< X-Amz-Request-Id: 1591887408636162435
<
* Connection #0 to host localhost left intact
<Error><Code>NoSuchKey</Code><Message>The specified key does not exist.</Message><Resource>/not/exist</Resource><RequestId>1591887408636143362</RequestId></Error>
```

### DIFF
--- a/weed/s3api/s3api_errors.go
+++ b/weed/s3api/s3api_errors.go
@@ -33,6 +33,7 @@ const (
 	ErrBucketAlreadyExists
 	ErrBucketAlreadyOwnedByYou
 	ErrNoSuchBucket
+	ErrNoSuchKey
 	ErrNoSuchUpload
 	ErrInvalidBucketName
 	ErrInvalidDigest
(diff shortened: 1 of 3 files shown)


DXist:
@chrislusf actually I'm not sure about 404 status code when deleting nonexisting object in AWS.

I'm doing tests with Minio S3 server and it returns 404. I don't have AWS account to check it.

DXist:
Actually no - Minio returns 204 No content.

Ceph S3 returns 204 as well - https://tracker.ceph.com/issues/7502.

So I'll remove handling of 404. The actual fix adds NoSuchKey code for GetObject/HeadObject

chrislusf:
Thanks! 

btw: How do you compare SeaweedFS S3 vs Minio vs Ceph S3?

DXist:
@chrislusf I don't have own experience here but my employer looks for Ceph S3 alternatives. Main pain points are operations burden and not good enough latency.

Issue #1370, opened 2020-06-22
File integrity problem - file no longer matches checksum after upload and download


ingardm:
**Describe the bug**
Discovered that a file uploaded to seaweed no longer matches the md5 checksum it had when uploading it. We used an internal migration tool that transferred a lot of files. We have observed a strange problem with one of the files where seemingly a byte has been flipped. If it is because of compression or memory error or whatever else we do not know.

**System Setup**
seaweed 1.81 96f7c65e
5x weed master + 5x weed volume
```
/usr/local/bin/weed -logtostderr master -mdir=/mnt/data-small/meta -ip=10.1.14.114 -defaultReplication=010 -volumePreallocate -volumeSizeLimitMB 30000 -garbageThreshold 0.3 -peers=10.1.14.110:9333,10.1.14.111:9333,10.1.14.112:9333,10.1.14.113:9333,10.1.14.114:9333
/usr/local/bin/weed -logtostderr -v=1 volume -compactionMBps=50 -port=8080 -mserver=10.1.14.110:9333,10.1.14.111:9333,10.1.14.112:9333,10.1.14.113:9333,10.1.14.114:9333 -dir=/mnt/data-small/data -index leveldb -max 0 -ip 10.1.14.114 -rack dn-381
```

**Expected behavior**
Upload a file, and be able to download it with full integrity

**Additional context**
I do now know what might have caused this problem, and its the first time I've seen it. The problem or maybe challenge is to verify the content of the file as close to writing it to disk as possible. Our migration tools will say something like this:
```
2020/06/18 02:58:36 upload complete [username:a08a64f587fcd043ce9450ce2c094a20] (2653398) 200 => RemoteWriteResponse{writtenMD5=a08a64f587fcd043ce9450ce2c094a20, writtenSize=2653398}
```
Which means that the checksum in transit has been verified to be a08a64f587fcd043ce9450ce2c094a20.

So a few questions:
1. Is there a way to get the md5 computed and returned to the client when the file is flushed to disk?
2. Is it possible to implement something maybe like we include a Content-MD5:a08a64f587fcd043ce9450ce2c094a20 header and then seaweed validates the data against that and returns an error code if it doesnt match?
3. Can you document a little bit more how the Etag solution works and maybe with some example code that verifies the file integrity using the Etag sent back by the volume server?

ingardm:
[...]

PR #1371, opened 2020-06-24
Optional md5 validation of uploads

Kimbsen:
We would like to add this due to #1370 

If the Content-MD5 header is set profiling shows the md5 compute uses a significant amount of the total cpu but as its optional it can be avoided.


chrislusf:
FYI: https://github.com/chrislusf/seaweedfs/issues/1374

Issue #1220, opened 2020-03-07
Docker Image for ARM (Raspberry PI)


mario-tux:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs

**Describe the bug**
I'm trying to deploy seaweedfs on a swarm with 3 nodes: two x86 servers and one raspberry pi 2 (to get a majority in case of a server fault). I'm was not able to launch the docker container on the rpi. In the Docker Hub builds page I can't see an ARM build. In report https://github.com/chrislusf/seaweedfs/issues/1162 you suggest to use `chrislusf/seaweedfs:dev` tag but it doesn't help me.

On the rpi:
```
# docker run chrislusf/seaweedfs -ti
standard_init_linux.go:211: exec user process caused "exec format error"
# docker run chrislusf/seaweedfs:dev -ti
standard_init_linux.go:211: exec user process caused "exec format error"
```
Both the commands launch `weed` command, without parameters, on the x86 servers.

thiscantbeserious:
I manually build an armhf binary just fine (armv7+) not sure about armv7 tought.

I'll shortly create an layerized multiarch docker-image that I'll upload to Docker Hub and send the Dockerfile as a PR to chris that will include arm(v7+) support as well.

thiscantbeserious:
Eh, sorry guys was a bit busy lately. Since Chris officially publishes ARM releases now and finished his CSI-Driver for Kubernetes I'm going to get back to this soon. In the meantime just use the ARM builds directly now. :)

Gronis:
There are binary releases for multiple architectures, but the docker image only builds for `amd64` (x86 64-bit) processors. The problem is that `chrislusf/seaweedfs` docker images uses `frolvlad/alpine-glibc`, which is also an `amd64` only build.

I forked the alpine-glibc and built a base image for `amd64`, `arm64`, `armv7`, and `armv6`. Then I forked `seaweedfs` and changed the Dockerfile somewhat to support more cpu architechtures. I have not tested `armv6` build, but the other builds seem to work. I do not have a running seaweed cluster as of now though so my testing is very shallow, but if you want to test the images, you can find the docker image here: https://hub.docker.com/r/gronis/seaweedfs/tags

Gronis:
Actually, I don't think that having glib adds anything since the build of seaweed is happening outside the container. I used alpine image directly and it seams to work as well. Will post a PR with a multiarch supported build.

chrislusf:
[...]

PR #1377, opened 2020-06-30
Support multiple architectures when building Dockerfile

Gronis:
Add support for the following cpu architectures when building the Dockerfile:
- i386
- arm64
- armv6
- armv7

Update readme with instructions to build and push all cpu architectures at the same time.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,15 +1,13 @@
-FROM frolvlad/alpine-glibc
+FROM gronis/alpine-glibc
 
-# Supercronic install settings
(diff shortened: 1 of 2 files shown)


Gronis:
Adresses issue #1220

Gronis:
I have pushed a multi-architecture build here https://hub.docker.com/r/gronis/seaweedfs
It would be even better if this could be supported in the official docker image.

Gronis:
Note: This new image uses alpine image directly rather than using a glibc supported image. I have not found any issues with it, but maybe there are some known issues that I'm unaware of.

PR #1389, opened 2020-07-14
shell: support configure volume ttl

iliul:
just compile pass, will test tomorrow


chrislusf:
To configure a volume, ttl is a required parameter? What if the volume
doesn’t need ttl?

Changing ttl also needs to let volume server to reload the volume.

PR #1406, opened 2020-07-28
add 5-byte large disk builds for arm and arm64

mwpeterson:
I have Odroid HC2s with 12TB drives in them, so running the 5-byte build of SeaweedFS seemed correct.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,14 @@ release: deps windows_build darwin_build linux_build bsd_build 5_byte_linux_buil
 	$(call build_large,windows,amd64,.exe)
 	$(call zip_large,windows,amd64,.exe)
 
+5_byte_arm_build: $(sources)
+	$(call build_large,linux,arm,)
+	$(call tar_large,linux,arm)
+
+5_byte_arm64_build: $(sources)
+	$(call build_large,linux,arm64,)
+	$(call tar_large,linux,arm64)
+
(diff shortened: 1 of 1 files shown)


chrislusf:
Do you also want to add it to the release?

Makefile
```diff
@@ -55,6 +55,14 @@ release: deps windows_build darwin_build linux_build bsd_build 5_byte_linux_buil
 	$(call build_large,windows,amd64,.exe)
 	$(call zip_large,windows,amd64,.exe)
 
+5_byte_arm_build: $(sources)
```
chrislusf, line 58: 32bit arm may not be that useful

Makefile
```diff
@@ -55,6 +55,14 @@ release: deps windows_build darwin_build linux_build bsd_build 5_byte_linux_buil
...
+	$(call build_large,linux,arm,)
+	$(call tar_large,linux,arm)
+
+5_byte_arm64_build: $(sources)
```
chrislusf, line 62: you can add this to line 43, so it can be part of the release process.

Makefile
mwpeterson: I've spent the morning with google trying to figure out why, but I don't think I have. :(
My guess is either:

- something with bit-shifting by 32 being undefined on 32-bit
- `[]byte` can only hold 4 bytes on 32-bit

I'm guessing large disk weed will work until it needs to use that 5th byte and then it'll panic or worse.

PR #1408, opened 2020-07-31
fix spark read s3 bug (sc.binaryFiles)

hilimd:
fix spark read s3 bug

```
sc.hadoopConfiguration.set("fs.s3a.access.key", "ak")
sc.hadoopConfiguration.set("fs.s3a.secret.key", "sk")
sc.hadoopConfiguration.set("fs.s3a.impl","org.apache.hadoop.fs.s3a.S3AFileSystem")
sc.hadoopConfiguration.set("fs.s3a.endpoint", "http://127.0.0.1:8333")

val rdd = sc.binaryFiles("s3a://f01/f02/*.zip")
rdd.count
```


weed/s3api/s3api_objects_list_handlers.go
```diff
@@ -125,9 +124,18 @@ func (s3a *S3ApiServer) listFilerEntries(bucket, originalPrefix string, maxKeys
...
 			if entry.IsDirectory {
 				if entry.Name != ".uploads" {
+					prefix = fmt.Sprintf("%s%s/", dir, entry.Name)
+
 					commonPrefixes = append(commonPrefixes, PrefixEntry{
-						Prefix: fmt.Sprintf("%s%s/", dir, entry.Name),
+						Prefix: prefix,
 					})
+
+					if delimiter != "/" {
```
chrislusf, line 133: in line 64, delimiter could not be "/".

chrislusf: maybe change it to
```
  if delimiter == "" {
```

PR #1433, opened 2020-08-26
stop send heartbeat before stop volume server

kmlebedev:


chrislusf:
What is the reason this is needed? There is already graceful shutdown to drain traffic.

weed/command/volume.go
```diff
@@ -227,6 +228,11 @@ func (v VolumeServerOptions) startVolumeServer(volumeFolders, maxVolumeCounts, v
 		fmt.Println("volume server has be killed")
 		var startTime time.Time
 
+		// Stop heartbeats
+		fmt.Println("stop send heartbeat and sleep %d sec", v.preStopSeconds)
+		volumeServer.SendHeartbeat = false
+		time.Sleep(time.Duration(*v.preStopSeconds) * time.Second)
+		fmt.Println("end sleep 20 sec")
```
chrislusf, line 235: need a fix for the number of seconds printed

PR #1445, opened 2020-09-03
new filer storage to elastic v7.

taozix:


chrislusf:
Thanks! Never thought of using Elastic as a filer store before. In what cases that people may prefer Elastic?

taozix:
we store huge photos and will use for search later directly, also self-defined metadata.

weed/filer2/elastic/v7/elastic_store.go
```diff
@@ -0,0 +1,295 @@
...
+}
+
+func (store *ElasticStore) Shutdown() {
+	store.client.Stop()
+}
+
+func getIndex(fullpath weed_util.FullPath) string {
+	path := strings.Split(string(fullpath), "/")
+	if len(path) > 1 {
+		return indexPrefix + path[1]
```
chrislusf, line 284: this is creating one index for each directory?

PR #1446, opened 2020-09-03
new filer store option to es7.

taozix:
1. can easily scale up the stateless filer server with Elastic cluster store .
2. treat Elastic-cluster as the MetaData service of the whole seaweedfs, we can search Elastic cluster Fast and directly using filename or self-defined metadata, and metadata management is more easy, specially use seaweedfs as the S3 object storage.
3. Elastic cluster also provide excellent put/get/query...performance, and no SPOF.

### DIFF
--- a/weed/filer/elastic/v7/elastic_store.go
+++ b/weed/filer/elastic/v7/elastic_store.go
@@ -7,7 +7,7 @@ import (
 	"math"
 	"strings"
 
-	"github.com/chrislusf/seaweedfs/weed/filer2"
+	"github.com/chrislusf/seaweedfs/weed/filer"
 	"github.com/chrislusf/seaweedfs/weed/glog"
 	"github.com/chrislusf/seaweedfs/weed/pb/filer_pb"
 	weed_util "github.com/chrislusf/seaweedfs/weed/util"
@@ -22,11 +22,11 @@ var (
 
 type ESEntry struct {
 	ParentId string `json:"ParentId"`
-	Entry    *filer2.Entry
+	Entry    *filer.Entry
 }
 
 func init() {
-	filer2.Stores = append(filer2.Stores, &ElasticStore{})
+	filer.Stores = append(filer.Stores, &ElasticStore{})
 }
 
 type ElasticStore struct {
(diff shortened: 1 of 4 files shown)


taozix:
new filer store option to es7, and search/list directly from es cluster.

taozix:
new filer store option to es7, and search/list directly from es cluster.

weed/filer/elastic/v7/elastic_store.go
```diff
@@ -0,0 +1,308 @@
...
+	return ctx, nil
+}
+func (store *ElasticStore) CommitTransaction(ctx context.Context) error {
+	return nil
+}
+func (store *ElasticStore) RollbackTransaction(ctx context.Context) error {
+	return nil
+}
+func (store *ElasticStore) KvDelete(ctx context.Context, key []byte) (err error) {
+	return filer.ErrKvNotImplemented
```
chrislusf, line 70: could you please help to implement the kv operations? I do not have an ES cluster to develop.

It will be used to for more advanced features, such as replication, support hard link, etc.

weed/filer/elastic/v7/elastic_store.go
```diff
@@ -0,0 +1,308 @@
...
+func (store *ElasticStore) GetName() string {
+	return "elastic7"
+}
+func (store *ElasticStore) Initialize(configuration weed_util.Configuration, prefix string) (err error) {
+	servers := configuration.GetString(prefix + "servers")
+	if servers == "" {
+		return fmt.Errorf("error elastic endpoints.")
+	}
+	store.maxPageSize = configuration.GetInt(prefix + "index.max_result_window")
+	if store.maxPageSize <= 0 {
```
chrislusf, line 46: make this optional? Default to some number high enough for most cases?

taozix: default is 10000, coded in scaffold cfg files.
it’s ok for most cases, but increase value will get better query performance.
etc, list the bucket which contains 10 million photos.

weed/filer/elastic/v7/elastic_store.go
taozix: of course, we will implement kv ops, but maybe next submits

weed/filer/elastic/v7/elastic_store.go
chrislusf: The instruction in scaffolded yaml file is not very clear to me.

To list 10 million photos, what is a good number? Let's say it is 1 million, so why not default to 1 million? What is the limiting factor?

Seems complicated and for experts only. Please treat it as optional here and provide a default value, instead of just return an error in line 47.

taozix: because default index.max_result_window in elastic is 10000(in production elastic, it’s much higher),if we query more than  10000 entries a time, es will return errors.if make filer value higher, we can query more entries in one elastic request, but be sure the elastic value more higher, or at least equals the filer value.

chrislusf:
Merged. Thanks!

PR #1449, opened 2020-09-04
ES backended filer support kv ops.

taozix:


chrislusf:
Is there any way to tell Elastic not to index the content?

https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html#enabled

chrislusf:
separate into an `elastic_store_kv.go` file

weed/filer/elastic/v7/elastic_store.go
```diff
@@ -66,15 +72,66 @@ func (store *ElasticStore) CommitTransaction(ctx context.Context) error {
...
 }
+
 func (store *ElasticStore) KvDelete(ctx context.Context, key []byte) (err error) {
-	return filer.ErrKvNotImplemented
+	id := fmt.Sprintf("%x", md5.Sum(key))
+	deleteResult, err := store.client.Delete().
+		Index(indexKV).
+		Type(indexType).
+		Id(id).
+		Do(context.Background())
```
chrislusf: there is already a `ctx` parameter. No need to use `context.Background()`. Same for other places.

weed/filer/elastic/v7/elastic_store.go
```diff
@@ -66,15 +72,66 @@ func (store *ElasticStore) CommitTransaction(ctx context.Context) error {
...
 	return nil
 }
+
 func (store *ElasticStore) KvDelete(ctx context.Context, key []byte) (err error) {
-	return filer.ErrKvNotImplemented
+	id := fmt.Sprintf("%x", md5.Sum(key))
+	deleteResult, err := store.client.Delete().
+		Index(indexKV).
+		Type(indexType).
+		Id(id).
```
chrislusf: why not use `key` here?

taozix: will cause problem if key contains '/'.

weed/filer/elastic/v7/elastic_store.go
taozix: it can be without indexing if totally no need to query.

chrislusf: Only need to fetch the value by the key.

weed/filer/elastic/v7/elastic_store.go
chrislusf: no need to worry about that.

PR #1479, opened 2020-09-22
Fix: remove the oversized state after compaction

LIBA-S:
The oversized volumes can not turn to writable state even if its size is within the writable limitsa after compaction.
.
This PR abstracts a struct to record the real state of a copyset.


chrislusf:
Thanks for the thoughtful PR!

I will merge this. 

Do you consider the case when a volume replication is changed? Maybe need another PR.

chrislusf:
Reverted this PR since it is breaking test topology.TestHandlingVolumeServerHeartbeat()

LIBA-S:
thanks, I will calibrate with caution.

> Reverted this PR since it is breaking test topology.TestHandlingVolumeServerHeartbeat()

LIBA-S:
Hi，I have completed the unit and integration tests.  And I will give serious consideration to the issues you mentioned.

> Thanks for the thoughtful PR!
> 
> I will merge this.
> 
> Do you consider the case when a volume replication is changed? Maybe need another PR.

chrislusf:
create a new PR

PR #1480, opened 2020-09-23
s3 copy and delete object

hilimd:
1.add S3 copy directory function
2.fixed the problem of empty directory when S3 deleted the directory


chrislusf:
> 1.add S3 copy directory function

Is this a new feature that AWS S3 does not have?

> 2.fixed the problem of empty directory when S3 deleted the directory

What is the problem?

PR #1508, opened 2020-10-05
We return etag using the same algorithm as aws s3

kmlebedev:
See [All about AWS S3 ETags](https://teppen.io/2018/06/23/aws_s3_etags/)

Necessary for
```
aws --endpoint-url http://localhost:8333/ s3 sync s3://my-us-west-2-bucket  s3://seaweedfs-bucket --source-region us-west-2
```

### DIFF
--- a/weed/filer/filechunks.go
+++ b/weed/filer/filechunks.go
@@ -1,13 +1,15 @@
 package filer
 
 import (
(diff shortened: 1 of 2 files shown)


chrislusf:
Good job! Thanks!

PR #1517, opened 2020-10-10
Run go test on github actions

fdns:
Currently the go workflows only try to run tests in the weed module (non-recursive), make it so it runs them recursively over the repository.


chrislusf:
@fdns Just ask: do you know how to make this work? https://github.com/marketplace/actions/github-action-publish-binaries

Need to build a latest binary on each commit for users to download.

PR #1537, opened 2020-10-15
add number of read only volumes to prometheus metrics

kmlebedev:
Need for alerting if a lot of read only volumes


chrislusf:
Could you please help to add corresponding changes to the grafana
dashboard?

PR #1547, opened 2020-10-20
Allow custom command arguments to override default values in docker

Suika:
Depending on a setup, 1GB volume limit is not the preferable size.
As weed can process multiple arguements and the later argument is the one to be used,
it is relatively easy to override the arguments defined in the entrypoint.

The arguments passed to the script will be shifted and appended at the end of the command. Since the first argument will fall into specific case, the argument can be predefined and the whole argument array shifted. Allowing to do "volume/server/master/etc $ARGS $@"


chrislusf:
Thanks! I did not know this nifty "shift" command. Nice!

PR #1566, opened 2020-10-27
return x-amz-tag-count header when GET object.

taozix:


chrislusf:
btw: what is your use case with tagging?

taozix:
we need object lifecycle :)

PR #1578, opened 2020-10-29
S3 bucket list, response with uploaded storageclass.

taozix:


weed/util/http_header.go
```diff
@@ -0,0 +1,141 @@
...
+
+	// Signature v2 related constants
+	AmzSignatureV2 = "Signature"
+	AmzAccessKeyID = "AWSAccessKeyId"
+
+	// Response request id.
+	AmzRequestID = "x-amz-request-id"
+
+	// Deployment id.
+	MinioDeploymentID = "x-minio-deployment-id"
```
chrislusf: please remove unused variables.

weed/util/http_header.go
```diff
@@ -0,0 +1,141 @@
+/*
```
chrislusf: Move this file to s3api folder

chrislusf: And remove all unused variables.

taozix: it seems will cause "import cycle" if move http_header to s3api.

chrislusf:
Create a folder there

PR #1596, opened 2020-11-03
S3 credentials store in filer

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1500


chrislusf:
I think this can be implemented similar to https://github.com/chrislusf/seaweedfs/wiki/Path-Specific-Configuration

The s3 configuration file can be stored on filer `/etc/s3.conf`. 

I can help to add a subscribe function to load the `/etc/s3.conf` file whenever it is changed.

kmlebedev:
It would be nice, for k8s could change only /etc/s3.conf config via configmap

chrislusf:
> It would be nice, for k8s could change only /etc/s3.conf config via configmap

You mean `/etc/s3.conf` on local disk, right?

kmlebedev:
> > It would be nice, for k8s could change only /etc/s3.conf config via configmap
> 
> You mean `/etc/s3.conf` on local disk, right?
No i was wrong, in k8s configmap file vi symlinks and s3.conf.json contains secrets, stored in vault and save on disk at the moment init container

chrislusf:
So it is better to store the `s3.conf` on filer? This allows multiple s3 instances to share the same config file, and reload the config upon changes. 

The downside is the `s3.conf` is visible to everyone.

kmlebedev:
Yes, and the config file might be worth encrypting

chrislusf:
Could you please add `s3.configure` similar to this?

https://github.com/chrislusf/seaweedfs/blob/master/weed/shell/command_fs_configure.go

kmlebedev:
@chrislusf   Can you tell me what to do with ?
```
Run cd weed; go build -v .
package github.com/chrislusf/seaweedfs/weed
	imports github.com/chrislusf/seaweedfs/weed/command
	imports github.com/chrislusf/seaweedfs/weed/s3api
	imports github.com/chrislusf/seaweedfs/weed/server
	imports github.com/chrislusf/seaweedfs/weed/shell
	imports github.com/chrislusf/seaweedfs/weed/s3api: import cycle not allowed
Error: Process completed with exit code 1.
```

chrislusf:
Need to break the import cycle. The`shell` package should not import `s3api`. You can create a separate package, e.g., `s3api/s3config`, and both `s3api` and `shell` use the the package.

chrislusf:
btw: 

- put file "s3.conf" under "/etc".
- add tests

chrislusf:
> * save content as a file, not extended attributes. We can add encryption/decryption later.

https://github.com/chrislusf/seaweedfs/blob/master/weed/shell/command_fs_configure.go has the example of saving to a file.

kmlebedev:
> > * save content as a file, not extended attributes. We can add encryption/decryption later.
> 
> https://github.com/chrislusf/seaweedfs/blob/master/weed/shell/command_fs_configure.go has the example of saving to a file.

I saw it seems not so reliable and requires additional backup efforts, unlike DB

chrislusf:
Later, small files less than a limit, such as 4KB, will be cached to the filer store for faster reads. 

We do not really need to optimize it now.

chrislusf:
See https://github.com/chrislusf/seaweedfs/commit/f4abd01adf9fb2ebdefa844b2a4f0aaf8eeb1ccc

kmlebedev:
@chrislusf Maybe we will already ? and in another PR I will continue s3 admin/iam api

chrislusf:
Great! I will merge this and continue from here. Thanks!

chrislusf:
One small thing: can you move "github.com/chrislusf/seaweedfs/weed/s3iam" to "github.com/chrislusf/seaweedfs/weed/s3api/s3iam"?

Also, remove the "Draft" status from this PR, so that I can merge it.

kmlebedev:
> One small thing: can you move "github.com/chrislusf/seaweedfs/weed/s3iam" to "github.com/chrislusf/seaweedfs/weed/s3api/s3iam"?

I deliberately did this, because in AWS  iam this is a separate service

PR #1610, opened 2020-11-11
prefer to read from volumes in this data center

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1608


chrislusf:
this is not ready, correct?

kmlebedev:
> this is not ready, correct?

yes it's not ready

chrislusf:
Currently the master already knows each volume id dc/rack/node. Just need to return the dc info.

https://github.com/chrislusf/seaweedfs/blob/3a5e4769cfe383ca5a3e3d4078b5e80120763f3b/weed/server/master_server_handlers.go#L73

kmlebedev:
@chrislusf i checked it works locally
without unnecessary grpc calls

PR #1650, opened 2020-12-04
security master use access.ui

kmlebedev:
Disable master api separately from UI


chrislusf:
disable ui access is good but not really "safe", right?

kmlebedev:
Disable ui access is safe, unlike api access. This PR allows disable api access and enable ui access is good

chrislusf:
Got it. How about just always enable the UI access, and the "DisableHttp" only applies to API access?

I want to avoid the number of options if possible.

kmlebedev:
> Got it. How about just always enable the UI access, and the "DisableHttp" only applies to API access?
> 
> I want to avoid the number of options if possible.

understand, let's do it

PR #1683, opened 2020-12-16
fix race condition when loading volumes concurrently

qieqieplus:
fix #1682

PR #1685, opened 2020-12-17
delete volume placeholder when error

qieqieplus:
Delete placeholder from volumes when NewVolume()  failed

Also fix:
[store.go#L259](weed/storage/store.go#L259)

```
	found, err := location.deleteVolumeById(vid)
	if found {
		if err == nil {
			glog.V(0).Infof("volume %d is deleted", vid)
		} else {
			glog.V(0).Infof("delete volume %d: %v", vid, err)
		}
	}
```

won't log err, since `found` is always false when `deleteVolumeById` returns an error

### DIFF
--- a/weed/storage/disk_location.go
+++ b/weed/storage/disk_location.go
@@ -112,10 +112,11 @@ func (l *DiskLocation) loadExistingVolume(fileInfo os.FileInfo, needleMapKind Ne
 	// load the volume
 	v, e := NewVolume(l.Directory, l.IdxDirectory, collection, vid, needleMapKind, nil, nil, 0, 0)
 	if e != nil {
+		_ = l.DeleteVolume(vid)
 		glog.V(0).Infof("new volume %s error %s", volumeName, e)
 		return false
 	}
-	
+
 	l.SetVolume(vid, v)
 
 	size, _, _ := v.FileStat()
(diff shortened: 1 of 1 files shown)


weed/storage/disk_location.go
```diff
@@ -207,16 +208,15 @@ func (l *DiskLocation) DeleteCollectionFromDiskLocation(collection string) (e er
...
-	v, ok := l.volumes[vid]
-	if !ok {
-		return
-	}
-	e = v.Destroy()
-	if e != nil {
-		return
+	v, found := l.volumes[vid]
+	if found {
+		if v != nil {
```
chrislusf, line 213: Why put nil in the map in the previous PR?

chrislusf (changes requested):
Related to https://github.com/chrislusf/seaweedfs/pull/1683

These two fixes are not needed. Need to find the original problem and prevent one volume being loaded twice.

weed/storage/disk_location.go
qieqieplus: explained in #1683

qieqieplus:
Still the other issue I mentioned in this pull request comment has not been addressed.

chrislusf:
Added a fix. How is it now?

qieqieplus:
LGTM

chrislusf:
Thanks!

PR #1727, opened 2021-01-04
fix #1726

### LINKED ISSUE #1726: memory related bugs of rocksdb implementation
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs

Hi Chris,
I maintained my own rocksdb branch previously, happy to find out it's now offically supported.
Here's some small suggestions for the current implementation.
1. Option objects created from `NewDefaultOptions` `NewDefaultReadOptions` `NewDefaultWriteOptions` contain cgo pointers need to be released explicitly.
2. rocksdb_store.go#L96 `entry.FullPath` leads to nil pointer dereference.
3. Rocksdb iterator related keys & values will be freed automatically upon iterator's destruction, so `iter.Close()` ensures no memory leak without `key.Free()` or `value.Free()`.

Other possible improvements:
1. Reuse default options when possible.
2. Use `db.Get`(and free by hand) instead of `db.GetBytes` to avoid memory copy when bytes are read only. Gorocksdb `charToByte` use reflect tricks by creating byte slice with underlying C char array to avoid char* to go []byte copy, while `C.GoBytes` performs memory copy.

qieqieplus:


chrislusf:
Thanks a lot! Any changes also needed for the kv operations?

qieqieplus:
Add another commit.
There's no guarantee that `KvGet` result is read-only, so leave it as it is for now.

weed/filer/rocksdb/rocksdb_store.go
```diff
@@ -11,17 +11,36 @@ import (
 	"github.com/chrislusf/seaweedfs/weed/glog"
 	"github.com/chrislusf/seaweedfs/weed/pb/filer_pb"
 	weed_util "github.com/chrislusf/seaweedfs/weed/util"
-	"github.com/tecbot/gorocksdb"
+	rocksdb "github.com/tecbot/gorocksdb"
```
chrislusf, line 14: this is the same as the current package name. 

seems the go compiler is ok with this. But it is better to have a different name. The original name should be fine.

PR #1728, opened 2021-01-05
Seaweed custom header are not visible to Vue or javascript

fuyouyshengwu:


weed/server/filer_server_handlers_read.go
```diff
@@ -100,6 +100,16 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request,
 		w.Header().Set(k, string(v))
 	}
 
+	//Add some custom headers that need to be exposed
+	seaweedHeaders := []string{}
+	for i, _ := range w.Header() {
+		if strings.Contains(i, "Seaweed") {
```
chrislusf: - Use has prefix "Seaweed-"
- Use variable name "header" instead of "i"

weed/server/filer_server_handlers_read.go
```diff
@@ -100,6 +100,16 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request,
 		w.Header().Set(k, string(v))
 	}
 
+	//Add some custom headers that need to be exposed
```
chrislusf: "need to be exposed" is not clear.

How about "Seaweed custom header are not visible to Vue or javascript?"

PR #1730, opened 2021-01-05
impl: TTL per entry for rocksdb; fix package name

qieqieplus:
Also remove `MaybeDecompressData` since no Compress performed in `InsertEntry`

Did a quick benchmark on insert gzipped or raw large random values. (rocksdb use snappy by default), 
for values < 10k, raw values are much faster,
for values > 100k, gzipped values are slightly faster, but really close (-10% ~ +15%)

### DIFF
--- a/weed/filer/rocksdb/rocksdb_ttl.go
+++ b/weed/filer/rocksdb/rocksdb_ttl.go
@@ -1,0 +1,41 @@
+//+build rocksdb
+
+package rocksdb
+
+import (
+	"time"
+
+	"github.com/tecbot/gorocksdb"
+
+	"github.com/chrislusf/seaweedfs/weed/filer"
+)
+
+type TTLFilter struct {
+	skipLevel0 bool
+}
+
+func NewTTLFilter() gorocksdb.CompactionFilter {
+	return &TTLFilter{
+		skipLevel0: true,
+	}
+}
+
+func (t *TTLFilter) Filter(level int, key, val []byte) (remove bool, newVal []byte) {
+	// decode could be slow, causing write stall
+	// level >0 sst can run compaction in parallel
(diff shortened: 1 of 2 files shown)


weed/filer/rocksdb/rocksdb_ttl.go
```diff
@@ -0,0 +1,41 @@
...
+func (t *TTLFilter) Filter(level int, key, val []byte) (remove bool, newVal []byte) {
+	// decode could be slow, causing write stall
+	// level >0 sst can run compaction in parallel
+	if t.skipLevel0 && level == 0 {
+		return false, val
+	}
+	entry := filer.Entry{}
+	if err := entry.DecodeAttributesAndChunks(val); err == nil {
+		if entry.TtlSec == 0 ||
+			entry.Crtime.Add(time.Duration(entry.TtlSec)*time.Second).Before(time.Now()) {
```
chrislusf: Before or After?

qieqieplus: after ;)

qieqieplus: By the way, it can be a lot faster if we can get ttl without deserialization. Any ideas?

chrislusf: Compaction speed is ok.

qieqieplus:
fix on my mobile phone. hope it doesn't break anything.

qieqieplus:
Just came up with that some data stored via `KvPut` could be purged by filter since they're not serialized entry.
Should we open up another db for kv or ignore decode error in filter?

chrislusf:
create a new db would be better.

chrislusf:
Please ignore decode errors for now.

PR #1738, opened 2021-01-07
fix rocksdb crash when list directory

qieqieplus:
rocksdb iterator `iter.Seek()` may reach EOF,
`iter.Next()` without check could cause segmentation violation.

How to reproduce:
List an invalid directory.

Also remove redundant limit check.

### DIFF
--- a/weed/filer/rocksdb/rocksdb_store.go
+++ b/weed/filer/rocksdb/rocksdb_store.go
@@ -182,17 +182,9 @@ func enumerate(iter *gorocksdb.Iterator, prefix, lastKey []byte, includeLastKey 
 	} else {
 		iter.Seek(lastKey)
 
-		if !includeLastKey {
-			key := iter.Key().Data()
-
-			if !bytes.HasPrefix(key, prefix) {
-				return nil
-			}
-
-			if bytes.Equal(key, lastKey) {
-				iter.Next()
-			}
-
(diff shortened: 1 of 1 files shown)


weed/filer/rocksdb/rocksdb_store.go
```diff
@@ -250,10 +242,6 @@ func (store *RocksDBStore) ListDirectoryPrefixedEntries(ctx context.Context, ful
 		if fileName == "" {
 			return true
 		}
-		limit--
```
chrislusf, line 253: why this is removed?

weed/filer/rocksdb/rocksdb_store.go
```diff
@@ -182,17 +182,9 @@ func enumerate(iter *gorocksdb.Iterator, prefix, lastKey []byte, includeLastKey
...
-			if !bytes.HasPrefix(key, prefix) {
-				return nil
-			}
-
-			if bytes.Equal(key, lastKey) {
-				iter.Next()
-			}
-
+		if iter.Valid() && !includeLastKey &&
+			bytes.Equal(iter.Key().Data(), lastKey) {
```
chrislusf: Please use the previous style to split into multiple `if`

chrislusf:
ok. I will merge it and fix the nested if.

PR #1743, opened 2021-01-14
MYSQL - Better Connection Pool Control

LazyDBA247-Anyvision:
added SetConnMaxLifetime setting (https://golang.org/pkg/database/sql/#DB.SetConnMaxLifetime)
so we can control the age of connection in the connection pool, keeping them young and alive.

updated k8s chart mysql setting to accumadate the new setting and set lower connections per filer (when having 3 filers, no need for 450 connections) and even 75 connection is more than enough.


chrislusf:
merged and fixed a trivial bug. https://github.com/chrislusf/seaweedfs/commit/fd1d8a2a0958e4256ac0bbd77a982c5fc2274546

PR #1761, opened 2021-01-27
replication to todays date directory

kmlebedev:


kmlebedev:
This is needed for an incremental backup.

chrislusf:
on second thought, I think it is better to avoid this feature. 

1. The deletions or updates would not work.
2. The configuration is a bit hard to understand.

kmlebedev:
> on second thought, I think it is better to remove this feature.
> 
> 1. The deletions or updates would not work.
I redo on entry mtime

chrislusf:
How about create one sink with simple "rsync" to local, and another sink with daily partitions?

chrislusf:
> 1. The deletions or updates would not work.
>    I redo on entry mtime

You may want to use entry ctime, which does not change for following updates after creation.

kmlebedev:
> How about create one sink with simple "rsync" to local, and another sink with daily partitions?

You can, you need to think about how to reduce the copy-paste of the code

PR #1762, opened 2021-01-27
replication to create time date directory

kmlebedev:
replace https://github.com/chrislusf/seaweedfs/pull/1761


weed/command/scaffold.go
```diff
@@ -353,9 +353,14 @@ directory = "/buckets"
 
 [sink.local]
```
chrislusf, line 354: how about use these two sink names, "local_incremental" and "local"?

kmlebedev: ok

kmlebedev:
done

chrislusf:
I think there are some errors. Let me merge and check.

chrislusf:
I updated the related code. Please help to check whether it makes sense.

kmlebedev:
> I updated the related code. Please help to check whether it makes sense.

https://github.com/chrislusf/seaweedfs/commit/da08402ba253fe5391f0152905f51f82c51fd527#commitcomment-46462627

PR #1794, opened 2021-02-10
restart replication if RabbitMQ connection closed

kmlebedev:
https://github.com/google/go-cloud/issues/2958


chrislusf:
wow, really? Fatal error just to restart...

kmlebedev:
> wow, really? Fatal error just to restart...
> 
> Probably we should switch to some other rabbit mq library.

So far, only so for rabbitpubsub 
Let's see what they say
https://github.com/google/go-cloud/issues/2958

kmlebedev:
```
filer_1      | I0210 11:59:59     1 filer_notify.go:57] notifying entry update /buckets/test/dir/zookeeper-by-prometheus_rev4.json
filer_1      | E0210 11:59:59     1 filer_notify.go:59] send message via Go CDK pubsub rabbit://swexchange: pubsub (code=FailedPrecondition): Exception (504) Reason: "channel/connection is not open"
```

PR #1823, opened 2021-02-20
feature: support command line arguments in a configuration file

bingoohuang:
direct command line is too long like:

`weed -logdir=../mlog master -mdir=../mdir -peers=192.168.126.5:3333,192.168.126.16:3333,192.168.126.18:3333 -port=3333 -defaultReplication=001`

after applying this feature, we can use a configuration file like:

```
# master.conf
logdir=../mlog
mdir=../mdir
peers=192.168.126.5:3333,192.168.126.16:3333,192.168.126.18:3333
port=3333
defaultReplication=001
```

then refer it like `weed master -config=master.conf`

Order of precedence of this flag, it is borrowed from https://github.com/namsral/flag with some customized for weed:

1. Command line options
2. Environment variables
3. Configuration file
4. Default values

So, if someone like command lines, old style is still there. If someone get boring with so many command line arguments, he can use a configuration file as an alternative. Why not!


chrislusf:
How to see the full list of command line options in effect?

chrislusf:
This breaks the help message for `weed -h`. It used to have
```
Use "weed help [command]" for more information about a command.

For Logging, use "weed [logging_options] [command]". The logging options are:
  -alsologtostderr
    	log to standard error as well as files (default true)
  -log_backtrace_at value
    	when logging hits line file:N, emit a stack trace
  -logdir string
    	If non-empty, write log files in this directory
  -logtostderr
    	log to standard error instead of files
  -stderrthreshold value
    	logs at or above this threshold go to stderr
  -v value
    	log level for V logs
  -vmodule value
    	comma-separated list of pattern=N settings for file-filtered logging
```

After this PR, the corresponding part became:
```
Use "weed help [command]" for more information about a command.

For Logging, use "weed [logging_options] [command]". The logging options are:
  -config string
    	config file
  -logdir string
    	If non-empty, write log files in this directory
```

bingoohuang:
It is fixed.

chrislusf:
Thanks a lot!

chrislusf:
Merged. It would be better if the config file content is printed out when the program starts.

bingoohuang:
Yes, some additional work can be do :

1. sample config files creation.
2. config get effected by command-line, env, or config file printing.
3. final config adopted overview printing.

And any suggestion more ?

chrislusf:
One bug with this approach is that the command line option can not duplicate with existing environment variables. For example, if `weed filer -path=xxx` is defined, but not specified when just running `weed filer`. The env variable $PATH would be used. This caused a big confusion.

It should only read env variable $WEED_CLI_XXXX, e.g., $WEED_CLI_PATH.

bingoohuang:
Yes, it is really a bug, I should set env prefix to a default "WEED_", a pr will be submit ASAP.

PR #1862, opened 2021-03-05
Proxy the master UI to the leader as well

Woellchen:


chrislusf:
How to just see the UI of one particular master?

Woellchen:
Ok maybe I am doing it wrong, but whenever I open the status page of a master that is not leader, the tables are all empty and i have to click on the leader link to access the status page from the leader.
I thought this was maybe overlooked because the `/` route is also proxied to the leader.
With this change I can always see a working status page no matter which master I access.

chrislusf:
Right. But sometimes there could be problem with the non-leader master and we need to check the UI.

Woellchen:
Ok, I am fine with that :) We can close this if you want.

PR #1883, opened 2021-03-10
s3 delete-objects pass OK for fail to delete non-empty folder

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1857


kmlebedev:
@chrislusf  Its ok ?

PR #1903, opened 2021-03-15
avoid http error: superfluous response.WriteHeader

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1838

### DIFF
--- a/weed/filer/stream.go
+++ b/weed/filer/stream.go
@@ -73,9 +73,11 @@ func StreamContent(masterClient wdclient.HasLookupFileIdFunction, w io.Writer, c
 
 func CheckAllChunkViews(chunkViews []*ChunkView, fileId2Url *map[string][]string, gErr *errgroup.Group) {
 	for _, chunkView := range chunkViews {
+		urlStrings := (*fileId2Url)[chunkView.FileId]
+		glog.V(9).Infof("Check chunk: %+v\n url: %v", chunkView, urlStrings)
 		gErr.Go(func() error {
 			_, err := retriedFetchChunkData(
-				(*fileId2Url)[chunkView.FileId],
+				urlStrings,
 				chunkView.CipherKey,
 				chunkView.IsGzipped,
 				chunkView.IsFullChunk(),
--- a/weed/filer/stream.go
+++ b/weed/filer/stream.go
@@ -14,7 +14,7 @@ import (
 	"github.com/chrislusf/seaweedfs/weed/wdclient"
 )
 
-func StreamContent(masterClient wdclient.HasLookupFileIdFunction, w io.Writer, chunks []*filer_pb.FileChunk, offset int64, size int64) error {
+func StreamContent(masterClient wdclient.HasLookupFileIdFunction, w io.Writer, chunks []*filer_pb.FileChunk, offset int64, size int64, isCheck bool) error {
 
 	glog.V(9).Infof("start to stream content for chunks: %+v\n", chunks)
 	chunkViews := ViewFromChunks(masterClient.GetLookupFileIdFunction(), chunks, offset, size)
@@ -34,15 +34,20 @@ func StreamContent(masterClient wdclient.HasLookupFileIdFunction, w io.Writer, c
 		fileId2Url[chunkView.FileId] = urlStrings
 	}
 
-	for idx, chunkView := range chunkViews {
-
-		urlStrings := fileId2Url[chunkView.FileId]
-
+	if isCheck {
 		// Pre-check all chunkViews urls
 		gErr := new(errgroup.Group)
-		if len(chunkViews) > 1 && idx == 0 {
-			CheckAllChunkViews(chunkViews[1:], &fileId2Url, gErr)
+		CheckAllChunkViews(chunkViews, &fileId2Url, gErr)
+		if err := gErr.Wait(); err != nil {
+			glog.Errorf("check all chunks: %v", err)
+			return fmt.Errorf("check all chunks: %v", err)
 		}
+		return nil
+	}
+
+	for _, chunkView := range chunkViews {
+
+		urlStrings := fileId2Url[chunkView.FileId]
 		data, err := retriedFetchChunkData(
 			urlStrings,
 			chunkView.CipherKey,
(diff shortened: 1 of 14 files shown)


kmlebedev:
> If WriteHeader is not called explicitly, the first call to Write
will trigger an implicit WriteHeader(http.StatusOK).
WriteHeader: Only one header may be written. Go does not currently!

See
https://github.com/golang/go/blob/master/src/net/http/server.go#L150

@chrislusf  Until I found a good solution on the fly, maybe you know better?
How to change the default httpCode for write() ?
https://github.com/golang/go/blob/master/src/net/http/server.go#L1601

kmlebedev:
@chrislusf I found an option for s3 to replace the http status OK with partial content

weed/filer/stream.go
```diff
@@ -33,16 +34,32 @@ func StreamContent(masterClient wdclient.HasLookupFileIdFunction, w io.Writer, c
...
-	for _, chunkView := range chunkViews {
+	for idx, chunkView := range chunkViews {
 
 		urlStrings := fileId2Url[chunkView.FileId]
 
-		data, err := retriedFetchChunkData(urlStrings, chunkView.CipherKey, chunkView.IsGzipped, chunkView.IsFullChunk(), chunkView.Offset, int(chunkView.Size))
+		// Pre-check all chunkViews urls
+		gErr := new(errgroup.Group)
+		if len(chunkViews) > 1 && idx == 0 {
+			CheckAllChunkViews(chunkViews[1:], &fileId2Url, gErr)
```
chrislusf: seems a lot of overhead here.

kmlebedev: > seems a lot of overhead here.

here is log, check requests with bytes=0-1 processed in parallel with the receipt of the first chunk
```
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,16bed1797b 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,1994fd057f bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1d032c3d04 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,176f4d5358 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /7,1c3133458c bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,1ba19850c6 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1f72f56cc2 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,21b3aabf73 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1e6d572d2b bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,22136b5ac6 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /6,258329f089 bytes=0-1
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,18e9cbf3f2
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /7,1a9119ffd1 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1d032c3d04 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1f72f56cc2 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,20a3a041ac 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,23f7514b68 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,24cade9bbf 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,176f4d5358 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,1994fd057f 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,1ba19850c6 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /7,1c3133458c 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /5,1e6d572d2b 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /4,21b3aabf73 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,22136b5ac6 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /6,258329f089 
volume_1  | I0315 18:29:33     1 volume_server_handlers_read.go:30] GET /3,1595545117 
```

chrislusf: so one request becomes two requests?

kmlebedev: yes it may slightly increase the utilization of the CPU. But on the other hand, the http connection warms up
Not as critical as, for example, enabling TLS

chrislusf: It is "free" when the system load is low, but it is costly when the load is high. 

It is a coding problem and should not rely on extra network round trip.

kmlebedev: Let's move behind the option then? 
In fact, s3 makes range requests, if they match the filler chunk size, then there will always be one request.
So it's more like a reassurance.

chrislusf: This does not worth an option. I think this is ok to return a 200 status code, promising a good response, but failed in the middle.

kmlebedev: > This does not worth an option. I think this is ok to return a 200 status code, promising a good response, but failed in the middle.

It is necessary to somehow distinguish normal responses from broken ones in the middle in monitoring. Here you can add a metric and follow it.
If there are many such requests, it will be possible to take action.

We are also looking forward to the opportunity to repeat the request for 5xx on the upstream

kmlebedev: There is also an option to move the check only for HEAD requests ?

kmlebedev:
@chrislusf I moved checks to HEAD request

kmlebedev:
Everything seems to be ready here
Monitoring by another issues

chrislusf:
> Everything seems to be ready here
> Monitoring by another issues

Seems this PR needs a rebase.

kmlebedev:
> Seems this PR needs a rebase.

@chrislusf This new PR https://github.com/chrislusf/seaweedfs/pull/1913

chrislusf:
> I moved checks to HEAD request

This caused re-reading the whole file. #2046 
I am going to remove it for now.

kmlebedev:
We need to check with loading one byte

PR #1910, opened 2021-03-16
s3 putToFiler maxMB size

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1908

### DIFF
--- a/weed/s3api/s3api_object_multipart_handlers.go
+++ b/weed/s3api/s3api_object_multipart_handlers.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"github.com/chrislusf/seaweedfs/weed/glog"
 	"github.com/chrislusf/seaweedfs/weed/s3api/s3err"
+	"math"
 	"net/http"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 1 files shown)


weed/s3api/s3api_object_multipart_handlers.go
```diff
@@ -201,8 +202,9 @@ func (s3a *S3ApiServer) PutObjectPartHandler(w http.ResponseWriter, r *http.Requ
 	}
 	defer dataReader.Close()
 
-	uploadUrl := fmt.Sprintf("http://%s%s/%s/%04d.part?collection=%s",
-		s3a.option.Filer, s3a.genUploadsFolder(bucket), uploadID, partID, bucket)
+	maxMB := int(math.Ceil(float64(r.ContentLength) / 1024 / 1024))
+	uploadUrl := fmt.Sprintf("http://%s%s/%s/%04d.part?collection=%s&maxMB=%d",
```
chrislusf, line 206: S3 upload part can be fairly big. Why not leave the filer to determine the chunk size?

kmlebedev: The size of the downloadable part of S3 is convenient for the user himself, but you can limit it to a reasonable size, for example, filer maxMB * 5

chrislusf: I prefer not to change here. Filer can handle the chunking transparently. S3 server should not need to worry about it.

kmlebedev: This is necessary in special cases, the maxMB code is on average less than the size of a s3 chunk, as a result of which several chunkviews must be delivered for one s3 request.
on average, the size of a s3 chunk is from 5mb to 50mb,  so i configured maxMB to 64.

chrislusf: but why need to have a large chunk instead of several smaller ones? The latter is more efficient to process.

kmlebedev: > but why need to have a large chunk instead of several smaller ones? The latter is more efficient to process.

It depends on the content, if I have a regular 100Gb source file I want the chunk size of at least 100 mb.

chrislusf: The number of chunks is not a big issue. Filer has a feature to merge every 1000 chunk ids into one chunk, and just reference that chunk id.

ddorian: @kmlebedev

> It depends on the content, if I have a regular 100Gb source file I want the chunk size of at least 100 mb.

full chunk goes in memory before it can be processed, so you will be limited with these big chunks, there's no streaming afaik.

PR #1933, opened 2021-03-23
add missingKeysThreshold for fsck

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/1930

### DIFF
--- a/weed/shell/command_volume_check_disk.go
+++ b/weed/shell/command_volume_check_disk.go
@@ -49,8 +49,7 @@ func (c *commandVolumeCheckDisk) Do(args []string, commandEnv *CommandEnv, write
 	slowMode := fsckCommand.Bool("slow", false, "slow mode checks all replicas even file counts are the same")
 	verbose := fsckCommand.Bool("v", false, "verbose mode")
 	applyChanges := fsckCommand.Bool("force", false, "apply the fix")
-	missingKeysThreshold := fsckCommand.Float64("missingKeysThreshold", 0.3, "repair when missing keys is not more than this limit")
-
+	repairThreshold := fsckCommand.Float64("repairThreshold", 0.3, "repair when missing keys is not more than this limit")
 	if err = fsckCommand.Parse(args); err != nil {
 		return nil
 	}
@@ -103,10 +102,10 @@ func (c *commandVolumeCheckDisk) Do(args []string, commandEnv *CommandEnv, write
 			}
 
 			// find and make up the differnces
-			if err := c.doVolumeCheckDisk(aDB, bDB, a, b, *verbose, writer, *applyChanges, *missingKeysThreshold); err != nil {
+			if err := c.doVolumeCheckDisk(aDB, bDB, a, b, *verbose, writer, *applyChanges, *repairThreshold); err != nil {
 				return err
 			}
-			if err := c.doVolumeCheckDisk(bDB, aDB, b, a, *verbose, writer, *applyChanges, *missingKeysThreshold); err != nil {
+			if err := c.doVolumeCheckDisk(bDB, aDB, b, a, *verbose, writer, *applyChanges, *repairThreshold); err != nil {
 				return err
 			}
 			replicas = replicas[1:]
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_check_disk.go
```diff
@@ -129,6 +131,12 @@ func (c *commandVolumeCheckDisk) doVolumeCheckDisk(subtrahend, minuend *needle_m
 	})
 
 	fmt.Fprintf(writer, "volume %d %s has %d entries, %s missed %d entries\n", source.info.Id, source.location.dataNode.Id, counter, target.location.dataNode.Id, len(missingNeedles))
+	missingNeedlesPercent := float64(len(missingNeedles)) / float64(counter)
```
chrislusf: the variable name is percentage, but the calculation is not.

weed/shell/command_volume_check_disk.go
```diff
@@ -49,6 +49,8 @@ func (c *commandVolumeCheckDisk) Do(args []string, commandEnv *CommandEnv, write
 	slowMode := fsckCommand.Bool("slow", false, "slow mode checks all replicas even file counts are the same")
 	verbose := fsckCommand.Bool("v", false, "verbose mode")
 	applyChanges := fsckCommand.Bool("force", false, "apply the fix")
+	missingKeysThreshold := fsckCommand.Float64("missingKeysThreshold", 0.3, "repair when missing keys is not more than this limit")
```
chrislusf: The parameter name is confusing whether it is a maximum limit or minimum limit without looking at the help message.

And it is confusing whether it is a percentage or not.

When this parameter is needed?

kmlebedev: > When this parameter is needed?

Since this command has been added to cronjob 
And it starts automatically, then I want to have protection against a possible software error in the case of an abnormal situation where the number of lost keys cannot exceed a reasonable limit.
If the limit is reached, it means that something went wrong and the job will end with an error and I can draw attention to this and come in manual mode to fix it
https://github.com/chrislusf/seaweedfs/blob/master/k8s/seaweedfs/templates/cronjob.yaml#L44

chrislusf: need a better name than `missingKeysThreshold`

kmlebedev: > need a better name than `missingKeysThreshold`

@chrislusf 
`repairThreshold` norm ?

chrislusf: `skipRepairThreshold`?

kmlebedev: > `skipRepairThreshold`?

In the sense then `stopRepairThreshold` 
Although from my point of view a pun
since Threshold already implies stopping the action

chrislusf: The meanings of `Threshold` in `garbageThreshold` and `repairThreshold` are opposite.

kmlebedev: > The meanings of `Threshold` in `garbageThreshold` and `repairThreshold` are opposite.

nonRepairThreshold ?

weed/shell/command_volume_check_disk.go
```diff
@@ -129,6 +130,12 @@ func (c *commandVolumeCheckDisk) doVolumeCheckDisk(subtrahend, minuend *needle_m
 	})
 
 	fmt.Fprintf(writer, "volume %d %s has %d entries, %s missed %d entries\n", source.info.Id, source.location.dataNode.Id, counter, target.location.dataNode.Id, len(missingNeedles))
+	missingNeedlesFraction := float64(len(missingNeedles)) / float64(counter)
```
chrislusf, line 138: counter == 0 ?

weed/shell/command_volume_check_disk.go
chrislusf: good!

weed/shell/command_volume_check_disk.go
```diff
@@ -129,6 +130,12 @@ func (c *commandVolumeCheckDisk) doVolumeCheckDisk(subtrahend, minuend *needle_m
...
+	if missingNeedlesFraction > nonRepairThreshold {
+		return fmt.Errorf(
+			"failed to start repair volume %d, percentage of missing keys is greater than the threshold: %.2f > %.2f",
```
LazyDBA247-Anyvision, line 141: @kmlebedev if just skipping why  "failed to start repair volume",  maybe  "skipping repair volume"
or did I misunderstood?

kmlebedev: > @kmlebedev if just skipping why "failed to start repair volume", maybe "skipping repair volume"
> or did I misunderstood?

Yes, check command exits with this error

weed/shell/command_volume_check_disk.go
kmlebedev: > counter == 0 ?

fixed

PR #1946, opened 2021-03-29
Adding Customisation - insertQuery for postgres/mysql 

LazyDBA247-Anyvision:
Added the ability to write your own insertQuery to support the UPSERT syntax most of the RDBMS in now day already support.

### DIFF
--- a/weed/command/scaffold.go
+++ b/weed/command/scaffold.go
@@ -102,6 +102,10 @@ enabled = false
 dir = "./filerrdb"					# directory to store rocksdb files
 
 [mysql]  # or memsql, tidb
+# Empty insertQuery will use the default insert query
+# example insertQuery can be for UPSERT syntax:
+# insertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
+insertQuery = ""
 # CREATE TABLE IF NOT EXISTS filemeta (
 #   dirhash     BIGINT               COMMENT 'first 64 bits of MD5 hash value of directory field',
 #   name        VARCHAR(1000) BINARY COMMENT 'directory or file name',
@@ -123,6 +127,10 @@ interpolateParams = false
 
 [mysql2]  # or memsql, tidb
 enabled = false
+# Empty insertQuery will use the default insert query
+# example insertQuery can be for UPSERT syntax:
+# insertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
+insertQuery = ""
 createTable = """
   CREATE TABLE IF NOT EXISTS ` + "`%s`" + ` (
     dirhash BIGINT,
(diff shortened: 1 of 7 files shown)


kmlebedev:
@LazyDBA247-Anyvision 
It seems like a bad idea to add it to the config `insertQuery`
It might be better to check the RDBMS version and choose the insert type automatically ?

weed/command/scaffold.go
```diff
@@ -102,6 +102,10 @@ enabled = false
 dir = "./filerrdb"					# directory to store rocksdb files
 
 [mysql]  # or memsql, tidb
+# Empty insertQuery will use the default insert query
+# example insertQuery can be for UPSERT syntax:
+# insertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
```
chrislusf, line 107: The example seems wrong. ``` ` + "`%s`" + ` ``` should not need the plus operator for the string concatenation.

LazyDBA247-Anyvision: that is because the scaffold uses \` and MySQL uses \` to quote the table name, but go uses it as string literal...

LazyDBA247-Anyvision:
@kmlebedev postgres / mysql driver fits many databases and for each, you may have some syntax differences.
until today the old way worked but caused 2 round trips from filer to DB and 2 table seeks. I am offering customization that needs to be verified and maybe even written by the owner/deployer.

chrislusf:
> @LazyDBA247-Anyvision
> It seems like a bad idea to add it to the config `insertQuery`
> It might be better to check the RDBMS version and choose the insert type automatically ?

There are many mysql/postgres variations.

weed/command/scaffold.go
LazyDBA247-Anyvision: @chrislusf  already exists and working in:
https://github.com/chrislusf/seaweedfs/blob/96c62bd34d4d2c99dea3164a9c062abaddb4eaab/weed/command/scaffold.go#L135

LazyDBA247-Anyvision:
@chrislusf that's exactly why I added it as an optional and the possibility to update the syntax to fit your specific variant.

kmlebedev:
@LazyDBA247-Anyvision I understand, but this raises the entry threshold for configuring inserts.
I believe it could have been made easier!

LazyDBA247-Anyvision:
@kmlebedev fully backward compatible and without breaking new deployment, we must add it as a custom/advanced setting. not sure I / We can start testing all possible Postgres (9.X -> 13.X) or cockroach/yugabytedb/etc... and not MySQL ones... so this is a special config for a DBA/Optimizer, as SW can continue to work without it.

kmlebedev:
> @kmlebedev fully backward compatible and without breaking new deployment, we must add it as a custom/advanced setting. not sure I / We can start testing all possible Postgres (9.X -> 13.X) or cockroach/yugabytedb/etc... and not MySQL ones... so this is a special config for a DBA/Optimizer, as SW can continue to work without it.

I disagree, since we have a problem https://github.com/chrislusf/seaweedfs/issues/1921
that is solved by this optimization, but in fact, only DBA can still fix

LazyDBA247-Anyvision:
@kmlebedev the syntax I added in scaffold.go, should work in most of the mysql/postgres deployment. so maybe the only change that is needed is to add to the exception/error in the insertEntry is to notify the user that a fix can be done by using upsert and a good/working example exists in the scaffold?

on vanilla  mysql/postgres (last years versions) that example work.

chrislusf:
Or how about just give user the option to "enableUpsert", and turn on by default?

LazyDBA247-Anyvision:
@chrislusf that sounds a good option, but auto-enabled sounds risky to me. 
maybe we can add a "test/verify" the upsert command on startup?
and if no exception is being made?  something like 2x insert/upsert to "fake" path /test  (like /buckets  /queues ).
that will verify the DB support the syntax, with an exception in the log to update the syntax if not working...

@kmlebedev  that's what you wanted, yes?

chrislusf:
> maybe we can add a "test/verify" the upsert command on startup?

That is safe but not really necessary. Upsert is command and most mysql/postgres dialect implementations should already support them. With defaulting enableUpsert to true, filers will fail fast if filer store does not support it, and not cause any issue.

Aytuar:
[https://postgrespro.ru/docs/postgresql/9.6/sql-insert?lang=en
](url)
`
Tip
It is often preferable to use unique index inference rather than naming a constraint directly using ON CONFLICT ON CONSTRAINT constraint_name. Inference will continue to work correctly when the underlying index is replaced by another more or less equivalent index in an overlapping way, for example when using CREATE UNIQUE INDEX ... CONCURRENTLY before dropping the index being replaced.
`

PR #1948, opened 2021-03-29
add enableUpsert=true

LazyDBA247-Anyvision:
and rename config to upsertQuery

### DIFF
--- a/weed/command/scaffold.go
+++ b/weed/command/scaffold.go
@@ -120,10 +120,9 @@ connection_max_idle = 2
 connection_max_open = 100
 connection_max_lifetime_seconds = 0
 interpolateParams = false
-# Empty insertQuery will use the default insert query
-# example insertQuery can be for UPSERT syntax:
-# insertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
-insertQuery = ""
+# if insert/upsert failing, you can disable upsert or update query syntax to match your RDBMS syntax:
+upsertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
+enableUpsert = true
 
 [mysql2]  # or memsql, tidb
 enabled = false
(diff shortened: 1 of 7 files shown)


weed/command/scaffold.go
```diff
@@ -120,10 +120,9 @@ connection_max_idle = 2
 connection_max_open = 100
 connection_max_lifetime_seconds = 0
 interpolateParams = false
-# Empty insertQuery will use the default insert query
-# example insertQuery can be for UPSERT syntax:
-# insertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
-insertQuery = ""
+# if insert/upsert failing, you can disable upsert or update query syntax to match your RDBMS syntax:
+upsertQuery = """INSERT INTO ` + "`%s`" + ` (dirhash,name,directory,meta) VALUES(?,?,?,?) ON DUPLICATE KEY UPDATE meta = VALUES(meta)"""
```
chrislusf, line 124: How about hide this customizable "upsertQuery" from the scaffold.go? This should be rarely needed, if our assumptions are correct. Adding two configuration options here for one feature seems confusing. We can put it to wiki if this is ever needed.

LazyDBA247-Anyvision: but we want to give the users to be able to customize the query, so how can they if we will remove it?

I would then remove the enableUpsert, and only use the upsertQuery, and setting it empty disables it.

ddorian: I believe you should just use upsert by default and remove insert in postgresql, there's no point keeping both.

LazyDBA247-Anyvision: not all PostgreSQL variants use the same SQL dialect. we need to have backward compatibility and the ability to update the query.

ddorian: @LazyDBA247-Anyvision please explain which variants ?

LazyDBA247-Anyvision: simple example:
cockroach db support:
```UPSERT INTO t (a, b, c) VALUES (1, 2, 3);```
but postgresql not.
postgresql and cockroach support:
```INSERT INTO t (a, b, c)
    VALUES (1, 2, 3)
    ON CONFLICT (a, b)
    DO UPDATE SET c = excluded.c; 
```
but cockroachdb not support:

```INSERT INTO t (a, b, c)
    VALUES (1, 2, 3)
    ON CONFLICT ON CONSTRAINT t_pkey
    DO UPDATE SET c = excluded.c;
 ```
or any unique index constraint, only on pkey, so not all RDBMS created equal or support same syntax, I am sure not going to cripple SW ability to support many compatible PG/MYSQL drivers with hardcoding "special" upsert syntax.
we must support customizing the upsert query, and even fall backing to regular insert command.

ddorian: >  UPSERT INTO t (a, b, c) VALUES (1, 2, 3);
 
This looks like a shortcut, no new functionality. No need to use this at all.

> or any unique index constraint, only on pkey, so not all RDBMS created equal or support same syntax, I am sure not going to cripple SW ability to support many compatible PG/MYSQL drivers with hardcoding "special" upsert syntax.
we must support customizing the upsert query, and even fall backing to regular insert command.

The current table only has the primary key and no other ones, so no need for this syntax either. Anyway.

LazyDBA247-Anyvision: I know :)
but just to show that not all support the same syntax.
need to keep the customization, I don't have a problem to have "default" syntax, but need to keep the option to change it.

PR #1975, opened 2021-04-08
IAM handlers

kmlebedev:
add DeleteUserPolicy


kmlebedev:
I will add more as needed, do not merge yet

kmlebedev:
@chrislusf That's all for now

PR #1976, opened 2021-04-08
fix entrypoint.sh and seaweedfs-compose.yml

stallion2303:
entrypoint.sh
- Default arguments are passed through the environment
- Added -filer argument to crontab
- Indent 2 spaces
seaweedfs-compose.yml
- Added default values for environment variables

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -1,73 +1,51 @@
 #!/bin/sh
 
-isArgPassed() {
-  arg="$1"
-  argWithEqualSign="$1="
-  shift
-  while [ $# -gt 0 ]; do
-    passedArg="$1"
-    shift
-    case $passedArg in
-    $arg)
-      return 0
-      ;;
-    $argWithEqualSign*)
-      return 0
-      ;;
-    esac
-  done
-  return 1
-}
+WEED=${WEED-"/usr/bin/weed"}
+WEED_DIR=${WEED_DIR-"/data"}
+WEED_VOLUME_PREALLOCATE=${WEED_VOLUME_PREALLOCATE-"-volumePreallocate"}
+WEED_MASTER_VOLUME_PREALLOCATE=${WEED_MASTER_VOLUME_PREALLOCATE-"-master.volumePreallocate"}
+WEED_VOLUME_SIZE_LIMIT_MB=${WEED_VOLUME_SIZE_LIMIT_MB-1024}
+WEED_MAX=${WEED_MAX-0}
+WEED_S3_DOMAIN_NAME=${WEED_S3_DOMAIN_NAME-""}
+WEED_S3_KEY_FILE=${WEED_S3_KEY_FILE-""}
+WEED_S3_CERT_FILE=${WEED_S3_CERT_FILE-""}
+
+WEED_MASTER=${WEED_MASTER-"localhost:9333"}
+WEED_FILER=${WEED_FILER-"localhost:8888"}
+FIX_REPLICATION_CRON_SCHEDULE=${FIX_REPLICATION_CRON_SCHEDULE-"*/7 * * * * *"}
+BALANCING_CRON_SCHEDULE=${BALANCING_CRON_SCHEDULE-"25 * * * * *"}
 
 case "$1" in
-
   'master')
-  	ARGS="-mdir=/data -volumePreallocate -volumeSizeLimitMB=1024"
-  	shift
-  	exec /usr/bin/weed master $ARGS $@
-	;;
-
+    ARGS="-mdir=$WEED_DIR $WEED_VOLUME_PREALLOCATE -volumeSizeLimitMB=$WEED_VOLUME_SIZE_LIMIT_MB"
+    shift
+    exec $WEED master $ARGS $@
+    ;;
   'volume')
-  	ARGS="-dir=/data -max=0"
-  	if isArgPassed "-max" "$@"; then
-  	  ARGS="-dir=/data"
-  	fi
-  	shift
-  	exec /usr/bin/weed volume $ARGS $@
-	;;
-
+    ARGS="-dir=$WEED_DIR -max=$WEED_MAX"
+    shift
+    exec $WEED volume $ARGS $@
+    ;;
   'server')
-  	ARGS="-dir=/data -volume.max=0 -master.volumePreallocate -master.volumeSizeLimitMB=1024"
-  	if isArgPassed "-volume.max" "$@"; then
-  	  ARGS="-dir=/data -master.volumePreallocate -master.volumeSizeLimitMB=1024"
-  	fi
- 	shift
-  	exec /usr/bin/weed server $ARGS $@
-  	;;
-
-  'filer')
-  	ARGS=""
-  	shift
-  	exec /usr/bin/weed filer $ARGS $@
-	;;
-
+    ARGS="-dir=$WEED_DIR -volume.max=$WEED_MAX $WEED_MASTER_VOLUME_PREALLOCATE -master.volumeSizeLimitMB=$WEED_VOLUME_SIZE_LIMIT_MB"
+    shift
+    exec $WEED server $ARGS $@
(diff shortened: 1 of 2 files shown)


docker/entrypoint.sh
```diff
@@ -1,73 +1,51 @@
...
   'server')
-  	ARGS="-dir=/data -volume.max=0 -master.volumePreallocate -master.volumeSizeLimitMB=1024"
-  	if isArgPassed "-volume.max" "$@"; then
-  	  ARGS="-dir=/data -master.volumePreallocate -master.volumeSizeLimitMB=1024"
-  	fi
- 	shift
-  	exec /usr/bin/weed server $ARGS $@
-  	;;
-
-  'filer')
```
chrislusf, line 48: "filer" is missing?

stallion2303: “filer” had no arguments, so I decided to delete it because will be executed on line "*)"

chrislusf: Got it. But I think it is better to leave it there to keep the option visible, easier to add other arguments later.

Suika:
As I'm already on this topic. The original `entrypoint.sh` is "kinda" strange and this PR makes it more of a pain than anything. Reason being as follows.

The original `entrypoint.sh` forces you to use certain flags that you can't remove. The example compose file implies that you would start the contaiener as you would weed on a normal machine.
i.e. `weed server OPTIONS` creating an illusion that it is 1-to-1, as if you were starting weed from the console. Which is actually not completely true.
The `entrypoint.sh` containers has a case for each master/volume/server/etc and each section has it's own pre-defined arguments, which will be preappeneded to whatever you write into the command.
i.e. `master -ip=seaweed-master -volumeSizeLimitMB=20000` would actually become `master -mdir=/data -volumePreallocate -volumeSizeLimitMB=1024 -ip=seaweed-master -volumeSizeLimitMB=20000`.
e932262b070ddcd426f98a2599fad0eec0972c37 was made so at least some arguments would be overwritten.

This PR is kinda starange that it starts mixing the environment variables with the commands in the `entrypoint.sh`.
On one side I get it that it would be really nice to have ENV for everything and simply be able to set them for every container and ignore the command line.
On the other side, it's a double effort because you still need to define the command in the compose. 
You need to figure out the ENV variables that are used for default values, which also should not clash with toml variables.
And should you want extra options, you will need to define them in the command. So in the end instead of one place, you start juggling with two places that need to be maintained.

Personally I find it nice if it is possible to configure everything using the ENV variables. Even better would be if all the arguments being set via the CLI would also be configurable via the `.toml`, then the whole step of building an intricate entrypoint would be avoided and whatever is passed via CLI, would override the `.toml` configuration.
But this kindal halfway mix of both ENV and commands is for me personally strange.

stallion2303:
Then I propose a scheme:

cat seaweedfs-compose.yml
services:
  master:
    image: chrislusf/seaweedfs # use a remote image
    ports:
      - 9333:9333
      - 19333:19333
    environment:
      WD_COMMAND: "server"
      WD_DIR: "/data"
      WD_VOLUME_MAX: 0
      WD_MASTER_VOLUMEPREALLOCATE:
      WD_MASTER_VOLUMESIZELIMITMB: 1024

entrypoint.sh like that
env -u WD_COMMAND \
 | sed '/^WD_/!d;s/^WD_//;s/_/./g;s/=$//' > weed.options
 
cat weed.options
MASTER.VOLUMESIZELIMITMB=1024
MASTER.VOLUMEPREALLOCATE
VOLUME.MAX=0
DIR=/data

exec weed $WD_COMMAND -options=weed.options

chrislusf:
> MASTER.VOLUMESIZELIMITMB=1024
> MASTER.VOLUMEPREALLOCATE
> VOLUME.MAX=0
> DIR=/data
> ```

the options are just same as CLI options and are case sensitive.
env variables are not case sensitive.

stallion2303:
Writing variables in the same case
```
    environment:
      WD_COMMAND: "server"
      WD_dir: "/data"
      WD_volume_max: 0
      WD_master_volumePreallocate: ""
      WD_master_volumeSizeLimitMB: 1024
```
cat weed.options
```
dir=/data
master.volumePreallocate
master.volumeSizeLimitMB=1024
volume.max=0
```

chrislusf:
env variables are all upper case.

stallion2303:
for fun
```
env -u WD_COMMAND \
 | sed '/^WD_/!d;s/^WD_//;s/_/./g;s/=$//' \
 | awk '{print tolower($0)}' \
 | sed 's/volumesizelimitmb/volumeSizeLimitMB/' \
 | sed 's/volumepreallocate/volumePreallocate/' > weed.options
```

Suika:
> 
> 
> > MASTER.VOLUMESIZELIMITMB=1024
> > MASTER.VOLUMEPREALLOCATE
> > VOLUME.MAX=0
> > DIR=/data
> 
> the `-options` file should have the same option names as CLI, which are case sensitive.
> env variables are not case sensitive.

>  env variables are not case sensitive.

Yeah no. Bash and sh variables are case sensitive. Same goes for docker ENV Variables, they are too case sensitive. Dunno about k8s and podman, but can't find the specific statement in their docs.
Unless I misunderstood as to what "env variables are not case sensitive" is referring to.

chrislusf:
sorry, wrong memory. Env variable dos not support "." dot.

stallion2303:
As a result, I came to.
If necessary, I can make a pull-request. If the solution is not interesting, then you can close it.

**entrypoint.sh**
```
#!/bin/sh

WD=${WD-"/usr/bin/weed"}
OPTIONS=${OPTIONS-"weed.options"}

WD_COMMAND=${WD_COMMAND-"help"}
WD_dir=${WD_DIR-"/data"}
WD_volumePreallocate=${WD_volumePreallocate-""}
WD_master_volumePreallocate=${WD_master_volumePreallocate-""}
WD_volumeSizeLimitMB=${WD_volumeSizeLimitMB-1024}
WD_master_volumeSizeLimitMB=${WD_master_volumeSizeLimitMB-1024}
WD_volume_max=${WD_volume_max-0}

WD_MASTER=${WD_MASTER-"localhost:9333"}
WD_FILER=${WD_FILER-"localhost:8888"}
FIX_REPLICATION_CRON_SCHEDULE=${FIX_REPLICATION_CRON_SCHEDULE-"*/7 * * * * *"}
BALANCING_CRON_SCHEDULE=${BALANCING_CRON_SCHEDULE-"25 * * * * *"}

case "$WD_COMMAND" in
  'cronjob')
    echo "Running Crontab:"
    cat <<-EOF | tee /crontab
	$FIX_REPLICATION_CRON_SCHEDULE echo "lock; volume.fix.replication; unlock" | $WD shell -master=$WD_MASTER -filer=$WD_FILER
	$BALANCING_CRON_SCHEDULE echo "lock; volume.balance -collection ALL_COLLECTIONS -force; unlock" | $WD shell -master=$WD_MASTER -filer=$WD_FILER
	EOF
    exec supercronic /crontab
    ;;
  *)
    env -u WD_COMMAND \
     | sed '/^WD_/!d;s/^WD_//;s/_/./g;s/=$//' > $OPTIONS

    if [ -s "$OPTIONS" ]; then
      exec $WD $WD_COMMAND -options=$OPTIONS
    else
      exec $WD $WD_COMMAND
    fi
    ;;
esac
```

**seaweedfs-compose.yml**
```
version: '2'

services:
  master:
    image: chrislusf/seaweedfs # use a remote image
    ports:
      - 9333:9333
      - 19333:19333
    environment:
      WD_COMMAND: "master"
      WD_dir: "/data"
      WD_volumePreallocate: ""
      WD_volumeSizeLimitMB: 1024
      WD_ip: "master"
...
```

Need to change the "latest" tag to support the -options

chrislusf:
There are 4 places to configure one thing: CLI, entry point hard coded values, entry point scripts, options.

Let's stick to the first 2 places, which in a sense is already confusing.

PR #1981, opened 2021-04-09
Fix If-Modified-Since behavior

mirdhyn:
- `s3api`: passthrough 304 response code from filer, even if `Content-Lenght` is not present
  this bug was leading to 404 responses when passing expired If-Modified-Since   
  
- `filer`: return 304 when `If-Modified-Since` **equal** `Last-Modified`, not only "later"
  as browser saves the time that server returns (using Last-Modified) and uses it with If-Modified-Since on next call
  (note that `volume` handler already behaves this way: https://github.com/chrislusf/seaweedfs/blob/master/weed/server/volume_server_handlers_read.go#L115 )

### DIFF
--- a/weed/server/filer_server_handlers_read.go
+++ b/weed/server/filer_server_handlers_read.go
@@ -79,7 +79,7 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request, 
 		w.Header().Set("Last-Modified", entry.Attr.Mtime.UTC().Format(http.TimeFormat))
 		if r.Header.Get("If-Modified-Since") != "" {
 			if t, parseError := time.Parse(http.TimeFormat, r.Header.Get("If-Modified-Since")); parseError == nil {
-				if t.After(entry.Attr.Mtime) {
+				if t.After(entry.Attr.Mtime) || t.Equal(entry.Attr.Mtime) {
 					w.WriteHeader(http.StatusNotModified)
 					return
 				}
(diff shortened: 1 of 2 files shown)


kmlebedev:
@mirdhyn  Can you write tests for this?

mirdhyn:
sure, I'll add them shortly.

weed/server/filer_server_handlers_read.go
```diff
@@ -79,7 +79,7 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request,
 		w.Header().Set("Last-Modified", entry.Attr.Mtime.UTC().Format(http.TimeFormat))
 		if r.Header.Get("If-Modified-Since") != "" {
 			if t, parseError := time.Parse(http.TimeFormat, r.Header.Get("If-Modified-Since")); parseError == nil {
-				if t.After(entry.Attr.Mtime) {
+				if t.After(entry.Attr.Mtime) || t.Equal(entry.Attr.Mtime) {
```
chrislusf, line 82: "not before" is simpler

chrislusf:
I will merge this first.

mirdhyn:
Thank you @chrislusf 
@kmlebedev sorry writting tests for that is harder than I expected, especially that there are no similar existing tests that I could rely on :) I will push tests once I figured out how to wire out everything properly.

kmlebedev:
> Thank you @chrislusf
> @kmlebedev sorry writting tests for that is harder than I expected, especially that there are no similar existing tests that I could base on :) I will push tests once I figured out how to wire out everything properly.

@mirdhyn  This is normal, since it is not possible to do such tests right away without rewriting functions into interfaces.

PR #2025, opened 2021-04-26
improvement for minFreeSpace argument

bingoohuang:
minFreeSpace support:

1. 1 to be as 1% percent value (digits that <= 100)
2. human readable size like 10G, 10GiB and etc.


chrislusf:
Need to keep the parameters consistent with the previous versions as much as possible.

weed/storage/disk_location.go
```diff
@@ -23,9 +23,10 @@ type DiskLocation struct {
 	DiskType               types.DiskType
 	MaxVolumeCount         int
 	OriginalMaxVolumeCount int
-	MinFreeSpacePercent    float32
-	volumes                map[needle.VolumeId]*Volume
-	volumesLock            sync.RWMutex
+	// MinFreeSpace limits the minimum free space (<=100 as percentage, > 100 as bytes)
+	MinFreeSpace float32
```
chrislusf: This is confusing. 

How about creating a type for this, e.g. MinimumFreeSpaceType, and put these logic inside, such as EmptyTo, ParseMinFreeSpace, etc.

bingoohuang: good idea

bingoohuang: code has been refactored to create a MinFreeSpace type to separate relative logics like parsing and judging.

weed/util/bytes.go
```diff
@@ -161,3 +165,94 @@ func NewBytesReader(b []byte) *BytesReader {
 		Reader: bytes.NewReader(b),
 	}
 }
+
+// EmptyTo returns to if s is empty.
+func EmptyTo(s, to string) string {
```
chrislusf, line 170: please move the 2 new functions to the new minfreespace.go file

PR #2030, opened 2021-04-27
auth use bucket wild cards

kmlebedev:


chrislusf:
Add some wiki doc for this?

PR #2035, opened 2021-04-28
fix aws style Etag for chunks

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2007

### DIFF
--- a/weed/filer/filechunks.go
+++ b/weed/filer/filechunks.go
@@ -2,7 +2,6 @@ package filer
 
 import (
 	"bytes"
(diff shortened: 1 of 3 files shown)


chrislusf:
Very nice idea!!! Thanks!

qieqieplus:
md5 is an iterative alogrithm, I don't think there's a way to get hashes for chunks and combined file at once. 
You need to Reset() md5 digest for every chunk in order to get correct checksum. But in that case you need another round to compute hash for the combined file.

PR #2045, opened 2021-05-05
fix concurrent vacuum & delete panic

qieqieplus:
When delete volume during compact, something like this could happen:
1. `CommitCompactVolume` is done
2. `DeleteVolume` triggered right after commit, holds `volumesLock`
3. `GetVolume` RLock() waits for `volumesLock` Unlock(), after that it returns nil
4. `IsReadOnly` panic for nil pointer dereference


qieqieplus:
I forgot to save crash stack trace, but it did happened a moment ago.

PR #2049, opened 2021-05-06
ahead of time volume assignment

qieqieplus:


weed/topology/volume_layout.go
```diff
@@ -324,11 +326,15 @@ func (vl *VolumeLayout) GetActiveVolumeCount(option *VolumeGrowOption) int {
 				if option.DataNode != "" && dn.Id() != NodeId(option.DataNode) {
 					continue
 				}
-				counter++
+				active++
+				info, _ := dn.GetVolumesById(v)
+				if float64(info.Size) > float64(vl.volumeSizeLimit)*option.Threshold() {
+					crowded++
+				}
```
qieqieplus, line 333: Not so sure if it's correct. @chrislusf PTAL thx

PR #2074, opened 2021-05-20
avoid http error: superfluous response.WriteHeader

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2046


weed/server/filer_server_handlers_read.go
```diff
@@ -131,6 +131,9 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request,
 
 	if r.Method == "HEAD" {
 		w.Header().Set("Content-Length", strconv.FormatInt(totalSize, 10))
+		processRangeRequest(r, w, totalSize, mimeType, func(writer io.Writer, offset int64, size int64) error {
```
chrislusf: The clients sending HEAD requests not necessarily want to check the chunks.  How to get around this?

kmlebedev: > The clients sending HEAD requests not necessarily want to check the chunks. How to get around this?

I don't understand, why then send a HEAD request? and why run out of it?
Since this is the only mechanism to understand before downloading a file in a stream, is everything good with it?

kmlebedev: Here you need to download a 100 gig file and 10500 chunks, and here it was loaded for 10 hours, and on 10499 chunk an error occurred due to the absence of a chunk, and at the same time a HEAD request that checks for the presence of only the first chunk works, why is such a check from a single chunk needed?

chrislusf: HEAD request may just want to get the file size, or modified time.

kmlebedev: > HEAD request may just want to get the file size, or modified time.

Yes, this is often needed for synchronization and it is better to check the liveness and return an error on HEAD than to return 200 and a broken file during synchronization

chrislusf: checking file health is different from checking file size. The file chunks may be offline for some reason. But it should not fail the HEAD operation.

kmlebedev: > checking file health is different from checking file size. The file chunks may be offline for some reason. But it should not fail the HEAD operation.

Well i realized

weed/server/volume_server_handlers_read.go
```diff
@@ -261,13 +261,10 @@ func writeResponseContent(filename, mimeType string, rs io.ReadSeeker, w http.Re
...
 	}
 
-	processRangeRequest(r, w, totalSize, mimeType, func(writer io.Writer, offset int64, size int64, httpStatusCode int) error {
+	processRangeRequest(r, w, totalSize, mimeType, func(writer io.Writer, offset int64, size int64) error {
 		if _, e = rs.Seek(offset, 0); e != nil {
 			return e
 		}
-		if httpStatusCode != 0 {
-			w.WriteHeader(httpStatusCode)
-		}
```
chrislusf, line 270: This PR focuses on stats. I added these code in this this and a few related places. I am not sure whether this is removed by mistake.

kmlebedev: ok

PR #2093, opened 2021-05-27
put-object and copy-object metadata patch

bennykwok:
put-object / copy-object metadata seem work after adding touch().
But I am not sure this method is correct or not.

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -102,6 +102,18 @@ func (s3a *S3ApiServer) PutObjectHandler(w http.ResponseWriter, r *http.Request)
 			return
 		}
 
+                // PATCH 
+                newEntry, newEntryErr := s3a.getEntry(s3a.option.BucketsPath, bucket+object)
+
+                if newEntryErr == nil {
+                   newEntry.Extended = weed_server.SaveAmzMetaData(r, newEntry.Extended, true)
+
+                   touchErr := s3a.touch(s3a.option.BucketsPath+"/"+bucket, object, newEntry)
+                   if touchErr != nil {
+                        glog.V(0).Infof("error: %v ", touchErr)
+                   }
+                }
+
 		setEtag(w, etag)
 	}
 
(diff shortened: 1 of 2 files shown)


weed/s3api/s3api_object_handlers.go
```diff
@@ -102,6 +102,18 @@ func (s3a *S3ApiServer) PutObjectHandler(w http.ResponseWriter, r *http.Request)
 			return
 		}
 
+                // PATCH 
+                newEntry, newEntryErr := s3a.getEntry(s3a.option.BucketsPath, bucket+object)
+
+                if newEntryErr == nil {
+                   newEntry.Extended = weed_server.SaveAmzMetaData(r, newEntry.Extended, true)
+
+                   touchErr := s3a.touch(s3a.option.BucketsPath+"/"+bucket, object, newEntry)
```
chrislusf: this would cause all uploads one extra "touch" operation.

bennykwok: um...you are right 
May be adding "len(newEntry.Extended) > 0" for checking?

weed/s3api/s3api_object_handlers.go 
                // PATCH
                newEntry, newEntryErr := s3a.getEntry(s3a.option.BucketsPath, bucket+object)

                if newEntryErr == nil {
                   newEntry.Extended = weed_server.SaveAmzMetaData(r, newEntry.Extended, true)

                   if len(newEntry.Extended) > 0 {
                      touchErr := s3a.touch(s3a.option.BucketsPath+"/"+bucket, object, newEntry)
                      if touchErr != nil {
                         glog.V(0).Infof("error: %v ", touchErr)
                      }
                   }
                }


Does s3a.getEntry case extra read ? May be the logic should put into putToFiler? If the AmzMetaData header len() > 0 then write the header into db during putToFiler.

I am new seaweedfs user, not really know about seaweedfs structure.

chrislusf:
Not needed with fix fb8036385ab71c8277a554aff06f937f2e5ab6d6

PR #2085, opened 2021-05-26
create fuse subcommand to use weed with mount

danielflira:
It add fuse subcommand that allow use weed as mount type.

It's related with issues: #1323 #1674

PR #2099, opened 2021-05-29
fix parameter multiple values

danielflira:
This PR change way that mount parameters are parsed allowing multi value parameters. Example:

mount -t weed fuse /mnt -o "filer='192.168.0.1:8888,192.168.0.2:8888',filer.path=/"

It is a pending feature from PR #2085.


chrislusf:
Thanks! Also add this example to the wiki?

PR #2105, opened 2021-06-02
Return artificial . and .. directories

Woellchen:
this adds support for `.` and `..` directories in fuse mounts. they properly point to the correct inodes of directories of themselves (`.`) and to their parents (`..`), also for the mount root + parent.
this especially solves some errors on linux. on mac it works already out of the box as either macos or osxfuse apparently adds those directories automatically.
this is tested on macos and linux (debian). i don't know if it makes sense to test this on other platforms too. maybe windows/freebsd?
i am happy to hear your thoughts :)

### DIFF
--- a/weed/command/mount_std.go
+++ b/weed/command/mount_std.go
@@ -5,14 +5,17 @@ package command
 import (
 	"context"
 	"fmt"
-	"github.com/chrislusf/seaweedfs/weed/storage/types"
 	"os"
 	"os/user"
 	"path"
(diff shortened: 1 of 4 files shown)


chrislusf:
For bazil.org/fuse, the mount root inode is hard coded to 1. Changing that would cause leaks.

Woellchen:
i have not changed this bevahior, the root inode is still 1

chrislusf:
Thanks!

PR #2127, opened 2021-06-15
Check for history file errors as well

nagy:
Check for errors when writing the shell history.


chrislusf:
thanks! did it cause some problem?

nagy:
No, I just spotted it while looking through the code.

PR #2130, opened 2021-06-16
avoid exit in case of errors on single volumes

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2129

### DIFF
--- a/weed/shell/command_volume_check_disk.go
+++ b/weed/shell/command_volume_check_disk.go
@@ -71,6 +71,7 @@ func (c *commandVolumeCheckDisk) Do(args []string, commandEnv *CommandEnv, write
 	defer aDB.Close()
 	defer bDB.Close()
 
+REPLICAS:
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_check_disk.go
```diff
@@ -96,18 +97,22 @@ func (c *commandVolumeCheckDisk) Do(args []string, commandEnv *CommandEnv, write
 
 			// read index db
 			if err := c.readIndexDatabase(aDB, a.info.Collection, a.info.Id, a.location.dataNode.Id, *verbose, writer); err != nil {
-				return err
+				fmt.Fprintln(writer, err.Error())
+				continue REPLICAS
```
chrislusf, line 101: How to avoid this kind of jumps?

jonaz: Extract to function and use return? :)

PR #2136, opened 2021-06-18
Update filer_server_handlers_write_autochunk.go

JustTestCode:
add filer post file out fid
Sometimes we need filer server for file management, and we also need fid for direct access and bypass the filer server, I think fid output is necessary


chrislusf:
Concatenating file ids into the Fid field is too hacky.

The file ids are not always the file content. For small files, the file content may be stored in filer store directly. For large files, the chunk can store file ids also. So exposing the storage detail is not ideal and hard to keep the API consistent.

Issue #2159, opened 2021-06-26
You should add missing `attr.Env = os.Environ()`


thephoenixofthevoid:
Here you asked why env variables are lost:
https://github.com/chrislusf/seaweedfs/blob/1cac2f2278b2344ac41ac882c7b21593f212d3bd/weed/command/fuse.go#L214

The issue must be due to missing:
```
attr.Env = os.Environ()
```
here https://github.com/chrislusf/seaweedfs/blob/1cac2f2278b2344ac41ac882c7b21593f212d3bd/weed/command/fuse.go#L198

Another way to do this is to explicitly set Env to `nil` 

It seems all env. vars are lost.

Need to delete the workaround if everything works:
https://github.com/chrislusf/seaweedfs/blob/1cac2f2278b2344ac41ac882c7b21593f212d3bd/weed/command/fuse.go#L215-217

chrislusf:
@thephoenixofthevoid send a PR?

cc @danielflira

danielflira:
The condition about environment was wrote before the condition about master process. I don't know if it is related with my system (ubuntu 20.04) but without this validation I got this error:

```
root@macubuntu:/tmp# mount -t weed fuse /mnt -o child
I0627 00:36:51 80808 mount_std.go:236] mount: fusermount: exec: "fusermount": executable file not found in $PATH
```
The parameter -o child prevent StartProcess, and run weedfuse in foreground.

If I remember well about this, mount command runs weedfuse  without full PATH variable,  when weedfuse command runs fusermount an "command not found" error happens.

danielflira:
I faked command /sbin/fake to use with mount:

```python3
#!/usr/bin/env python3

import os
import sys

print(os.environ.get("PATH", "environment var PATH does not exists"))
input()

sys.exit(0)
```
Running mount with fake command:

```sh
root@macubuntu:/tmp# mount -t fuse.fake fuse /mnt -o child
environment var PATH does not exists
```
The faked command with Bash:

```sh
#!/bin/bash

echo "${PATH}"
read

exit 0
```

Result:

```sh
root@macubuntu:/tmp# mount -t fuse.fake fuse /mnt -o child
/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
```

I think that bash set PATH variable on his own.

thephoenixofthevoid:
[...]

PR #2160, opened 2021-06-26
Fix: Loosing environment variables at StartProcess

thephoenixofthevoid:
Related to #2159

Let's continue discussions here

### DIFF
--- a/weed/command/fuse.go
+++ b/weed/command/fuse.go
@@ -195,7 +195,9 @@ func runFuse(cmd *Command, args []string) bool {
 		arg0 := os.Args[0]
 		argv := append(os.Args, "-o", "child")
 
-		attr := os.ProcAttr{}
+		attr := os.ProcAttr{}		
+		attr.Env = os.Environ() 
+		
 		child, err := os.StartProcess(arg0, argv, &attr)
 
 		if err != nil {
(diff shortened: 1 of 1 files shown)


weed/command/fuse.go
```diff
@@ -212,9 +219,9 @@ func runFuse(cmd *Command, args []string) bool {
 	}
 
 	// I don't know why PATH environment variable is lost
```
chrislusf: clean up the comments?

weed/command/fuse.go
```diff
@@ -195,7 +195,14 @@ func runFuse(cmd *Command, args []string) bool {
 		arg0 := os.Args[0]
 		argv := append(os.Args, "-o", "child")
 
-		attr := os.ProcAttr{}
+		attr := os.ProcAttr{}		
+		attr.Env = os.Environ() 
+		// By default attr.Env is [] ie empty set of env vars
+		// not nil and not os.Environ()
+		
+		//attr.Dir = "."
```
chrislusf: Are these needed? If not, please clean up.

thephoenixofthevoid: Including line 204 or not? Line 204 allows for enabling output from child process to check what envs are passed (literally none (zero) were passed). I will leave it here:
```
attr.Files = []*os.File{ nil, os.Stdout, os.Stderr }
```
and remove in the code.

thephoenixofthevoid: Also will remove these:
```
// I don't know why PATH environment variable is lost
// if err := os.Setenv("PATH", "/bin:/sbin:/usr/bin:/usr/sbin:/usr/local/bin:/usr/local/sbin"); err != nil {
//	panic(fmt.Errorf("setenv: %s", err))
// }
```

PR #2181, opened 2021-07-05
extract embed toml example to separate files

bingoohuang:
example toml files can be edited in favor of separate file with syntax highlighting.


bingoohuang:
this PR is created by the latest master code to separate embed scaffold sample configuration files.

PR #2184, opened 2021-07-05
Change java client MIN_BUFFER_SIZE to 1MB

jbarotin:
I propose to reduce min buffer size to 1MB for java client.


chrislusf:
Context:

In the java client, I noticed in the ByteBufferPool
 object that there is a minimum buffer size of 8MB, why it's not possible to reduce size of th buffer ?
I'm working with a lot of smalls files (about 10 KB) in parallel, It makes my application "Out Of Memory" with 4G RAM allocated in xmx with G1GC.

PR #2185, opened 2021-07-06
show RemoteVolumes/EcVolumes only if it is not empty

bingoohuang:
make volume ui more compact


chrislusf:
I think this PR is OK, and I do not want to waste your effort. But later please discuss about it if you want to make such changes.

It is better to show the page framework, rather than let new users guess where the content is for other volumes.

PR #2198, opened 2021-07-16
cassandra: Use LocalOne instead of One consistency

nivekuil:
This should nearly always be better

### DIFF
--- a/weed/filer/cassandra/cassandra_store.go
+++ b/weed/filer/cassandra/cassandra_store.go
@@ -124,7 +124,7 @@ func (store *CassandraStore) FindEntry(ctx context.Context, fullpath util.FullPa
 	var data []byte
 	if err := store.session.Query(
 		"SELECT meta FROM filemeta WHERE directory=? AND name=?",
-		dir, name).Consistency(gocql.One).Scan(&data); err != nil {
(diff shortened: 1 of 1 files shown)


chrislusf:
How about making it configurable?

nivekuil:
I don't really see the point -- when would you ever want to do cross-dc reads?  Should the default LocalQuorum also be configurable? The java driver has a lot of options that can be configured through a file (reference.conf), so maybe that could work if the go driver supported it.

nivekuil:
Not sure if this is what you meant, but it would make a lot of sense to be able to toggle between LOCAL_ONE and LOCAL_QUORUM on reads. Currently the behavior is eventually consistent and read repairs are never run (unless you have probabilistic repairs on, which at least scylla has deprecated). With LOCAL_QUORUM the filer store will be strongly consistent, which seems like maybe a better default.

chrislusf:
Usually when read with LOCAL_ONE, the writes should be ALL.

When read with LOCAL_QUORUM, the writes can be LOCAL_QUORUM.

Current LOCAL_ONE is not correct, since the write is LOCAL_QUORUM.

nivekuil:
Should we just let all queries LOCAL_QUORUM then, with no optional for eventual consistency? It'll be a big change for users but is probably the best default.

chrislusf:
yes

PR #2200, opened 2021-07-19
weed shell: fs.rm remove file or directory

combineads:
Add rm.fs command.

### DIFF
--- a/weed/shell/command_fs_rm.go
+++ b/weed/shell/command_fs_rm.go
@@ -1,0 +1,67 @@
+package shell
+
+import (
+	"context"
+	"fmt"
+	"io"
+
+	"github.com/chrislusf/seaweedfs/weed/pb/filer_pb"
+	"github.com/chrislusf/seaweedfs/weed/util"
+)
+
+func init() {
+	Commands = append(Commands, &commandFsRm{})
+}
+
+type commandFsRm struct {
+}
+
+func (c *commandFsRm) Name() string {
+	return "fs.rm"
+}
+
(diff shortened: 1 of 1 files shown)


weed/shell/command_fs_rm.go
```diff
@@ -0,0 +1,67 @@
...
+		request := &filer_pb.DeleteEntryRequest{
+			Directory:            targetDir,
+			Name:                 targetName,
+			IgnoreRecursiveError: true,
+			IsDeleteData:         true,
+			IsRecursive:          true,
+			IsFromOtherCluster:   false,
+			Signatures:           nil,
+		}
+		_, err = client.DeleteEntry(context.Background(), request)
```
chrislusf, line 57: this just delete the one file or directory meta data.

It does not delete the file chunks, and it does not recursively delete directory children.

combineads: If I use IsDeleteData: true, IsRecursive: true , wouldn't data and subdirectories be deleted?

chrislusf: sorry, you are correct and my memory is wrong.

chrislusf:
Merged. Please add some checking to see whether the to-be-removed item exists or not. Here is what I did.

```
> fs.rm -h
remove: /-h
```

chrislusf:
Also, adding some support to differentiate directory?
```
fs.rm -r dir
```

combineads:
OK. I'll add -r and -f options.

PR #2217, opened 2021-07-24
Add the webp image type

combineads:
Add the webp image type without encoding


chrislusf:
There was a similar PR for webp before, but was rolled back because it was not compiling on some OS.

Could you please check that?

combineads:
This is a library developed with go, so there seems to be no problem. I've tried this on Mac and Linux.

Unfortunately, this library doesn't include an encoder, and other libraries that use libwebp were not used because there were difficult to install due to dependency issues.

combineads:
Please refer to https://cs.opensource.google/go/x/image/+/a66eb644:webp/decode.go.

PR #2266, opened 2021-08-17
Create command_volume_recycle.go

fc-coder:
Recycle files according to cluster usage


chrislusf:
1. Go fmt.
2. Should require locking, similar to https://github.com/chrislusf/seaweedfs/blob/master/weed/shell/command_volume_balance.go#L65

LazyDBA247-Anyvision:
@fc-coder any update? sound interesting...

kmlebedev:
> @fc-coder any update? sound interesting...

I don't understand what is the use?

kmlebedev:
@fc-coder ping

kmlebedev:
While the refactoring realized that this is a useless command, it tries to delete the tail of volumes of data files when the disk is full

PR #2288, opened 2021-08-28
feature(cross-platform): add multi-arch support to release

DragonStuff:
Currently, it looks like there aren't any builds for PPC or Z-kernel platforms. The project is also missing multiple arches from the release system.

Below are the missing platforms added as part of this PR:
- `arm64`
- `386`

Below are the missing platforms not added to this PR due to compatibility issues ([see here](https://github.com/chrislusf/seaweedfs/pull/2288/checks?sha=46e78e9b3ef7b3a48553411ff56ef2dcf54ea5b3)):
- `ppc64le`
- `s390x`

This MR allows the team to support future architectures with minimal work if/when they decide to.

Supported Go and Buildx architectures confirmed against:

- https://gist.github.com/asukakenji/f15ba7e588ac42795f421b48b8aede63#a-list-of-valid-goarch-values
- https://github.com/logband/github-buildx-stats/runs/3448399619?check_suite_focus=true

Changes:
- Added a new cross-platform binary test build to cover that if there are any issues in the future.
- Added the new architectures to the release workflow.
- Added the new architectures into the Dockerfile so they are supported once a new release is built. This includes architectures with compatibility issues, since they may be added in the future.
- The container builds (test / latest / release) use matrixes.

This should make development a lot easier for newcomers to the project! 🎉

### DIFF
--- a/.github/workflows/binary_test.yml
+++ b/.github/workflows/binary_test.yml
@@ -13,23 +13,15 @@ jobs:
     strategy:
       matrix:
         goos: [linux, windows, darwin, freebsd, netbsd, openbsd]
-        goarch: [amd64, arm, arm64, 386, ppc64le, s390x]
+        goarch: [amd64, arm, arm64, 386]
         exclude:
           - goarch: arm
             goos: darwin
           - goarch: 386
             goos: darwin
-          - goarch: ppc64le
-            goos: darwin
-          - goarch: s390x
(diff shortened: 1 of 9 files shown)


DragonStuff:
@chrislusf Ready for review! 🙇

This needs a new release to be created before the container build test will work, and the new platforms will be added.

PR #2292, opened 2021-08-29
fix(ci): update latest to only run once

DragonStuff:
Further update from #2291, apply the same logic to `latest` images.

Also includes:
- a quick fix to never push to Docker Hub from pull requests on test jobs
- a quick fix to never run the test jobs on anything other than `pull_request`

Test is fine to stay as-is since it's not pushing.


DragonStuff:
@chrislusf Ready for review. Also made a seperate PR #2293 for some CI job naming cleanup.

PR #2310, opened 2021-09-05
Detect ipv6 addresses

nivekuil:
This allows ipv6 addresses to be found for e.g. volume server IP.  Note that this may be a breaking change if the ipv6 came first in whatever order this is traversed. I think most setups would have the routable interfaces earlier in the iterator order, and any ipv6s are more likely to be routable than ipv4. It also might make sense to have additional checks for interface local/link local as documented here https://pkg.go.dev/net#IP.IsInterfaceLocalMulticast, though this patch at least works for me (pod has a routable ipv6 + a private, node-local ipv4)

### DIFF
--- a/weed/util/network.go
+++ b/weed/util/network.go
@@ -24,7 +24,7 @@ func DetectedHostAddress() string {
 
 		for _, a := range addrs {
 			if ipNet, ok := a.(*net.IPNet); ok && !ipNet.IP.IsLoopback() {
-				if ipNet.IP.To4() != nil {
+				if ipNet.IP.To4() != nil || ipNet.IP.To16() != nil {
 					return ipNet.IP.String()
 				}
 			}


nivekuil:
@chrislusf for the problem you ran into, it looks like it picked up a link local address (`fe80::/10`). I think it would make sense just to skip these explicitly, in the same way loopback is already skipped. Technically this is an ipv4 concept too (`169.254/16` range) but it's hardly ever used, only when the interface can get no other IP. Though I'm not sure why it would have bound to `fe80::/1`.. is that the `docker0` interface?

chrislusf:
Just on my laptop.

I fixed the related code a bit. Please help to check it since I don't have any IPv6 env.

nivekuil:
One thing I noticed is that the master ui seems to show a badly formatted host/port:

image

Seems to still work though

chrislusf:
do you mean the second url does not format well? It is from publicUrl, and is basically free text.

PR #2354, opened 2021-10-04
Add Redis3

chrislusf:
Redis3 directory size is not limited by the Redis per entry size limit.


banjiaojuhao:
Why not using zset to store entries instead of protobuf?
Inserting entries into zset avoids reading the whole directory from redis and write back them.
Splitting and merging zset can be done by [lua script in redis](https://redis.io/commands/eval) ([lua script in go-redis](https://redis.uptrace.dev/guide/lua-scripting.html#getting-started)), which is atomic and more performant.

chrislusf:
The code is using zest

PR #2374, opened 2021-10-12
Let volume server can use io_uring API

blacktear23:
Let weed volume server can use io_uring API. When io_uring api fails it will fallback to syscall. 

Note: This feature is experimental.

usage:

```
weed volume -iouring
```


weed/storage/backend/io_driver_others.go
```diff
@@ -0,0 +1,16 @@
...
+	"fmt"
+	"os"
+)
+
+var (
+	errUnsupportIODriver = fmt.Errorf("Unsupport IO driver")
+)
+
+func NewIOUringDriver(file *os.File) (IODriver, error) {
+	return nil, errUnsupportIODriver
```
chrislusf, line 15: Delegate to the file if not linux?

blacktear23: The IODriver create function will catch this error and fallback to SyscallIODriver.

blacktear23:
I just run a apache bench mark test for some large file such as 22MB file. And syscall will get average 771MB/s, iouring-go will get average 785MB/s.

```
ab -c 50 -n 100 -k http://127.0.0.1:8888/xxxxxx
```

qieqieplus:
IMHO without io-uring integration into the golang runtime, using io-uring is very likely to lead to performance degradation (unless users have access to go runtime scheduler).

PR #2379, opened 2021-10-13
S3 detect mime type

kmlebedev:
Detect mime type on before uploading to filer

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -102,7 +102,7 @@ func (s3a *S3ApiServer) PutObjectHandler(w http.ResponseWriter, r *http.Request)
 			return
 		}
 		setEtag(w, etag)
-		if mime != "" {
+		if mime != "" && mime != "application/octet-stream" {
 			if err := s3a.setMime(s3a.option.BucketsPath+"/"+bucket, object, mime); err != nil {
 				glog.Error(err)
 			}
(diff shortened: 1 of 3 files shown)


weed/s3api/s3api_object_handlers.go
```diff
@@ -383,34 +388,42 @@ func (s3a *S3ApiServer) putToFiler(r *http.Request, uploadUrl string, dataReader
...
 	resp, postErr := client.Do(proxyReq)
 
 	if postErr != nil {
 		glog.Errorf("post to filer: %v", postErr)
-		return "", s3err.ErrInternalError
+		return "", "", s3err.ErrInternalError
 	}
 	defer resp.Body.Close()
 
+	bufferBytes := bufferRead.Bytes()
```
chrislusf, line 399: This will make a full copy of the data. Is there anyway to have something similar to both TeeReader and LimitedReader, which can just peek into the first 512 bytes, but not more than necessary?

kmlebedev: No, I spent a lot of time sorting through the options for how to read the first 512 bytes
And it turned out to be optimal in that place to get a full copy of the bytes when reading the body after transferring, since bytes are needed for get md5 check sum

chrislusf: memory consumption will slow down the upload, especially during any benchmark. Reading the whole file should be avoided. (same thing should apply to other places, to be improved.)

PR #2381, opened 2021-10-14
refactor: move from io/ioutil to io and os package

Juneezee:
The `io/ioutil` package has been deprecated in Go 1.16 (See https://golang.org/doc/go1.16#ioutil). Since SeaweedFS has upgraded to Go 1.17 (e64339e8c0f8c713ec78e336cb6f333d5a27868f), this PR replaces the existing `io/ioutil` functions with their new definitions in `io` and `os` packages.

### DIFF
--- a/test/s3/basic/basic_test.go
+++ b/test/s3/basic/basic_test.go
@@ -2,14 +2,15 @@ package basic
 
 import (
 	"fmt"
+	"io"
+	"os"
+	"strings"
+	"testing"
+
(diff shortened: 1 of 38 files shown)


Juneezee:
> Thanks for taking the efforts to do this! Is there some tool to do this, or it's all manual effort?

@chrislusf Hi! It requires some manual effort and some search assistance from an IDE (I use GoLand).

PR #2427, opened 2021-11-09
Fix: potencial iam identities data race

Bl1tz23:
Hello! 
I guess I found a potential data race condition with `iam.identities`.
Method `loadS3ApiConfiguration` contains assignment to identities slice which is not safe. Cause it might be read in the same time by other goroutines (server handlers etc). This PR must fix this issue.

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -24,7 +24,7 @@ type Iam interface {
 }
 
 type IdentityAccessManagement struct {
-	m sync.Mutex
+	m sync.RWMutex
 
 	identities []*Identity
 	domain     string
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks for your effort! Did you notice any problems with this or just think
there might be a problem?

It’s just a copy on write operation and should not have any problems.

Bl1tz23:
> Thanks for your effort! Did you notice any problems with this or just think there might be a problem? It’s just a copy on write operation and should not have any problems.

It would be hard to catch the race condition while runtime with `-race` flag, but I'm sure it's exactly the case. 
It's unsafe to assign a variable (no matter what type it is) without any synchronization when other goroutines are reading it.

For example:
```
s := []int{1, 2, 3, 4, 5}

ctx, cancel := context.WithTimeout(context.TODO(), 10*time.Second)
defer cancel()

done := make(chan struct{})

go func() {
	for {
		select {
		case <-ctx.Done():
			done <- struct{}{}
			return
		default:
			s = []int{6, 7, 8, 9, 10}
		}
	}
}()

go func() {
	for range s {
	}
}()
	
<-done
```
This code contains similar race condition.

chrislusf:
Good to know! Not so sure. Any documents on this? If so, there should be
some locks on reading also?

Bl1tz23:
> Good to know! Not so sure. Any documents on this? If so, there should be some locks on reading also?

Yes, that's true, I've just added a RWMutex for concurrent-safe reading. 
Here's the docs that describes the problem: https://golang.org/doc/articles/race_detector

PR #2438, opened 2021-11-11
Volume configure replication by collection pattern

ekhvalov:


weed/shell/command_volume_configure_replication.go
```diff
@@ -106,3 +108,19 @@ func (c *commandVolumeConfigureReplication) Do(args []string, commandEnv *Comman
 
 	return nil
 }
+
+func getVolumeFilter(replicaPlacement *super_block.ReplicaPlacement, volumeId uint32, collectionPattern string) func(message *master_pb.VolumeInformationMessage) bool {
+	replicaPlacementInt32 := uint32(replicaPlacement.Byte())
+	if collectionPattern != "" {
```
chrislusf: users may be confused and specify both volumeId and collectionPattern.

it's better to prioritize volumeId first.

PR #2451, opened 2021-11-15
Fix meta aggregation problems

Dando-Real-ITA:
- For mount and filer.meta.backup
- Fixes unlimited reconnection loop on ReadPersistedLogBuffer fail, allowing to progress to use memory read. Can happen if some logs are missing, and always fails for the latest log while it is in memory
- Added more waits before retries


weed/server/filer_grpc_server_sub_meta.go
```diff
@@ -39,12 +39,10 @@ func (fs *FilerServer) SubscribeMetadata(req *filer_pb.SubscribeMetadataRequest,
 	var readInMemoryLogErr error
 
 	for {
-
 		glog.V(4).Infof("read on disk %v aggregated subscribe %s from %+v", clientName, req.PathPrefix, lastReadTime)
-
 		processedTsNs, readPersistedLogErr = fs.filer.ReadPersistedLogBuffer(lastReadTime, eachLogEntryFn)
 		if readPersistedLogErr != nil {
-			return fmt.Errorf("reading from persisted logs: %v", readPersistedLogErr)
+			glog.V(4).Infof("read on disk %v subscribe %s from %+v: %v", clientName, req.PathPrefix, lastReadTime, readPersistedLogErr)
```
chrislusf: the error could be a retry-able error. Just skipping it may cause more problems later.

Dando-Real-ITA:
I found other problems:
- Infinite loop on the same metadata subscription until filer is restarted
```
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
log_buffer.go:213] resume with last flush time: 2021-12-09 16:35:15.381594584 +0000 UTC
masterclient.go:125] filer masterClient failed to receive from cloud_server:9333: rpc error: code = Unavailable desc = closing transport due to: connection error: desc = "error reading from server: EOF", received prior goaway: code: NO_ERROR
masterclient.go:158] + filer cloud_filer:8888 leader:true
filer_grpc_server_sub_meta.go:237] + listener filer:cloud_filer:8888@10.222.0.26:56698
filer_grpc_server_sub_meta.go:88]  filer:cloud_filer:8888@10.222.0.26:56698 local subscribe / from 2021-12-09 20:49:00.840459361 +0000 UTC
```

- After a while ( hours ) a metadata listener is subscribed, it starts to receive multiple parallel metadata changes instead of 1. It is clearly visible by using filer meta backup, while the effect on mount is to slow down write speed from 300 Mb/s to < 1Mb/s

image

chrislusf:
Thanks for the error. For the tight loop, I added https://github.com/chrislusf/seaweedfs/commit/2d6fcdf83ae4d033e3457f4722feaeea761b1a02

chrislusf:
your change is also good. How to separate this change from other commits?

Dando-Real-ITA:
I guess cherry picking, otherwise diff the branches and get only the portions needed

Dando-Real-ITA:
Unfortunately there is still the problem with multiple parallel metadata connections
image

And a spam of resume in the filer
image

I'll try to investigate where multiple metadata connections are created and try to force to have only 1

Dando-Real-ITA:
What is still unclear for me, is why when there is a file change, in filer I get a lot of `rpc errors: transport is closing` and then log_buffer resume messages.
The connection should be in keep-alive between mount/filer.meta.backup and filer, is it not?

PR #2500, opened 2021-12-09
rename operation returns the correct old parent path

paochiang:
hello, i have tried to fix issue: https://github.com/chrislusf/seaweedfs/issues/2499, hope it is helpful.


chrislusf:
Nice fix! This is a hard bug and I wonder how you debugged it. :)

paochiang:
I did it in a dumb way. First compare with a working version: https://github.com/chrislusf/seaweedfs/commit/789465d99e9a14bdd7424d62a5a252e101d14cc5 , then do a lot of testing.

PR #2510, opened 2021-12-14
helm add auditLogConfig option

kmlebedev:


k8s/helm_charts2/values.yaml
```diff
@@ -273,9 +273,6 @@ filer:
   # Limit sub dir listing size (default 100000)
   dirListLimit: 100000
 
-  # Turn off directory listing
-  disableDirListing: false
```
chrislusf, line 277: This is an unrelated change.

kmlebedev: @chrislusf  This is a duplicate of IDE highlighted
https://github.com/chrislusf/seaweedfs/blob/master/k8s/helm_charts2/values.yaml#L263

PR #2511, opened 2021-12-14
feat: Added a volume heartbeat graph to grafana dashboard

ErikCoder07:
In the current dashboard there is no way to know the status of
volume servers, as prometheus only accepts push information, if
any server volume goes offline, the last information sent will
be read continuously by grafana.

Added a graph that calculates the delta between the information
of the last push timestamp with the timestamp of 15 seconds ago,
if the timestamp has not been changed, server state will be 0,
if volume server sent any information the server state will be 1.

-This graph will have to be modified if the time interval of pushs
is modified, a suggestion for improvement is to send the interval
in seconds along with the metrics.

-Only local tests were run, maybe if the latency between pushgateway
and volume server is too high, the delta calculation could be
affected

grafana_heartbeat_example
.

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,7 +501,7 @@ git clone https://github.com/chrislusf/seaweedfs.git
 Step 3: download, compile, and install the project by executing the following command
 
 ```bash
-make install
+cd seaweedfs/weed && make install
 ```
 
 Once this is done, you will find the executable "weed" in your `$GOPATH/bin` directory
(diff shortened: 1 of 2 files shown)


chrislusf:
Nice. Could this be merged with existing dashboards?

ErikCoder07:
Yes, I used the "seaweedfs/k8s/helm_charts2/dashboards/seaweedfs-grafana-dashboard.json" as a base, just added the graph in the volume server information. I created a new file because as I mentioned in the commit, it may be that in a situation of use that takes a little more than 15 seconds to push the metrics, in this case the delta calculation may not work, in the print I sent the volumes are in the same environment than the master, if everything is ok even with external volume servers it can be merged!

chrislusf:
thanks and merged. I adjusted the file name to be consistent. Could you please add a README.md file there also?

PR #2515, opened 2021-12-16
Added Cache-Control header

Dando-Real-ITA:
Tuned for HLS streams


chrislusf:
this looks good. but the parameters would likely need customization and this feature is better handled in nginx.

PR #2530, opened 2021-12-22
filer server: allow upload file to specific dataNode

banjiaojuhao:
"dataNode" parameter is not supported in upload files to Filer now: [Filer server POST/PUT/GET files](https://github.com/chrislusf/seaweedfs/wiki/Filer-Server-API#postputget-files)
This PR adds support of action like: `curl -F file=@report.js "http://localhost:8888/javascript/?dataNode=node1"`


chrislusf:
could you also adjust the related wiki page?

banjiaojuhao:
> could you also adjust the related wiki page?

updated

PR #2536, opened 2021-12-26
redis2 filer store: use lua script for atomicity

banjiaojuhao:
> ### [Atomicity of scripts](https://redis.io/commands/eval#atomicity-of-scripts)
>Redis uses the same Lua interpreter to run all the commands. Also Redis guarantees that **a script is executed in an atomic way**: no other script or Redis command will be executed while a script is being executed. This semantic is similar to the one of MULTI / EXEC. From the point of view of all the other clients the effects of a script are either still not visible or already completed.
However this also means that executing slow scripts is not a good idea. It is not hard to create fast scripts, as the script overhead is very low, but if you are going to use slow scripts you should be aware that while the script is running no other client can execute commands.

### DIFF
--- a/weed/filer/redis2/universal_redis_store.go
+++ b/weed/filer/redis2/universal_redis_store.go
@@ -21,11 +21,22 @@ const (
 type UniversalRedis2Store struct {
 	Client                  redis.UniversalClient
 	superLargeDirectoryHash map[string]bool
+	supportLuaScript        bool
 }
 
 func (store *UniversalRedis2Store) isSuperLargeDirectory(dir string) (isSuperLargeDirectory bool) {
 	_, isSuperLargeDirectory = store.superLargeDirectoryHash[dir]
 	return
+}
+
+func (store *UniversalRedis2Store) detectSupportLuaScript(ctx context.Context) bool {
+	result, err := redis.NewScript("return 'hello world'").Run(ctx, store.Client, nil).Result()
+	store.supportLuaScript = err == nil && result == "hello world"
+	return store.supportLuaScript
+}
+
+func (store *UniversalRedis2Store) isSupportLuaScript() bool {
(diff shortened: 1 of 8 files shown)


chrislusf:
The redis2 is used by many Redis API compatible key-value stores, which usually do not support Lua.

This change would reduce the compatibility. How about moving it to a different "redis_lua" version, or adding an option to toggle this feature?

banjiaojuhao:
codes are added to detect lua script support

weed/filer/redis2/universal_redis_store.go
```diff
@@ -20,13 +21,24 @@ const (
...
 }
 
 func (store *UniversalRedis2Store) isSuperLargeDirectory(dir string) (isSuperLargeDirectory bool) {
 	_, isSuperLargeDirectory = store.superLargeDirectoryHash[dir]
 	return
 }
 
+func (store *UniversalRedis2Store) detectSupportLuaScript(ctx context.Context) bool {
+	result, err := redis.NewScript("return 'hello world'").Run(ctx, store.Client, nil).Result()
+	store.supportLuaScript = err == nil && result == "hello world"
```
chrislusf, line 34: Add a log here.

chrislusf:
thanks! I think this `isSupportLuaScript()` works. However, the lua script logic and the Go logic are mixed together with "if/else". The lua logic needs to always be 100% consistent with Go logic. This is not easy to test or refactor.

Please change to `redis2_lua`. In the future, you can even add transactional support there.
Or the current `redis2` can switch to pipeline to achieve the same goal.

chrislusf:
One additional benefit of this PR is to reduce one network trip to Redis server, which can improve performance quite a bit.

I would like this to be a separate Redis filer store implementation so that all Redis and variants with Lua support can use this.

Issue #158, opened 2015-06-29
How to make file server secure?


truonguit2010:
Hi,
I am writing a mobile application, I allow my application write/read file to/from server. But I don't know how to make write operation secure. I want that logged use can do write operation. How can I do that?

Thank you.

chrislusf:
There is a plan to add json web token for write operations, but not ready yet.

Davidx7:
A good alternative would be to serve weedfs content trough a nginx server, limit weedfs to localhost and add authentication to nginx, until @chrislusf adds tokens

truonguit2010:
Thank you. I will consider that @Davidx7 . Are there any config file to config the server?

janober:
You could also have a look at: http://getkong.org (uses under the hood also nginx)

mingqlin:
@chrislusf is json web token  ready?

chrislusf:
Not getting time to finish it yet.

jjn2009:
I'm pretty interested in this as well, it would make using seaweed in production much easier.

chrislusf:
RFC for JWT in SeaweedFS:

Purpose: Volume servers need to ensure the write operations are valid.

Not covered: Securing operations to control Volume server. These should be done via 2-way SSL.

Approach: 
1. When starting master servers and volume servers, they should have a common secret key option.
2. If the secret key is specified, when a client requests a fileId from master server, the master server should also return a JWT, specifying the permission to upload content to a volume server.
2. The volume server should check for JWT for all write operations.

krzysztof-kabala:
+1

brstgt:
Whats the exact purpose?
In MOST cases, end users don't have direct access to vol servers and IMHO they never should. E.g. we use a public API, do a lot of stuff in our backend and as one part of the whole upload process, a file is uploaded to a volserver internally. But there is a lot more to an upload like associating the uploaded file and it's file-id with some other entities like a database entry that stores more metadata or counting stats, checking a sophisticated auth/permission system, sending out notifications and a lot more.
So I don't see many situations where a standalone auth system makes so much sense.
By the way you also could realize that with a redis-based auth in nginx. You can do this in nginx with a few lines of lua and still keep seaweed simple and fast.

jjn2009:
The point of JWT is that the auth is not stand alone, you delegate short lived privileges to clients with short lived tokens via your main application server or some service specifically meant for authentication to other services which may or may not have access to the database with authentication or session information.

[...]

PR #2543, opened 2021-12-29
 FEATURE: add JWT to HTTP endpoints of Filer and use them in S3 Client 

skurfuerst:
# Description

- one JWT for reading and one for writing, analogous to how the JWT
  between Master and Volume Server works
- I did not implement IP `whiteList` parameter on the filer

Additionally, because http_util.DownloadFile now sets the JWT,
the `download` command should now work when `jwt.signing.read` is
configured. By looking at the code, I think this case did not work
before.

## Docs to be adjusted after a release

Page `Amazon-S3-API`:

```
# Authentication with Filer

You can use mTLS for the gRPC connection between S3-API-Proxy and the filer, as
explained in [Security-Configuration](Security-Configuration) -
controlled by the `grpc.*` configuration in `security.toml`.

Starting with version XX, it is also possible to authenticate the HTTP
operations between the S3-API-Proxy and the Filer (especially
uploading new files). This is configured by setting
`jwt.jwt.filer_signing.key` and `jwt.filer_signing.read.key` in
`security.toml`.

With both configurations (gRPC and JWT), it is possible to have Filer
and S3 communicate in fully authenticated fashion; so Filer will reject
any unauthenticated communication.
```

Page `Security Overview`:

```
The following items are not covered, yet:

- master server http REST services

Starting with version XX, the Filer HTTP REST services can be secured
with a JWT, by setting `jwt.filer_signing.key` and
`jwt.filer_signing.read.key` in `security.toml`.

...

Before version XX: "weed filer -disableHttp", disable http operations, only gRPC operations are allowed. This works with "weed mount" by FUSE. It does **not work** with the [S3 Gateway](Amazon S3 API), as this does HTTP calls to the Filer.
Starting with version XX: secured by JWT, by setting `jwt.filer_signing.key` and `jwt.filer_signing.read.key` in `security.toml`. **This now works with the [S3 Gateway](Amazon S3 API).**

...

# Securing Filer HTTP with JWT

To enable JWT-based access control for the Filer,

1. generate `security.toml` file by `weed scaffold -config=security`
2. set `jwt.filer_signing.key` to a secret string - and optionally `jwt.filer_signing.read.key` as well to a secret string
3. copy the same `security.toml` file to the filers and all S3 proxies.

If `jwt.filer_signing.key` is configured: When sending upload/update/delete HTTP operations to a filer server, the request header `Authorization` should be the JWT string (`Authorization: Bearer [JwtToken]`). The operation is authorized after the filer validates the JWT with `jwt.filer_signing.key`.

If `jwt.filer_signing.read.key` is configured: When sending GET or HEAD requests to a filer server, the request header `Authorization` should be the JWT string (`Authorization: Bearer [JwtToken]`). The operation is authorized after the filer validates the JWT with `jwt.filer_signing.read.key`.

The S3 API Gateway reads the above JWT keys and sends authenticated
HTTP requests to the filer.
```

Page `Security Configuration`:

```
(update scaffold file)

...

[jwt.filer_signing]
key = "blahblahblahblah"

[jwt.filer_signing.read]
key = "blahblahblahblah"
```

Resolves: #158

### DIFF
--- a/test/s3/compatibility/.gitignore
+++ b/test/s3/compatibility/.gitignore
@@ -1,0 +1,2 @@
+/s3-tests
+/tmp
--- a/test/s3/compatibility/Dockerfile
+++ b/test/s3/compatibility/Dockerfile
@@ -1,0 +1,11 @@
+# the tests only support python 3.6, not newer
+FROM ubuntu:latest
+
+RUN apt-get update && DEBIAN_FRONTEND=noninteractive TZ=Etc/UTC apt-get install -y git-core sudo tzdata
+RUN git clone https://github.com/ceph/s3-tests.git
+WORKDIR s3-tests
+
+# we pin a certain commit
+RUN git checkout 9a6a1e9f197fc9fb031b809d1e057635c2ff8d4e
+
+RUN ./bootstrap
--- a/test/s3/compatibility/README.md
+++ b/test/s3/compatibility/README.md
@@ -1,0 +1,13 @@
+# Running S3 Compatibility tests against SeaweedFS
+
+This is using [the tests from CephFS](https://github.com/ceph/s3-tests).
+
+## Prerequisites
+
+- have Docker installed
+- this has been executed on Mac. On Linux, the hostname in `s3tests.conf`  needs to be adjusted.
+
+## Running tests
+
+- `./prepare.sh` to build the docker image
+- `./run.sh` to execute all tests
--- a/test/s3/compatibility/prepare.sh
+++ b/test/s3/compatibility/prepare.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+set -ex
+
+docker build  --progress=plain  -t s3tests .
--- a/test/s3/compatibility/run.sh
+++ b/test/s3/compatibility/run.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+
+set -ex
+
+killall -9 weed || echo "already stopped"
+rm -Rf tmp
+mkdir tmp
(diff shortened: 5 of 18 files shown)


skurfuerst:
Hey @chrislusf,

IMHO this is ready except for testing. I'll test this in detail in the upcoming days, before it can be merged. 

I think it can be already thoroughly reviewed.

All the best, 
Sebastian

chrislusf:
Looks good. Can you move the filer jwt checking to this separate file instead of mixing it with existing logic?

https://github.com/chrislusf/seaweedfs/blob/master/weed/server/filer_server_handlers.go

skurfuerst:
@chrislusf good catch! Changed the code, what do you think about it?

I think I am now also securing the "proxy to volume servers" case properly - so all cases should be covered now.

As said, I'll test in detail in the upcoming days and then report back :) When this all works in practice, I think this would be ready to merge.

All the best,
Sebastian

skurfuerst:
Hey @chrislusf during testing, I found some tiny bugs which are fixed now.

BTW: I added a script to execute a S3 conformance test suite based on https://github.com/ceph/s3-tests - to be sure that I did not break anything regarding S3 compatibility before and after this change.

IMHO this is ready to merge now :)

All the best, and a happy new year,
Sebastian

chrislusf:
Thanks for the great work! Merged.

chrislusf:
Please help to update the wiki also. @skurfuerst

skurfuerst:
@chrislusf sure, planned this for today. Wiki is all updated now :)  Thanks again for your great work, it's a pleasure to work with you!

PR #2614, opened 2022-01-25
asyncVolumeTierMove

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2591

### DIFF
--- a/weed/shell/command_volume_tier_move.go
+++ b/weed/shell/command_volume_tier_move.go
@@ -24,6 +24,7 @@ type commandVolumeTierMove struct {
 	activeServers     map[pb.ServerAddress]struct{}
 	activeServersLock sync.Mutex
 	activeServersCond *sync.Cond
+	allLocations      []location
 }
 
 func (c *commandVolumeTierMove) Name() string {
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_tier_move.go
```diff
@@ -68,27 +71,57 @@ func (c *commandVolumeTierMove) Do(args []string, commandEnv *CommandEnv, writer
...
+	for len(volumeIds) > 0 {
+		for _, vid := range volumeIds {
+			volumeIdChan <- vid
+			if *volumesPerStep == 0 {
+				wg.Wait()
+			}
+		}
+		if *volumesPerStep > 0 {
+			wg.Wait()
+			err, volumeIds = c.collectAllVolumeIds(commandEnv, fromDiskType, toDiskType, *collectionPattern, *fullPercentage, *quietPeriod, *volumesPerStep)
```
chrislusf: The info collected here may be stale. Need to adjust the local data structure directly.

kmlebedev: @chrislusf Can you explain why I don't understand

chrislusf: "collectAllVolumeIds" will read from the master, which will wait for the volume servers to report their volumes. If the volume servers are slow to report, the info will be stale. 

For example, in some cases the "collectAllVolumeIds" may get the same info as the last time. The same volume may be picked again. But actually the volume has already been moved.

Issue #2591, opened 2022-01-13
Parallel migration between tiers?


divanikus:
We've been using tiered storage for a quite time already, even in production. But as our storage grows, we have faced the bottleneck in the current scheme. Hot storage is filling up quite fast, and as far as `volume.tier.move ` move one volume at time, `master` script simply can't keep up with that pace. I've already disabled almost all other operations on cluster and forced script rerun to every 30 minutes, but still. 
So it would be cool to be able to move volumes to another tier in parallel. Say, I have 3 ssd servers and 6 hdd servers. I think I can move up to 3 volumes at time without greatly affecting performance. Could this be implemented or I have missed something and it can be done already?

chrislusf:
The algo is all in this file: https://github.com/chrislusf/seaweedfs/blob/master/weed/shell/command_volume_tier_move.go#L86

Probably you can try to adjust the algo. Need to avoid one server moving more than one volume at any moment.

divanikus:
I'm afraid I'm not that familiar with Go, but I'll give it a shot. My pretty much straightforward approach can be maintaining a list of currently running `[from, to]` migrations and scanning for pairs which has neither `from` nor `to` from the previous list. If such pair found, just wind up another migration right away and restart scan. If no pairs found, wait till previous tasks finish. Also I can think of an option which explicitly limits the maximum number of running tasks. 

Let me now if I'm designing something stupid or not implementable.

divanikus:
I came up with a little bit different approach, please review it.

PR #2618, opened 2022-01-26
async volumeTierMove

divanikus:
Attempt to solve the issue #2591 
I'm using channels and keyed mutexes to avoid copying from the same or to the same server at once.

### DIFF
--- a/weed/shell/command_volume_tier_move.go
+++ b/weed/shell/command_volume_tier_move.go
@@ -20,10 +20,17 @@ func init() {
 	Commands = append(Commands, &commandVolumeTierMove{})
 }
 
+type volumeTierMoveJob struct {
+	src pb.ServerAddress
+	vid needle.VolumeId
+}
+
 type commandVolumeTierMove struct {
-	activeServers     map[pb.ServerAddress]struct{}
-	activeServersLock sync.Mutex
-	activeServersCond *sync.Cond
+	activeServers sync.Map
+	queues map[pb.ServerAddress]chan volumeTierMoveJob
+	//activeServers     map[pb.ServerAddress]struct{}
+	//activeServersLock sync.Mutex
+	//activeServersCond *sync.Cond
 }
 
 func (c *commandVolumeTierMove) Name() string {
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_tier_move.go
```diff
@@ -82,15 +87,87 @@ func (c *commandVolumeTierMove) Do(args []string, commandEnv *CommandEnv, writer
...
+			wg.Done()
+		}(dst, c.queues[destServerAddress], *applyChange)
+	}
+
 	for _, vid := range volumeIds {
-		if err = c.doVolumeTierMove(commandEnv, writer, vid, toDiskType, allLocations, *applyChange); err != nil {
+		if err = c.doVolumeTierMove(commandEnv, writer, vid, toDiskType, allLocations); err != nil {
 			fmt.Printf("tier move volume %d: %v\n", vid, err)
 		}
+		allLocations = rotateDataNodes(allLocations)
```
chrislusf, line 134: This PR is in the right direction. Thanks! It is already excellent since you are just started with Go.

The algo needs some changes. Here the destinations are not "weighted". Some could be "more empty" than others, although they are already the top-k empty destinations.

How about calling "keepDataNodesSorted()" again after planning each move and adjust the estimated volume count?

divanikus: The problem with calling `keepDataNodesSorted()` on each turn is within those "more empty" volumes. Say if we have a server with 1300 volumes and another with 1400, the first one would be choosen on each turn. This practically enqueue all migration job to just one server, turning the algo back to fully sequential.  This is why I'm calling `rotate`, just to ensure that we choose next server on the next turn.

The algo by itself is pretty much dumb, because two neighbor queues might have a job with the same source server effectivily blocking each other and turning it sequentual again. So we just pushing jobs to queues and hoping for the best. Idk if we can improve it somehow, at least for single pass algo. May be we could sort vids somehow to distribute them better to queues. 

Also, I'm unsure about increasing of the volume count in the enqueueing cycle, because they are also updated after succesful move routines. Could led to some edge cases with "almost full" destinations.

chrislusf: This algo let all nodes in the "allLocations" to receive about equal number of volumes. So some of the less empty nodes may have more than average number of volumes.

Should be ok to let the more empty nodes to have more volumes to move into. It will take more time to run. But it is better than waiting for the next pass?

divanikus: My point is that if you say just added a new empty volume server to a destination tier, all jobs would be enqueued to it, making the whole migration process fully sequentual again. This is why I'm forcing the algo to pick next server on the next turn. Can you suggest how to make it weighted? 
I think we can add tier filter to `volume.balance` to address that problem though.

chrislusf: > added a new empty volume server to a destination tier, all jobs would be enqueued to it

This is the ideal way to place the volumes, correct? Waiting for the `volume.balance` is ok, but not so efficient.

divanikus: Yeah, maybe. But it kills the whole point of parallelism, at list for a couple of my setups. One of them has no replicas enabled and for some reason usually have like 10 to 100 volume difference between servers. As far as it is also suffering from slow migrations, real parallelism most likely won't happen anytime soon. Adding a new server and waiting for it to fill up to 1500 volumes would also be painfully slow.

chrislusf: I see. You want to parallelize it before reaching the balanced state. Make a lot of sense!

chrislusf: I am going to merge this,and adjust some minor things, such as variable names, etc.

PR #2632, opened 2022-02-04
add s3api error for copy in file, not directory

lapshin-vitaly:
weed/s3api/s3err/s3api_errors.go

When user try to copy some file into another file, he received only "500 Internal server error". This error not informative.

Add error - "ErrExistingObjectIsFile", when user try to copy file into file, he receive "409 Existing object is a file"

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -436,10 +436,14 @@ func setEtag(w http.ResponseWriter, etag string) {
 }
 
 func filerErrorToS3Error(errString string) s3err.ErrorCode {
-	if strings.HasPrefix(errString, "existing ") && strings.HasSuffix(errString, "is a directory") {
+	switch {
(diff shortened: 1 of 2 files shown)


chrislusf:
Thanks for the PR! Should the destination file be overwritten?

lapshin-vitaly:
Thanks for comment. Destination file should n't be overwritten. Overwrite file should be clearly for user

chrislusf:
https://stackoverflow.com/questions/30246784/aws-s3-replace-file-atomically

this link suggested AWS S3 does overwrite. Do you have some official doc on this?

lapshin-vitaly:
I mean the next case:

I put file ./foofile into s3://bucketname/foofile
Than  put ./barfile into s3://bucketname/barfile

We have two files in bucketname 
Than I copy s3 cp s3://bucketname/foofile s3://bucketname/barfile/foofile and sw-s3 return only 500 Internal server error
In this case it should return 409 error and not overwrite or something else

Issue #2605, opened 2022-01-21
2.86 - needs fsck to drop missing filer listings from filers


Radtoo:
In Seaweedfs 2.86 (linux 8000GB amd64) you can use "weed shell" "volume.fsck -findMissingChunksInFiler". It detects missing files from volumes that are listed in the filers. Breaking volume .dat files in some way will generally result in some or a lot of files getting listed there.

But while this command can certainly find problems, it does not seem capable of fixing - removing entries from filers that were found to be missing from volumes. The missing files remain listed and produce various errors on attempted access depending on what accesses them through what method. On FUSE mounts various software will terminate on hitting an error like this, it's an unexpected type of I/O error.

I am thinking command_volume_fsck.go could use some kind of pendant to "-reallyDeleteFromVolume" for use with "-findMissingChunksInFiler" to just tell filer servers they should remove the detected entries from filers.

Radtoo:
I started trying but did not get very far. I hit the question how to sanely get a handle on the live filer(s), whether to use filer.doDeleteFileIds to handle the deletions, whether item.fileKey is a reliable identifier or if the paths should be used, and such.

If you too are interested in fixing this, maybe you can trivially see what could be added here?
https://github.com/Radtoo/seaweedfs/blob/f2d71a22b5237432c1ee92b9aa327adff9a08ef1/weed/shell/command_volume_fsck.go#L399

chrislusf:
The original intention was to just print out the affected files, which can be deleted by some other tools. For example, a simple shell script to go through them and send http delete requests.

chrislusf:
One file with multiple chunks will show up multiple times. So there would need another step to dedup.

This can be trivially done via some linux shell scripts. We can add an option to output the list of files to a file. And let the user to send http delete requests.

Radtoo:
>  For example, a simple shell script to go through them and send http delete requests.

This is harder than expected even now that you pointed at HTTP delete requests. After disabling JWT, I tried sending "curl -vX DELETE http://ip:port/volume,id" with the same volume,id as in "volume.fsck -findMissingChunksInFiler".

Nothing changes regardless if this is sent to a master (permanent redirect 308), volume (not found 404) or filer (200 OK - but it's OK in pretty much all instances).

Is it simply the wrong command to send? What would you use to instruct a filer to delete a record?

> The original intention was to just print out the affected files, which can be deleted by some other tools.

[...]

PR #2640, opened 2022-02-06
Added basic deletion capabilities to fsck.

Radtoo:
With your helpful feedback from #2605 , I added/fixed the option to delete missing files volumes<->in filer to volume.fsck. It does successfully get rid of the errors that volume.fsck can catch.

What wasn't implemented from the ideas in #2605 is an option to write lists of affected files for weed external review/consumption.

PS: I had to drop a small functionality that dropped a whole volume rather than sending filewise deletion requests when all entries were orphaned. It did not really work (haven't fully established the reason, but my guess was that not all additional replicas on all volume servers were removed). However "only" doing the usual filewise deletion seems fast enough and has to be used for 1+ files anyhow, so I went with just that.

### DIFF
--- a/weed/shell/command_volume_fsck.go
+++ b/weed/shell/command_volume_fsck.go
@@ -64,7 +64,7 @@ func (c *commandVolumeFsck) Do(args []string, commandEnv *CommandEnv, writer io.
 	verbose := fsckCommand.Bool("v", false, "verbose mode")
 	findMissingChunksInFiler := fsckCommand.Bool("findMissingChunksInFiler", false, "see \"help volume.fsck\"")
 	findMissingChunksInFilerPath := fsckCommand.String("findMissingChunksInFilerPath", "/", "used together with findMissingChunksInFiler")
-	applyPurging := fsckCommand.Bool("purge", false, "<expert only!> after detection, delete missing data from volumes / delete missing file entries from filer")
+	applyPurging := fsckCommand.Bool("reallyDeleteFromVolume", false, "<expert only!> after detection, delete missing data from volumes / delete missing file entries from filer")
 	purgeAbsent := fsckCommand.Bool("purgeAbsent", false, "<expert only!> delete missing file entries from filer if the corresponding volume is missing for any reason, please ensure all still existing/expected volumes are connected! used together with findMissingChunksInFiler")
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_fsck.go
```diff
@@ -61,7 +64,8 @@ func (c *commandVolumeFsck) Do(args []string, commandEnv *CommandEnv, writer io.
 	verbose := fsckCommand.Bool("v", false, "verbose mode")
 	findMissingChunksInFiler := fsckCommand.Bool("findMissingChunksInFiler", false, "see \"help volume.fsck\"")
 	findMissingChunksInFilerPath := fsckCommand.String("findMissingChunksInFilerPath", "/", "used together with findMissingChunksInFiler")
-	applyPurging := fsckCommand.Bool("reallyDeleteFromVolume", false, "<expert only> delete data not referenced by the filer")
+	applyPurging := fsckCommand.Bool("purge", false, "<expert only!> after detection, delete missing data from volumes / delete missing file entries from filer")
```
chrislusf: I want to keep the "ugly" long name "reallyDeleteFromVolume" so that it is clear what it does and that it is a dangerous command.

Just "purge" or "purgeAbsent", I can not tell what they do.

Radtoo: I thought this nomenclature would be more obvious ("purge" the "missing chunks in filer"), but it seems like it could be just me. So absolutely, lets go with what you prefer!

I am suspecting -purgeAbsent is equally hard to understand? Or does that make more sense?

chrislusf: Same for "purgeAbsent". How about "purgeIncompleteFiles"?

Radtoo: Hm. It doesn't exactly purge "incomplete" files, it purges -maybe we equally use "reallyDelete"?- filer entries when the respective volumes are gone via a file deletion request.

Maybe something like -reallyDeleteFilerEntries if we go with a very verbose flag?

chrislusf: "-reallyDeleteFilerEntries" sounds good enough to scare away non-experts. :)

Radtoo: Cheers to to even scarier flags in the future! :)

PR #2644, opened 2022-02-07
updated needle with fsync

kmlebedev:


kmlebedev:
may help to avoid https://github.com/chrislusf/seaweedfs/issues/2630

PR #2646, opened 2022-02-08
graceful stop cluster http server default timeout 60 sec

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2630


chrislusf:
nit. The value is 150 instead of 60 as in the title.

kmlebedev:
> nit. The value is 150 instead of 60 as in the title.

This is a save to stop two http and one grpc

PR #2652, opened 2022-02-09
[FEAT] Support extensible gRPC authentication mechanisms.

garenchan:
### Background

One security issue we currently face is that seaweedfs can be mounted on any machine as long as the IP address is known. 
Of course we can solve this problem by limiting the source IP with iptables, but this is not an optimal approach. 
So I implemented extensible gRPC authentication mechanisms.

### Usage

Servers and clients can modify the `security.toml` file as follows to enable gRPC authentication. 
Currently, only the `bearer` authentication mechanism is supported. 
Client authenticates by carrying an :authorization header in the metadata, e.g., authorization: "bearer [token]"

```
[grpc.auth]
mech = "bearer"
token = "<non-empty string>"
```

Servers and clients can also configure the gRPC authentication based on environment variables, such as:

```
WEED_GRPC_AUTH_MECH="bearer" WEED_GRPC_AUTH_TOKEN="<non-empty string>" weed mount -dir=/mnt/
``` 

When servers enable `bearer` authentication mechanism, the clients must be configured with the same token as the servers; otherwise, the clients will fail to mount seaweedfs.


chrislusf:
Thanks for the initiative and such a big PR!

I was thinking the clients can just have its own security.toml file? What is the benefit from this PR?

Also, this would need some wiki changes if merged. Need to explain the feature better with some scenarios and examples.

garenchan:
@chrislusf, there seems to be no way to configure `security.toml` to prevent unauthorized users from mounting seaweedfs. In other words, as long as the unauthorized user has the IP address of filer and can access it, he can read and write files at will. This is too dangerous for us.

I tried to solve the problem using gRPC mutual TLS, but it didn't work. I configured the private key and certificate on the server side. But the client can still mount seaweedfs without configuring the private key and certificate. Mutual TLS doesn't seem to be mandatory.

By supporting GRPC authentication, we can prevent unauthorized users from mounting seaweedfs. For example, if the client does not get the token, it will fail to mount.

chrislusf:
```
# use this for any place needs a grpc client
# i.e., "weed backup|benchmark|filer.copy|filer.replicate|mount|s3|upload"
[grpc.client]
cert = ""
key = ""
```

This was meant for the clients to use. Does it not working for you?

garenchan:
I just tested it again. It works. I may have misconfigured it earlier.

I'm sorry to have taken up your valuable time. I'll close this PR and make changes to fix [the problem](https://github.com/chrislusf/seaweedfs/issues/951).

By the way, can you tell me how to configure certificate and key in seaweedfs-csi-driver 1.0.6?

PR #2653, opened 2022-02-09
weed: added optional feature to call replication fix on io error

zerospiel:
This change adds a new block in `master.toml` config with enabling a feature to call weed shell's command `volume.fix.replication` automatically on getting any `i/o error` from a volume.

Disabled by default.

This is a totally optional feature that can be used while there were i.e. something like HDD failure in a rack, that will cause getting several `input/output error` from different volumes within the data node in this case.


zerospiel:
@chrislusf can you comment over this PR, please?

zerospiel:
Actually, I don't get why is this a bad idea.

- it's turned off by default;
- the code does not spawn any child processes, so nothing to leak in case of OS' resources;
- the only way it can go the wrong way is that `volume.fix.replication` possibility to infinite loop but for that case, there is a timeout and this is the same behavior that can be in `weed shell`

zerospiel:
Yes, it's only for a case when there was a disaster with one or more volumes (HDD failure i.e.). 

I don't mind that someone will enable this without understanding the case. Though, this will be about no overhead (except one more gRPC client conn) if disabled. 

Anyway, the `input/output error` for a volume is a very rare case

chrislusf:
I prefer no automatic the system admin reactions. Many types of errors could be transient, or even "normal". Being too smart may just lead to unexpected consequences.

For this case, the system can collect metrics on how healthy the system is. For example, each volume server can store data on how many IO errors on each volume recently. 

The admin scripts can regularly check the volume servers and disable the error prone volumes.

This separation gives more flexibility on how the metrics is calculated, and what the reactions should be.

PR #2661, opened 2022-02-12
Make heartbeat interval and election timeout of masters configurable.

garenchan:
### Problem

Election timeout is fixed at 10 seconds. This means that when the leader dies, it takes 10 to 20 seconds to elect a new leader.
That may be too long for some people.  If the network between masters is good, I think a short election timeout is appropriate.

### Solution

Make heartbeat interval and election timeout of masters configurable. Therefore, we can configure them according to our actual situation.

### Usage

Specified on the command line, e.g.:

```shell
weed server --master.electionTimeout=3s --master.heartbeatInterval=300ms
```

Or specified in the configuration file, e.g:

```
# Run server with: weed server -options=server.conf
# The content of server.conf:
master.electionTimeout=3s
master.heartbeatInterval=300ms
```

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -44,6 +44,8 @@ type MasterOptions struct {
 	metricsIntervalSec *int
 	raftResumeState    *bool
 	metricsHttpPort    *int
+	heartbeatInterval  *time.Duration
+	electionTimeout    *time.Duration
 }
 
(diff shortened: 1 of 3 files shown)


garenchan:
@chrislusf, I have modified it as you suggested.

Issue #2359, opened 2021-10-06
Failed to write to local disk: volume size limit exceeded


Robinysh:
**Describe the bug**
I'm getting some `Failed to write to local disk: volume size limit exceeded` after a few hours of data transfer with `weed filer.copy`. Deleting the broken volume should work for now but the same error comes around every few hours.
 
**Relevant logs**
`weed server -options=/etc/xdg/seaweedfs/server.conf -ip=node1` log:
```
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 store_replicate.go:48] failed to write to local disk: volume size limit 34359738368 exceeded! current size is 34359738285
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 common.go:67] response method:POST URL:/92,f1a83423c5dffd with httpStatus:500 and JSON:{"name":"1329.obj-1","size":4194304,"error":"failed to write to local disk: volume size limit 34359738368 exceeded! current size is 34359738285","eTag":"12e79c95","mime":"text/plain; charset=utf-8"}
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 store_replicate.go:48] failed to write to local disk: volume size limit 34359738368 exceeded! current size is 34359738285
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 common.go:67] response method:POST URL:/92,f1a85d77684546 with httpStatus:500 and JSON:{"name":"1345.mtl","size":8773,"error":"failed to write to local disk: volume size limit 34359738368 exceeded! current size is 34359738285","eTag":"95772b04","mime":"text/plain; charset=utf-8"}
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 filer_deletion.go:57] deleting fileIds len=68 error: delete fileId 76,edca7ad421b2af: volume 76 not found on node1:8080
Oct 06 11:33:09 node1 weed[2538376]: I1006 11:33:09 38376 store_replicate.go:48] failed to write to local disk: volume size limit 34359738368 exceeded! current size is 34359738285
[...]

PR #2673, opened 2022-02-15
filer_copy: retry when uploading file

banjiaojuhao:
#2359 
There is no retrying in `weed filer.copy` command now.

### DIFF
--- a/weed/command/filer_copy.go
+++ b/weed/command/filer_copy.go
@@ -457,11 +457,11 @@ func (worker *FileCopyWorker) uploadFileInChunks(task FileCopyTask, f *os.File, 
 				wg.Done()
 				<-concurrentChunks
 			}()
-			// assign a volume
-			var assignResult *filer_pb.AssignVolumeResponse
-			var assignError error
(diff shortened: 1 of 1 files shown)


weed/command/filer_copy.go
```diff
@@ -457,64 +457,66 @@ func (worker *FileCopyWorker) uploadFileInChunks(task FileCopyTask, f *os.File,
...
-					}
-					if assignResult.Error != "" {
-						return fmt.Errorf("assign volume failure %v: %v", request, assignResult.Error)
-					}
-					return nil
+			for retryI := 0; retryI < 3; retryI++ {
+				// assign a volume
+				var assignResult *filer_pb.AssignVolumeResponse
+				var assignError error
+				err := util.Retry("assignVolume", func() error {
```
chrislusf: this is another retry inside.

Please use util.Retry for uploading to the volume servers only.

chrislusf:
this should not be needed any more. Added a better retry approach in https://github.com/seaweedfs/seaweedfs/commit/6c8822f269c1dbddcc80383137417b5f071652b0

PR #2678, opened 2022-02-15
filer: fix uninterruptible ListDirectoryEntries calls

Bl1tz23:
Hey!
This PR contains fix for uninterruptible ListDirectoryEntries calls (the reason of high RPS in the actualStore).


chrislusf:
Even though this trick works, I feel this is not fixing some real issue. I will leave this PR here, until the real reason is found.

chrislusf:
could it be related to MongoDB read consistency setting?

PR #2680, opened 2022-02-16
healthz check to avoid drain node with last replicas

kmlebedev:


LazyDBA247-Anyvision:
@kmlebedev Hi, can the new healthz route be used as a k8s liveness check?
what will a new "empty" volume server response will be?

kmlebedev:
> @kmlebedev Hi, can the new healthz route be used as a k8s liveness check? what will a new "empty" volume server response will be?

No you can not. i only use for startupprobe

LazyDBA247-Anyvision:
in general /healthz route is saved for liveness (usually) 
and we sure need /healthz in volume server, running /status as an /healthz is taking too many resources.

kmlebedev:
> in general /healthz route is saved for liveness (usually) and we sure need /healthz in volume server, running /status as an /healthz is taking too many resources.

I agree, but so far I don’t see that you can check

PR #2690, opened 2022-02-20
filer_web: support moving entry

banjiaojuhao:
add "moving entry" to filer web server:
```bash
> curl -X POST 'http://localhost:8888/path/to/dst_file?from=/path/2/src_file&pretty=yes'
```


chrislusf:
Thanks! Please also added it to the wiki.

banjiaojuhao:
done

PR #2694, opened 2022-02-21
filer_http: support uploading file with offset

banjiaojuhao:
```bash
# upload data to "offset" position of some_file
> curl -F file=@report.js "http://localhost:8888/javascript/some_file?offset=10"
```

### DIFF
--- a/weed/server/filer_server_handlers_write_autochunk.go
+++ b/weed/server/filer_server_handlers_write_autochunk.go
@@ -126,10 +126,6 @@ func (fs *FilerServer) doPutAutoChunk(ctx context.Context, w http.ResponseWriter
 	return
 }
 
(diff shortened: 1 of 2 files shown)


chrislusf:
This is nice. But what are the possible use cases?

banjiaojuhao:
> This is nice. But what are the possible use cases?

When uploading large files, user can upload one smaller block in every request, then re-upload blocks that failed to be uploaded according to file metadata.

chrislusf:
This API seems simpler than what https://news.ycombinator.com/item?id=27019287 mentioned https://tools.ietf.org/id/draft-wright-http-partial-upload-01.html

chrislusf:
This could be used to resume uploading a large file, or upload in chunks.

PR #2698, opened 2022-02-23
fix another return value bug

guo-sj:


qieqieplus:
This is a feature called "named return values"
Check out: https://go.dev/tour/basics/7

guo-sj:
@qieqieplus Yes, you are right. Thanks for pointing that.

PR #2700, opened 2022-02-23
filer: support shallow copy file

banjiaojuhao:
This can be used to take snapshots for files and directories.
```bash
# copy files and directories
> curl -X POST 'http://localhost:8888/path/to/dst_file?from=/path/2/src_file&deleteSource=false&pretty=yes'
# move files and directories
> curl -X POST 'http://localhost:8888/path/to/dst_file?from=/path/2/src_file&deleteSource=true&pretty=yes'

### DIFF
--- a/weed/server/filer_copy.go
+++ b/weed/server/filer_copy.go
@@ -1,0 +1,219 @@
+package weed_server
+
+import (
+	"context"
+	"fmt"
+	"github.com/chrislusf/seaweedfs/weed/operation"
+	"strings"
+	"time"
+
+	"github.com/chrislusf/seaweedfs/weed/filer"
+	"github.com/chrislusf/seaweedfs/weed/glog"
+	"github.com/chrislusf/seaweedfs/weed/pb/filer_pb"
+	"github.com/chrislusf/seaweedfs/weed/util"
+)
+
+func (fs *FilerServer) Copy(ctx context.Context, stream filer_pb.SeaweedFiler_StreamRenameEntryServer, src, dst string, so *operation.StorageOption, deleteSource bool) (err error) {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by chrislusf** [comment]
why this is needed? Two entries will point to the same set of file chunks. What should happen if one of the file is deleted?

**Comment by chrislusf** [comment]
In `weed mount`, symlink is supported. We can use the same mechanism to support read symlink-ed files.

**Comment by banjiaojuhao** [comment]
> why this is needed? Two entries will point to the same set of file chunks. What should happen if one of the file is deleted?

This can be used to save the state of some large files and directories very quickly.
I should also make delete files not deleting chunks. Then delete chunks by GC periodically.

**Comment by chrislusf** [comment]
If the source file is deleted, should the chunks be deleted? If so, the copied files would not be complete. If not, a reference counter is needed?

**Comment by banjiaojuhao** [comment]
> If the source file is deleted, should the chunks be deleted? If so, the copied files would not be complete. If not, a reference counter is needed?

use reference counter or scan unused chunks periodically in background.
```python
def delete_unused_chunks():
  used_chunks = get_all_chunks_from_all_entry()
  ts = get_timestamp() - 12.hours()
  for chunk in all_chunks:
    if chunk not in used_chunks and chunk.create_time < ts:
      delete(chunk)
```

chrislusf:
This would not scale when there are billions of files.

banjiaojuhao:
> 

Is reference counter acceptable?

chrislusf:
> > 
> 
> Is reference counter acceptable?

Maybe. But it is better not to include these complicated features, which can be very hard to support when combined with other features.

chrislusf:
The hard link implementation in "weed mount" is most similar to this. It already has a reference counter. But this may have problem with other features, such as replicating to remote sites.

PR #2729, opened 2022-03-06
filer: support get metadata with resolved manifest chunk

banjiaojuhao:
Some chunks in the metadata is manifest, which is not usable for client to get chunk checksum.
```bash
curl  -H "Accept: application/json" "http://localhost:8888/weed?metadata&resolveManifest=true&pretty=yes"
```

### DIFF
--- a/weed/server/filer_server_handlers_read.go
+++ b/weed/server/filer_server_handlers_read.go
@@ -119,7 +119,16 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
-	if r.URL.Query().Has("metadata") {
+	query := r.URL.Query()
+	if query.Get("metadata") == "true" {
+		if query.Get("resolveManifest") == "true" {
+			if entry.Chunks, _, err = filer.ResolveChunkManifest(
+				fs.filer.MasterClient.GetLookupFileIdFunction(),
+				entry.Chunks, 0, int64(entry.Size())); err != nil {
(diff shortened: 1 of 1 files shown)


weed/server/filer_server_handlers_read.go
```diff
@@ -119,7 +119,16 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if r.URL.Query().Has("metadata") {
+	query := r.URL.Query()
+	if query.Get("metadata") == "true" || query.Has("metadata") {
+		if query.Get("resolveManifest") == "true" || query.Has("resolveManifest") {
```
chrislusf: these query.Has() logic is wrong. If `metadata=false`, this would still work.

Please test the code before submitting PR.

banjiaojuhao: 😂sorry

banjiaojuhao:
fixed @chrislusf

weed/server/filer_server_handlers_read.go
```diff
@@ -119,7 +119,16 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request)
 		return
 	}
 
-	if r.URL.Query().Has("metadata") {
+	query := r.URL.Query()
+	if query.Has("metadata") {
```
chrislusf: Please be consistent and use "metadata=true". Please also fix the wiki.

banjiaojuhao: > Please be consistent and use "metadata=true". Please also fix the wiki.

@chrislusf done

PR #2741, opened 2022-03-10
fix:support some databases which not support 'IF NOT EXISTS' keyword.

zhuqunzhou:
support some databases(such as:[BaikalDB)](https://github.com/baidu/BaikalDB) which not support `IF NOT EXISTS` keyword in the `create table` command.

wecom-temp-4df2e3090c3a103d08643e383a802190

### DIFF
--- a/weed/filer/mysql2/mysql2_store.go
+++ b/weed/filer/mysql2/mysql2_store.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/chrislusf/seaweedfs/weed/filer"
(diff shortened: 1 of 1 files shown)


chrislusf:
Interesting! How does SeaweedFS run on BaikalDB?

zhuqunzhou:
> Interesting! How does SeaweedFS run on BaikalDB?

testing

chrislusf:
how is the test going?

zhuqunzhou:
> how is the test going?

Everything is fine except need to create table like `filemeta` first.Because`baikalDB` create tables is asynchronous.
From the perspective of performance,data consistency and others,we choose `baikalDB`  to store the filer metadata at last.

chrislusf:
Thanks! It'll be great if you can blog about this. Not many people have the opportunity to try with it and experience the performance.

PR #2749, opened 2022-03-11
filer: refactor fullpath to handle invalid path

banjiaojuhao:
I refactored `fullpath.go`. 
It supports:
1. Directory literal (ends with /) switch and detect. (no need to detect, add, remove `/` suffix directly anymore)
2. Replace invalid UTF-8 characters with `?`.
3. Handle `.` `..` and `//` in path.

Then refactored some codes related with `fullpath`.


chrislusf:
The pr is too big. Need to split into smaller ones.

PR #2752, opened 2022-03-12
filer: refactor fullpath

banjiaojuhao:
#2749
Refactor fullpath and add test case.

### DIFF
--- a/weed/util/fullpath.go
+++ b/weed/util/fullpath.go
@@ -2,7 +2,7 @@ package util
 
 import (
 	"os"
-	"path/filepath"
+	"path"
 	"strings"
 )
 
@@ -18,7 +18,7 @@ func NewFullPath(name ...string) FullPath {
 }
 
 func (fp FullPath) DirAndName() (string, string) {
-	dir, name := filepath.Split(clearName(string(fp)))
+	dir, name := path.Split(clearName(string(fp)))
 	if dir == "/" {
 		return dir, name
 	}
@@ -26,26 +26,6 @@ func (fp FullPath) DirAndName() (string, string) {
 		return "/", ""
 	}
 	return dir[:len(dir)-1], name
-}
-
-func (fp FullPath) ToDir() FullPath {
-	return fp.Child("./")
-}
-
-func (fp FullPath) ToFile() FullPath {
-	file := fp.Child(".")
-	if string(file) == "/" {
-		return "."
-	}
-	return file
-}
-
-func (fp FullPath) IsFile() bool {
-	return !fp.IsDir()
-}
-
-func (fp FullPath) IsDir() bool {
-	return strings.HasSuffix(string(fp), "/")
 }
 
 func (fp FullPath) Name() string {
@@ -85,19 +65,18 @@ func (fp FullPath) AsInode(fileMode os.FileMode) uint64 {
 }
 
 func Join(names ...string) string {
-	newName := clearName(filepath.ToSlash(filepath.Join(names...)))
-	if newName != "/" && strings.HasSuffix(filepath.ToSlash(strings.Join(names, "")), "/") {
-		newName += "/"
-	}
-	return newName
+	return clearName(path.Join(names...))
 }
 
 func clearName(name string) string {
 	name = strings.ToValidUTF8(name, "?")
-	slashed := strings.HasSuffix(name, "/")
-	name = filepath.Clean(name)
-	if name != "/" && slashed {
-		name += "/"
+	name = strings.ReplaceAll(name, "\\", "/")
+	name = path.Clean(name)
+	if name == "." {
+		name = "/"
+	}
+	if !strings.HasPrefix(name, "/") {
+		name = "/" + name
 	}
 	return name
 }
--- a/weed/util/fullpath.go
+++ b/weed/util/fullpath.go
@@ -6,15 +6,19 @@ import (
 	"strings"
 )
 
+// FullPath will keep the literal passed in.
+// file literal: /path/to/file
+// directory literal: /path/
+// change to file literal: fp.Child(".") or NewFullPath(name, ".")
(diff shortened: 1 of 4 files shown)


weed/util/fullpath.go
```diff
@@ -6,15 +6,19 @@ import (
 	"strings"
 )
 
+// FullPath will keep the literal passed in.
+// file literal: /path/to/file
+// directory literal: /path/
```
chrislusf, line 10: The previous assumption was to have `/path/to/file` format always. The fullpath will not end with "/" unless it is root.

banjiaojuhao: Scenarios like below need path literal to represent a path.
https://github.com/chrislusf/seaweedfs/blob/b40d252761e92387b281b727d490088ad0bfdab5/weed/server/filer_server_handlers_write_autochunk.go#L147-L160

chrislusf: Current logic works well. What does it need now?

banjiaojuhao: For seaweed, it's better to have a **uniform** way (using this fullpath "class", which is more clean and readable) rather than string operation (which is more error-prone) to handle path (**resolve special dir like "." and ".."; replace invalid utf-8 characters with "?"**).

And in fullpath, we need to **distinguish directories from files**. For example, user upload file to dir instead of file by http request. (upload xxx.txt to /x/y/). Supporting dir literal will not save files uploaded to dir into file (save to /x/y mistakenly).

chrislusf: The current uniform way is to use `/parent/dir/entry_name`, where entry_name can be a file or a directory.

Supporting both `/x/y` and `/x/y/` will be confusing. There are no sure way to tell `/x/y` is a directory or not. Having a `/x/y/` will only add more confusion.

The path resolving, cleaning or distinguishing problem can be solved before the path "entering" the system.

banjiaojuhao: Almost all handlers use fullpath to resolve path. So fullpath is a "common place" to add path resolve.

> The path resolving, cleaning or distinguishing problem can be solved **before the path "entering" the system.**

If we choose other way, there is no such "common place".
For http server, we can add code like `r.URL.Path = clearPath(r.URL.Path)` in `filer_server_handlers.go` after line 16.
But for grpc server, there is no such "common place". We have to clear path in every handle respectively.
https://github.com/chrislusf/seaweedfs/blob/f70c1e449b9f80545ff7e6c1c88f2c2e27e4d0e8/weed/server/filer_server_handlers.go#L16-L25

chrislusf: The change is trying too many things and no clear assumption. This will add more confusion.

banjiaojuhao: > The current uniform way is to use `/parent/dir/entry_name`, where entry_name can be a file or a directory.
> 
> Supporting both `/x/y` and `/x/y/` will be confusing. There are no sure way to tell `/x/y` is a directory or not. Having a `/x/y/` will only add more confusion.
> 
> The path resolving, cleaning or distinguishing problem can be solved before the path "entering" the system.

OK! Follow this uniform and use `NewFullPath` manually to clean path before use the path.

weed/util/fullpath.go
```diff
@@ -64,19 +44,18 @@ func (fp FullPath) AsInode(unixTime int64) uint64 {
...
 func clearName(name string) string {
 	name = strings.ToValidUTF8(name, "?")
-	slashed := strings.HasSuffix(name, "/")
-	name = filepath.Clean(name)
-	if name != "/" && slashed {
-		name += "/"
+	name = strings.ReplaceAll(name, "\\", "/")
+	name = path.Clean(name)
+	if name == "." {
+		name = "/"
```
chrislusf, line 55: this seems wrong.

weed/util/fullpath.go
```diff
@@ -64,19 +44,18 @@ func (fp FullPath) AsInode(unixTime int64) uint64 {
...
+	return clearName(path.Join(names...))
 }
 
 func clearName(name string) string {
 	name = strings.ToValidUTF8(name, "?")
-	slashed := strings.HasSuffix(name, "/")
-	name = filepath.Clean(name)
-	if name != "/" && slashed {
-		name += "/"
+	name = strings.ReplaceAll(name, "\\", "/")
```
chrislusf, line 52: this is a new logic I am not sure. There could be cases where a non-linux env has a back slash.

banjiaojuhao: Should the file separator saved in FilerStore always be "/"? Or OS-dependent?

weed/util/fullpath.go
banjiaojuhao: Yes, it should add `/` prefix before clean the path.

weed/util/fullpath.go
chrislusf: It could also be some start of escaping characters. Do not add logic here.

weed/util/fullpath.go
```diff
@@ -1,19 +1,23 @@
...
+// FullPath will keep the literal passed in.
+// file literal: /path/to/file
+// directory literal: /path/
+// change to file literal: fp.Child(".") or NewFullPath(name, ".")
+// change to directory literal: fp.Child("./") or NewFullPath(name, "./")
 type FullPath string
 
-func NewFullPath(dir, name string) FullPath {
-	return FullPath(dir).Child(name)
+func NewFullPath(name ...string) FullPath {
```
chrislusf, line 15: Why need to support array here? This was not intended as a replacement for Join.

chrislusf:
For this PR, I feel confused about the intension. If still needed, please split into more specific PR.

banjiaojuhao:
> For this PR, I feel confused about the intension. If still needed, please split into more specific PR.

Maybe a simple `CleanPath` function is enough.

PR #2758, opened 2022-03-15
it is safer to discard requests that have reached the concurrent limit

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2755

### DIFF
--- a/weed/server/volume_server_handlers.go
+++ b/weed/server/volume_server_handlers.go
@@ -1,6 +1,7 @@
 package weed_server
 
 import (
+	"fmt"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


weed/server/volume_server_handlers.go
```diff
@@ -52,9 +53,11 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
 		// wait until in flight data is less than the limit
 		contentLength := getContentLength(r)
 		vs.inFlightUploadDataLimitCond.L.Lock()
```
chrislusf, line 55: If the wait is removed, the lock/unlock is not needed either.

chrislusf:
fixed with 0562fceb9925f6c2cd13b68f24d29dd1fd4c3e5e

PR #2760, opened 2022-03-16
Add mTLS support for both master and volume http server.

bercknash:
This adds TLS support to the master server, and mTLS support to both the master and volume server conditional on specifying the correct certificate files in security.toml.  If this is acceptable, I'm happy to also update the necessary documentation.

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -1,15 +1,17 @@
 package command
 
 import (
-	"github.com/chrislusf/raft/protobuf"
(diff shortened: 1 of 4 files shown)


chrislusf:
Looks good! Currently the existing clients are only using http to connect to volume and master servers. Is this intended for use cases without filer?

bercknash:
Correct.  We can't currently do grpc, but need mTLS.

PR #2761, opened 2022-03-17
filer: support uploading file without needEnsureParentDir

zhuqunzhou:
Mostly,the parent dir has already exists or i can make the parent dir manually in advance.
Use the parameter,i can reduce one `metadata` request on per `upload` request.

### DIFF
--- a/weed/filer/filer.go
+++ b/weed/filer/filer.go
@@ -151,7 +151,7 @@ func (f *Filer) RollbackTransaction(ctx context.Context) error {
 	return f.Store.RollbackTransaction(ctx)
 }
 
-func (f *Filer) CreateEntry(ctx context.Context, entry *Entry, o_excl bool, isFromOtherCluster bool, signatures []int32) error {
+func (f *Filer) CreateEntry(ctx context.Context, entry *Entry, o_excl bool, isFromOtherCluster bool, signatures []int32,needEnsureParentDir bool) error {
 
 	if string(entry.FullPath) == "/" {
 		return nil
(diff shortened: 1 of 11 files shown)


chrislusf:
This `needEnsureParentDir` is default to true, which is current behavior.

I am confused. Do you want to ensure parent directory or not? If so, it is already current behavior.

zhuqunzhou:
> This `needEnsureParentDir` is default to true, which is current behavior.
> 
> I am confused. Do you want to ensure parent directory or not? If so, it is already current behavior.

By default,i want to ensure parent dir exists,It makes `weedfs` particularly easy to use. 
But,I use `curl -F file=@report.js "http://localhost:8888/javascript/some_file?ensureParentDir=false` in my product environment to reduce the number of metadata requests.

In my case,I use `BaikalDB` to store the metadata, about `7ms` to upload a file,but `3ms` used to request the metadata service.The parameter can improve performance by about 14%(7ms -> 6ms)

chrislusf:
Please do these:
* go fmt your changes.
* add to wiki about this API changes and why. https://github.com/chrislusf/seaweedfs/wiki/Filer-Server-API

zhuqunzhou:
> 



> Please do these:
> 
> * go fmt your changes.
> * After the PR is merged, add to wiki about this API changes and what is the benefit, as you explained here. https://github.com/chrislusf/seaweedfs/wiki/Filer-Server-API

done

chrislusf:
I changed the code and doc to use `skipCheckParentDir`, which defaults to false.

PR #2763, opened 2022-03-17
filer: support uploading file skip check expired entry

zhuqunzhou:
1.Like [2761](https://github.com/chrislusf/seaweedfs/pull/2761).The `skipCheckExpiredEntry` parameter also can reduce one `metadata` request per upload request.
2.fix the bug: Calculation error of method `skipCheckParentDirEntry`


chrislusf:
the proto file change is not needed.

weed/server/filer_server_handlers_write_autochunk.go
```diff
@@ -154,7 +158,7 @@ func (fs *FilerServer) saveMetaData(ctx context.Context, r *http.Request, fileNa
 			path += fileName
 		}
 	} else {
-		if fileName != "" {
+		if fileName != "" && !skipCheckExpiredEntry(r) {
```
chrislusf: this is a wrong name. There are no expired entry here.

weed/pb/filer_pb/filer.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
-// 	protoc-gen-go v1.26.0
```
chrislusf: no needed.

famosss: ok

chrislusf:
I want to avoid having too many "skipXXX" options.

This can be avoided by uploading to "http://filer:port/path/to/dir/", instead of "http://filer:port/path/to/dir/file"?

zhuqunzhou:
> I want to avoid having too many "skipXXX" options.
> 
> This extra lookup can be avoided by uploading to `http://filer:port/path/to/dir/`, instead of `http://filer:port/path/to/dir/file`?

Ok. Please `close` it.

PR #2813, opened 2022-03-25
always clear previous log to avoid server is promotable

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2804

### DIFF
--- a/weed/server/raft_server.go
+++ b/weed/server/raft_server.go
@@ -81,10 +81,11 @@ func NewRaftServer(option *RaftServerOption) (*RaftServer, error) {
 	transporter := raft.NewGrpcTransporter(option.GrpcDialOption)
 	glog.V(0).Infof("Starting RaftServer with %v", option.ServerAddr)
 
+	// always clear previous log to avoid server is promotable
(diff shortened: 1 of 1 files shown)


weed/server/raft_server.go
```diff
@@ -81,10 +81,11 @@ func NewRaftServer(option *RaftServerOption) (*RaftServer, error) {
 	transporter := raft.NewGrpcTransporter(option.GrpcDialOption)
 	glog.V(0).Infof("Starting RaftServer with %v", option.ServerAddr)
 
+	// always clear previous log to avoid server is promotable
+	os.RemoveAll(path.Join(s.dataDir, "log"))
```
chrislusf, line 85: why not set option.RaftResumeState to false when starting the master?

kmlebedev: > why not set option.RaftResumeState to false when starting the master?

But it seems to me, on the contrary, the default value of RaftResumeState should be set to true.

This allows, after turning off all the master servers, to get the topology and MaxVolumeId
https://github.com/chrislusf/raft/pull/4/files

PR #2815, opened 2022-03-25
Added checkBucket to PutObject and NewMultipartUpload.

pavelzhurov:
Without these checks user can unintentionally create bucket, while putting object.


kmlebedev:
@pavelzhurov Why ErrNoSuchBucketPolicy ?

pavelzhurov:
@kmlebedev My logic was the following: a bucket isn't able to have any Bucket Policy, so for any bucket we can return that there is no such policy. I guess the same logic is applied for GetBucketPolicy.

kmlebedev:
> @kmlebedev My logic was the following: a bucket isn't able to have any Bucket Policy, so for any bucket we can return that there is no such policy. I guess the same logic is applied for GetBucketPolicy.

This is the correct logic, but some s3 client applications may not agree with you.

pavelzhurov:
@kmlebedev Agree, changed response function to WriteEmptyError with 204 code

chrislusf:
The lady bucket creation was intentional. Adding an extra check will slow down writes.

chrislusf:
Better avoid too many options. How about adding an in memory cache of existing buckets, and only check the bucket if not in the cache?

kmlebedev:
It seems that at the start of s3 it is not difficult to get a list of packages
@chrislusf But it is not clear how best to update, it seems that you need a subscription to the filer for deletion and addition, approximately how it works on the replicator now. Or is there an easier option?

chrislusf:
Metadata event subscription has some delays. 

A TTL-Ed cache should work.

pavelzhurov:
@chrislusf but what about putting object immediately after bucket creation? If we use TTLed cache, we should wait at most TTL seconds to put object, because cache hasn't been updated yet. And updating it on PutBucket event won't help, because we can have more than one replica of s3api.

pavelzhurov:
@chrislusf  Yeah, I got it. The problem lies down in the cache update process. Suppose we've initialised it at the startup process, and then periodically updating it using TTL as a sync period value, or on each create bucket event. In first case, after bucket creation, we have to wait until next sync to upload an object in this bucket. In second one, cache will be updated only in the replica which handled PutBucket request.
The only option I see here, is to make TTL very small, but it still will produce occasional errors (especially if user automates buckets/objects creation process).
Maybe there is some other approach which I missed?

go.mod
```diff
@@ -180,6 +180,7 @@ require (
 	github.com/jcmturner/dnsutils/v2 v2.0.0 // indirect
 	github.com/jcmturner/goidentity/v6 v6.0.1 // indirect
 	github.com/jcmturner/rpc/v2 v2.0.3 // indirect
+	github.com/jellydator/ttlcache/v3 v3.0.0
```
chrislusf: could you use github.com/karlseguin/ccache ? it is already in the use.

weed/command/s3.go
```diff
@@ -49,6 +51,7 @@ func init() {
 	s3StandaloneOptions.tlsCertificate = cmdS3.Flag.String("cert.file", "", "path to the TLS certificate file")
 	s3StandaloneOptions.metricsHttpPort = cmdS3.Flag.Int("metricsPort", 0, "Prometheus metrics listen port")
 	s3StandaloneOptions.allowEmptyFolder = cmdS3.Flag.Bool("allowEmptyFolder", true, "allow empty folders")
+	s3StandaloneOptions.bucketsCacheTTL = cmdS3.Flag.Uint("bucketsCacheTTL", 600, "TTL of buckets cache entries in seconds. default is 600.")
```
chrislusf: No need for an extra option. Just use a reasonable number.

imfantuan:
Is `s3a.exists` better than `s3a.checkBucket` here?

imfantuan:
> @kmlebedev Agree, changed response function to WriteEmptyError with 204 code

How about `NoSuchBucket` or `InvalidRequest `?

kmlebedev:
> How about `NoSuchBucket` or `InvalidRequest`

No need to change handles from cors and policy as they are not implemented

chrislusf:
How about put this option into the filer.toml file? Similar to this?

```
[filer.options]
# with http DELETE, by default the filer would check whether a folder is empty.
# recursive_delete will delete all sub folders and files, similar to "rm -Rf"
recursive_delete = false
```

pavelzhurov:
@chrislusf Isn't this file reserved for filer options only? If so, we can't put it there, since the check is occurred before s3api calls filer.

kmlebedev:
> How about put this option into the filer.toml file? Similar to this?
> 
> ```
> [filer.options]
> # with http DELETE, by default the filer would check whether a folder is empty.
> # recursive_delete will delete all sub folders and files, similar to "rm -Rf"
> recursive_delete = false
> ```

S3 server does not have a configuration file, it seems to create it?

PR #2828, opened 2022-03-28
Fix delete all "Seaweed-" prefixed tagging error

guo-sj:


weed/server/filer_server_handlers_tagging.go
```diff
@@ -81,8 +81,10 @@ func (fs *FilerServer) DeleteTaggingHandler(w http.ResponseWriter, r *http.Reque
 	// parse out tags to be deleted
 	toDelete := strings.Split(r.URL.Query().Get("tagging"), ",")
 	deletions := make(map[string]struct{})
-	for _, deletion := range toDelete {
-		deletions[deletion] = struct{}{}
```
chrislusf, line 85: You can also just fix by checking `deletion == ""` or not.

guo-sj: I've updated it accordingly.

PR #2839, opened 2022-03-29
Enable meta aggregator for internal storages only

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2545


chrislusf:
The metadata aggregator is also needed in many other features, e.g., metadata backup, cross-cluster replication, weed mount, etc.

kmlebedev:
> The metadata aggregator is also needed in many other features, e.g., metadata backup, cross-cluster replication, weed mount, etc.

Maybe then make a parameter in the config, where aggregation will be enabled by default?

chrislusf:
this will also impact basic stuff, such as "s3.configure" would not apply to other s3 instances unless restarted.

PR #2857, opened 2022-03-31
fix server address for grpcPort

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2847


chrislusf:
the grpc port is customizable.

PR #2868, opened 2022-04-04
hashicorp raft

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2804


chrislusf:
The logic is mixed together. How about adding some interfaces instead of if/else for the raft server checking?

Please separate into 2 PRs. The first do the refactoring, and the second implement the hashicorp raft.

kmlebedev:
There will be no long-term profit from this, if the hashicorp experiment is successful, then the old raft will need to be removed. 

So for the transitional period, this is quite acceptable and I don’t want to overload the logic with another interface, and after removing the old raft, deep refactoring will be needed.

Therefore, you do not want to do refactoring for the sake of refactoring prematurely.

chrislusf:
ok. let me know when this PR is stable enough.

weed/server/raft_hashicorp.go
```diff
@@ -25,7 +73,7 @@ func NewHashicorpRaftServer(option *RaftServerOption) (*RaftServer, error) {
 	}
 
 	c := raft.DefaultConfig()
-	c.LocalID = raft.ServerID(s.serverAddr) // TODO maybee the IP:port address will change
+	c.LocalID = raft.ServerID(s.serverAddr.String()) // TODO maybee the IP:port address will change
```
chrislusf: Need to use the raw string instead of String() function. 
String() returns server + http port only.

weed/server/raft_hashicorp.go
```diff
@@ -66,32 +114,17 @@ func NewHashicorpRaftServer(option *RaftServerOption) (*RaftServer, error) {
...
-			if peer == option.ServerAddr {
-				continue
-			}
-			cfg.Servers = append(cfg.Servers, raft.Server{
-				Suffrage: raft.Voter,
-				ID:       raft.ServerID(peer),
-				Address:  raft.ServerAddress(peer.ToGrpcAddress()),
-			})
-		}
+	if option.RaftBootstrap || len(s.RaftHashicorp.GetConfiguration().Configuration().Servers) == 0 {
```
chrislusf, line 150: Need to wait some random time, in case all servers do this at the same time.

kmlebedev: A good decision would be to take a lock to the one who is the first

chrislusf:
The raft algorithm is basically the same. The library is updated, which is nice. But it may not "solve" some issues. Just be careful not to introduce new problems and take more time to test.

weed/server/raft_hashicorp.go
kmlebedev: add wait random time
```
I0405 10:39:21     1 raft_hashicorp.go:141] Bootstrapping idx: 0 sleep: 0s new cluster: {Servers:[{Suffrage:Voter ID:master2:9335 Address:master2:19335} {Suffrage:Voter ID:master0:9333 Address:master0:19333} {Suffrage:Voter ID:master1:9334 Address:master1:19334}]}

I0405 10:39:20     1 raft_hashicorp.go:141] Bootstrapping idx: 1 sleep: 357.129207ms new cluster: {Servers:[{Suffrage:Voter ID:master0:9333 Address:master0:19333} {Suffrage:Voter ID:master2:9335 Address:master2:19335} {Suffrage:Voter ID:master1:9334 Address:master1:19334}]}

I0405 10:39:20     1 raft_hashicorp.go:141] Bootstrapping idx: 2 sleep: 832.2263ms new cluster: {Servers:[{Suffrage:Voter ID:master1:9334 Address:master1:19334} {Suffrage:Voter ID:master2:9335 Address:master2:19335} {Suffrage:Voter ID:master0:9333 Address:master0:19333}]}
```

kmlebedev:
> The raft algorithm is basically the same. The library is updated, which is nice. But it may not "solve" some issues. Just be careful not to introduce new problems and take more time to test.

@chrislusf 
I've already tested
1. create a new cluster
2. reboot the entire cluster with restored max volume id
3. reboot the entire cluster with a change in peers

kmlebedev:
@chrislusf PR deployed in testing, everything is fine

chrislusf:
Shell commands should be optional, not required for adding/removing servers.

kmlebedev:
> Shell commands should be optional, not required for adding/removing servers.

I'm currently working on it, but somehow OnPeerUpdate is not being called obviously

chrislusf:
This 100 should not matter.

kmlebedev:
> This 100 should not matter.

@chrislusf exactly ?
This is bug, see https://go.dev/play/p/qXWzGiL0HPt

kmlebedev:
> I do not understand why need to remove old raft server. Is it not working?

Overcame. I wrote to a channel that did not have a reader, so the onPeerUpdate function was blocked accordingly

weed/server/raft_hashicorp.go
```diff
@@ -0,0 +1,182 @@
...
+				peerLeader := string(s.serverAddr)
+				existsPeerName := make(map[string]bool)
+				for _, server := range s.RaftHashicorp.GetConfiguration().Configuration().Servers {
+					if string(server.ID) == peerLeader {
+						continue
+					}
+					existsPeerName[string(server.ID)] = true
+				}
+				for _, peer := range s.peers {
+					if peer.String() == peerLeader || existsPeerName[peer.String()] {
```
chrislusf: need to avoid using serverAddress.String() here. Use the full ServerAddress string.

chrislusf:
LGTM in general. Thanks for the effort!

* Need to use the full ServerAddress as the identifier.
* May use "cluster.raft.{ps,add,remove}" instead of "raft.server.{add,remove}" and "raft.cluster.ps".

kmlebedev:
@kmlebedev Done

kungf:
hi, @chrislusf @kmlebedev It's time to remove the old raft? It was so confused about two rafts left in the code！
which one should I choose？

kmlebedev:
@kungf Now it seems that both raft are working fine, but changing peers without deleting the state is supported only in hashicorp raft.

kungf:
@kmlebedev thanks for your reply！ Can I upgrade raft to hashicorp in old cluster safely with lost snapshot？

kmlebedev:
> @kmlebedev thanks for your reply！ Can I upgrade raft to hashicorp in old cluster safely with lost snapshot？

No, the update must be performed while the service is stopped. And run the cluster in RO mode (with blocking on the ingress side of PUT and POST requests) until all volumes are read up to 15 seconds.

PR #2879, opened 2022-04-06
remove Redundant type conversion and use strings.TrimSuffix to enhance readability

leyou240:
remove Redundant type conversion and use strings.TrimSuffix to enhance readability


chrislusf:
Thanks! Good to learn this!

PR #2885, opened 2022-04-07
fix: master lose some volumes

binbinshi:
how to reproduce the error ; 

1.  start a volume and a master  
2. Add code in the for loop of master_grpc_server.go, the for loop return after some times;   The purpose is to enter the defer func of SendHeartbeat;
3.  debug or time.Sleep in defer func , now restart volume， repeat register to master ;
4.  ms.topo will remove the dataNode , but at the same time volume register in topo ;
5. topo has the dn, but the dn's volumes not removed and sync to filer and master;
6. volume-server heartbeat sync is normal but fiiler and master doesn't know the volumes info ; 
7. so some volumes are lose can't get .

### DIFF
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -113,6 +113,9 @@ func (ms *MasterServer) SendHeartbeat(stream master_pb.Seaweed_SendHeartbeatServ
 		}
 
 		if len(heartbeat.Volumes) > 0 || heartbeat.HasNoVolumes {
+			dcName, rackName := ms.Topo.Configuration.Locate(heartbeat.Ip, heartbeat.DataCenter, heartbeat.Rack)
+			ms.Topo.DataNodeRegistration(dcName, rackName, dn)
+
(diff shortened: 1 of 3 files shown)


weed/server/master_grpc_server.go
```diff
@@ -113,6 +113,9 @@ func (ms *MasterServer) SendHeartbeat(stream master_pb.Seaweed_SendHeartbeatServ
 		}
 
 		if len(heartbeat.Volumes) > 0 || heartbeat.HasNoVolumes {
+			dcName, rackName := ms.Topo.Configuration.Locate(heartbeat.Ip, heartbeat.DataCenter, heartbeat.Rack)
+			ms.Topo.DataNodeRegistration(dcName, rackName, dn)
```
chrislusf, line 117: Also need to do it this for EcVolumes?

binbinshi: ecShardTickChan is 17 times larger than volumeTickChan so I don't think it's necessary.

PR #2886, opened 2022-04-07
Change filer-ui to bootstrap style

blacktear23:
Change filer-ui to bootstrap style.

As seaweedfs already contains bootstrap css so it is easy to change the UI style to bootstrap style.


chrislusf:
Can the "Rename" and "Delete" be changed to some icon? Similar to the folder icon size.

blacktear23:
@chrislusf what about using bootstrap's icon instead of gif for button icons?

chrislusf:
that would be better.

PR #2900, opened 2022-04-11
avoid breaking loop in cleanupUploads if error is empty

kmlebedev:


weed/shell/command_s3_clean_uploads.go
```diff
@@ -82,7 +84,7 @@ func (c *commandS3CleanUploads) cleanupUploads(commandEnv *CommandEnv, writer io
 		fmt.Fprintf(writer, "purge %s\n", deleteUrl)
 
 		err = util.Delete(deleteUrl, "")
-		if err != nil {
+		if err != nil && err.Error() != "" {
```
chrislusf, line 95: the additional check is necessary?

kmlebedev: > the additional check is necessary?

@chrislusf It is needed in order to be able to delete a file from a filer without a jwt token, but it is not deleted from the volume and return empty error string. Strange behavior, why files are deleted from filer without token ?

kmlebedev: I use jwt token on filler

```
> s3.clean.uploads -timeAgo 2h
purge http://fast-api.s3.svc:9090/buckets/cloud/.uploads/b470e03b-4851-48f5-9935-67649eda2f28?recursive=true&ignoreRecursiveError=true
> fs.ls /buckets/cloud/.uploads/
```
but s3.clean.upload remove files without jwt token

PR #2901, opened 2022-04-11
check object name and uploadID when processing  multiupload

guol-fnst:
When initializing multipartUploading , a specified uploadid will be assigned.
But we can still upload part of object successfully if a wrong object name combined with right upload id are specified.
Actually a error should be returned in this situation.

Here is a example response when wrong object name or wrong upload id are specified.

> PUT /bucket1/nnnpp?uploadId=48134035-a1e8-4308-a3b4-872373405e3f&partNumber=1 HTTP/1.1
> User-Agent: curl/7.29.0
> Host: 193.168.140.103:29000
> Accept: */*
> Date: Fri, 08 Apr 2022 09:12:59 +0000
> Authorization: AWS 1VL1TGB3NMTB6DKQBSN8:xwuT9u22zjzuz+Qi9sHROj231/A=
> Content-Length: 176
> Expect: 100-continue
>
< HTTP/1.1 404 Not Found
< Accept-Ranges: bytes
< Content-Length: 349
< Content-Type: application/xml
< Server: SeaweedFS S3
< X-Amz-Request-Id: 1649408623833001509
< Date: Fri, 08 Apr 2022 09:03:43 GMT
< Connection: close
<
* Closing connection 0
<Error><Code>NoSuchUpload</Code><Message>The specified multipart upload does not exist. The upload ID may be invalid, or the upload may have been aborted or completed.</Message><Resource>/bucket1/nnnpp</Resource><RequestId>1649408623832912777</RequestId><Key>nnnpp</Key><BucketName>bucket1</BucketName></Error>

### DIFF
--- a/weed/s3api/s3api_object_multipart_handlers.go
+++ b/weed/s3api/s3api_object_multipart_handlers.go
@@ -70,6 +70,11 @@ func (s3a *S3ApiServer) CompleteMultipartUploadHandler(w http.ResponseWriter, r 
 
 	// Get upload id.
 	uploadID, _, _, _ := getObjectResources(r.URL.Query())
+	err := s3a.chkUploadID(bucket, object, uploadID)
+	if err != nil {
+		s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchUpload)
+		return
+	}
 
 	response, errCode := s3a.completeMultipartUpload(&s3.CompleteMultipartUploadInput{
 		Bucket:   aws.String(bucket),
@@ -94,6 +99,11 @@ func (s3a *S3ApiServer) AbortMultipartUploadHandler(w http.ResponseWriter, r *ht
 
 	// Get upload id.
 	uploadID, _, _, _ := getObjectResources(r.URL.Query())
+	err := s3a.chkUploadID(bucket, object, uploadID)
+	if err != nil {
+		s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchUpload)
+		return
+	}
 
 	response, errCode := s3a.abortMultipartUpload(&s3.AbortMultipartUploadInput{
 		Bucket:   aws.String(bucket),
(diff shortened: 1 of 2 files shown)


chrislusf:
this will work, but add another metadata lookup.

We can hash the object key into an upload id, and just verify the hash.

chrislusf:
this will work, but add another metadata lookup.

We can hash the object key into an upload id, and just verify the hash.

weed/s3api/s3api_object_multipart_handlers.go
```diff
@@ -205,7 +206,13 @@ func (s3a *S3ApiServer) PutObjectPartHandler(w http.ResponseWriter, r *http.Requ
...
 
 	uploadID := r.URL.Query().Get("uploadId")
-	err := s3a.chkUploadID(bucket, object, uploadID)
+	exists, err := s3a.exists(s3a.genUploadsFolder(bucket), uploadID, true)
+	if !exists {
+		s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchUpload)
+		return
+	}
+
+	err = s3a.chkUploadID(object, uploadID)
```
chrislusf: change `chkUploadID` to `checkUploadId`?

weed/s3api/s3api_object_multipart_handlers.go
```diff
@@ -205,7 +206,13 @@ func (s3a *S3ApiServer) PutObjectPartHandler(w http.ResponseWriter, r *http.Requ
 	bucket, object := xhttp.GetBucketAndObject(r)
 
 	uploadID := r.URL.Query().Get("uploadId")
-	err := s3a.chkUploadID(bucket, object, uploadID)
+	exists, err := s3a.exists(s3a.genUploadsFolder(bucket), uploadID, true)
```
chrislusf: this seems unnecessary. If the request can have the correct uploadId, it is unlikely the uploadId does not exist.

PR #2902, opened 2022-04-11
refactor and rm old raft

kmlebedev:


weed/server/master_grpc_server.go
```diff
@@ -262,7 +262,7 @@ func (ms *MasterServer) broadcastToClients(message *master_pb.KeepConnectedRespo
 
 func (ms *MasterServer) informNewLeader(stream master_pb.Seaweed_KeepConnectedServer) error {
 	leader, err := ms.Topo.Leader()
-	if err != nil {
+	if err == nil {
```
chrislusf: ?

kmlebedev: > ?

Thanks, that's a typo

kmlebedev:
@chrislusf Seems like more than enough time has passed to make sure the hashicorp raft is stable. I suggest replacing it

kmlebedev:
@chrislusf  If I understand correctly, then this is an intermediate step?
https://github.com/seaweedfs/seaweedfs/pull/3519
Or do you want to keep the old raft completely?

PR #2932, opened 2022-04-18
avoid no such cpd or cpx file

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2928


chrislusf:
Under what condition the cpd or cpx files do not exist but it is a valid compaction?

PR #2974, opened 2022-04-26
waite volume being closed during compression idx

kmlebedev:


weed/storage/volume.go
```diff
@@ -195,6 +195,10 @@ func (v *Volume) Close() {
 	v.dataFileAccessLock.Lock()
 	defer v.dataFileAccessLock.Unlock()
 	if v.nm != nil {
+		for v.isCompacting {
```
chrislusf, line 198: With this code, when the volume is committing, it can not close current volume.

kmlebedev: This is the point, since closing the volumes causes when the volume server is stopped. But it seems it's worth taking it to the stop level above.

chrislusf:
going to merge this and adjust to use `v.isCommitCompacting`

PR #2991, opened 2022-04-30
Comment out information in issue template

remram44:
This will avoid unintentionally mentioning the example issues.


chrislusf:
this was intentional

PR #2996, opened 2022-05-01
ydb filer store

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/2942


kmlebedev:
@chrislusf  Checked filer starts here

kmlebedev:
@chrislusf  I'm done, all tests passed except https://github.com/chrislusf/seaweedfs/issues/3011

weed/filer/ydb/readme.md
```diff
@@ -0,0 +1,26 @@
+## YDB
+
+database: https://github.com/ydb-platform/ydb
+
+go driver: https://github.com/ydb-platform/ydb-go-sdk
+
+options:
+
+```
+[ydb]
```
chrislusf, line 10: Need to add a section to filer.toml example to scaffold.

Also, the compiled binary will be how much larger after this PR? If it is too large, e.g., more than 5MB larger, we may need to do something similar to etcd to compile it conditionally.

kmlebedev: > Need to add a section to filer.toml example to scaffold.
> 
> Also, the compiled binary will be how much larger after this PR? If it is too large, e.g., more than 5MB larger, we may need to do something similar to etcd to compile it conditionally.

without ydb store
```
 ls -lh weed 
-rwxr-xr-x  1 lebedev_k  staff    74M May  3 23:15 weed
```

with ydb store
```
-rwxr-xr-x  1 lebedev_k  staff    78M May  3 22:57 weed
```

exactly 5 mb, like looking into the water :)

add build tag

chrislusf:
you may need to add one empty file to the ydb folder, to ensure compilation when "ydb" tag is not defined.

kmlebedev:
>  you may need to add one empty file to the ydb folder, to ensure compilation when "ydb" tag is not defined.

done

chrislusf:
Thanks! Please add it to the wiki also. https://github.com/chrislusf/seaweedfs/wiki/Filer-Stores

weed/filer/ydb/ydb_store.go
```diff
@@ -0,0 +1,403 @@
...
+		environ.WithEnvironCredentials(ctx),
+	}
+	if poolSizeLimit > 0 {
+		opts = append(opts, ydb.WithSessionPoolSizeLimit(poolSizeLimit))
+	}
+	if dsn == "" {
+		dsn = os.Getenv("YDB_CONNECTION_STRING")
+	}
+	store.DB, err = ydb.Open(ctx, dsn, opts...)
+	if err != nil || store.DB == nil {
```
rekby, line 92: You can check only err != nil.
ydb.Open will return exactly one non nil value

weed/filer/ydb/ydb_store.go
```diff
@@ -0,0 +1,403 @@
...
+		table.ValueParam("$dir_hash", types.Int64Value(util.HashStringToLong(*shortDir))),
+		table.ValueParam("$directory", types.UTF8Value(*shortDir)),
+		table.ValueParam("$start_name", types.UTF8Value(startFileName)),
+		table.ValueParam("$prefix", types.UTF8Value(prefix+"%")),
+		table.ValueParam("$limit", types.Uint64Value(uint64(limit))),
+	)
+	err = store.doTxOrDB(ctx, query, queryParams, roTX, func(res result.Result) error {
+		var name string
+		var data []byte
+		for res.NextResultSet(ctx) {
```
rekby, line 255: DB will return exactly one result set for the query.

YDB return more than one result set for few select queries at one:
```
SELECT 1;
SELECT 2;
```

or from scan query query (other queries type with other api request)

weed/filer/ydb/ydb_store.go
```diff
@@ -0,0 +1,403 @@
...
+			for res.NextRow() {
```
rekby, line 256: YDB has limit for return rows (default - 1000).
The client code must check Truncated() flag to detect if response truncated (and select next page/return error).

```
res.CurrentResultSet().Truncated()
```

weed/filer/ydb/ydb_queries.go
```diff
@@ -0,0 +1,85 @@
...
+		DECLARE $name AS Utf8;
+		DECLARE $meta AS String;
+		DECLARE $expire_at AS Optional<uint32>;
+
+		UPSERT INTO ` + asql.DEFAULT_TABLE + `
+			(dir_hash, name, directory, meta, expire_at)
+		VALUES
+			($dir_hash, $name, $directory, $meta, $expire_at);`
+
+	updateQuery = `
```
rekby, line 22: for insert query you use UPDATE - it preserve values of non inserted colums.
for update query you use REPLACE - it will reset of values in non inserted columns.  It is only differenct between UPDATE and REPLACE.
And it only differences between the queries. 

1. Now both queries write all table columns and it equals - do you need different queries?

2. If in feature you will add columns to the table, but bot in the request - update query will reset other columns values, but insert query will preserve it. May be need to swap methods and use UPSERT for update query, REPLACE to insert?

weed/filer/ydb/ydb_store_kv.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, filer.ErrKvNotFound
+	}
+
+	return value, nil
+}
+
+func (store *YdbStore) KvDelete(ctx context.Context, key []byte) (err error) {
+	dirStr, dirHash, name := abstract_sql.GenDirAndName(key)
+	return store.DB.Table().Do(ctx, func(ctx context.Context, s table.Session) (err error) {
+		_, _, err = s.Execute(ctx, rwTX, *withPragma(&store.tablePathPrefix, insertQuery),
```
rekby, line 68: insertQuery has 4 non optional parameters, you send two only.
Query will fail if not send values for non optional parameters.

chrislusf:
Please add some tests similar to `weed/filer/etcd/etcd_store_test.go` in https://github.com/chrislusf/seaweedfs/commit/d5fe5d12b74d65a65138de62e9d98cfb3bac14de

weed/filer/ydb/ydb_store.go
kmlebedev: ok

weed/filer/ydb/ydb_store.go
kmlebedev: Understood

weed/filer/ydb/ydb_queries.go
kmlebedev: Chose 1) both

weed/filer/ydb/ydb_store_kv.go
kmlebedev: Thanks, this is copy&past. replaced to deleteQuery

weed/filer/ydb/ydb_store.go
kmlebedev: @rekby How to get this page if the request already has a limit and a start name ?
Perhaps there are regular means without changing the select query ?
```
		SELECT name, meta
		FROM filemeta
		WHERE dir_hash = $dir_hash AND directory = $directory and name >= $start_name and name LIKE $prefix
		ORDER BY name ASC LIMIT $limit;`
```

kmlebedev:
> add some tests

done https://github.com/chrislusf/seaweedfs/pull/3043

PR #3021, opened 2022-05-08
Update README.md

jazeem-azeez-aicorp:
It seems there is a port error in documentation. 
service was listening to 8888 by documentation was pointing to 8333


chrislusf:
Did you run into any problems?

PR #3040, opened 2022-05-11
fixes from comments

kmlebedev:
https://github.com/chrislusf/seaweedfs/pull/2996

### DIFF
--- a/weed/filer/ydb/ydb_store.go
+++ b/weed/filer/ydb/ydb_store.go
@@ -28,6 +28,7 @@ import (
 
 const (
 	defaultDialTimeOut = 10
+	maxRowsInQuery     = 1000 // Limit number of rows in query results https://cloud.yandex.com/en-ru/docs/ydb/concepts/limits-ydb
 )
 
 var (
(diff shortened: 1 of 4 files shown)


weed/filer/ydb/ydb_store.go
```diff
@@ -242,17 +238,39 @@ func (store *YdbStore) ListDirectoryPrefixedEntries(ctx context.Context, dirPath
...
-	)
-	err = store.doTxOrDB(ctx, query, queryParams, roTX, func(res result.Result) error {
-		var name string
-		var data []byte
-		for res.NextResultSet(ctx) {
+	truncated := true
+	eachEntryFuncIsNotBreake := true
+	shortLimit := limit
+	if limit > maxRowsInQuery {
+		shortLimit = maxRowsInQuery * 2
```
rekby, line 245: You don't need maxRowsInQuery const.

You can set limit as restLimit in each query (independent of it size), then check truncated flag.
Use 1000 as const is bad idea - the limit can change in feature or in self hosted DB. In that case - code can break (get maxRowsInQuery*2 rows without truncate flag (because query limit).

weed/filer/ydb/ydb_store.go
```diff
@@ -242,17 +238,39 @@ func (store *YdbStore) ListDirectoryPrefixedEntries(ctx context.Context, dirPath
...
+	entryCount := int64(0)
+	for truncated && eachEntryFuncIsNotBreake {
+		if lastFileName != "" {
+			startFileName = lastFileName
+			if includeStartFile {
+				query = withPragma(tablePathPrefix, listDirectoryQuery)
+			}
+		}
+		restLimit := limit - entryCount
+		if maxRowsInQuery > restLimit {
```
rekby, line 256:
```
if shortLimit > restLimit {
...
```
?

PR #3047, opened 2022-05-13
fix the problem of metadata and tagging loss when files are copied

shichanglin5:
by adding processing of metadata and tagging in s3 api CopyObject (judging whether to copy or overwrite according to the directive header)


chrislusf:
Looks good. Add some tests?

shichanglin5:
Thanks for reminding. Method test has been added

PR #3056, opened 2022-05-16
avoid delete volume replica if collection mismatch

kmlebedev:


weed/shell/command_volume_fix_replication.go
```diff
@@ -200,6 +200,13 @@ func (c *commandVolumeFixReplication) deleteOneVolume(commandEnv *CommandEnv, wr
 			}
 		}
 
+		for _, volumeReplica := range replicas {
+			if volumeReplica.info.Collection != replica.info.Collection {
+				fmt.Fprintf(writer, "skip delete volume %d as collection %s is mismatch: %s\n", replica.info.Id, replica.info.Collection, volumeReplica.info.Collection)
+				continue
```
chrislusf: Is this tested? This only breaks out of the for loop in line 203. Nothing is skipped.

kmlebedev: > Is this tested? This only breaks out of the for loop in line 203. Nothing is skipped.

In the process of testing, fixed

PR #3057, opened 2022-05-16
fix build binaries with rocksdb

kmlebedev:


chrislusf:
there is a rocksdb specific docker already.

rocksdb needs OS having the right libraries installed. Just building a binary is not enough.

Issue #3069, opened 2022-05-17
Benchmark result seems not good, how can I make optimization


ningfdx:
**Describe**
I deploy a SeaweedFS cluster, 4 machine with 4*8TB HDD disk, 3 mster + 3filer + 4 ** 4=16 volume server
And when I finish deploy and run some test on it, I found the io performance is wired.
The read rate is very slow.
![Screenshot from 2022-05-17 11-35-17](https://user-images.githubusercontent.com/17705801/168723421-2d0babc4-0e02-4014-985f-146bbd0970ec.png)

![Screenshot from 2022-05-17 11-05-53](https://user-images.githubusercontent.com/17705801/168720276-5798b043-8828-428d-a30c-9db55e851852.png)

**System Setup**
./weed master -mdir=/seaweedfs/master -ip={{ansible_host}} -port=9333 -peers={{masters_ip_str}}
./weed volume -dir=/data -ip={{ansible_host}} -max=0 -port=808{{item}} -mserver={{masters_ip_str}}
./weed filer -ip={{ansible_host}} -port=8888 -master={{masters_ip_str}}
fuse mount command is weed mount -filer=192.168.1.173:8888,192.168.1.172:8888,192.168.1.171:8888 -dir=/storage/adfs -filer.path=/testcase -collection=test-collection

![Screenshot from 2022-05-17 11-08-15](https://user-images.githubusercontent.com/17705801/168721877-39986144-754a-4be0-a915-1823846b64e4.png)

- OS version: Ubuntu 20.04.1 LTS
- output of `weed version` version 30GB 3.02 a4ca3ed1f0c2bcde1d36bb48f81dedc3e1924679 linux amd64
- if using filer, show the content of `filer.toml`
![Screenshot from 2022-05-17 11-10-34](https://user-images.githubusercontent.com/17705801/168720641-e7535312-c9da-4fc8-a8f9-fcdd73c34a67.png)

**Expected behavior**
Expect the benchmark be normal like https://github.com/chrislusf/seaweedfs/wiki/FUSE-Mount

**Additional context**
To avoid machine reason affect result, I pick another SSD machine, run the follw command to test io performance, got the same result.
./weed server -dir="/data/fs" -filer -s3 -volume.concurrentUploadLimitMB=1024 -volume.concurrentDownloadLimitMB=1024 -volume.compactionMBps=1024 -volume.fileSizeLimitMB=1024 -volume.max=0
[...]

PR #3074, opened 2022-05-18
fix: sequence read rate is slow when use filer.ChunkReadAt.ReadAt

ningfdx:
### When I use mount fuse to test the io rate, I got a wired result. The reading sequence is very slow。#3069 
Screenshot from 2022-05-17 16-14-11

### So i analyze some pprof of the weed mount program. Got this: 
Screenshot from 2022-05-18 17-59-50
Screenshot from 2022-05-18 18-00-19

### Pull Request Description
  The filer.(*ChunkReadAt).ReadAt cost too much time, because every read request call the filer.fetchChunkRange. 

  I read the code of filer.(*ChunkReadAt).ReadAt, and found the reason is filer.ChunkReadAt.readerPattern.isStreaming was set to false, and can not be true forever. 

  That make the every fs.Read() make a ChunkRange request to volume server, so the read rate is very slow.

  Compare the read speed after this pull request:
Screenshot from 2022-05-18 19-11-15

  And I use fio test the random read 4K, the performance is same as origin version weed.

### DIFF
--- a/weed/filer/filechunk_manifest.go
+++ b/weed/filer/filechunk_manifest.go
@@ -2,6 +2,7 @@ package filer
 
 import (
 	"bytes"
+	"errors"
 	"fmt"
 	"github.com/chrislusf/seaweedfs/weed/wdclient"
 	"io"
@@ -115,7 +116,12 @@ func fetchWholeChunk(bytesBuffer *bytes.Buffer, lookupFileIdFn wdclient.LookupFi
 	return nil
 }
 
+var ErrNilFuncNotPermitted = errors.New("nil func is not permitted")
+
 func fetchChunkRange(buffer []byte, lookupFileIdFn wdclient.LookupFileIdFunctionType, fileId string, cipherKey []byte, isGzipped bool, offset int64) (int, error) {
+	if lookupFileIdFn == nil {
+		return 0, ErrNilFuncNotPermitted
+	}
 	urlStrings, err := lookupFileIdFn(fileId)
 	if err != nil {
 		glog.Errorf("operation LookupFileId %s failed, err: %v", fileId, err)
--- a/weed/filer/reader_at.go
+++ b/weed/filer/reader_at.go
@@ -98,10 +98,9 @@ func (c *ChunkReadAt) Close() error {
 
 func (c *ChunkReadAt) ReadAt(p []byte, offset int64) (n int, err error) {
 
-	c.readerPattern.MonitorReadAt(offset, len(p))
-
 	c.readerLock.Lock()
 	defer c.readerLock.Unlock()
+	c.readerPattern.MonitorReadAt(offset, len(p))
 
 	// glog.V(4).Infof("ReadAt [%d,%d) of total file size %d bytes %d chunk views", offset, offset+int64(len(p)), c.fileSize, len(c.chunkViews))
 	return c.doReadAt(p, offset)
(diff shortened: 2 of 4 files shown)


weed/filer/filechunk_manifest.go
```diff
@@ -115,7 +116,12 @@ func fetchWholeChunk(bytesBuffer *bytes.Buffer, lookupFileIdFn wdclient.LookupFi
 	return nil
 }
 
+var ErrNilFuncNotPermitted = errors.New("nil func is not permitted")
+
 func fetchChunkRange(buffer []byte, lookupFileIdFn wdclient.LookupFileIdFunctionType, fileId string, cipherKey []byte, isGzipped bool, offset int64) (int, error) {
+	if lookupFileIdFn == nil {
```
chrislusf, line 122: this is not related?

weed/filer/reader_at.go
```diff
@@ -98,10 +98,9 @@ func (c *ChunkReadAt) Close() error {
 
 func (c *ChunkReadAt) ReadAt(p []byte, offset int64) (n int, err error) {
 
-	c.readerPattern.MonitorReadAt(offset, len(p))
-
 	c.readerLock.Lock()
 	defer c.readerLock.Unlock()
+	c.readerPattern.MonitorReadAt(offset, len(p))
```
chrislusf, line 103: why this is needed?

weed/filer/reader_at_test.go
```diff
@@ -163,3 +163,86 @@ func TestReaderAt1(t *testing.T) {
...
+	}
+
+	testReadAt(t, readerAt, 0, 10, 10, io.EOF)
+	testReadAt(t, readerAt, 1, 45, 9, io.EOF)
+	testReadAt(t, readerAt, 42, 8, 0, io.EOF)
+	testReadAt(t, readerAt, 3, 6, 6, nil)
+
+	// we don't implement rand read api in test, if c.readerPattern.IsRandomMode() == true, test should get this error
+	readerAt.readerPattern = NewReaderPattern()
+	testReadAtWithMonitorReadAt(t, readerAt, 2, 8, 0, ErrNilFuncNotPermitted)
```
chrislusf, line 235: the tests only check errors are valid. How the data are checked?

ningfdx: Well, This pr aims at the judgement of the RandomRead in SeaweedFS.Mount client.

I read the file read flow of SeaweedFS.Mount, The single read range [start, end] flow is:
```
1. if the read op is a random read, just fetch VolumeServer over HTTP request with Header 'Range: bytes=start-end', return
2. if the read op is not a random read, caching the hole chunk from VolumeServer, return [start, end] content.
```

#### To check the data, I should write a unit test to verify the range interface and file cache interface of the volume server, I think seaweedfs/weed/filer is not the suitable place to add the unit test.

So I write the unit test to cover the case of "Corrective Judgment", and this unit test is not aims at detect the error of ErrNilFuncNotPermitted.
The error ErrNilFuncNotPermitted will never happen beside unit test. 

This error is only use in unit test to detect the flow of code branch execution
```
1. func testReadAtWithMonitorReadAt() will return ErrNilFuncNotPermitted if a doReadAt() is a random read.
2. func testReadAtWithMonitorReadAt() will return nil/io.EOF if a doReadAt() is a none random read.
```
Please let me know if my understanding is wrong :)

weed/filer/reader_at.go
ningfdx: the MonitorReadAt is not concurrent safe, I think call this func after lock is high Robustness.
By the way, I don't know whether fuse will call multi read at same time, could this happen？

chrislusf:
I used your code and added ebd6097e6f7c60599e198e5444cdcb353b6328b7

Issue #2755, opened 2022-03-14
[bug:volume] stuck replication on all volume servers (critical)


kmlebedev:
**Describe the bug**

All file upload stopped due to volume replication
There may have been network problems

logs without  errors

many
```
s3-fast-api/harbor-registry/docker/registry/v2/repositories/proxy_docker-elastic-co/elasticsearch/elasticsearch/_uploads/c46580d0-225a-4549-a666-768f15856839/startedat | 504 | 60.001, 60.001, 60.000 | PUT
```

after one  volume-7 shutdown
```
upload_content.go:109] uploading to http://fast-volume-7.s3:8080/1259,0145c8e8c2a09536?ts=1647241384&ttl=&type=replicate: upload 47854270 108 bytes to http://fast-volume-7.s3:8080/1259,0145c8e8c2a09536?ts=1647241384&ttl=&type=replicate: Post "http://fast-volume-7.s3:8080/1259,0145c8e8c2a09536?ts=1647241384&ttl=&type=replicate": EOF
```

debug log from api
```
I0315 10:12:51     1 s3api_object_handlers.go:55] PutObjectHandler logs/dcdp/indices/ZABdOyHIRVqT0WKXw5GcZQ/2/__Y3d4e0C2THW7s_ma1UpErQ
I0315 10:12:51     1 filer_server_handlers_write_upload.go:51] received byte buffer 1
```

**System Setup**
```
2.90
```
![Снимок экрана 2022-03-15 в 14 06 52](https://user-images.githubusercontent.com/9497591/158343849-b6f1548a-e75b-485b-bed2-356f3541d02f.png)

kmlebedev:
the problem repeated again tried to restart all the server filler did not help
But the restart of one server volume out of 20 helped

kmlebedev:
Problem reaching concurrentUploadLimitMB limit
This is option dangerous and causes a domino effect

chrislusf:
For concurrentUploadLimitMB, can we exclude the replication from the concurrentUploadLimitMB counting?

kmlebedev:
> For concurrentUploadLimitMB, can we exclude the replication from the concurrentUploadLimitMB counting?

@chrislusf Yes, i was thinking the same thing.
I ran into several problems:
1.  Exclude the replication from the concurrentUploadLimitMB
2. All requests are not processed after hitting concurrentUploadLimitMB for up to several minutes
[...]

PR #3081, opened 2022-05-20
exclude replication from the concurrentUploadLimitMB

paochiang:


kmlebedev:
@paochiang If you think waiting is better, but not safer, then add a time limit to the wait.

kmlebedev:
@paochiang If you think waiting is better, but not safer, then add a time limit to the wait.

kmlebedev:
@paochiang If you think waiting is better, but not safer, then add a time limit to the wait.

paochiang:
@kmlebedev How about removing the client timeout? 
> Since waiting can easily lead to an infinite waiting loop on all instances. 

I think if cannot write the data successfully, the client should wait forever. All requests sent to the volume server should ensure that they are finally written successfully, and volume server doesn't need to care whether the client connection is closed.

kmlebedev:
> @kmlebedev How about removing the client timeout?
> 
> > Since waiting can easily lead to an infinite waiting loop on all instances.
> 
> I think if cannot write the data successfully, the client should wait forever. All requests sent to the volume server should ensure that they are finally written successfully, and volume server doesn't need to care whether the client connection is closed.

Nobody forces the client to wait forever, and if something goes wrong, it is better to always quickly return the error, and the client himself will flexibly adapt and retry when he needs to and will wait as long as he needs

kmlebedev:
There is also often a limit on the number of client connections on ingresses(nginx as HTTP load balancer)

paochiang:
> @paochiang If you think waiting is better, but not safer, then add a time limit to the wait.

How long should this time limit?

kmlebedev:
> > @paochiang If you think waiting is better, but not safer, then add a time limit to the wait.
> 
> How long should this time limit?

I think this limit should be configurable, but the default is given that the upstream http balancer waits for a response no more than 60 seconds

paochiang:
@kmlebedev modified. I will test it in my env.

kmlebedev:
> @kmlebedev modified. I will test it in my env.

lgtm

paochiang:
> It is also worth checking on load tests to avoid #2755
> 
Verified locally, it's ok.

In addition, in my own environment, with new volume server, reading and writing operations look normal.
image

kmlebedev:
> > It is also worth checking on load tests to avoid #2755
> 
> Verified locally, it's ok.
> 
> In addition, in my own environment, with new volume server, reading and writing operations look normal. image

Looks like testing concurrentDownloadLimit with get requests

paochiang:
> Looks like testing concurrentDownloadLimit with get requests

It is an AI training task that reads and writes data normally, not testing `Concurrentdownloadlimit`

chrislusf:
Prefer not to add a timer for each write request.

paochiang:
modified,  change to lazy timeout check

weed/server/volume_server_handlers.go
```diff
@@ -56,20 +57,33 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
 		vs.guard.WhiteList(vs.DeleteHandler)(w, r)
 	case "PUT", "POST":
 
-		// wait until in flight data is less than the limit
 		contentLength := getContentLength(r)
-
+		startTime := time.Now()
+		vs.inFlightUploadDataLimitCond.L.Lock()
```
chrislusf: too many locks. The lock scope is also too big.

why not use the original atomic.AddInt64 operations?

paochiang: modified

kmlebedev:
@paochiang How are things here?

paochiang:
Waiting for review comments or the code to be merged. I don't know what else needs to be modified.

weed/server/volume_server_handlers.go
```diff
@@ -56,20 +57,29 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
...
-		if vs.concurrentUploadLimit != 0 && r.URL.Query().Get("type") != "replicate" &&
-			atomic.LoadInt64(&vs.inFlightUploadDataSize) > vs.concurrentUploadLimit {
-			err := fmt.Errorf("reject because inflight upload data %d > %d", vs.inFlightUploadDataSize, vs.concurrentUploadLimit)
-			glog.V(1).Infof("too many requests: %v", err)
-			writeJsonError(w, r, http.StatusTooManyRequests, err)
-			return
+		if r.URL.Query().Get("type") != "replicate" {
+			startTime := time.Now()
+			vs.inFlightUploadDataLimitCond.L.Lock()
+			for vs.concurrentUploadLimit != 0 && vs.inFlightUploadDataSize > vs.concurrentUploadLimit {
```
chrislusf: * move this `vs.concurrentUploadLimit != 0` out of the lock

paochiang: modified, moved to` if r.URL.Query().Get("type") != "replicate" {`  this line.

weed/server/volume_server_handlers.go
```diff
@@ -56,20 +57,29 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
...
+				}
+				glog.V(4).Infof("wait because inflight upload data %d > %d", vs.inFlightUploadDataSize, vs.concurrentUploadLimit)
+				vs.inFlightUploadDataLimitCond.Wait()
+			}
+			vs.inFlightUploadDataLimitCond.L.Unlock()
 		}
 		atomic.AddInt64(&vs.inFlightUploadDataSize, contentLength)
 		defer func() {
 			atomic.AddInt64(&vs.inFlightUploadDataSize, -contentLength)
+			vs.inFlightUploadDataLimitCond.Signal()
```
chrislusf: This also need to be conditional?

paochiang: It does not need.
image

chrislusf: It's still an unnecessary call.

paochiang: Got it.
`if r.URL.Query().Get("type") != "replicate" && vs.concurrentUploadLimit != 0 ` or `if vs.concurrentUploadLimit != 0`， which condition  is better?

weed/server/volume_server_handlers.go
kmlebedev: It seems that Or logic is needed

paochiang: > It seems that Or logic is needed

`if r.URL.Query().Get("type") != "replicate" || vs.concurrentUploadLimit != 0 {` ? 
 It cannot exclude replication writing. It should be `and` logic.

weed/server/volume_server_handlers.go
paochiang: modified, added `if vs.concurrentUploadLimit != 0 {`

weed/server/volume_server_handlers.go
kmlebedev: ok

kmlebedev:
This must be checked on load tests with replication. Set myself a task

PR #3102, opened 2022-05-24
mount: fix partial ReadDir results

coelho:

See: https://github.com/chrislusf/seaweedfs/issues/3097

1. There seems to be no guarantee that just because we added a `dirEntry` in Fuse, that the caller has to read it. The caller can advance the `input.Offset` as they please. This results in a partial result if we do not respect the offset.
2. The meta_cache async `EnsureVisited` will definitely cause duplicate entries on first cache and can cause undefined application behavior.
3. The meta_cache can cache a directory that had an error from the Filer, thus creating a potential partial cache.

1. We store the previous read as `entryStream`. It is a window that is cleared as the offset advances. (We assume the offset can not go backwards, and if it does, it will trigger a scan from the beginning)
2. The meta_cache async `EnsureVisited` has been reverted.
3. The meta_cache will only cache a directory if there is no error.

Thank you!

### DIFF
--- a/weed/mount/weedfs_dir_read.go
+++ b/weed/mount/weedfs_dir_read.go
@@ -12,9 +12,22 @@ import (
 
 type DirectoryHandleId uint64
 
+const (
+	directoryStreamBaseOffset = 2 // . & ..
+)
+
 type DirectoryHandle struct {
-	isFinished    bool
-	lastEntryName string
+	isFinished        bool
+	entryStream       []*filer.Entry
+	entryStreamOffset uint64
+}
+
+func (dh *DirectoryHandle) reset() {
+	*dh = DirectoryHandle{
+		isFinished:        false,
+		entryStream:       []*filer.Entry{},
+		entryStreamOffset: directoryStreamBaseOffset,
+	}
 }
 
 type DirectoryHandleToInode struct {
@@ -37,10 +50,8 @@ func (wfs *WFS) AcquireDirectoryHandle() (DirectoryHandleId, *DirectoryHandle) {
 
 	wfs.dhmap.Lock()
 	defer wfs.dhmap.Unlock()
-	dh := &DirectoryHandle{
-		isFinished:    false,
-		lastEntryName: "",
-	}
+	dh := new(DirectoryHandle)
+	dh.reset()
 	wfs.dhmap.dir2inode[DirectoryHandleId(fh)] = dh
 	return DirectoryHandleId(fh), dh
 }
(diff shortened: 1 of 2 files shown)


weed/mount/weedfs_dir_read.go
```diff
@@ -182,37 +177,60 @@ func (wfs *WFS) doReadDirectory(input *fuse.ReadIn, out *fuse.DirEntryList, isPl
...
+				// early terminated
+				return fuse.OK
+			}
 		}
 	}
 
+	var err error
+	if err = meta_cache.EnsureVisited(wfs.metaCache, wfs, dirPath, nil); err != nil {
+		glog.Errorf("dir ReadDirAll %s: %v", dirPath, err)
+	}
```
chrislusf, line 225: need to handle this err

weed/mount/weedfs_dir_read.go
```diff
@@ -37,10 +50,8 @@ func (wfs *WFS) AcquireDirectoryHandle() (DirectoryHandleId, *DirectoryHandle) {
 
 	wfs.dhmap.Lock()
 	defer wfs.dhmap.Unlock()
-	dh := &DirectoryHandle{
-		isFinished:    false,
-		lastEntryName: "",
-	}
+	dh := new(DirectoryHandle)
+	dh.reset()
```
chrislusf, line 54: remove this line?

weed/mount/weedfs_dir_read.go
```diff
@@ -51,11 +62,8 @@ func (wfs *WFS) GetDirectoryHandle(dhid DirectoryHandleId) *DirectoryHandle {
 	if dh, found := wfs.dhmap.dir2inode[dhid]; found {
 		return dh
 	}
-	dh := &DirectoryHandle{
-		isFinished:    false,
-		lastEntryName: "",
-	}
-
+	dh := new(DirectoryHandle)
+	dh.reset()
```
chrislusf, line 66: remove this line?

weed/mount/weedfs_dir_read.go
coelho: necessary to set `entryStreamOffset` to the constant default
do you want me to just inline that instead?

weed/mount/weedfs_dir_read.go
```diff
@@ -182,37 +177,60 @@ func (wfs *WFS) doReadDirectory(input *fuse.ReadIn, out *fuse.DirEntryList, isPl
...
 
-	if !hasData {
-		listErr := wfs.metaCache.ListDirectoryEntries(context.Background(), dirPath, dh.lastEntryName, false, int64(math.MaxInt32), func(entry *filer.Entry) bool {
-			return processEachEntryFn(entry, false)
-		})
-		if listErr != nil {
-			glog.Errorf("list meta cache: %v", listErr)
-			return fuse.EIO
+	var lastEntryName string
+	if input.Offset >= directoryStreamBaseOffset {
```
chrislusf: this If condition seems to be always true

coelho: this commit fixes that: https://github.com/chrislusf/seaweedfs/pull/3102/commits/1fabbe8a25f8cd4183589f728dcf9a5031905b2d

weed/mount/weedfs_dir_read.go
coelho: resolved in https://github.com/chrislusf/seaweedfs/pull/3102/commits/0e6e72d46263a7041606a28edbd8285b4bbf900c

coelho:
All issues addressed @chrislusf

PR #3150, opened 2022-06-07
feat(filer.sync): Add a specified offset for synchronization between filer cluster

qzhello:
Sometimes we need to synchronize from a certain point in time, and the operations before this point in time can be discarded, which is so convenient that we don't have to go to store to change offset.

Just add two configurations to filer.sync.
We will continue to call the "getOffset" method after using it once.


weed/command/filer_sync.go
```diff
@@ -65,6 +67,8 @@ func init() {
 	syncOptions.bProxyByFiler = cmdFilerSynchronize.Flag.Bool("b.filerProxy", false, "read and write file chunks by filer B instead of volume servers")
 	syncOptions.aDebug = cmdFilerSynchronize.Flag.Bool("a.debug", false, "debug mode to print out filer A received files")
 	syncOptions.bDebug = cmdFilerSynchronize.Flag.Bool("b.debug", false, "debug mode to print out filer B received files")
+	syncOptions.aOffset = cmdFilerSynchronize.Flag.Int("a.offset", 0, "synchronization from offset on filer A")
```
chrislusf: Not clear what's the unit.

How about using a relative time similar to the "timeAgo" in other places?

qzhello: Thank you for reminding me.

I do need to explain the unit. The unit is seconds in this code and i want to change it to a millisecond timestamp.  

There is also something wrong with my naming. Many people don't know that offset is a timestamp.

I want to change it like this.
Eg.  **a.syncFromTs,  b.syncFromTs** 

It is similar to the consumption point of rocketmq: **CONSUME_FROM_TIMESTAMP**



I still don't recommend using "TimeAge".
We usually know the start time of synchrionization before filer.sync starts. 
This will result in an inaccurate synchronization point.

weed/command/filer_sync.go
```diff
@@ -65,6 +67,8 @@ func init() {
 	syncOptions.bProxyByFiler = cmdFilerSynchronize.Flag.Bool("b.filerProxy", false, "read and write file chunks by filer B instead of volume servers")
 	syncOptions.aDebug = cmdFilerSynchronize.Flag.Bool("a.debug", false, "debug mode to print out filer A received files")
 	syncOptions.bDebug = cmdFilerSynchronize.Flag.Bool("b.debug", false, "debug mode to print out filer B received files")
+	syncOptions.aSyncFromTs = cmdFilerSynchronize.Flag.Int("a.syncFromTs", 0, "synchronization from timestamp on filer A. The unit is millisecond")
```
chrislusf: How about just use fromTsMs instead of syncFromTs?

qzhello: I agree with you.
It's easier to understand with the unit.

weed/command/filer_sync.go
```diff
@@ -147,6 +153,13 @@ func doSubscribeFilerMetaChanges(clientId int32, grpcDialOption grpc.DialOption,
 		return err
 	}
 
+	if *fromTsMs > 0 {
+		// convert to nanosecond
+		sourceFilerOffsetTsNs = int64(*fromTsMs * 1000_000)
+		// use it only once
+		*fromTsMs = 0
```
chrislusf: This is too hacky and not easy-to-read.

qzhello: Yes, it's not easy-to-read.
It was meant to initialize offset by "fromTsMs". 
Now extract it and become a method and do not change the "doSubscribeFilerMetaChanges".
How about this?

weed/command/filer_sync.go
```diff
@@ -35,6 +36,8 @@ type SyncOptions struct {
 	bDiskType       *string
 	aDebug          *bool
 	bDebug          *bool
+	aFromTsMs       *int
```
chrislusf: the type should be *int64

weed/command/filer_sync.go
```diff
@@ -126,6 +146,26 @@ func runFilerSynchronize(cmd *Command, args []string) bool {
...
 }
 
+func initOffsetFromTsMs(grpcDialOption grpc.DialOption, sourceFiler pb.ServerAddress, targetFiler pb.ServerAddress, fromTsMs int) error {
+	if fromTsMs <= 0 {
+		return nil
+	}
+	// convert to nanosecond
+	fromTsNs := int64(fromTsMs * 1000_000)
+	// read source filer signature
+	sourceFilerSignature, sourceErr := replication.ReadFilerSignature(grpcDialOption, sourceFiler)
```
chrislusf, line 156: please try to avoid repeating the code.

chrislusf: you can change the old logic and only read the filer signatures when the sync process starts.

qzhello: OK, there are too many commit. I'll change the edition and mention pr again. I'll turn this off first. Thanks.

PR #3159, opened 2022-06-10
perf: Optimized volume handling duplicateUUID logic to avoid quitting…

shichanglin5:

Under normal circumstances, there will be no problems, but when the
master is debugged in the local environment, the volume client cannot
communicate with the master normally, so the sendHeartBeat logic is
restarted, and a new connection is created to report the heartbeat. If
the master has not cleared the uuid of the volume at this time, then The
master will respond to volume duplicateUUIDS, and the volume service
will exit, but in fact the uuid of the volume is not duplicated

In the previous logic, a judgment is added to determine whether there is indeed a duplicate uuid

# Checks

### DIFF
--- a/weed/storage/volume.go
+++ b/weed/storage/volume.go
@@ -293,7 +293,6 @@ func (v *Volume) collectStatus() (maxFileKey types.NeedleId, datFileSize int64, 
 	fileCount = uint64(v.nm.FileCount())
 	deletedCount = uint64(v.nm.DeletedCount())
 	deletedSize = v.nm.DeletedSize()
-	fileCount = uint64(v.nm.FileCount())
 
 	return
 }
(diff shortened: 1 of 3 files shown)


weed/storage/volume.go
```diff
@@ -293,7 +293,6 @@ func (v *Volume) collectStatus() (maxFileKey types.NeedleId, datFileSize int64,
 	fileCount = uint64(v.nm.FileCount())
 	deletedCount = uint64(v.nm.DeletedCount())
 	deletedSize = v.nm.DeletedSize()
-	fileCount = uint64(v.nm.FileCount())
```
chrislusf, line 296: what is this deletion?

shichanglin5: fileCount Repeat with line 293

guol-fnst:
@shichanglin5  At what situation the master did not cleare the uuid of the volume?
When volume server exists, master would remove uuid of the volume.

PR #3163, opened 2022-06-10
stats master_replica_placement_mismatch

kmlebedev:

Alerting if the number of replicas does not match the given ones

Add a metric about the discrepancy between the number of replicas given

# Checks


chrislusf:
Need to also update the grafana dashboard.

PR #3168, opened 2022-06-13
fix: Anonymous access cannot be configured by deleting all permission…

shichanglin5:

fix: Anonymous access cannot be configured by deleting all permissions, but anonymous access can be accessed after restarting, in this case "One-way, no switching" is broken

Sometimes it is convenient for testing such as test environments to enable anonymous access by removing all permissions, for example using JMeter

Remove the if judgment so that when isAuthEnabled can be changed from true to false

# Checks

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -137,9 +137,7 @@ func (iam *IdentityAccessManagement) loadS3ApiConfiguration(config *iam_pb.S3Api
 	iam.m.Lock()
 	// atomically switch
 	iam.identities = identities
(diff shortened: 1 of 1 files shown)


chrislusf:
this basically reverted the fix to prevent sometimes the authentication was disabled.

shichanglin5:
After the service is started, anonymous access is allowed if there is no configuration permission; if user permission is configured, subsequent requests must go through authentication; but even if all the configured permissions are cleared, anonymous access is not allowed. If you restart s3api at this time, then The restarted s3api can be accessed anonymously, which sometimes confuses users. If you want the entire cluster to take effect, you need to restart all s3s; I think it is better to give control to the user

PR #3180, opened 2022-06-14
Add filer command line parameter to let Filer UI show/hide directory delete button

blacktear23:

As #3149 mentioned, show delete directory button is danger. This PR will provide a filer command line parameter to let end user control the delete button show or not. 

Provide `-ui.deleteDir` parameter for filer to control filer UI show or hide the delete button for directory. Default is true so if user want to disable delete directory button just add `-ui.deleteDir=false` parameter when start filer server.

# Checks

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -83,7 +83,7 @@ func init() {
 	f.debug = cmdFiler.Flag.Bool("debug", false, "serves runtime profiling data, e.g., http://localhost:<debug.port>/debug/pprof/goroutine?debug=2")
 	f.debugPort = cmdFiler.Flag.Int("debug.port", 6060, "http port for debugging")
 	f.localSocket = cmdFiler.Flag.String("localSocket", "", "default to /tmp/seaweedfs-filer-<port>.sock")
-	f.showUIDirectoryDelete = cmdFiler.Flag.Bool("ui.deleteDir", false, "enable filer UI show delete directory button")
+	f.showUIDirectoryDelete = cmdFiler.Flag.Bool("ui.deleteDir", true, "enable filer UI show delete directory button")
 
 	// start s3 on filer
 	filerStartS3 = cmdFiler.Flag.Bool("s3", false, "whether to start S3 gateway")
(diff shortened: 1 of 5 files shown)


weed/command/filer.go
```diff
@@ -82,6 +83,7 @@ func init() {
 	f.debug = cmdFiler.Flag.Bool("debug", false, "serves runtime profiling data, e.g., http://localhost:<debug.port>/debug/pprof/goroutine?debug=2")
 	f.debugPort = cmdFiler.Flag.Int("debug.port", 6060, "http port for debugging")
 	f.localSocket = cmdFiler.Flag.String("localSocket", "", "default to /tmp/seaweedfs-filer-<port>.sock")
+	f.showUIDirectoryDelete = cmdFiler.Flag.Bool("ui.deleteDir", false, "enable filer UI show delete directory button")
```
chrislusf: how about defaulting to true?

blacktear23: no problem. Do you think the parameter name change to `ui.disableDeleteDir` is acceptable?

chrislusf: "deleteDir" is ok.

blacktear23: So user should only using `ui.deleteDir=false` parameter to disable delete directory button. That seems strange. As normal usage, bool parameter always using false as default.

weed/command/filer.go
```diff
@@ -82,6 +83,7 @@ func init() {
 	f.debug = cmdFiler.Flag.Bool("debug", false, "serves runtime profiling data, e.g., http://localhost:<debug.port>/debug/pprof/goroutine?debug=2")
 	f.debugPort = cmdFiler.Flag.Int("debug.port", 6060, "http port for debugging")
 	f.localSocket = cmdFiler.Flag.String("localSocket", "", "default to /tmp/seaweedfs-filer-<port>.sock")
+	f.showUIDirectoryDelete = cmdFiler.Flag.Bool("ui.deleteDir", true, "enable filer UI show delete directory button")
```
chrislusf, line 86: the server.go also need to add this option.

PR #3182, opened 2022-06-15
feat(filer.sync): add metricsServer in filer.sync.

qzhello:
For synchronization between clusters, we used filer.sync.
We want to monitor the synchronization of offset .
We may need to subtract these two timestamps to observe synchronization.

Metrics include:
(1) the offset of the filer.sync
(2) the last send timestamp of the filer subscription

(1) I added these two metrics.
(2) Started the metrics server in filer.sync.
(3) Change the format of "clientName".
# Checks

### DIFF
--- a/weed/command/filer_sync.go
+++ b/weed/command/filer_sync.go
@@ -216,7 +216,7 @@ func doSubscribeFilerMetaChanges(clientId int32, grpcDialOption grpc.DialOption,
 	}
 
 	var lastLogTsNs = time.Now().Nanosecond()
-	var clientName = fmt.Sprintf("syncTo_%s_From_%s", string(targetFiler), string(sourceFiler))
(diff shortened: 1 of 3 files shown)


weed/command/filer_sync.go
```diff
@@ -210,14 +216,17 @@ func doSubscribeFilerMetaChanges(clientId int32, grpcDialOption grpc.DialOption,
 	}
 
 	var lastLogTsNs = time.Now().Nanosecond()
+	var clientName = fmt.Sprintf("syncTo_%s_From_%s", string(targetFiler), string(sourceFiler))
```
chrislusf: use from ~ to, not to ~ from

qzhello: That's right. Thank you for reminding me.

PR #3201, opened 2022-06-20
force close gRPC client closed connections

kmlebedev:

We are trying to close the permanently closed gRPC client connection

# Checks

### DIFF
--- a/weed/pb/grpc_client_server.go
+++ b/weed/pb/grpc_client_server.go
@@ -144,6 +144,12 @@ func WithGrpcClient(streamingMode bool, fn func(*grpc.ClientConn) error, address
 		defer grpcConnection.Close()
 		executionErr := fn(grpcConnection)
 		if executionErr != nil {
+			// Try force close closed client connection https://github.com/chrislusf/seaweedfs/issues/2802
+			if strings.Contains(executionErr.Error(), "connection closed") {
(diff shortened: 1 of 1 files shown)


weed/pb/grpc_client_server.go
```diff
@@ -144,6 +144,12 @@ func WithGrpcClient(streamingMode bool, fn func(*grpc.ClientConn) error, address
 		defer grpcConnection.Close()
```
chrislusf, line 144: The connection is closed here. The extra close is not needed?

kmlebedev: This is the best idea I've come up with so far - this is to make sure there are no errors when closing the connection.

chrislusf: right. but closing twice is not necessary.

kmlebedev: > right. but closing twice is not necessary.

Now I'm more inclined to the fact that the firewall does not allow to keep very long sessions and closes it on its side, but I still don't know how to recreate the session on the application side

chrislusf: there used to be a 10 hour connection time limit. May need to add it back.

chrislusf: reverted by 41dfe27102b16ae1fd9cc6f17243d2599295f297

PR #3212, opened 2022-06-21
avoid concurrent re-creation of connections

kmlebedev:
https://github.com/chrislusf/seaweedfs/pull/3201

avoid concurrent re-creation of connections

https://github.com/grpc/grpc-go/issues/5449

re-creation of connections in random time

# Checks


chrislusf:
the MaxConnectionAge by default already has 10% jitter.

```
        // MaxConnectionAge is a duration for the maximum amount of time a
	// connection may exist before it will be closed by sending a GoAway. A
	// random jitter of +/-10% will be added to MaxConnectionAge to spread out
	// connection storms.
	MaxConnectionAge [time](https://pkg.go.dev/time).[Duration](https://pkg.go.dev/time#Duration) // The current default value is infinity.
```

kmlebedev:
ok

PR #3226, opened 2022-06-23
Advancedtls

kmlebedev:

close connections

credential reloading

# Checks

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,3 +9,6 @@ install:
 
 full_install:
 	cd weed; go install -tags "elastic gocdk sqlite ydb"
+
+test:
+	cd weed; go test -tags "elastic gocdk sqlite ydb" -v ./...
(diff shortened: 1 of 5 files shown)


kmlebedev:
Verified via ```make dev_tls``` and and works on stage layer with duration 24h for cert Issuer
```
kubectl -n s3 describe certificates.cert-manager.io fast-cert
  Duration:  24h0m0s
  Issuer Ref:
    Kind:        ClusterIssuer
    Name:        vault-issuer
  Renew Before:  6h0m0s
  Secret Name:   fast-cert
```

chrislusf:
After this PR, is the 10 hour max connection age still needed or not related?

kmlebedev:
> After this PR, is the 10 hour max connection age still needed or not related?

Stell needed, for recconnet with new cert

chrislusf:
I am not clear exactly why the re-establishing the connection is necessary.

kmlebedev:
> I am not clear exactly why the re-establishing the connection is necessary.

Let's remove max connection age - it seems to be needed more for security reasons

https://github.com/chrislusf/seaweedfs/issues/2589#issuecomment-1166467411

PR #3228, opened 2022-06-24
When the connection with the leader is disconnected, the vidMap shoul…

shichanglin5:

> filer_server_handlers_read.go:244] failed to stream content /osss3/tests3/172.24.79.234-1MB-gmwqfrr13y-942622: volume 4134 not found
> E0623 02:35:19     8 filer_server_handlers_read.go:244] failed to stream content /osss3/tests3/172.24.17.91-1MB-5lwlhmbmwt-932442: volume 4162 not found
> E0623 02:35:19     8 filer_server_handlers_read.go:244] failed to stream content /osss3/tests3/172.24.79.234-1MB-ab4rmdzlr9-942619: volume 4146 not found

The above problem is triggered whenever the grpc connection expires or the leader changes, causing the filer to process read requests normally

When the connection with the leader is disconnected, the vidMap should not be reset immediately, but should be reset after connecting to a new leader to avoid the problem of "volume xx not found" every time the grpc connection of the leader node is disconnected

reset vidMap only when masterClient successfully connected to new leader

# Checks

### DIFF
--- a/weed/wdclient/masterclient.go
+++ b/weed/wdclient/masterclient.go
@@ -127,22 +127,21 @@ func (mc *MasterClient) tryConnectToMaster(master pb.ServerAddress) (nextHintedL
 		}
 		glog.V(1).Infof("%s.%s masterClient Connected to %v", mc.FilerGroup, mc.clientType, master)
 
-		//reset vidMap only when masterClient successfully connected to new leader
 		resp, err := stream.Recv()
 		if err != nil {
 			glog.V(0).Infof("%s.%s masterClient failed to receive from %s: %v", mc.FilerGroup, mc.clientType, master, err)
 			stats.MasterClientConnectCounter.WithLabelValues(stats.FailedToReceive).Inc()
 			return err
 		}
-		if resp.VolumeLocation != nil {
-			//immediately check whether it is the leader
-			if resp.VolumeLocation.Leader != "" {
-				glog.V(0).Infof("redirected to leader %v", resp.VolumeLocation.Leader)
-				nextHintedLeader = pb.ServerAddress(resp.VolumeLocation.Leader)
-				stats.MasterClientConnectCounter.WithLabelValues(stats.RedirectedToleader).Inc()
-				return nil
-			}
+
+		// check if it is the leader to determine whether to reset the vidMap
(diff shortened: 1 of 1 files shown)


weed/wdclient/masterclient.go
```diff
@@ -126,9 +125,26 @@ func (mc *MasterClient) tryConnectToMaster(master pb.ServerAddress) (nextHintedL
...
 		glog.V(1).Infof("%s.%s masterClient Connected to %v", mc.FilerGroup, mc.clientType, master)
+
+		//reset vidMap only when masterClient successfully connected to new leader
+		resp, err := stream.Recv()
+		if err != nil {
+			glog.V(0).Infof("%s.%s masterClient failed to receive from %s: %v", mc.FilerGroup, mc.clientType, master, err)
+			stats.MasterClientConnectCounter.WithLabelValues(stats.FailedToReceive).Inc()
+			return err
+		}
+		if resp.VolumeLocation != nil {
```
chrislusf: this if clause is not needed.

shichanglin5: > this if clause is not needed.

If the connected master is the leader, the returned response may not contain volumeLocation, and directly judging isLeader will cause a null pointer; the leader will broadcast node addition events after each new client is added. This response does not contain volumeLocation, in this case, directly Judgment  isLeader will result in a null pointer error

weed/server/master_grpc_server.go:246
```
       if !ms.Topo.IsLeader() {
		return ms.informNewLeader(stream)
	}
	peerAddress := pb.ServerAddress(req.ClientAddress)
	// buffer by 1 so we don't end up getting stuck writing to stopChan forever
	stopChan := make(chan bool, 1)

	clientName, messageChan := ms.addClient(req.FilerGroup, req.ClientType, peerAddress)
	for _, update := range ms.Cluster.AddClusterNode(req.FilerGroup, req.ClientType, peerAddress, req.Version) {
		ms.broadcastToClients(update)
	}
```

weed/cluster/cluster.go:100
```		
               return []*master_pb.KeepConnectedResponse{
			{
				ClusterNodeUpdate: &master_pb.ClusterNodeUpdate{
					NodeType: nodeType,
					Address:  string(address),
					IsAdd:    true,
				},
			},
		}
```

PR #3233, opened 2022-06-26
Add TiKV Filerstore back

blacktear23:

Update TiKV client version and add one PC support. After update TiKV client version the go library conflict problem is gone.

Update TiKV client version.

# Checks


chrislusf:
how much larger the compiled version will be with and without the tikv support?
can you make it part of `make full_install` as how YDB/RocksDB works?

blacktear23:
Run `go build` without TiKV: 73MB
Run `go build -tags "tikv"` with TiKV: 81MB
Increase 8MB.

Run `go build -ldflags "-s -w"` without TiKV: 58MB
Run `go build -ldflags "-s -w" -tags "tikv"` with TiKV: 64MB
Increase 6MB

chrislusf:
need to add to `https://github.com/chrislusf/seaweedfs/blob/master/.github/workflows/go.yml` also

chrislusf:
thanks! Please also adjust https://github.com/chrislusf/seaweedfs/wiki/Filer-Stores

blacktear23:
Updated. :)

PR #3239, opened 2022-06-27
feat(filer.sync): add parallel synchronization function

qzhello:
One filerSink synchronization is too slow. I intend to synchronize with multiple filerSink.  
This increases the processing speed.

(1) Split data and execution order.
Split by pathname.
I'm going to use a tree structure, and extract the affected event path. 
The affected set is placed separately in a worker thread. This ensures the correct send order.
The top level of operation will put the children into a collection, and the speed increase is relatively small.

(2) Multiple sending threads
Create multiple FilerSink during initialization to assign to worker thread.
This is the report that I created multiple 1kb files in my unit test.
```
Add 100 folder and Each folder contains 100 1kb files. As it is a stand-alone machine, it has a certain impact.
sync cost: 190s
async cost: 165s [parallel number: 10, batch: 500, period:15s]
async cost: 140s [parallel number: 10, batch: 1000, period:10s]
async cost: 140s [parallel number: 20, batch: 1000, period:20s]
```

(3) Related parameters
* a.parallelNum, b.parallelNum: 
    Number of synchronization workers. The default value is 1, using the original logic.
* a.parallelBatchSize, b.parallelBatchSize: 
    Number of events handled at a time. This parameter takes effect only if x is greater than 1.
* parallelMaxInterval:  
    Maximum processing interval.

(4) Fix bug: nanosecond timestamp display
the value of lastLogTsNs should be used  ```time.Now().UnixNano()``` instead of ```time.Now().Nanosecond```

# Checks

### DIFF
--- a/weed/command/filer_sync.go
+++ b/weed/command/filer_sync.go
@@ -22,27 +22,32 @@ import (
 )
 
 type SyncOptions struct {
-	isActivePassive *bool
-	filerA          *string
-	filerB          *string
-	aPath           *string
-	bPath           *string
-	aReplication    *string
-	bReplication    *string
-	aCollection     *string
-	bCollection     *string
-	aTtlSec         *int
-	bTtlSec         *int
-	aDiskType       *string
-	bDiskType       *string
-	aDebug          *bool
-	bDebug          *bool
-	aFromTsMs       *int64
-	bFromTsMs       *int64
-	aProxyByFiler   *bool
-	bProxyByFiler   *bool
-	metricsHttpPort *int
-	clientId        int32
+	isActivePassive     *bool
+	filerA              *string
+	filerB              *string
+	aPath               *string
+	bPath               *string
+	aReplication        *string
+	bReplication        *string
+	aCollection         *string
+	bCollection         *string
+	aTtlSec             *int
+	bTtlSec             *int
+	aDiskType           *string
+	bDiskType           *string
+	aDebug              *bool
+	bDebug              *bool
+	aFromTsMs           *int64
+	bFromTsMs           *int64
+	aProxyByFiler       *bool
+	bProxyByFiler       *bool
+	aParallelNum        *int
+	bParallelNum        *int
+	aParallelBatchSize  *int64
+	bParallelBatchSize  *int64
+	parallelMaxInterval *string
+	metricsHttpPort     *int
+	clientId            int32
 }
 
 var (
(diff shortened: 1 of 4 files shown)


weed/command/filer_sync.go
```diff
@@ -198,33 +212,70 @@ func doSubscribeFilerMetaChanges(clientId int32, grpcDialOption grpc.DialOption,
...
+			persistEventFns = append(persistEventFns, persistEventFn)
+		}
+
+		c := ParallelSyncMetadataCache{persistEventFns: persistEventFns, l: []*filer_pb.SubscribeMetadataResponse{}, ParallelNum: parallelNum,
+			ParallelBatchSize: parallelBatchSize, sourceFiler: sourceFiler.String(), targetFiler: targetFiler.String()}
+
+		// parameter range limit
+		if c.ParallelNum < 1 {
+			c.ParallelNum = 1
+		} else if uint8(c.ParallelNum) > ^uint8(0) {
```
chrislusf: what is this?

weed/command/filer_sync_parallel.go
```diff
@@ -0,0 +1,264 @@
...
+	go func(c *ParallelSyncMetadataCache) {
+		for range time.Tick(d) {
+			select {
+			case <-stop:
+				close(stop)
+				return
+			default:
+				if len(c.l) > 0 && time.Since(time.UnixMilli(c.lastTime)) >= d {
+					counter := doPersistEventParallel(c)
+					// it's not necessary to set offset successfully. Just wait until next time.
```
chrislusf: disagree. do not skip error checking.

weed/command/filer_sync_parallel.go
```diff
@@ -0,0 +1,264 @@
...
+				fmt.Printf("%s skipping %s change to %v\n", targetFiler, sourceFiler, message)
+				return nil, 1
+			}
+		}
+		return persistEventParallelSyncFunc(c, resp)
+	}, 3*time.Second, setOffsetFunc)
+}
+
+var runParallelSyncScheduledTask = func(cache *ParallelSyncMetadataCache, stop chan struct{}, d time.Duration, setOffsetFunc func(counter int64, lastTsNs int64) error) {
+	go func(c *ParallelSyncMetadataCache) {
```
chrislusf: do not start a goroutine here.

chrislusf:
how does it handle out of order updates caused by the parallel operations?

weed/command/filer_sync.go
qzhello: I want to limit the range of parameters here. It's really not a good place. Is there any place to write? Or delete it？

weed/command/filer_sync_parallel.go
qzhello: I think it will come in next time, so I didn't deal with it here. 
Then I'll do this, If setOffset goes wrong, I'll try setOffset again.

weed/command/filer_sync_parallel.go
qzhello: This scheduled task to deal with events that have not been handled for a long time. I called it with a method "runParallelSyncScheduledTask".

qzhello:
Hello, For the problem of  parallel operations.
First, take down a batch of **SubscribeMetadataResponse**. I split the path names and form a tree structure, record the subscript for the changed paths, and finally extract a subscript set (Include subpath subscript) according to the affected top directory, take the event from the original order table to execute, and they will be put into a worker thread to ensure the order.

weed/command/filer_sync.go
chrislusf: 1. why choosing a fancy way to write a simple comparison?
2. this would not work at all

qzhello:
Changes:
(1)  Adjusted some parameter names and method names.
(2)  Adjusted the way to send events. Put the event into the channel and consume it with goroutine.
(3)  Sending events depends on two conditions: batchSize and waitTime.

* The whole process is that after the event is put into the channel.
* Events are handled by the consuming goroutine, and one of the two conditions of batchSize or waitTime is triggered.
* Construct a tree structure by the path name of the event .
* Traverse all affected nodes and get event subscripts.
* The events are placed in the worker thread and sent to the remote end.

PR #3256, opened 2022-06-30
fixed `volume xx not found` caused by missing VolumeLocation events

shichanglin5:

When the requested master node is the leader, `VolumeLocation` or `ClusterNodeUpdate` may be returned here. If it is `VolumeLocation`, the update will be performed while resetting the vidMap, otherwise the event will be lost

This bug was introduced in #3228 

Processed separately for `VolumeLocation` response

# Checks

### DIFF
--- a/weed/wdclient/masterclient.go
+++ b/weed/wdclient/masterclient.go
@@ -168,15 +168,19 @@ func (mc *MasterClient) tryConnectToMaster(master pb.ServerAddress) (nextHintedL
 		}
 
 		// check if it is the leader to determine whether to reset the vidMap
-		if resp.VolumeLocation != nil && resp.VolumeLocation.Leader != "" && string(master) != resp.VolumeLocation.Leader {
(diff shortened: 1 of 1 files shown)


weed/wdclient/masterclient.go
```diff
@@ -168,15 +168,19 @@ func (mc *MasterClient) tryConnectToMaster(master pb.ServerAddress) (nextHintedL
...
+			if resp.VolumeLocation.Leader != "" && string(master) != resp.VolumeLocation.Leader {
+				glog.V(0).Infof("master %v redirected to leader %v", master, resp.VolumeLocation.Leader)
+				nextHintedLeader = pb.ServerAddress(resp.VolumeLocation.Leader)
+				stats.MasterClientConnectCounter.WithLabelValues(stats.RedirectedToleader).Inc()
+				return nil
+			}
+			mc.vidMap = newVidMap("")
+			mc.updateVidMap(resp)
+		} else {
+			mc.vidMap = newVidMap("")
```
chrislusf, line 181: move the duplicated lines before the if/else clause

PR #3260, opened 2022-07-01
feat(filer.sync): Parallel synchronization

qzhello:
One filerSink synchronization is too slow. I intend to synchronize with multiple filerSink.
This increases the processing speed.

This pr was closed by me
https://github.com/chrislusf/seaweedfs/pull/3239

Changes:
(1) Adjusted some parameter names and method names.
(2) Adjusted the way to send events. Put the event into the channel and consume it with goroutine.
(3) Sending events depends on two conditions: batchSize and waitTime.

The whole process is that after the event is put into the channel.
Events are handled by the consuming goroutine, and one of the two conditions of batchSize or waitTime is triggered.
Construct a tree structure by the path name of the event .
Traverse all affected nodes and get event subscripts.
The events are placed in the worker thread and sent to the remote end.

# Checks


chrislusf:
I am not able to understand the code. It's trying to parallel the operations, but I am not sure how it can ensure the operation ordering for the same file. The tests have nothing for it either.

qzhello:
Operations on the same file will be sent in a goroutine, which will not affect its execution order. These goroutines will wait until all goroutines have finished processing the batch of assigned events. That is, operations on the same file cannot be assigned to other goroutines. For those unrelated operations (there is no common upper directory to be operated), they will be assigned to different goroutines.

This will make the complexity lower, but if there is an operation on the upper directory in a batch of events, and the file operations in this directory are the majority, there will be no performance improvement, the events in this directory will be processed in the same goroutine.

chrislusf:
I simply could not understand your code. There are no comments on major functions or data structures. The code structure is confusing to me.

chrislusf:
filer.sync is paralleized in 1a4bf0dcb5852c62070fe9627b7e63c5d55bb460

PR #3264, opened 2022-07-02
add metric support and code optimize for s3 circuit breaker

shichanglin5:

add metric support and code optimize for s3 circuit breaker

1. All test cases are based on a CircuitBreaker instance to cover the configuration modification case
2. Optimized test cases for higher readability
3. Add the processing of `limitType` to support operations on limitType like `s3.circuitBreaker -type MB -delete`
4. `command_s3_circuitbreaker.go` only checks the value, and the configuration parsing is placed in `s3_circuitbreaker.go`, so that the value seen when viewing the configuration is consistent with the configured value
5. add metrics for circuit breaker

# Checks
- [yes ] I have added unit tests if possible.


weed/s3api/s3api_circuit_breaker.go
```diff
@@ -103,14 +115,17 @@ func (cb *CircuitBreaker) Limit(f func(w http.ResponseWriter, r *http.Request),
...
 			return
 		}
+
 		s3err.WriteErrorResponse(w, r, errCode)
+		stats.S3RequestLimitCounter.WithLabelValues(action).Inc()
 	}, Action(action)
 }
 
-func (cb *CircuitBreaker) limit(r *http.Request, bucket string, action string) (rollback []func(), errCode s3err.ErrorCode) {
+func (cb *CircuitBreaker) limit(r *http.Request, bucket string, action string) (limitAction string, rollback []func(), errCode s3err.ErrorCode) {
```
chrislusf: how about create a struct as
```
type TrafficLimit struct{
  count int64
  size int64
}
```

This way can avoid creating the concatenated the keys for each request, and are more type-safe.

shichanglin5: looks good, i will try it

shichanglin5: > how about create a struct as
> 
> ```
> type TrafficLimit struct{
>   count int64
>   size int64
> }
> ```
> 
> This way can avoid creating the concatenated the keys for each request, and are more type-safe.



> how about create a struct as
> 
> ```
> type TrafficLimit struct{
>   count int64
>   size int64
> }
> ```
> 
> This way can avoid creating the concatenated the keys for each request, and are more type-safe.



> how about create a struct as
> 
> ```
> type TrafficLimit struct{
>   count int64
>   size int64
> }
> ```
> 
> This way can avoid creating the concatenated the keys for each request, and are more type-safe.

..., I didn't understand your idea, I'm curious how to find the corresponding limit configuration based on bucket, action, limitType without concatenating keys.

shichanglin5: ..., I didn't understand your idea, I'm curious how to find the corresponding limit configuration based on bucket, action, limitType without concatenating keys @chrislusf

shichanglin5: What I'm thinking of is to put `limitation` and `count` in a struct `TrafficLimit`, so that finding the struct by key directly completes the operation (increasing or decrementing the counter) without having to look up the counter again

shichanglin5: > how about create a struct as
> 
> ```
> type TrafficLimit struct{
>   count int64
>   size int64
> }
> ```
> 
> This way can avoid creating the concatenated the keys for each request, and are more type-safe.

make it clearer ?

chrislusf: Basically use a struct instead of map to store values if possible.

there could be a `map[bucket]TrafficLimit`. But `action, limitType` should be members within the `TrafficLimit` struct.

weed/s3api/s3api_circuit_breaker.go
```diff
@@ -9,23 +10,78 @@ import (
...
+	"github.com/chrislusf/seaweedfs/weed/stats"
 	"github.com/gorilla/mux"
 	"net/http"
-	"sync"
+	"strconv"
+	"strings"
 	"sync/atomic"
 )
 
+type TrafficLimits map[string]*BucketTrafficLimit
```
chrislusf, line 21: how about using a struct as such instead of a map here?
```
type TrafficLimits {
  requestCount  int64
  throughputMB int64
}
```

shichanglin5: > how about using a struct as such instead of a map here?
> 
> ```
> type TrafficLimits {
>   requestCount  int64
>   throughputMB int64
> }
> ```

没这样做是因为目前 bucket 限流的 action 和 limitType 的值有多个，limitType 有两种：`Count` 和 `MB`，但是`action` 可能是 `Read`,`Write`,`Admin`,`Tagging`，这样下来 action 和 limitType 的组合会比较多；所以如果采用这种方式的话，需要在代码里面就会针对不同的 action 判断取不同的成员变量，如果action判断多了，会让代码显得乱；

不过实际限流用到的就只有 `Read` 和 `Write`，要不就只支持这两种类型的限流；对于读（下载）请求，只有请求数 `Count` 的限制，对于写请求，支持并发 请求数和 吞吐量的的限制，这样的话，结构体变成下面这样：
```go
type TrafficLimits struct {
	readRequestCountTraffic  *BucketTrafficLimit
	writeRequestCountTraffic *BucketTrafficLimit
	writeThroughputTraffic   *BucketTrafficLimit
}

type BucketTrafficLimit struct {
	max   int64
	count int64
}
```

你怎么看 @chrislusf

weed/pb/s3.proto
```diff
@@ -31,5 +31,5 @@ message S3CircuitBreakerConfig {
 
 message S3CircuitBreakerOptions {
     bool enabled=1;
-    map<string, int64> actions = 2;
+    map<string, string> actions = 2;
```
chrislusf, line 34: use explicit limits instead of untyped map:
```
read_count
read_bytes
write_count
write_bytes
list_count
tagging_count
```

PR #3267, opened 2022-07-04
fix s3cfgLock

guo-sj:

The position of s3cfgLock is inappropriate.

Put s3cfgLock to an appropriate position.

# Checks

### DIFF
--- a/weed/iamapi/iamapi_management_handlers.go
+++ b/weed/iamapi/iamapi_management_handlers.go
@@ -36,6 +36,7 @@ var (
 		rand.NewSource(time.Now().UnixNano()))
 	policyDocuments = map[string]*PolicyDocument{}
 	policyLock      = sync.RWMutex{}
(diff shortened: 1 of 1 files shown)


weed/iamapi/iamapi_management_handlers.go
```diff
@@ -412,11 +413,11 @@ func (iama *IamApiServer) DoActions(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	values := r.PostForm
-	var s3cfgLock sync.RWMutex
 	s3cfgLock.RLock()
```
chrislusf, line 416: the lock seems unnecessary

guo-sj: Why is that?

chrislusf: fixed in f87952a411339f2404cac2e771e6ae16644b97ca

PR #3279, opened 2022-07-07
Filer upload from URL

blacktear23:

Support upload a file to filer from URL. And provide filer UI a button `Upload From URL` to upload a file from URL.

Without this feature user should download a file to local and then upload it to filer. After provide this features user can just input a download URL and confirm, filer server can handle the rest.

Just check query string with `source_url` field and using HTTP download stream to `autoChunk` logic. The downloaded file will inherate the URL response's content type and content length.

# Usage

Make a `POST` or `PUT` request with `from.url` query parameter.

```
curl -XPOST "http://[FILER_IP]:[FILER_PORT]/path/alpine-standard-3.16.0_rc1-x86_64.iso?from.url=http://mirrors.edge.kernel.org/alpine/latest-stable/releases/x86_64/alpine-standard-3.16.0_rc1-x86_64.iso"
```
This also provide an option for enable or disable this feature: `-uploadFromUrl`, default set to false. If want use this feature just start filer server with `-uploadFromUrl` option.

```
./weed filer -uploadFromUrl
```

# Checks

### DIFF
--- a/weed/server/filer_server_handlers_write_autochunk.go
+++ b/weed/server/filer_server_handlers_write_autochunk.go
@@ -3,7 +3,6 @@ package weed_server
 import (
 	"context"
 	"fmt"
-	"github.com/chrislusf/seaweedfs/weed/s3api/s3_constants"
 	"io"
 	"net/http"
 	"os"
@@ -11,6 +10,8 @@ import (
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/chrislusf/seaweedfs/weed/s3api/s3_constants"
 
 	"github.com/chrislusf/seaweedfs/weed/filer"
 	"github.com/chrislusf/seaweedfs/weed/glog"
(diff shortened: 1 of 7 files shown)


weed/server/filer_server_handlers_write_autochunk.go
```diff
@@ -69,6 +70,56 @@ func (fs *FilerServer) autoChunk(ctx context.Context, w http.ResponseWriter, r *
 	}
 }
 
+func (fs *FilerServer) autoChunkWithOtherStream(ctx context.Context, w http.ResponseWriter, r *http.Request, stream io.Reader, contentLength int64, contentType string, so *operation.StorageOption) {
```
chrislusf: what is "other"?

blacktear23:
I just using seaweedfs as NAS server for file store and sharing between devices. So this is why I am glad to refine the filer UI and add some convenience features.

And for this feature, it provide a method to sync file to filer from other cloud or web site. Filer UI is just the first user of this feature.

PR #3298, opened 2022-07-11
volume server evacuate to target server

kmlebedev:

Since the volume server is physically located on one node, I want to move it to another physical node

Add option target server

# Checks

### DIFF
--- a/weed/shell/command_volume_server_evacuate.go
+++ b/weed/shell/command_volume_server_evacuate.go
@@ -47,7 +47,7 @@ func (c *commandVolumeServerEvacuate) Do(args []string, commandEnv *CommandEnv, 
 
 	vsEvacuateCommand := flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 	volumeServer := vsEvacuateCommand.String("node", "", "<host>:<port> of the volume server")
-	c.targetServer = *vsEvacuateCommand.String("target", "", "<host>:<port> of target volume")
+	targetServer := vsEvacuateCommand.String("target", "", "<host>:<port> of target volume")
 	skipNonMoveable := vsEvacuateCommand.Bool("skipNonMoveable", false, "skip volumes that can not be moved")
 	applyChange := vsEvacuateCommand.Bool("force", false, "actually apply the changes")
 	retryCount := vsEvacuateCommand.Int("retry", 0, "how many times to retry")
@@ -62,6 +62,9 @@ func (c *commandVolumeServerEvacuate) Do(args []string, commandEnv *CommandEnv, 
 
 	if *volumeServer == "" {
 		return fmt.Errorf("need to specify volume server by -node=<host>:<port>")
+	}
+	if *targetServer != "" {
+		c.targetServer = *targetServer
 	}
 	for i := 0; i < *retryCount+1; i++ {
 		if err = c.volumeServerEvacuate(commandEnv, *volumeServer, *skipNonMoveable, *applyChange, writer); err == nil {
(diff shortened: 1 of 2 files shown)


weed/shell/command_volume_server_evacuate.go
```diff
@@ -47,6 +48,7 @@ func (c *commandVolumeServerEvacuate) Do(args []string, commandEnv *CommandEnv,
 
 	vsEvacuateCommand := flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 	volumeServer := vsEvacuateCommand.String("node", "", "<host>:<port> of the volume server")
+	volumeRack := vsEvacuateCommand.String("rack", "", "rack for then volume servers")
```
chrislusf: typo

Is the rack of the source or target volume server?

kmlebedev: > Is the rack of the source or target volume server?

source

kmlebedev:
@chrislusf I seem to have done

weed/shell/command_volume_server_evacuate.go
```diff
@@ -184,7 +204,7 @@ func (c *commandVolumeServerEvacuate) moveAwayOneEcVolume(commandEnv *CommandEnv
 	return
 }
 
-func moveAwayOneNormalVolume(commandEnv *CommandEnv, volumeReplicas map[uint32][]*VolumeReplica, vol *master_pb.VolumeInformationMessage, thisNode *Node, otherNodes []*Node, applyChange bool) (hasMoved bool, err error) {
+func (c *commandVolumeServerEvacuate) moveAwayOneNormalVolume(commandEnv *CommandEnv, volumeReplicas map[uint32][]*VolumeReplica, vol *master_pb.VolumeInformationMessage, thisNode *Node, otherNodes []*Node, applyChange bool) (hasMoved bool, err error) {
```
chrislusf: this does not need to be instance method.

weed/shell/command_volume_server_evacuate.go
```diff
@@ -80,71 +90,84 @@ func (c *commandVolumeServerEvacuate) volumeServerEvacuate(commandEnv *CommandEn
...
-	topologyInfo, _, err := collectTopologyInfo(commandEnv, 0)
+	c.topologyInfo, _, err = collectTopologyInfo(commandEnv, 0)
 	if err != nil {
 		return err
 	}
 
-	if err := c.evacuateNormalVolumes(commandEnv, topologyInfo, volumeServer, skipNonMoveable, applyChange, writer); err != nil {
+	go func() {
+		for {
+			if topologyInfo, _, err := collectTopologyInfo(commandEnv, 5*time.Minute); err != nil {
```
chrislusf: This is not a good design. 

* The info would not be deterministic. It is also hard to debug.
* Each time this command is run, a new goroutine will start and run forever until the `weed shell` exits.

kmlebedev: > This is not a good design.

@chrislusf  Would it be better if the topology update happens synchronously for the next volum if the evacuation lasted more than 5 minutes?

chrislusf: updating synchronously is more deterministic.

weed/shell/command_volume_server_evacuate.go
```diff
@@ -134,18 +111,34 @@ func (c *commandVolumeServerEvacuate) volumeServerEvacuate(commandEnv *CommandEn
...
 		return fmt.Errorf("%s is not found in this cluster", volumeServer)
 	}
 
 	// move away normal volumes
+	ticker := time.NewTicker(topologyInfoUpdateInterval)
 	for _, thisNode := range thisNodes {
 		for _, diskInfo := range thisNode.info.DiskInfos {
+			if applyChange {
+				select {
+				case <-ticker.C:
```
chrislusf: no timer needed if explicitly updating the topology.

kmlebedev: > no timer needed if explicitly updating the topology.

don't want to update often
if the network is fast or the files are not large

chrislusf: you can use `collectTopologyInfo(commandEnv, 0)` and set a non-zero number for the second parameter.

kmlebedev: > you can use `collectTopologyInfo(commandEnv, 0)` and set a non-zero number for the second parameter.

I tried like this without checking applyChange
When restarted, it will be updated twice and will have to be run asynchronously so as not to wait for sleep inside collectTopologyInfo

it’s better to completely remove the select case

chrislusf: How about waiting a few seconds after each successful move? I prefer not to use the timer since it is not predictable. Some volume may take longer than 5 minutes, some may take 5 seconds. The topology is not always updated.

kmlebedev: > How about waiting a few seconds after each successful move? I prefer not to use the timer since it is not predictable. Some volume may take longer than 5 minutes, some may take 5 seconds. The topology is not always updated.

ok. i remove ticker not important

PR #3303, opened 2022-07-12
speeding up loading volumes

guol-fnst:
Loading volumes is very slow when there are billions of objects

increase concurrency workers

# Checks

### DIFF
--- a/weed/storage/disk_location.go
+++ b/weed/storage/disk_location.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


chrislusf:
This only reduces the concurrency, right?

guol-fnst:
> This only reduces the concurrency, right?

No, on my server(CPU:intel 4114) , workers can be set up to 40.
And on my VM, works will be set to 10 at least.

kmlebedev:
@guol-fnst Maybe you should use something like GOMAXPROCS ?

guol-fnst:
> @guol-fnst Maybe you should use something like GOMAXPROCS ? I use a server with 64 cores, but 4 cores are given for virtualization k8s chart example
> 
> ```
>         env:
>           - name: GOMAXPROCS
>             valueFrom:
>               resourceFieldRef:
>                 resource: limits.cpu
> ```

I got what you mean. Using runtime.GOMAXPROCS(-1) instead of runtime.NumCPU() to make number of workers to be customized by user, expercially in docker env ,right?
Just like this
```
workerNum := runtime.GOMAXPROCS(-1) 
if workerNum <= 10 {
workerNum = 10
}
l.concurrentLoadingVolumes(needleMapKind, workerNum)
```

chrislusf:
GOMAXPROCS() This call will go away when the scheduler improves.

guol-fnst:
> GOMAXPROCS() This call will go away when the scheduler improves.
> 
> https://pkg.go.dev/runtime

How about using Getenv("GOMAXPROCS") to get customized numeber of worker, or if not set, use max  cpu number or 10.

chrislusf:
looks good.

guol-fnst:
> looks good.

I'll submite a new PR later

guol-fnst:
@kmlebedev 
Do you have any different opinions about my previous solution?

kmlebedev:
> @kmlebedev Do you have any different opinions about my previous solution?

no, thanks. this good

PR #3305, opened 2022-07-13
Check whether there is a duplicate accessKey when modifying iam

### LINKED ISSUE #3304: accesskey and secretKey are configured correctly but responds AccessDenied
## reproduce
1. First execute the following command to create two buckets a and b, and then configure the same accessKey and secretKey, different user and bucket2. 
```shell
s3.bucket.create -name a
s3.bucket.create -name b
s3.configure -user a -buckets a -actions Read,Write,List -access_key=a -secret_key=a  -apply
s3.configure -user b -buckets b -actions Read,Write,List -access_key=a -secret_key=a  -apply
```

2. Then upload files to bucket a and b, only one of them will succeed, and the other will respond AccessDenied
```xml
<Error>
    <Code>AccessDenied</Code>
    <Message>Access Denied.</Message>
    <Resource>/b/sdfasfdasdfsa.jpg</Resource>
    <RequestId>1657680620516186256</RequestId>
    <Key>sdfasfdasdfsa.jpg</Key>
    <BucketName>b</BucketName>
</Error>
```

## The cause of the bug
``` at weed/s3api/auth_credentials.go:151 ```
When executing lookup, as long as the same accessKey is matched, it will return. If the same accessKey exists and does not belong to the current Identity, the authentication will fail.
```go
func (iam *IdentityAccessManagement) lookupByAccessKey(accessKey string) (identity *Identity, cred *Credential, found bool) {

	iam.m.RLock()
	defer iam.m.RUnlock()
	for _, ident := range iam.identities {
		for _, cred := range ident.Credentials {
			// println("checking", ident.Name, cred.AccessKey)
			if cred.AccessKey == accessKey {
				return ident, cred, true
			}
		}
	}
	glog.V(1).Infof("could not find accessKey %s", accessKey)
	return nil, nil, false
}
```

shichanglin5:

fix #3304 

check when save

# Checks

### DIFF
--- a/weed/shell/command_s3_configure.go
+++ b/weed/shell/command_s3_configure.go
@@ -171,7 +171,7 @@ func (c *commandS3Configure) Do(args []string, commandEnv *CommandEnv, writer io
 			if userName, found := accessKeySet[cred.AccessKey]; !found {
 				accessKeySet[cred.AccessKey] = ident.Name
 			} else {
-				return errors.New(fmt.Sprintf("duplicate accessKey: %s, already configured in user[%s]", cred.AccessKey, userName))
+				return errors.New(fmt.Sprintf("duplicate accessKey[%s], already configured in user[%s]", cred.AccessKey, userName))
 			}
 		}
 	}
--- a/weed/shell/command_s3_configure.go
+++ b/weed/shell/command_s3_configure.go
@@ -2,6 +2,7 @@ package shell
 
 import (
 	"bytes"
+	"errors"
 	"flag"
 	"fmt"
 	"github.com/chrislusf/seaweedfs/weed/filer"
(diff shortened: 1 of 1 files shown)


guo-sj:
What about [Static Configuration](https://github.com/chrislusf/seaweedfs/wiki/Amazon-S3-API)? In this situation, the https://github.com/chrislusf/seaweedfs/issues/3304 still exists.

shichanglin5:
> What about [Static Configuration](https://github.com/chrislusf/seaweedfs/wiki/Amazon-S3-API#static-configuration)? In this situation, the #3304 still exists.

Yes, I was going to extract this code into a s3api method and call it in `weed/s3api/auth_credentials.go:118` and `command_s3_configure.go`, but there is a circular reference problem with s3api and the shell, so I didn't do that
```shell
package github.com/chrislusf/seaweedfs/weed
	imports github.com/chrislusf/seaweedfs/weed/command
	imports github.com/chrislusf/seaweedfs/weed/iamapi
	imports github.com/chrislusf/seaweedfs/weed/s3api
	imports github.com/chrislusf/seaweedfs/weed/server
	imports github.com/chrislusf/seaweedfs/weed/shell
	imports github.com/chrislusf/seaweedfs/weed/s3api: import cycle not allowed
```

PR #3309, opened 2022-07-14
optimiz concurrency

guol-fnst:
user can customize number of workers via env "GOMAXPROCS"

optimiz  maxproc for docker env

add ENV "GOMAXPROCS" 

# Checks

### DIFF
--- a/weed/storage/disk_location.go
+++ b/weed/storage/disk_location.go
@@ -217,6 +217,10 @@ func (l *DiskLocation) loadExistingVolumes(needleMapKind NeedleMapKind) {
 			glog.Warningf("failed to set worker number from GOMAXPROCS , set to default:10")
 		}
 		workerNum = num
+	} else {
+		if workerNum <= 10 {
+			workerNum = 10
+		}
 	}
 	l.concurrentLoadingVolumes(needleMapKind, workerNum)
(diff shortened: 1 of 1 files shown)


weed/storage/disk_location.go
```diff
@@ -208,8 +209,14 @@ func (l *DiskLocation) concurrentLoadingVolumes(needleMapKind NeedleMapKind, con
...
-		workerNum = 10
+	val, ok := os.LookupEnv("GOMAXPROCS")
+	if ok {
+		num, err := strconv.Atoi(val)
+		if err != nil || num < 1 {
+			num = 10
+			glog.Warningf("failed to set worker number from GOMAXPROCS , set to default:10")
+		}
+		workerNum = num
 	}
```
chrislusf, line 224: If not found, limit to a default 10 with previous logic.

guol-fnst: So in this situation,  the worker number is set to 10 if "GOMAXPROCS" not found,  and never set to NumCPU()?

guol-fnst: > If not found, limit to a default 10 with previous logic.

So in this situation, the worker number is set to 10 if "GOMAXPROCS" not found, and never set to NumCPU()?

chrislusf:
use previous logic

```
	workerNum := runtime.NumCPU()
	if workerNum <= 10 {
		workerNum = 10
	}
```

PR #3317, opened 2022-07-15
update /etc/iam/identity.json in s3 static configuration

guo-sj:

s3 [static configuration](https://github.com/chrislusf/seaweedfs/wiki/Amazon-S3-API#static-configuration) operation can not update the `/etc/iam/identity.json` which may confuse the weed administrator.

Add update logic.

# Checks

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -71,7 +71,7 @@ func NewIdentityAccessManagement(option *S3ApiServerOption) *IdentityAccessManag
 		domain: option.DomainName,
 	}
 	if option.Config != "" {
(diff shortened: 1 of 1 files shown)


chrislusf:
I suspect this may cause more confusion. Users may not know which one to update.

guo-sj:
Dynamic configuration could run `s3.configure` to write s3 configuration to /etc/iam/identity.json, whereas static configuration does not. This pr aims to keep the consistency of s3 configuration. Whichever configuration method you choose, the weed administrator could run `s3.configure` to read and update the configuration with ease.

chrislusf:
The right fix may be to add a warning when using `s3.configure` and a static s3 configuration is provided.

PR #3318, opened 2022-07-15
Use fallback if urls are not found

kmlebedev:

https://github.com/chrislusf/seaweedfs/issues/3310

Normally resolve urls at the moment of leader change

Normally resolve urls at the moment of leader change
https://github.com/chrislusf/seaweedfs/issues/3310

# Checks


guol-fnst:
I did some tests on our server. If the master leader was killed, Slave needed 5-15 seconds to became leader.
So  in this situation whether 2.55 seconds seems to be a little short?

PR #3325, opened 2022-07-18
split s3 static configuration and dynamic configuration

guo-sj:

If the s3 server is using a static configuration, then give an error message when the weed administrator tries to execute `s3.configure` in the weed shell.

In the static configure situation, where the s3 server is started with the command `weed s3 -config=config.json`, the server writes the content of `config.json` to `/etc/iam/static.conf`. The `s3.configure` command will check if the `/etc/iam/static.conf` is existed. If the file does exist, which shows that it's a static configure situation, the proper error message is printed. Otherwise, the `s3.configure` will be executed correctly.

In the dynamic configure situation, the s3 server tries to remove `/etc/iam/static.conf` when it starts so that the `s3.configure` could do its job.

# Checks

### DIFF
--- a/weed/filer/filer_conf.go
+++ b/weed/filer/filer_conf.go
@@ -4,10 +4,11 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
+
 	"github.com/chrislusf/seaweedfs/weed/pb"
 	"github.com/chrislusf/seaweedfs/weed/wdclient"
 	"google.golang.org/grpc"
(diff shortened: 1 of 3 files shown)


chrislusf:
Creating static.conf is not a good idea. What if there are multiple S3 instances? The static configurations could even be different.

guo-sj:
OK, you are right. Could you give some hint about this problem?

chrislusf:
No quick solution here.

Issue #2334, opened 2021-09-10
s3 : does supports permission control of directories under buckets   


huamr:
I see that THE S3 gateway supports bucket-level permission control.Can directories under buckets also be supported？

chrislusf:
does AWS S3 support this?

huamr:
> does AWS S3 support this?

![1631261917](https://user-images.githubusercontent.com/28745710/132823516-d00de025-2cbc-4830-8e0d-99ebb71afc83.png)

huamr:
@ 
Owner
chrislusf

chrislusf:
ok. Will do.

PR #3335, opened 2022-07-19
add supplement for access control of the directories under buckets

guo-sj:

#2334

Because Seaweedfs already support the function of access control of the directories under buckets, see the `canDo` function: 
https://github.com/chrislusf/seaweedfs/blob/f90926572103c0c6d2dbf025b9e09e62d2137b23/weed/s3api/auth_credentials.go#L316
what I do is some trivial refinement:
1. refine the APIs `GetUserPolicy` and `GetActions` to give users a more AWS policy style output
~~2. change s3.configure flag `buckets` to `resource_paths` which indicates to users that you can give all AWS support paths here. It's consistent with [the static configuration](https://github.com/chrislusf/seaweedfs/wiki/Amazon-S3-API#actions-with-wildcard).~~

# Checks

### DIFF
--- a/weed/iamapi/iamapi_management_handlers.go
+++ b/weed/iamapi/iamapi_management_handlers.go
@@ -227,6 +227,7 @@ func (iama *IamApiServer) PutUserPolicy(s3cfg *iam_pb.S3ApiConfiguration, values
 	return resp, fmt.Errorf("%s: the user with name %s cannot be found", iam.ErrCodeNoSuchEntityException, userName)
 }
 
+// https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetUserPolicy.html
 func (iama *IamApiServer) GetUserPolicy(s3cfg *iam_pb.S3ApiConfiguration, values url.Values) (resp GetUserPolicyResponse, err error) {
 	userName := values.Get("UserName")
 	policyName := values.Get("PolicyName")
@@ -244,12 +245,12 @@ func (iama *IamApiServer) GetUserPolicy(s3cfg *iam_pb.S3ApiConfiguration, values
 		policyDocument := PolicyDocument{Version: policyDocumentVersion}
 		statements := make(map[string][]string)
 		for _, action := range ident.Actions {
-			// parse "Read:EXAMPLE-BUCKET"
+			// parse "Read:my-bucket/shared/*"
 			act := strings.Split(action, ":")
 
 			resource := "*"
 			if len(act) == 2 {
-				resource = fmt.Sprintf("arn:aws:s3:::%s/*", act[1])
+				resource = fmt.Sprintf("arn:aws:s3:::%s", act[1])
 			}
 			statements[resource] = append(statements[resource],
 				fmt.Sprintf("s3:%s", MapToIdentitiesAction(act[0])),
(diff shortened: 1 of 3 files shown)


chrislusf:
this is confusing. the access key and secret key would not apply to the paths.

guo-sj:
Ok, I've changed it back.

kmlebedev:
Let's run the unit tests before the merge iam ?
https://github.com/ceph/s3-tests
```
S3TEST_CONF=your.conf ./virtualenv/bin/nosetests s3tests_boto3.functional.test_iam

```

guo-sj:
@kmlebedev I've done IAM test with https://github.com/ceph/s3-tests. I've test three IAM APIs, which are all APIs this PR has effects on, and all of them got passed. Here is my screenshot:

image

The reason why I don't test all the IAM APIs is that the `s3tests_boto3.functional.test_iam` includes a few test cases that the Seaweedfs does not support yet, i.e. `test_get_user_policy_from_multiple_policies`, these cases could lead to failures and errors.

Besides, the `s3-tests` provides only one port in its configure file, which means the `s3` and `iam` utilities should be running in the same port, while Seaweedfs needs two ports. So I have to let both of them run in the port `8333` and write a redirect_port function to redirect the `iam` requests to `iam` default port `8111`.

I wonder if this is what you want? Before I create this PR, I tested so many times, using `boto3`, `aws cli` and `go test`. I don't really understand why you think the `s3_tests` is necessary?

kmlebedev:
@guo-sj I wonder if this is what you want? Before I create this PR, I tested so many times, using boto3, aws cli and go test. I don't really understand why you think the s3_tests is necessary?

This is more of a recommendation as when the time comes I will add an iam test similar to ```make s3tests```
Thanks

kmlebedev:
```
I have a feeling that the logic will break on the part of s3
```

kmlebedev:
Let's add tests right away, shall we?
https://github.com/chrislusf/seaweedfs/blob/master/weed/iamapi/iamapi_test.go#L141

guo-sj:
> Let's add tests right away, shall we?

Sure. What's your test point?

kmlebedev:
> > Let's add tests right away, shall we?
> 
> Sure. What's your test point?

for 
```
// Parse "arn:aws:s3:::my-bucket/shared/*"
```

kmlebedev:
> > I have a feeling that the logic will break on the part of s3
> 
> I don't understand, your feeling? Not any evidence or proof?
> 
> If you don't like it, show your design or code or advice, rather than sarcasm.

I'm sorry it's my duty, I know this place is short of tests.

guo-sj:
> // Parse "arn:aws:s3:::my-bucket/shared/*"
and
// Parse "arn:aws:s3:::my-bucket/*"
// Parse "arn:aws:s3:::my-bucket-prefix-*"

I've added these cases and here is the Github test result.

https://github.com/chrislusf/seaweedfs/runs/7445070659?check_suite_focus=true#step:6:351

guo-sj:
> I'm sorry it's my duty, I know this place is short of tests.

That's fine.

chrislusf:
@guo-sj we are not native English speakers, but all trying the best to improve SeaweedFS. No need to feel offensed because of actual words used. (And I do not find any "sarcasm" in the words either.)

guo-sj:
> @guo-sj we are not native English speakers, but all trying the best to improve SeaweedFS. No need to feel offensed because of actual words used. (And I do not find any "sarcasm" in the words either.)

Ok, sorry to say that. I apologize for that. @kmlebedev @chrislusf

guo-sj:
So what's your point of view about this PR @chrislusf 

Any modification needing to be done? Or If you don't like it feel free to close it.

chrislusf:
@kmlebedev is more familiar with S3 tests. I will let him confirm that the tests are passing.

PR #3340, opened 2022-07-20
avoid duplicated actions in PutUserPolicy

vanhtuan0409:

PutUserPolicy append `Actions` that already existed in user identity. Causing the identity ended with duplicated actions

Add logic to filter our existed actions

# Checks


vanhtuan0409:
one concern this is whether `PUT` should append to the identity's action list or should completely replaced it
AWS document does don't provide a clear instruction but to my experience, it should replace the action list

PR #3348, opened 2022-07-22
optimiz: master ui will render data in order

ningfdx:
Master UI http://localhost:9333/ change the order every time I refresh the page.
That make me feel bad.

change the store var type from map to []struct, then do sort to keep order

### DIFF
--- a/weed/topology/data_center.go
+++ b/weed/topology/data_center.go
@@ -31,25 +31,24 @@ func (dc *DataCenter) GetOrCreateRack(rackName string) *Rack {
 	return rack
 }
 
-type DataCenterMap struct {
-	Id    NodeId    `json:"Id"`
-	Racks []RackMap `json:"Racks"`
+type DataCenterInfo struct {
+	Id    NodeId     `json:"Id"`
+	Racks []RackInfo `json:"Racks"`
 }
 
-func (dc *DataCenter) ToMap() DataCenterMap {
-	m := DataCenterMap{}
-	m.Id = dc.Id()
-	var racks []RackMap
+func (dc *DataCenter) ToInfo() (info DataCenterInfo) {
+	info.Id = dc.Id()
+	var racks []RackInfo
 	for _, c := range dc.Children() {
 		rack := c.(*Rack)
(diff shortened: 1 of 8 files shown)


weed/topology/data_center.go
```diff
@@ -30,15 +31,24 @@ func (dc *DataCenter) GetOrCreateRack(rackName string) *Rack {
 	return rack
 }
 
-func (dc *DataCenter) ToMap() interface{} {
-	m := make(map[string]interface{})
-	m["Id"] = dc.Id()
-	var racks []interface{}
+type DataCenterMap struct {
```
chrislusf: how about calling this `DataCenterInfo`? 

Same for other `xxxMap` objects.

weed/topology/data_center.go
```diff
@@ -30,15 +31,24 @@ func (dc *DataCenter) GetOrCreateRack(rackName string) *Rack {
...
+	Id    NodeId    `json:"Id"`
+	Racks []RackMap `json:"Racks"`
+}
+
+func (dc *DataCenter) ToMap() DataCenterMap {
```
chrislusf: Change `ToMap` to `ToInfo`?

Same for other places.

weed/topology/topology_map.go
```diff
@@ -1,17 +1,32 @@
...
+)
+
+type TopologyMap struct {
+	Max         int64           `json:"Max"`
+	Free        int64           `json:"Free"`
+	DataCenters []DataCenterMap `json:"DataCenters"`
+	Layouts     []interface{}   `json:"Layouts"`
+}
 
 func (t *Topology) ToMap() interface{} {
```
chrislusf: Change the return object type?

ningfdx: OK, fix in new commit.

weed/topology/data_center.go
ningfdx: OK, fix in new commit.

weed/topology/data_center.go
ningfdx: OK, fix in new commit.

PR #3350, opened 2022-07-22
Solve the problem that `LookupFileId` lookup urls is empty due to lea…

### LINKED ISSUE #3327: [wdclient] adding some TTL to the "old" volume location
> This assumes the leader and the followers hold the source of truth of volume locations.
> If the follower becomes a leader, but a volume server does not come alive, the follower may wrongly think it knows where the volume is.
> Currently the topology is already synced to the followers in wdclient, but was reset when the leader fails. We may fix this by adding some TTL to the "old" volume location.

https://github.com/chrislusf/seaweedfs/issues/3310

shichanglin5:

fix #3327 
Solve the problem that `LookupFileId` lookup urls is empty due to leader switching

The vidMap structure is modified to a linked list structure (the length is limited to 5). When the vidMap is reset, the current vidMap is added to the new vidMap as a cache node. When the query locations is empty, the cache node is searched to avoid problems when the master switches leaders.

# Checks

### DIFF
--- a/weed/wdclient/vid_map.go
+++ b/weed/wdclient/vid_map.go
@@ -40,6 +40,7 @@ type vidMap struct {
 	ecVid2Locations map[uint32][]Location
 	DataCenter      string
 	cursor          int32
+	cache           *vidMap
 }
 
 func newVidMap(dataCenter string) vidMap {
@@ -119,17 +120,29 @@ func (vc *vidMap) GetVidLocations(vid string) (locations []Location, err error) 
 }
 
 func (vc *vidMap) GetLocations(vid uint32) (locations []Location, found bool) {
+	glog.V(4).Infof("~ lookup volume id %d: %+v ec:%+v", vid, vc.vid2Locations, vc.ecVid2Locations)
+	locations, found = vc.getLocations(vid)
+	if found && len(locations) > 0 {
+		return locations, found
+	}
+
+	if vc.cache != nil {
+		return vc.cache.GetLocations(vid)
+	}
+
+	return nil, false
+}
+
+func (vc *vidMap) getLocations(vid uint32) (locations []Location, found bool) {
 	vc.RLock()
 	defer vc.RUnlock()
-
-	glog.V(4).Infof("~ lookup volume id %d: %+v ec:%+v", vid, vc.vid2Locations, vc.ecVid2Locations)
 
 	locations, found = vc.vid2Locations[vid]
 	if found && len(locations) > 0 {
 		return
 	}
 	locations, found = vc.ecVid2Locations[vid]
-	return locations, found && len(locations) > 0
+	return
 }
 
 func (vc *vidMap) addLocation(vid uint32, location Location) {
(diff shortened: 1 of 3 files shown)


shichanglin5:
In order to avoid the vidmap cache being empty due to the failure of the master reconnection immediately after the successful reconnection, I set the length of the linked list to 5 by default.

kmlebedev:
Great job.
Only I did not understand what is the point of storing 5 linked list , if, as I understand it, it always refers to the current and cache?

shichanglin5:
> Great job.
> Only I did not understand what is the point of storing 5 linked list , if, as I understand it, it always refers to the current and cache?

If `locations` is empty, `GetLocations` will recursively execute until the cache is nil, that is, traverse the linked list

weed/wdclient/vid_map.go
```diff
@@ -177,6 +190,10 @@ func (vc *vidMap) addEcLocation(vid uint32, location Location) {
 }
 
 func (vc *vidMap) deleteLocation(vid uint32, location Location) {
+	if vc.cache != nil {
+		vc.cache.deleteLocation(vid, location)
+	}
+
```
shichanglin5, line 196: The delete operation will be performed on all cache nodes

weed/wdclient/vid_map.go
```diff
@@ -119,17 +120,29 @@ func (vc *vidMap) GetVidLocations(vid string) (locations []Location, err error)
 }
 
 func (vc *vidMap) GetLocations(vid uint32) (locations []Location, found bool) {
+	glog.V(4).Infof("~ lookup volume id %d: %+v ec:%+v", vid, vc.vid2Locations, vc.ecVid2Locations)
+	locations, found = vc.getLocations(vid)
```
shichanglin5, line 124: Extract `getLocations` to avoid recursive locks caused by recursive execution

weed/wdclient/vid_map_test.go
```diff
@@ -59,6 +61,71 @@ func TestLocationIndex(t *testing.T) {
...
+	//After delete `abc`, cache nodes are searched
+	deleteLoc := Location{Url: newUrl}
+	mc.deleteLocation(uint32(length), deleteLoc)
+	locations, found = mc.GetLocations(uint32(length))
+	if found && locations[0].Url != strconv.FormatInt(int64(length), 10) {
+		t.Fatalf("urls of vid[%d] not expected", length)
+	}
+
+	//lock: concurrent test
+	go func() {
```
chrislusf: This does not look right. The test does not wait for the goroutines to finish.

shichanglin5: right, fixed

kmlebedev:
> > Great job.
> > Only I did not understand what is the point of storing 5 linked list , if, as I understand it, it always refers to the current and cache?
> 
> If `locations` is empty, `GetLocations` will recursively execute until the cache is nil, that is, traverse the linked list

Got it

weed/wdclient/vid_map.go
chrislusf: If addition is also applied to cache, we just need one layer of cache.

PR #3371, opened 2022-07-27
delete disk_size metrics when collection deleted

askeipx:
[3341](https://github.com/chrislusf/seaweedfs/issues/3341)

[3341](https://github.com/chrislusf/seaweedfs/issues/3341)

[3341](https://github.com/chrislusf/seaweedfs/issues/3341)

# Checks

### DIFF
--- a/weed/stats/metrics.go
+++ b/weed/stats/metrics.go
@@ -17,6 +17,16 @@ import (
 	"github.com/prometheus/client_golang/prometheus/push"
 )
 
+// Readonly volume types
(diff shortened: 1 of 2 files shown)


kmlebedev:
Maybe then we will remove all the metrics for the remote collection?

kmlebedev:
@askeipx It seems to be worth adding the function of deleting metrics for UnmountVolume in the same place as s.DeletedVolumesChan

PR #3372, opened 2022-07-27
[filer] sync/backup/replicate exclude directories to sync on filer

kmlebedev:

There are buckets of many temporary files that do not need to be synchronized or backup

Add option with exclude directories to sync on filer

# Checks


kmlebedev:
PR rolled out to a test cluster

PR #3374, opened 2022-07-28
fix: Buckets are not created and deleted correctly on the filer with …

famosss:
Buckets are not created and deleted correctly on the filer with the same signature when they are created and deleted

When the bucket is created, the corresponding table cannot be created

Remove signature verification

# Checks


chrislusf:
how to reproduce this problem? Are files failing to persist?

famosss:
> how to reproduce this problem? Are files failing to persist?

The table creation of `BaikalDB` is asynchronous, and it takes about 15s to complete the creation. If the table is not built in advance, the first few uploads will go wrong

Issue #1953, opened 2021-03-30
Application state and health information


Suika:
**Describe the bug**
At the moment in the seaweed there is no way to know if a specific part of the stack is really healthy.

A single master is started. The moment it is up and the `/cluster/status` is reachable, the master can be considered online and healthy.
But if a cluster of 3+ masters is started, how do I know that every single master is online an healthy? And when is the cluster healthy?
How do I know that the cluster finished deciding who will be the leader?

Volumes are also kinda strange. As it is now, when a volume is started it loads all indices into memory and once it is done it connects to the master. This is the point where it becomes online and "healthy" in the master.
But if you ask the volume server itself, there is no real way to figure out if the volume server is online and healthy. Once the volume starts the api is reachable. 
There is no indication that the volume server is still loading the indices to memory or when it is done. For example:
- Loading indices (unhealthy/booting)
- Can't connect to master (unhealthy)
- Indices loaded [and connected to master] (healthy)

If the state is unknown, it can happen that not all volume servers are up in the cluster before the filers/s3 come up and the clients start writing to seaweed. 
Which can for example cause a creation of volume ids that already exists, but have not been reported to master because the volume server is still "booting".
And as of now, there is no easy way to fix this type of problem (kinda). You are even more screwed if the volume ids have different collections that their counterparts.

The filer is kinda ok at the moment, but would be nice to have a dedicated info about it's state. Since if the HTTP port of filer is reachable, it can be considered online and healthy, I hope.
What I'm after is health status of the containers. To have the abiltiy to properly start a cluster in the order it should: master(s) > volume(s) > filer(s) > S3/dav/...
By proxy, some kind of status indication can also be used in montoring of the application. Allowing to know when something is going awry.

**System Setup**
- Debian 9, Docker
- `version 8000GB 2.34 9672f9e linux amd64`

**Expected behavior**
An ability to query the status of single components of seaweed, not via master or the cluster but also via single process.

PR #3385, opened 2022-07-30
Add healthy indicator for raft status

Woellchen:
This adds a `/cluster/healthz` route to the master servers. It returns a status code based on raft that can be used as a health check. This might solve part of #1953.
I made it similar to the `/healthz` route of the volume server api

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -2,18 +2,20 @@ package command
 
 import (
 	"fmt"
-	"golang.org/x/exp/slices"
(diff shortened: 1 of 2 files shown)


chrislusf:
need to document this in https://github.com/seaweedfs/seaweedfs/wiki/Master-Server-API

Woellchen:
like this? https://github.com/seaweedfs/seaweedfs/wiki/Master-Server-API#check-system-status-1

PR #3388, opened 2022-08-01
optimiz commitig compact

guol-fnst:
Commiting compact will cost too much time ,expecially when there are billions of objects.
It will block reading, updating and deleting operations.

#How are we solving the problem?
Loading most of Needlemap in compacting volumes, and loading the rests in commiting compact.
Now the time of  commiting compact  can be possibly reduced to  seconds or minutes, which is tolerable.

# Checks


weed/storage/volume_vacuum.go
```diff
@@ -392,7 +412,7 @@ func (v *Volume) copyDataAndGenerateIndexFile(dstName, idxName string, prealloca
 	return nm.SaveToIdx(idxName)
 }
 
-func copyDataBasedOnIndexFile(srcDatName, srcIdxName, dstDatName, datIdxName string, sb super_block.SuperBlock, version needle.Version, preallocate, compactionBytePerSecond int64, progressFn ProgressFunc) (err error) {
+func copyDataBasedOnIndexFile(v *Volume, srcDatName, srcIdxName, dstDatName, datIdxName string, sb super_block.SuperBlock, version needle.Version, preallocate, compactionBytePerSecond int64, progressFn ProgressFunc) (err error) {
```
chrislusf: change to an instance method?

weed/storage/volume_loading.go
```diff
@@ -133,9 +133,21 @@ func (v *Volume) load(alsoLoadIndex bool, createDatIfMissing bool, needleMapKind
...
+				if v.tmpNm != nil {
+					glog.V(0).Infof("loading compact index %s ", v.FileName(".idx"))
+					v.tmpNm.indexFile = indexFile
+					stat, e := indexFile.Stat()
+					if e != nil {
+						glog.Fatalf("stat file %s: %v", indexFile.Name(), err)
+						err = e
+					}
+					v.tmpNm.indexFileOffset = stat.Size()
+					v.nm = v.tmpNm
```
chrislusf: set v.tmpNm to nil

weed/storage/volume_vacuum.go
```diff
@@ -463,6 +482,47 @@ func copyDataBasedOnIndexFile(srcDatName, srcIdxName, dstDatName, datIdxName str
...
+		}
+		indexFile, err := os.OpenFile(datIdxName, os.O_RDWR|os.O_CREATE, 0644)
+		if err != nil {
+			glog.Errorf("cannot open Volume Index %s: %v", datIdxName, err)
+			return err
+		}
+		defer indexFile.Close()
 
+		//can be optimized, filling nm in  oldNm.AscendingVisit
+		v.tmpNm, err = doOffsetLoading(indexFile, 0, nm)
```
chrislusf: close previous v.tmpNm

weed/storage/volume_loading.go
chrislusf: v.nm.Close() if v.nm is not nil

weed/storage/volume_vacuum.go
chrislusf: the same logic can be applied to leveldb indexes. Please consider refactoring.

guol-fnst: Sure,  I'll refactor these logic and then repen this PR again.
Now I convert this PR to draft.

weed/storage/volume_vacuum.go
guol-fnst: done

chrislusf:
also, watch out for memory leaks. Need to test and verify the memory usage would not grow under different scenarios.

guol-fnst:
> also, watch out for memory leaks. Need to test and verify the memory usage would not grow under different scenarios.

Got it. I'll watch out the memory usage.

weed/storage/volume_loading.go
guol-fnst: done

weed/storage/volume_vacuum.go
guol-fnst: I have refactored and tested on my server  on both memory index and leveldb index.

weed/storage/volume.go
```diff
@@ -25,6 +25,8 @@ type Volume struct {
 	Collection         string
 	DataBackend        backend.BackendStorageFile
 	nm                 NeedleMapper
+	tmpNm              *NeedleMap
```
chrislusf: why one NeedleMapper is enough for both cases, but the temp needle mapper needs to have different implementation?

guol-fnst: NeedleMapper is a interface.  tmpNm and tmpLnm are point of NeedleMap,  they have to be modified to store necessary information when vacuuming and committing volume, and make NeedleMapper(v.nm) to piont to them at last.
If I was wrong or this solution could be optimized, please let me know.I'll refactor it.

chrislusf:
Need to add tests.

guol-fnst:
> Need to add tests.

Ok, I'll add test cases later

guol-fnst:
> Need to add tests.

There are test case in volume_vacuum_test.go already. So I think if these tests passed, it would prove the new logic works fine.
BTW it lacks of cases for leveldb index scenarios.So I add new test cases for leveldb,

guol-fnst:
Hi @chrislusf 
When I doing tests, I notice that the test case "TestCompaction" used more time than before.
Before PR(https://github.com/seaweedfs/seaweedfs/pull/3426) was merge, this case only took 0.7 second. But now this case takes 7 seconds.

Due  to the new sync methord,  the following function takes several times to write data file. 
https://github.com/seaweedfs/seaweedfs/blob/cb476a53ffe8861565f4b8e40e985ce7bac81fe9/weed/storage/volume_vacuum.go#L302
Mabye the performance of seaweedfs will decrease seriouly not only vcacuuming

chrislusf:
I still think the incremental loading can be refactored into an interface for mem and leveldb. Current code may work, but not easy to maintain.

guol-fnst:
> I still think the incremental loading can be refactored into an interface for mem and leveldb. Current code may work, but not easy to maintain.

Ok，I'll try to refactot it.

weed/storage/volume.go
guol-fnst: Done

guol-fnst:
@chrislusf
Done. Now only one needlemapper interface  is used

weed/storage/volume.go
```diff
@@ -25,8 +25,7 @@ type Volume struct {
 	Collection         string
 	DataBackend        backend.BackendStorageFile
 	nm                 NeedleMapper
-	tmpNm              *NeedleMap
-	tmpLnm             *LevelDbNeedleMap
+	tmpNm              NeedleMapper
```
chrislusf: You can add an interface, e.g., TempNeedleMapper,
```
type TempNeedleMapper interface {
    NeedleMapper
    some_additional_functions()
}
```

guol-fnst: Sorry maybe I don't really get your point.
Do I just replace NeedleMapper with TempNeedleMapper. If so, no more modification will be need,  the current logic is OK.
Or  may I need to implement "some_additional_functions()" for incremental loading?

chrislusf: Use it to avoid reflection

guol-fnst: Got it.
I'll refactor it again

guol-fnst: Done

weed/storage/needle_map_leveldb.go
```diff
@@ -235,3 +235,100 @@ func (m *LevelDbNeedleMap) Destroy() error {
...
+	}
+	defer db.Close()
+
+	//set watermark
+	stat, e := indexFile.Stat()
+	if e != nil {
+		return e
+	}
+	watermark := stat.Size() / watermarkBatchSize
+	e = setWatermark(db, uint64(watermark))
```
chrislusf: Should it set the watermark after successfully load completion time, not before it starts?

guol-fnst: I think the order is not important.
No matter set it on the begaining or the end, it equals to entry counts / 10000. And entry counts stays same when begain to load volume.

guol-fnst: I may make a mistake. Wait a minute, I'll fix it

chrislusf: what if the loading has error?

guol-fnst: Remove the .cpldb folder if error happens

guol-fnst: Done

weed/storage/needle_map_leveldb.go
```diff
@@ -330,5 +327,17 @@ func (m *LevelDbNeedleMap) DoOffsetLoading(v *Volume, indexFile *os.File, startF
 	} else {
 		return NeedleMapMetricFromIndexFile(indexFile, &mm, startFrom)
 	}
+
+	//set watermark
+	stat, err := indexFile.Stat()
+	if err != nil {
+		return err
+	}
+	watermark := (stat.Size() / types.NeedleMapEntrySize / watermarkBatchSize) * watermarkBatchSize
```
chrislusf: seems this bug was not caught by the tests. Can you add tests to address this?

guol-fnst: Sure, I'll add the test case for this bug.

guol-fnst: Done
Add test, also improve logic.

guol-fnst: Another question, is it ok to use reflect in test case?
Besides I plan to add more tests to ensure data security.

guol-fnst: Now it's ready for review.
The statistics methord for file count is different between memory and leveldb index, which is confusing.
If use leveldb, entry count(idxSize/entrySize) = file count + deleted count.
If use memory, the above equationis is not  satisfied.

guol-fnst:
Hi @chrislusf.
I have fixed the bug for metric information and improved error processing logic.
 Now It's the same to before compared with original logic.

Also I found a differenct of statistical methord betweent memory and levelDB index. May be it's a bug?
In memory scenario,  FileCounter = old files counts + new file counts + updated file counts
https://github.com/seaweedfs/seaweedfs/blob/58dd8803224dd5340c54cecf843363fa2570905f/weed/storage/needle_map_memory.go#L38-L42
In levelDB scenario,  FileCounter = all uniqe entris  count in index files.
https://github.com/seaweedfs/seaweedfs/blob/58dd8803224dd5340c54cecf843363fa2570905f/weed/storage/needle_map_metric.go#L108-L110

Also you can check this through running test in volume_vacuum_test.go in my  PR.

chrislusf:
Thanks! I have fixed the code accordingly in c7892bc7c4a5df48cc1db946243573b3ded3d711

guol-fnst:
> Thanks! I have fixed the code accordingly in [c7892bc](https://github.com/seaweedfs/seaweedfs/commit/c7892bc7c4a5df48cc1db946243573b3ded3d711)

I'll modify my test case accordingly later.

weed/storage/volume_vacuum_test.go
```diff
@@ -107,13 +106,15 @@ func testCompaction(t *testing.T, needleMapKind NeedleMapKind) {
 		if realWatermark != watermark {
 			t.Fatalf("testing watermark failed")
 		}
-		if int(deleteCount+v.FileCount()+v.DeletedCount()) != beforeCommitFileCount+afterCommitFileCount*2 {
+		// logic deletedcount is different between memmory and leveldb. So this check is not common.
```
chrislusf: Is this comment still valid?

guol-fnst: Yes, it's valid. In memory scenario, the check is invalid.
Because the deleted count of statistical methord betweent memory and levelDB index is different too.
I used to try to modify them to make it stay same, but faild due to bloomfilter
So I check this only in leveldb scenario.
https://github.com/seaweedfs/seaweedfs/blob/3c8e95eeb33a97a094b6854ae928bbb1a767e1f6/weed/storage/needle_map_memory.go#L42-L50

chrislusf: ok. Please explain it in the comment.

weed/storage/needle_map_metric.go
```diff
@@ -91,11 +91,10 @@ func (mm *mapMetric) MaybeSetMaxFileKey(key NeedleId) {
 	}
 }
 
-func newNeedleMapMetricFromIndexFile(r *os.File) (mm *mapMetric, err error) {
-	mm = &mapMetric{}
+func NeedleMapMetricFromIndexFile(r *os.File, mm *mapMetric) error {
```
chrislusf: this is not a public function.

weed/storage/volume_vacuum.go
```diff
@@ -327,7 +341,11 @@ func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldI
 		}
 	}
 
-	return nil
+	if v.needleMapKind == NeedleMapInMemory {
```
chrislusf: why this is needed if v.tmpNm already has different implementations?

weed/storage/volume_vacuum.go
```diff
@@ -163,6 +167,7 @@ func (v *Volume) CommitCompact() error {
 	if e = v.load(true, false, v.needleMapKind, 0); e != nil {
 		return e
 	}
+	glog.V(0).Infof("Finish commiting volume %d", v.Id)
```
chrislusf: change log level

weed/storage/volume_vacuum.go
```diff
@@ -41,7 +42,7 @@ func (v *Volume) Compact(preallocate int64, compactionBytePerSecond int64) error
 	if v.MemoryMapMaxSizeMb != 0 { //it makes no sense to compact in memory
 		return nil
 	}
-	glog.V(3).Infof("Compacting volume %d ...", v.Id)
+	glog.V(0).Infof("Compacting volume %d ...", v.Id)
```
chrislusf: change log level

weed/storage/volume_vacuum.go
```diff
@@ -261,6 +270,12 @@ func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldI
 	if idx, err = os.OpenFile(newIdxFileName, os.O_RDWR, 0644); err != nil {
 		return fmt.Errorf("open idx file %s failed: %v", newIdxFileName, err)
 	}
+
+	stat, err := idx.Stat()
```
chrislusf, line 283: move this section below `defer idx.Close()`

weed/storage/volume_vacuum.go
```diff
@@ -327,7 +341,11 @@ func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldI
 		}
 	}
 
-	return nil
+	if v.needleMapKind == NeedleMapInMemory {
+		return v.tmpNm.DoOffsetLoading(nil, idx, uint64(idxSize)/types.NeedleMapEntrySize)
```
chrislusf, line 345: I am confused by the logic. It is reading a new idx file from its end, and then process the index entries. So it will read zero index entries?

weed/storage/volume_vacuum.go
guol-fnst: We need to do twice DoOffsetLoading:
1) Loading the  new .idx to tmpNm when copyDataAndGenerateIndexFile. At the same time new updated or deleted files may be written to the same volume.
2)  Loading the new .idx files from offset to the tmpNm. The offset is the start for .idx of new files when doing copyDataAndGenerateIndexFile.

After 1 and 2 , an intact needlemap can be generated.

chrislusf: With the different implementation for the temp needle mapper interface, the code should not need this `if v.needleMapKind == NeedleMapInMemory` checking for different leveldb and memory implementation.

guol-fnst: Yes, this check is not needed. At first I misunderstand your meaning.
I forgot to remove the check after I used  interface instead of  reflactor.

weed/storage/volume_vacuum_test.go
guol-fnst: In memory scenario if a file was deleted  many times, the DeletionCounter will be added only once because the code will check the old needle's size.

 In leveldb scenario if a file was deleted  many times, the DeletionCounter will be added every time, because the code doesn't check the old needle's size.

Bye the way, I'll remove the test for DeletionCounter because of the above reason.

weed/storage/volume_vacuum.go
guol-fnst: Done

weed/storage/volume_vacuum.go
guol-fnst: Done

weed/storage/volume_vacuum.go
guol-fnst: Done

weed/storage/volume_vacuum.go
guol-fnst: Done

weed/storage/needle_map_metric.go
guol-fnst: Done

weed/storage/volume_vacuum_test.go
guol-fnst: And multipal deleting happend when doSomeWritesDeletes was called serveral times in test cases.
Here is the code for dealing with multipal deleting in leveldb scenario.
https://github.com/seaweedfs/seaweedfs/blob/676e27c589a99691ba78f010a067898d89f764ac/weed/storage/needle_map_metric.go#L110-L120

guol-fnst: @chrislusf Is this a bug and needs to be fixed? Or it doesn't matter?

guol-fnst:
@chrislusf Is this PR ok? Should I do more work ?

weed/storage/needle_map_memory.go
```diff
@@ -79,3 +83,49 @@ func (nm *NeedleMap) Destroy() error {
...
+		}
+	}()
+	nm.indexFile = indexFile
+	stat, err := indexFile.Stat()
+	if err != nil {
+		glog.Fatalf("stat file %s: %v", indexFile.Name(), err)
+		return err
+	}
+	nm.indexFileOffset = stat.Size()
+	v.nm = nm
```
chrislusf, line 105: Between line 90 and line 105, `v.nm` is nil. If request comes in, there would be nil pointer exception.

Same for `weed/storage/needle_map_leveldb.go` UpdateNeedleMap implementation.

guol-fnst: When loading volume after commit compact no request will be handled.
https://github.com/seaweedfs/seaweedfs/blob/10414fd81c2744b2b031f4d2f8a935b614d702f1/weed/storage/volume_vacuum.go#L113-L116

guol-fnst: When loading a new volume, no request for this volume will be handled too.

Issue #3369, opened 2022-07-27
[vacuum] Perform a vacuum in a gentle mode


kmlebedev:
**Describe the bug**
A short-term significant increase in delays on volume servers during POST requests coinciding in time with the start of the vacuum

**System Setup**
```
3.18
```

**Expected behavior**
It might be worth making a reasonable gap between start Committing vacuum
It is also possible to add a randomizer +/- 3 minutes of launch

**Screenshots**
delays in 12:01
![Screenshot 2022-07-27 at 13 31 01](https://user-images.githubusercontent.com/9497591/181200601-385a17fd-718e-425d-abe1-2be48a09daba.png)

Volume server mostly delays on POST requests 
![Screenshot 2022-07-28 at 12 18 43](https://user-images.githubusercontent.com/9497591/181445217-4d0cf56e-7f8c-48eb-b6d8-092c4aa5022e.png)
S3 server mostly delays on  PUT requests to vacuum bucket

![Screenshot 2022-07-28 at 12 26 46](https://user-images.githubusercontent.com/9497591/181446700-a6b5cb2a-4160-4910-93a9-696a5abdad88.png)

start vacuum
![Screenshot 2022-07-27 at 13 16 04](https://user-images.githubusercontent.com/9497591/181199129-1dcacb7a-ffbe-4196-880b-04ec4754fab0.png)

LA5m avg low
![Screenshot 2022-07-27 at 21 42 11](https://user-images.githubusercontent.com/9497591/181303020-d5e3a604-fd21-4d98-999f-3ecd25ec9edd.png)

chrislusf:
need to space out for the same volume id

guol-fnst:
These days I am trying to reduce the time of locking volume. Then I found some questions.
There are severay steps  to vacuuming volume
1) remove volums from writable list , now this  volume shouldn't be written.
https://github.com/chrislusf/seaweedfs/blob/ca836568ac3df79373225bf6b98334e641194cfc/weed/topology/topology_vacuum.go#L67
2) CompactVolume, make copies of .dat and .idx, then create new .dat and .idx
https://github.com/chrislusf/seaweedfs/blob/ca836568ac3df79373225bf6b98334e641194cfc/weed/server/volume_grpc_vacuum.go#L27
3.1) batchVacuumVolumeCommit. First, lock the whole volume to avoid reading and writing.
[...]

PR #3390, opened 2022-08-01
vacuum stats of percent util LA

kmlebedev:

I would like to have more information about loading the system at the time of the vacuum

https://github.com/seaweedfs/seaweedfs/issues/3369

Since in the future, when the threshold value is exceeded, you can also slow down the vacuum

# Checks


kmlebedev:
```
I0801 10:44:24.344908 topology_vacuum.go:92 1 vacuum 1539 on fast-volume-1:8080 processed 719464584 bytes, loadAvg 17%
```

weed/topology/topology_vacuum.go
```diff
@@ -138,6 +139,8 @@ func (t *Topology) batchVacuumVolumeCommit(grpcDialOption grpc.DialOption, vl *V
 		} else {
 			glog.V(0).Infof("Complete Committing vacuum %d on %s", vid, dn.Url())
 		}
+		// small space out for the same volume id https://github.com/seaweedfs/seaweedfs/issues/3369
+		time.Sleep(3 * time.Second)
```
chrislusf: this is a separate issue. How to do it needs discussion. The value would not work for large clusters.

weed/pb/volume_server.proto
```diff
@@ -147,6 +147,7 @@ message VacuumVolumeCompactRequest {
 }
 message VacuumVolumeCompactResponse {
     int64 processed_bytes = 1;
+    uint32 percent_load_avg_1m = 2;
```
chrislusf: use "float64 load_avg_1m"

percent is how the value should be rendered.

chrislusf (changes requested):

weed/pb/volume_server.proto
kmlebedev: I thought about it, but float64 is definitely not suitable, since you need an integer from 0 to 100 with mathematical rounding

weed/topology/topology_vacuum.go
kmlebedev: ok

weed/server/volume_grpc_vacuum.go
```diff
@@ -29,11 +34,16 @@ func (vs *VolumeServer) VacuumVolumeCompact(req *volume_server_pb.VacuumVolumeCo
...
 	nextReportTarget := reportInterval
-
+	fs, fsErr := procfs.NewDefaultFS()
 	var sendErr error
 	err := vs.store.CompactVolume(needle.VolumeId(req.VolumeId), req.Preallocate, vs.compactionBytePerSecond, func(processed int64) bool {
 		if processed > nextReportTarget {
 			resp.ProcessedBytes = processed
+			if fsErr == nil && numCPU > 0 {
+				if fsLa, err := fs.LoadAvg(); err == nil {
+					resp.PercentLoadAvg_1M = uint32(math.Round(fsLa.Load1 * 100 / float64(numCPU)))
```
chrislusf: remove percent, rounding and multiply.

weed/topology/topology_vacuum.go
```diff
@@ -89,7 +89,8 @@ func (t *Topology) batchVacuumVolumeCompact(grpcDialOption grpc.DialOption, vl *
 							return recvErr
 						}
 					}
-					glog.V(0).Infof("%d vacuum %d on %s processed %d bytes", index, vid, url, resp.ProcessedBytes)
+					glog.V(0).Infof("%d vacuum %d on %s processed %d bytes, loadAvg %v%%",
+						index, vid, url, resp.ProcessedBytes, resp.PercentLoadAvg_1M)
```
chrislusf: adjust format and multiply by 100.

weed/server/volume_grpc_vacuum.go
kmlebedev: It made sense to take utilization as a percentage, so that there would be no confusion with the original LA

weed/pb/volume_server.proto
```diff
@@ -147,6 +147,7 @@ message VacuumVolumeCompactRequest {
 }
 message VacuumVolumeCompactResponse {
     int64 processed_bytes = 1;
+    float percent_load_avg_1m = 2;
```
chrislusf: float is not valid. Need to use float32

chrislusf: change percent_load_avg_1m to load_avg_1m

weed/topology/topology_vacuum.go
```diff
@@ -89,7 +89,8 @@ func (t *Topology) batchVacuumVolumeCompact(grpcDialOption grpc.DialOption, vl *
 							return recvErr
 						}
 					}
-					glog.V(0).Infof("%d vacuum %d on %s processed %d bytes", index, vid, url, resp.ProcessedBytes)
+					glog.V(0).Infof("%d vacuum %d on %s processed %d bytes, loadAvg %.0f%%",
```
chrislusf: format should be `%.02f%%`

weed/pb/volume_server.proto
kmlebedev: > float is not valid. Need to use float32

protoc thinks differently
Screenshot 2022-08-01 at 21 25 44

kmlebedev: > change percent_load_avg_1m to load_avg_1m

ok

Issue #3083, opened 2022-05-20
[hashicorp raft] bug: sometimes peers are deleted after master change 


kmlebedev:
**Describe the bug**

sometimes but regularly peers are deleted after master change 

**System Setup**
```
3.04
```
```
weed master  -resumeState=true -raftHashicorp
```

**Expected behavior**

after changing the master, the list of peers was unchanged

**Additional context**
raft logs:

```
May 20, 2022 @ 12:39:21.877	2022-05-20T07:39:21.877Z [WARN]  raft: failed to contact: server-id=slow-master-0.dc2:9333 time=317.196538ms	slow-master-1	node9-dc1
May 20, 2022 @ 12:39:21.999	2022-05-20T07:39:21.999Z [WARN]  raft: heartbeat timeout reached, starting election: last-leader-addr=slow-master-1.dc1:9333 last-leader-id=	slow-master-0	node9-dc2
May 20, 2022 @ 12:39:22.001	2022-05-20T07:39:22.001Z [WARN]  raft: rejecting vote request since we have a leader: from=slow-master-0.dc2:9333 leader=slow-master-1.dc1:9333 leader-id=""	slow-master-0	node3-dc2
May 20, 2022 @ 12:39:22.002	2022-05-20T07:39:22.000Z [WARN]  raft: rejecting vote request since we have a leader: from=slow-master-0.dc2:9333 leader=slow-master-1.dc1:9333 leader-id=slow-master-1.dc1:9333	slow-master-1	node9-dc1
May 20, 2022 @ 12:39:22.164	2022-05-20T07:39:22.163Z [WARN]  raft: failed to contact: server-id=slow-master-0.dc2:9333 time=602.853871ms	slow-master-1	tech-paas-k8s-node9-dcdp
May 20, 2022 @ 12:39:22.199	2022-05-20T07:39:22.198Z [ERROR] raft: peer has newer term, stopping replication: peer="{Voter slow-master-0.dc2:9333 slow-master-0.dc2:19333}"	slow-master-1	node9-dc1
May 20, 2022 @ 12:39:22.754	2022-05-20T07:39:22.754Z [WARN]  raft: heartbeat timeout reached, starting election: last-leader-addr=slow-master-1.dc1:9333 last-leader-id=	slow-master-0	node3-dc1
[...]

PR #3338, opened 2022-07-19
rollback over onPeerUpdate implementation of automatic clean-up of failed servers in favor of synchronous ping

kmlebedev:
https://github.com/chrislusf/seaweedfs/issues/3083

# What problem are we solving?

https://github.com/chrislusf/seaweedfs/issues/3083

# How are we solving the problem?

https://github.com/chrislusf/seaweedfs/issues/3083

# Checks
- [ok] I have added unit tests if possible.
- [ok] I will add related wiki document changes and link to this PR after merging.

Issue #1820, opened 2021-02-19
Failover master servers doesn't work


paochiang:
**Describe the bug**
 Failover master servers doesn't work , after the master server failover, can't read and write normally.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
```
# setup mastern (replication=001)
root@dev-4:~/swfs# weed master -port=9333 -mdir=/tmp/master0 -defaultReplication=001 -peers=localhost:9333,localhost:9334,localhost:9335 -ip=localhost
root@dev-4:~/swfs# weed master -port=9334 -mdir=/tmp/master1 -defaultReplication=001 -peers=localhost:9333,localhost:9334,localhost:9335 -ip=localhost
root@dev-4:~/swfs# weed master -port=9335 -mdir=/tmp/master2 -defaultReplication=001 -peers=localhost:9333,localhost:9334,localhost:9335 -ip=localhost

# setup volume
root@dev-4:~/swfs# weed volume -port=8080 -dir=/tmp/volume0 -ip=localhost -mserver=localhost:9333,localhost:9334,localhost:9335
root@dev-4:~/swfs# weed volume -port=8081 -dir=/tmp/volume1 -ip=localhost -mserver=localhost:9333,localhost:9334,localhost:9335
root@dev-4:~/swfs# weed volume -port=8082 -dir=/tmp/volume2 -ip=localhost -mserver=localhost:9333,localhost:9334,localhost:9335

# setup filer
root@dev-4:~/swfs# weed filer -port=8888  -ip=localhost -master=localhost:9333,localhost:9334,localhost:9335

# setup weed mount (disable cache)
root@dev-4:~/swfs# weed mount -filer=localhost:8888 -cacheCapacityMB=0 -cacheDir=/tmp/cache001 -dir=/root/swfs/mnt

# copy a directory
root@dev-4:~/swfs# cp -r /root/mydir /root/swfs/mnt/

# shut down master0 (9333)
use `ctrl + C`

# copy directory to host
cp -r /root/swfs/mnt/mydir /root/mydircopy
get error:
cp: error reading '/root/swfs/mnt/mydir/file.txt': Input/output error

#write file
cp /root/swfs/testfile.txt /root/swfs/mnt/
get error:
cp: failed to close '/root/swfs/mnt/testfile.txt': Input/output error
```

weed mount log:
```
This is SeaweedFS version 30GB 2.23 318a3d2 linux amd64
[...]

PR #3398, opened 2022-08-03
remove old raft servers if they don't answer to pings for too long

askeipx:

https://github.com/seaweedfs/seaweedfs/issues/3083

Ping masters and remove stalling ones

# Checks


chrislusf:
Prefer not to merge this. This duplicates with liveness checking via grpc connections.

This is also different from pinging the raft server when they disconnect.

askeipx:
> 

I assume you propose to ping servers only when master receives delete event in OnPeerUpdate? I don't mind doing it that way but it doesn't really change behaviour. Master pings others ~only once per minute so it doesn't create big load there. Can you describe what you don't like more specifically?

chrislusf:
> Can you describe what you don't like more specifically?

This duplicates with liveness checking via grpc connections.

askeipx:
Are you fine with pinging after master receives delete event in OnPeerUpdate?

chrislusf:
> Are you fine with pinging after master receives delete event in OnPeerUpdate?

that should be fine. Need to add a timeout for the ping operation.

askeipx:
So I should refactor all grpc_client_server.With{ClientType}Client methods to pass grpc.WaitForReady option?
So let's check some examples: if I lose 2 masters at the same time and I want to delete them from my raft servers list after 72 minutes, I will stuck on pinging first master for 72 minutes and only then will start to ping master2? or do you want to create new goroutine for each master that has been deleted? Basically that was my concern when I was making my implementation of ping

askeipx:
Which liveness check are you talking about btw? Didn't find it, where is it called from?

chrislusf:
> 



> Which liveness check are you talking about btw? Didn't find it, where is it called from?

see `weed/pb/grpc_client_server.go`

chrislusf:
> So I should refactor all grpc_client_server.With{ClientType}Client methods to pass grpc.WaitForReady option?
> So let's check some examples: if I lose 2 masters at the same time and I want to delete them from my raft servers list after 72 minutes, I will stuck on pinging first master for 72 minutes and only then will start to ping master2? or do you want to create new goroutine for each master that has been deleted? Basically that was my concern when I was making my implementation of ping

see my comment in https://github.com/seaweedfs/seaweedfs/pull/3338

askeipx:
I've seen all your comments there but it doesn't help me with my question. Timed ping request won't really work if master won't be able to create TCP connection. So I need WaitForReady + context WithTimeout to create what you want. It doesn't help me with `two masters going down at the same time` question however.
I will make some edits to this pr and will see if I made what you wanted later, I guess.

chrislusf:
the events are processed one by one. there are no updates at the same time.

askeipx:
So goroutine that gets KeepConnectedResponse will be blocked until master is pinged successfully?

askeipx:
I don't really see how it's better than pings from separate goroutine

weed/pb/grpc_client_server.go
```diff
@@ -65,15 +66,17 @@ func NewGrpcServer(opts ...grpc.ServerOption) *grpc.Server {
...
 	// opts = append(opts, grpc.WithBlock())
 	// opts = append(opts, grpc.WithTimeout(time.Duration(5*time.Second)))
 	var options []grpc.DialOption
+
 	options = append(options,
 		// grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithDefaultCallOptions(
 			grpc.MaxCallSendMsgSize(Max_Message_Size),
 			grpc.MaxCallRecvMsgSize(Max_Message_Size),
+			grpc.WaitForReady(waitForReady),
```
chrislusf, line 79: how about we just use wait for ready for all the grpc calls?

askeipx: I don't understand how it's gonna be helpful. If some server goes down and we try to wait for connection with it without timeout, then goroutine that made this call will be blocked. Why do you think it's appropriate?
I work on this project for a short time and don't know all specifics, but will WaitForReady on every grpc call really work?

kmlebedev: > how about we just use wait for ready for all the grpc calls?

We already tried it. :) 
https://github.com/seaweedfs/seaweedfs/pull/1577/files

askeipx: I don't understand what has changed since https://github.com/seaweedfs/seaweedfs/issues/1820 where this option was removed. You still insist on enabling WaitForReady for all grpc calls?

askeipx:
Can we merge this?

weed/server/master_server.go
```diff
@@ -358,5 +361,27 @@ func (ms *MasterServer) OnPeerUpdate(update *master_pb.ClusterNodeUpdate, startF
 				hashicorpRaft.ServerID(peerName),
 				hashicorpRaft.ServerAddress(peerAddress.ToGrpcAddress()), 0, 0)
 		}
+	} else {
+		pb.WithMasterClient(false, peerAddress, ms.grpcDialOption, true, func(client master_pb.SeaweedClient) error {
+			ctx, cancel := context.WithTimeout(context.TODO(), time.Minute*72)
```
chrislusf, line 364: why it is 72 minutes?

weed/server/master_server.go
```diff
@@ -358,5 +361,27 @@ func (ms *MasterServer) OnPeerUpdate(update *master_pb.ClusterNodeUpdate, startF
...
+			defer cancel()
+			if _, err := client.Ping(ctx, &master_pb.PingRequest{Target: string(peerAddress), TargetType: cluster.MasterType}); err != nil {
+				glog.V(0).Infof("master %s didn't respond to pings. remove raft server", peerName)
+				if err := ms.MasterClient.WithClient(false, func(client master_pb.SeaweedClient) error {
+					_, err := client.RaftRemoveServer(context.Background(), &master_pb.RaftRemoveServerRequest{
```
chrislusf, line 369: this is under when `!update.IsAdd`, which means it is acting upon a node deletion event.

This client is is connecting to the master, which is itself, to send a peer removal request, which will come back as another node deletion event.

Seems a dead loop?

weed/server/master_server.go
askeipx: I took it from RaftServerRemovalTime from line 37 https://github.com/kmlebedev/seaweedfs/commit/6cfbfb084941e213f38a83a763fd359cc6611108#diff-1f7b377a963eee3cdc38f3f4b7c168638a43855c5e3b03e7fcc00504dba4389aL37 . The reasoning behind this is to remove server that didn't recover for too long.
I can make this number configurable if you would like it this way

weed/server/master_server.go
askeipx: Just checked it again - RaftRemoveServer doesn't provoke OnPeerUpdate notification. It is used only to remove master from raft configuration

askeipx:
@chrislusf do you have any other comments?

chrislusf:
This branch has conflicts that must be resolved

weed/server/master_server.go
chrislusf: Merged. 

@kmlebedev 
Why we need to waitForReady here? If the connection has problem, shouldn't we declare this server is disconnected? 

The 72 minute timeout seems also too long.

kmlebedev: > The 72 minute timeout seems also too long.

@chrislusf This period should be long enough to not have the effect of temporarily taking the server out for maintenance, such as kernel upgrades, disk replacement.

chrislusf: It can be added back if it comes alive, right?

kmlebedev: > It can be added back if it comes alive, right?

Yes, but removing one node from the raft for a long time results in a quorum violation, which makes it impossible to choose a leader. Or you need to immediately check for a quorum.

PR #3402, opened 2022-08-04
refactor FilerRequest metrics

kmlebedev:

I would like more understanding of what happens when chunks are uploaded

# Checks


weed/server/filer_server_handlers.go
```diff
@@ -16,11 +16,12 @@ import (
 func (fs *FilerServer) filerHandler(w http.ResponseWriter, r *http.Request) {
 
 	start := time.Now()
-
+	stats.FilerRequestCounter.WithLabelValues(r.Method).Inc()
```
chrislusf: moving here would double count for "proxy" mode.

weed/server/filer_server_handlers.go
```diff
@@ -16,7 +16,6 @@ import (
 func (fs *FilerServer) filerHandler(w http.ResponseWriter, r *http.Request) {
 
 	start := time.Now()
-	stats.FilerRequestCounter.WithLabelValues(r.Method).Inc()
 	defer func() {
 		stats.FilerRequestHistogram.WithLabelValues(r.Method).Observe(time.Since(start).Seconds())
```
chrislusf: the defer should also move with the previous change.

chrislusf:
Merging. Need to also adjust the grafana panels.

PR #3426, opened 2022-08-10
Refactor for Sync method

blacktear23:

Restore Sync operation back.

For different system using different method to sync file data:

* For Linux using `fdatasync` make less IO operation (ignore sync metadata)
* For MacOS default is using `fcntl` with `F_BARRIERFSYNC`, this will resolve MacOS with ARM CPU version sync data too slow issue. As golang default implementation,  File.Sync is using `fcntl(fd, F_FULLFSYNC)` call which will flush all file system dirty data to disk and this is very very slow with MacOS running on ARM cpu. Using `F_BARRIERFSYNC` will make ARM CPU version's Mac running fast.
* For other OS just using `File.Sync`

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/backend/disk_file.go
+++ b/weed/storage/backend/disk_file.go
@@ -1,10 +1,11 @@
 package backend
 
 import (
-	"github.com/seaweedfs/seaweedfs/weed/glog"
-	. "github.com/seaweedfs/seaweedfs/weed/storage/types"
 	"os"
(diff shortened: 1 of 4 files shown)


blacktear23:
@chrislusf this PR should fix the issue of 52580743b9de1931b149e1df40e27048d99ca3bf mentioned.

PR #3428, opened 2022-08-11
add upload response

lyang24:
on filer server ui when we upload file there is no response

add js trigger output response on ui when request is done

# How is the PR tested?

tested on localhost, page did refresh after upload

# Checks

### DIFF
--- a/weed/server/filer_ui/filer.html
+++ b/weed/server/filer_ui/filer.html
@@ -290,8 +290,7 @@
         var xhr = new XMLHttpRequest();
         xhr.onreadystatechange = function() {
             if (xhr.readyState == XMLHttpRequest.DONE) {
-                var resp = JSON.parse(xhr.responseText);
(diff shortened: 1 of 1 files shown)


weed/server/filer_ui/filer.html
```diff
@@ -288,6 +288,12 @@ <h1>
     function uploadFile(file, i) {
         var url = window.location.href;
         var xhr = new XMLHttpRequest();
+        xhr.onreadystatechange = function() {
+            if (xhr.readyState == XMLHttpRequest.DONE) {
+                var resp = JSON.parse(xhr.responseText);
+                alert(`uploaded file: ${resp.name} with size: ${resp.size}`);
```
chrislusf: * I prefer just refreshing the page instead of adding an alert that users need to click away.
* What if there are multiple files uploaded?

lyang24: fixed

Issue #3421, opened 2022-08-08
Master maintenance script isn't invoked after enabling raft.


divanikus:
**Describe the bug**
After enabling raft I see no indications of maintenance script is ever invoked. It was working before.

**System Setup**
- `master.toml`
```
[master.maintenance]
# periodically run these scripts are the same as running them from 'weed shell'
scripts = """
  lock
  volume.tier.move -fullPercent=0 -fromDiskType=ssd -toDiskType=hdd -quietFor=18h -force
  volume.deleteEmpty -quietFor=24h
  volume.fix.replication
  unlock
"""
sleep_minutes = 10
```
- output of `weed version`: 
```
version 30GB 3.18 475185fb723b50e05f0afe39f7f529302bcee11a linux amd64
```

**Expected behavior**
Maintenance script should be running at least every 10 minutes

**Additional context**
I'm running 3 masters, on all of them i have added `-raftHashicorp` option to startup params.

lyang24:
Hi which linux version are you running with? and can you please share the command to repro?

chrislusf:
@lyang24 The linux version should not matter. 
Also, I think @kmlebedev is working on this issue.

kmlebedev:
The problem did not reproduce, but I found something else

```
I0811 11:11:35.223814 masterclient.go:189 master master0:9333 redirected to leader master2:9335
I0811 11:11:35.223991 masterclient.go:151 .master masterClient Connecting to master master2:9335
I0811 11:11:35.228544 masterclient.go:176 .master masterClient Connected to master2:9335
I0811 11:11:35.230840 masterclient.go:185 masterClient received volume_location:{url:"volume1:8080"  public_url:"localhost:8080"  leader:"master2:9335"  data_center:"dc1"} 
I0811 11:11:35.231017 masterclient.go:248 updateVidMap() resp.VolumeLocation.DataCenter dc1
I0811 11:11:35.231049 masterclient.go:199 masterClient replace master  to master2:9335
I0811 11:11:35.231265 masterclient.go:226 + .master master1:9334 leader:false
I0811 11:11:35.511379 masterclient.go:226 + .filer 172.18.0.8:8888 leader:true
I0811 11:11:39.109420 masterclient.go:97 .adminShell masterClient WaitUntilConnected()  and map[master1:9334:master1:9334]
I0811 11:11:49.110501 masterclient.go:97 .adminShell masterClient WaitUntilConnected()  and map[master1:9334:master1:9334]
[...]

PR #3431, opened 2022-08-11
avoid infinite loop WaitUntilConnected()

kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/3421

https://github.com/seaweedfs/seaweedfs/issues/3421

https://github.com/seaweedfs/seaweedfs/issues/3421

# How is the PR tested?

# Checks

### DIFF
--- a/weed/wdclient/masterclient.go
+++ b/weed/wdclient/masterclient.go
@@ -93,7 +93,10 @@ func (mc *MasterClient) GetMasters() map[string]pb.ServerAddress {
 }
 
 func (mc *MasterClient) WaitUntilConnected() {
-	for mc.currentMaster == "" {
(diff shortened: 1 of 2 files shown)


weed/wdclient/masterclient.go
```diff
@@ -93,7 +93,10 @@ func (mc *MasterClient) GetMasters() map[string]pb.ServerAddress {
 }
 
 func (mc *MasterClient) WaitUntilConnected() {
-	for mc.currentMaster == "" {
+	for {
```
chrislusf, line 96: no logical changes here, right?

kmlebedev: Yes

Issue #2792, opened 2022-03-23
[bug:s3] slash in front of filename should not return 500 (minor)


kmlebedev:
**Describe the bug**

minor 
s3cmd logs
```
WARNING: Upload failed: /old_file//3ea3479c-a798-46c4-9a83-b4d305043620 (500 (InternalError): We encountered an internal error, please try again.)
WARNING: Waiting 9 sec...
```

```
filer_server_handlers_read.go:231] failed to stream content /buckets/files/old_file/3ea3479c-a798-46c4-9a83-b4d305043620: read chunk: http://fast-volume-7:8080/1507,02b3d69219e49b00?readDeleted=true: 404 Not Found
s3api_object_handlers.go:411] upload to filer response read 500: unexpected EOF
```

**System Setup**
```
2.90
```

chrislusf:
what is the upload command to reproduce this?

krestkrest:
I'm reproducing this bug in `local-dev-compose` with the following `aws s3api` commands:
```
aws --endpoint-url=http://localhost:8333 s3api create-bucket --bucket testbucket
aws --endpoint-url=http://localhost:8333 s3api put-object --bucket testbucket --key testdir//testfile --body ./testfile.obj
aws --endpoint-url=http://localhost:8333 s3api put-object --bucket testbucket --key //testdir/testfile --body ./testfile.obj
```
The first command successfully creates a bucket, the second and the third command return:
```
An error occurred (InternalError) when calling the PutObject operation (reached max retries: 2): We encountered an internal error, please try again.
```
And I see the following logs:
```
filer_1   | I0815 09:20:00.434039 filer_server_handlers_read.go:98 Not found /buckets/testbucket/testfile2/testfile3: filer: no entry is found in filer store
s3_1      | E0815 09:20:00.434653 s3api_object_handlers.go:434 failing to read upload to http://filer:8888/buckets/testbucket/testfile2//testfile3 : 

filer_1   | I0815 09:23:46.617542 filer_server_handlers_read.go:98 Not found /buckets/testbucket/testfile2/testfile3: filer: no entry is found in filer store
s3_1      | E0815 09:23:46.617913 s3api_object_handlers.go:434 failing to read upload to http://filer:8888/buckets/testbucket//testfile2/testfile3 : 
```
[...]

PR #3442, opened 2022-08-15
Remove duplicate slashes in object path to prevent 500 errors

krestkrest:

https://github.com/seaweedfs/seaweedfs/issues/2792

The current implementation does PUT/POST/DELETE requests from s3 API to filer with duplicate slashes in object path (if duplicate slashes were in object path). This leads to 301 Moved Permanently response from filer with `Location` that has duplicate slashes removed. When HTTP client in s3 API does second request to the redirected URL, it changes `PUT` method to `GET`. The workaround cleans the object path from duplicate slashes to prevent 301 redirect.

# How is the PR tested?

Tested in `local-dev-compose.yml`, the command sequence is the following:
```
aws --endpoint-url=http://localhost:8333 s3api create-bucket --bucket testbucket
aws --endpoint-url=http://localhost:8333 s3api put-object --bucket testbucket --key testdir//testfile --body ./testfile.obj
aws --endpoint-url=http://localhost:8333 s3api put-object --bucket testbucket --key //testdir/testfile --body ./testfile.obj
```
Added unit tests.

# Checks
- [✓] I have added unit tests if possible.
- [✓] I will add related wiki document changes and link to this PR after merging.

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -130,9 +130,30 @@ func urlPathEscape(object string) string {
 	return strings.Join(escapedParts, "/")
 }
 
+func removeDuplicateSlashes(object string) string {
+	result := strings.Builder{}
+	result.Grow(len(object))
+
+	isLastSlash := false
+	for _, r := range object {
+		switch r {
+		case '/':
+			if !isLastSlash {
+				result.WriteRune(r)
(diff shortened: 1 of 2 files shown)


kmlebedev:
This seems to be a bad idea, since it's just part of the key and it's not clear why it needs to be removed.

PR #3458, opened 2022-08-17
feat(weed.move): Copy and move: add a speed limit parameter

qzhello:
When using the shell for move and copy, we want to limit the speed, not limited to the default parameter (compactionBytePerSecond) of the volume server.
We will evaluate the movement speed of different collections to ensure safe and fast range.

Increase the rate limit parameter when requesting, if it is 0, use the default configuration of the volume server.
```
message VolumeCopyRequest {
    uint32 volume_id = 1;
    string collection = 2;
    string replication = 3;
    string ttl = 4;
    string source_data_node = 5;
    string disk_type = 6;
    // here
    int64 compaction_byte_per_second = 7;
}
```

# Checks

### DIFF
--- a/weed/mount/weedfs_file_copy_range.go
+++ b/weed/mount/weedfs_file_copy_range.go
@@ -65,22 +65,41 @@ func (wfs *WFS) CopyFileRange(cancel <-chan struct{}, in *fuse.CopyFileRangeIn) 
 		return 0, fuse.EISDIR
 	}
 
-	// cannot copy data to an overlapping range of the same file
-	offInEnd := in.OffIn + in.Len - 1
-	offOutEnd := in.OffOut + in.Len - 1
-
-	if fhIn.inode == fhOut.inode && in.OffIn <= offOutEnd && offInEnd >= in.OffOut {
-		return 0, fuse.EINVAL
-	}
-
 	glog.V(4).Infof(
-		"CopyFileRange %s fhIn %d -> %s fhOut %d, %d:%d -> %d:%d",
+		"CopyFileRange %s fhIn %d -> %s fhOut %d, [%d,%d) -> [%d,%d)",
 		fhIn.FullPath(), fhIn.fh,
 		fhOut.FullPath(), fhOut.fh,
-		in.OffIn, offInEnd,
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_move.go
```diff
@@ -55,6 +55,7 @@ func (c *commandVolumeMove) Do(args []string, commandEnv *CommandEnv, writer io.
 	sourceNodeStr := volMoveCommand.String("source", "", "the source volume server <host>:<port>")
 	targetNodeStr := volMoveCommand.String("target", "", "the target volume server <host>:<port>")
 	diskTypeStr := volMoveCommand.String("disk", "", "[hdd|ssd|<tag>] hard drive or solid state drive or any tag")
+	compactionBytePerSecond := volMoveCommand.Int64("compactionBytePerSecond", 0, "limit the speed of copy")
```
chrislusf: this is not compaction. how about `ioBytePerSecond`? (Same applies to the variable names.)

weed/server/volume_grpc_copy.go
```diff
@@ -108,7 +108,14 @@ func (vs *VolumeServer) VolumeCopy(req *volume_server_pb.VolumeCopyRequest, stre
 		nextReportTarget := reportInterval
 		var modifiedTsNs int64
 		var sendErr error
-		if modifiedTsNs, err = vs.doCopyFile(client, false, req.Collection, req.VolumeId, volFileInfoResp.CompactionRevision, volFileInfoResp.DatFileSize, dataBaseFileName, ".dat", false, true, func(processed int64) bool {
+		var compactionBytePerSecond = func() int64 {
```
chrislusf: why need the closure here?

weed/shell/command_volume_move.go
qzhello: I agree with you.

weed/server/volume_grpc_copy.go
qzhello: I'm sorry. What does that mean?

chrislusf: the wrapping function is not necessary

PR #3473, opened 2022-08-21
Fix hanging reads in chunk cacher

Woellchen:
Sometimes when an unexpected error occurs the cacher would set an error and return.
However, it would not broadcast the condition signal in that case, therefore leaving the goroutine that runs readChunkAt stuck forever.

I figured that the condition is unnecessary because readChunkAt is acquiring a lock that is still held by the cacher goroutine anyway.
Callees of startCaching have to wait for a WaitGroup which makes sure that readChunkAt can't acquire the lock before startCaching.
This way readChunkAt can execute normally and check for the error.

# How is the PR tested?
Have `retriedFetchChunkData` return an unexpected error, like io.EOF, or anything else.
Can happen when the filer disappears and a file is read, for example.

# Checks

### DIFF
--- a/weed/filer/reader_cache.go
+++ b/weed/filer/reader_cache.go
@@ -2,11 +2,12 @@ package filer
 
 import (
 	"fmt"
+	"sync"
+	"time"
+
 	"github.com/seaweedfs/seaweedfs/weed/util/chunk_cache"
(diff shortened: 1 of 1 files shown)


chrislusf:
This is much more elegant! Thanks!

PR #3478, opened 2022-08-22
feat(weed.move): add a speed limit parameter of moving files

qzhello:
Refer:
https://github.com/seaweedfs/seaweedfs/pull/3458

add a new parameter of moving files: io_byte_per_second
```
message VolumeCopyRequest {
    uint32 volume_id = 1;
    string collection = 2;
    string replication = 3;
    string ttl = 4;
    string source_data_node = 5;
    string disk_type = 6;
    int64 io_byte_per_second = 7;
}
```

# How is the PR tested?

# Checks


weed/server/volume_grpc_copy.go
```diff
@@ -108,7 +108,8 @@ func (vs *VolumeServer) VolumeCopy(req *volume_server_pb.VolumeCopyRequest, stre
 		nextReportTarget := reportInterval
 		var modifiedTsNs int64
 		var sendErr error
-		if modifiedTsNs, err = vs.doCopyFile(client, false, req.Collection, req.VolumeId, volFileInfoResp.CompactionRevision, volFileInfoResp.DatFileSize, dataBaseFileName, ".dat", false, true, func(processed int64) bool {
+		throttler := util.NewWriteThrottler(req.IoBytePerSecond)
```
chrislusf: use the default value if zero

qzhello: Do you mean to add a parameter to VolumeServer and give a default value?

chrislusf: vs.compactionBytePerSecond should be the default value

PR #3495, opened 2022-08-23
fix coredump

guol-fnst:
fix coredump for https://github.com/seaweedfs/seaweedfs/commit/676e27c589a99691ba78f010a067898d89f764ac


guol-fnst:
The "ce.locker.RequestLock()" requires a master server in test case.Or it'll will coredump

guol-fnst:
The "ce.locker.RequestLock()" requires a master server in test case.Or it'll will coredump

PR #3503, opened 2022-08-24
Add raft server access mutex to avoid races

Woellchen:
When starting up the raft server and setting the instance in `Topology.RaftServer` / `Topology.HashicorpRaft` there are race conditions because other goroutines already try to access the reference, which is `nil` in the beginning.

Add an RWMutex to synchronize access.

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -206,11 +206,13 @@ func startMaster(masterOption MasterOptions, masterWhiteList []string) {
 	if !*m.raftHashicorp {
 		go func() {
 			time.Sleep(timeSleep)
-			if ms.Topo.RaftServer.Leader() == "" && ms.Topo.RaftServer.IsLogEmpty() && isTheFirstOne(myMasterAddress, peers) {
-				if ms.MasterClient.FindLeaderFromOtherPeers(myMasterAddress) == "" {
-					raftServer.DoJoinCommand()
-				}
+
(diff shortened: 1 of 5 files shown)


Woellchen:
I have covered all accesses to both `Topo.RaftServer` and `Topo.HashicorpRaft` with locks. Or am I missing something? The raft server instance is only ever set once.

kmlebedev:
> raft server instance is only ever set once

Yes, raft server instance is only ever set once.

Woellchen:
I do only synchronise the access to `Topology.RaftServer` / `Topology.HashicorpRaft`, not to the raft server itself. What else am I missing?

kmlebedev:
> I do only synchronise the access to `Topology.RaftServer` / `Topology.HashicorpRaft`, not to the raft server itself. Other references to the same raft server do not need to be synchronised. Only the references in `Topology`, this is where race conditions are happening.


@Woellchen I don't like this approach, to wrap it around with locks to the fullest because of a single write operation at startup.
Might be worth fixing the root of the problem?

Woellchen:
Thanks! I was fixing this race condition because it caused race conditions on startup, yes. It happens only once and only at startup, but it happens.
I will take your advise and see if I can fix the root cause without too much refactoring! Will let you know about the outcome :)

chrislusf:
thanks! Less code is better.

PR #3504, opened 2022-08-24
Avoid race conditions when opening and working on files in parallel in FUSE mounts

Woellchen:
Avoid race conditions when opening and working on files in parallel in FUSE mounts.

Avoid re-setting the same instance of `filer_pb.Entry` to `FileHandle.entry` on every `Open` call.
Once a file has been opened for the first time, the entry is set and does not change anymore.

# Checks

### DIFF
--- a/weed/mount/filehandle_map.go
+++ b/weed/mount/filehandle_map.go
@@ -1,8 +1,9 @@
 package mount
 
 import (
-	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 	"sync"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 )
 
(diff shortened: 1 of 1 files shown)


weed/mount/filehandle_map.go
```diff
@@ -49,7 +50,9 @@ func (i *FileHandleToInode) AcquireFileHandle(wfs *WFS, inode uint64, entry *fil
 	} else {
 		fh.counter++
 	}
-	fh.entry = entry
+	if fh.entry != entry {
```
chrislusf: I do not think this makes any difference though.

Woellchen: I think I have a better solution now. Will include this in another PR, I found some more issues I will fix along.

PR #3505, opened 2022-08-24
Fix race conditions during in-flight size checks

Woellchen:
Fix race conditions during in-flight size checks. Therefore, it should be more accurate now.

Atomically checking for sizes.

# How is the PR tested?
Upload and download bigger chunks of data in parallel, for example like this:

`echo {1..100} | xargs -P100 -n1 sh -c 'dd if=/dev/zero bs=1m count=1 2>/dev/null | tee mount/patrick2 | tee mount/patrick3 | tee mount/patrick4 | tee mount/patrick5 >/dev/null'; md5sum <(cat mount/patrick2) <(cat mount/patrick3) <(cat mount/patrick4) <(cat mount/patrick5)`

# Checks

### DIFF
--- a/weed/server/volume_server_handlers.go
+++ b/weed/server/volume_server_handlers.go
@@ -45,15 +45,17 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
 	case "GET", "HEAD":
 		stats.ReadRequest()
 		vs.inFlightDownloadDataLimitCond.L.Lock()
-		for vs.concurrentDownloadLimit != 0 && atomic.LoadInt64(&vs.inFlightDownloadDataSize) > vs.concurrentDownloadLimit {
+		inFlightDownloadSize := atomic.LoadInt64(&vs.inFlightDownloadDataSize)
+		for vs.concurrentDownloadLimit != 0 && inFlightDownloadSize > vs.concurrentDownloadLimit {
 			select {
 			case <-r.Context().Done():
 				glog.V(4).Infof("request cancelled from %s: %v", r.RemoteAddr, r.Context().Err())
 				return
 			default:
-				glog.V(4).Infof("wait because inflight download data %d > %d", vs.inFlightDownloadDataSize, vs.concurrentDownloadLimit)
+				glog.V(4).Infof("wait because inflight download data %d > %d", inFlightDownloadSize, vs.concurrentDownloadLimit)
 				vs.inFlightDownloadDataLimitCond.Wait()
 			}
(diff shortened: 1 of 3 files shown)


chrislusf:
Prefer to reduce channel usage for each request. 

Channel is not faster than signals and this piece of logic is not so easy to understand.

weed/server/volume_server_handlers.go
```diff
@@ -36,55 +36,76 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
...
+
+		inFlightDownloadSize := atomic.LoadInt64(&vs.inFlightDownloadDataSize)
+		if vs.concurrentDownloadLimit != 0 && inFlightDownloadSize > vs.concurrentDownloadLimit {
+			for {
+				glog.V(4).Infof("wait because inflight download data %d > %d", inFlightDownloadSize, vs.concurrentDownloadLimit)
+				select {
+				case <-r.Context().Done():
+					glog.V(4).Infof("request cancelled from %s: %v", r.RemoteAddr, r.Context().Err())
+					go func() {
+						<-vs.inFlightDownloadWaitCh
```
chrislusf: I could not follow the logic easily here. Why an extra goroutine is needed, which is much heavier than atomic operations.

Maybe do some benchmark to see whether this is faster.

weed/server/volume_server_handlers.go
```diff
@@ -36,55 +36,76 @@ func (vs *VolumeServer) privateStoreHandler(w http.ResponseWriter, r *http.Reque
...
 		}
+
 		atomic.AddInt64(&vs.inFlightUploadDataSize, contentLength)
 		defer func() {
 			atomic.AddInt64(&vs.inFlightUploadDataSize, -contentLength)
-			if vs.concurrentUploadLimit != 0 {
-				vs.inFlightUploadDataLimitCond.Signal()
+			select {
+			case vs.inFlightUploadWaitCh <- struct{}{}:
+			default:
```
chrislusf: with the default clause, this means the `vs.inFlightUploadWaitCh <- struct{}{}` may not actually happen if blocked?

Woellchen: Yes, just like `cond.Signal()`

weed/server/volume_server_handlers.go
Woellchen: This is needed in case the request was canceled, to make sure the incoming signal is drained from the channel. But if you don't like this I can make it `Cond` again. I just thought channels are more idiomatic.

Woellchen:
I reverted to `sync.Cond` and only fixed the specific race conditions

PR #3519, opened 2022-08-26
raft hashicorp enable by default

kmlebedev:

https://github.com/seaweedfs/seaweedfs/pull/2902

https://github.com/seaweedfs/seaweedfs/pull/2902

# How is the PR tested?

# Checks


chrislusf:
I did some research here. https://github.com/hashicorp/raft does not support joint consensus. 

https://github.com/hashicorp/raft/search?q=joint&type=issues

So it can only add or remove members one at a time.

chrislusf:
For all go implementation here https://raft.github.io/, seems only https://github.com/etcd-io/etcd/tree/main/raft can really support joint consensus. It is also used by commercial product, e.g., cockroachdb, tikv, etc.

Joint consensus was also mentioned in this CockroachDB paper https://dl.acm.org/doi/pdf/10.1145/3318464.3386134

We should go with this library.

chrislusf:
There are some `weed shell` command changes to go with this PR, which may be affected by future etcd raft implementation.

You will have more flexibility to adjust the commands by not moving this implementation to default.

PR #3522, opened 2022-08-26
add target dev_race to MakeFile

kmlebedev:

local dev with -race

add target dev_race

# How is the PR tested?

# Checks


kmlebedev:
https://github.com/golang/go/issues/54695

PR #3523, opened 2022-08-26
replace exec Leader func to MaybeLeader witout retry

kmlebedev:

We do not wait for a few seconds of information about the leader where it is not needed

replace exec Leader func to MaybeLeader witout retry
# How is the PR tested?

# Checks


chrislusf:
Removing retry can help what problem?

Issue #3515, opened 2022-08-25
[volume] DATA RACE on Created Volume id


kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/3507

```
volume_1   | I0825 10:41:53.044853 store.go:143 In dir /data adds volume:74 collection:yournamehere-ndnlnnt6gnh6x06j-123 replicaPlacement:000 ttl:
volume_1   | I0825 10:41:53.045166 volume_loading.go:140 loading memory index /data/yournamehere-ndnlnnt6gnh6x06j-123_74.idx to memory
volume_1   | I0825 10:41:53.048117 store.go:147 add volume 74
volume_1   | I0825 10:41:53.048595 volume_grpc_client_to_master.go:172 volume server volume:8080 adds volume 74
master_1   | I0825 10:41:53.049780 volume_layout.go:391 Volume 74 becomes writable
master_1   | I0825 10:41:53.050830 volume_growth.go:245 Created Volume 74 on topo:DefaultDataCenter:DefaultRack:volume:8080
volume_1   | ==================
volume_1   | WARNING: DATA RACE
volume_1   | Write at 0x00c000890e58 by goroutine 172:
volume_1   |   github.com/seaweedfs/seaweedfs/weed/storage.(*Volume).Version()
volume_1   |       /go/src/github.com/seaweedfs/seaweedfs/weed/storage/volume.go:101 +0x14a
volume_1   |   github.com/seaweedfs/seaweedfs/weed/storage.(*Volume).syncWrite()
volume_1   |       /go/src/github.com/seaweedfs/seaweedfs/weed/storage/volume_write.go:96 +0x69
volume_1   |   github.com/seaweedfs/seaweedfs/weed/storage.(*Volume).writeNeedle2()
volume_1   |       /go/src/github.com/seaweedfs/seaweedfs/weed/storage/volume_write.go:117 +0x5a4
volume_1   |   github.com/seaweedfs/seaweedfs/weed/storage.(*Store).WriteVolumeNeedle()
volume_1   |       /go/src/github.com/seaweedfs/seaweedfs/weed/storage/store.go:363 +0xd5
volume_1   |   github.com/seaweedfs/seaweedfs/weed/topology.ReplicatedWrite()
volume_1   |       /go/src/github.com/seaweedfs/seaweedfs/weed/topology/store_replicate.go:48 +0x272
[...]

PR #3527, opened 2022-08-26
Fix a few data races when reading files in mount

Woellchen:
Fix a few data races when reading files in mount, especially with concurrent operations.

Employ locks and atomic operations to update states.

# Checks

### DIFF
--- a/weed/filer/reader_cache.go
+++ b/weed/filer/reader_cache.go
@@ -3,6 +3,7 @@ package filer
 import (
 	"fmt"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/seaweedfs/seaweedfs/weed/util/chunk_cache"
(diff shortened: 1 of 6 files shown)


Woellchen:
Follows up #3504

Woellchen:
Should fix #3515

chrislusf:
Does this PR fix the issue with QEMU images?

Woellchen:
Unfortunately that was not enough to make it work. But I'll keep trying :)

chrislusf:
When writing, the write requests are synchronized by a lock. I feel it should have been a channel with large enough buffers, or a list ordered by request time stamps, to ensure ordering.

Woellchen:
Yes, I saw the locks. However, I think that "ordering" anything in the filesystem will solve any problem. It's up to the applications that use the filesystem do synchronize their actions, IF they use concurrency at all.
I rather have the feeling something is not 100% working with even single-threaded operations. I would like to find out the exact root cause of it. Still on it.

chrislusf:
We could start with creating a test to reproduce the problem, very likely multithreaded read and write with overlapping sections.

chrislusf:
this seems a good test https://github.com/seaweedfs/seaweedfs/issues/3535

Issue #3518, opened 2022-08-26
I lost files when I uploaded them in bulk


dzh6699:
**Describe the bug**
Files uploaded in batches are lost

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
- OS version
readhat 7 
- output of `weed version`
3.23
- if using filer, show the content of `filer.toml`
default
**Expected behavior**

**Screenshots**

Ready to upload,   11 files
![image](https://user-images.githubusercontent.com/29142176/186805609-4e550a19-74b6-424e-87fe-3ca4184d8675.png)

uploading
![image](https://user-images.githubusercontent.com/29142176/186805662-399890f1-7f66-43e3-ba41-af5b7b18d50b.png)

result ,  6 files

![image](https://user-images.githubusercontent.com/29142176/186805837-04768802-0f32-4e46-a4bc-728c82d58a62.png)

chrislusf:
seems something wrong with the javascript in the filer UI.

chrislusf:
@blacktear23 could you please take a look here?

blacktear23:
@dzh6699 after you uploaded, wait some seconds then use shift+f5 to refresh, then can you see the lost files?

dzh6699:
> @dzh6699 after you uploaded, wait some seconds then use shift+f5 to refresh, then can you see the lost files?

no, i can't

dzh6699:
> @dzh6699 after you uploaded, wait some seconds then use shift+f5 to refresh, then can you see the lost files?

![image](https://user-images.githubusercontent.com/29142176/187069189-944023c2-7139-4d3e-8c45-4b5cff558bf4.png)

dzh6699:
![image](https://user-images.githubusercontent.com/29142176/187069232-349fa37e-53eb-4381-9e60-edd93009b185.png)

blacktear23:
@dzh6699 It should be fixed after next release.

PR #3534, opened 2022-08-28
Fix reload page before all file uploaded bug

blacktear23:

Fix issue #3518 

Do not reload page before all file uploaded.

# How is the PR tested?

Test manually, upload many files include big and small files.

# Checks

### DIFF
--- a/weed/server/filer_ui/filer.html
+++ b/weed/server/filer_ui/filer.html
@@ -288,12 +288,12 @@
     function uploadFile(file, i) {
         var url = window.location.href;
         var xhr = new XMLHttpRequest();
+        var fileName = file.name;
         xhr.onreadystatechange = function() {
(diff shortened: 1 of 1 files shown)


blacktear23:
This issue is introduced by 986daecbfea12bffd4eef5b601108639428c0e7a

blacktear23:
Actually `loadend` event should be triggered by browser after the upload request finish (success or error). And `loadend` event is supported by modern browser such as Chrome, Edge, Firefox, Safari and so on. And for now IE is not supported by Microsoft so I cannot find whether IE support it. Maybe some version of IE cannot trigger the `loadend` event and the page not refreshed after all file uploaded.

chrislusf:
Thanks for the help!

IE is dead. No need to worry about it.

PR #3539, opened 2022-08-29
avoid race conditions access to  SuperBlock.Version

kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/3515

https://github.com/seaweedfs/seaweedfs/issues/3515

https://github.com/seaweedfs/seaweedfs/issues/3515

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/volume.go
+++ b/weed/storage/volume.go
@@ -36,6 +36,7 @@ type Volume struct {
 	super_block.SuperBlock
 
 	dataFileAccessLock    sync.RWMutex
+	superBlockAccessLock  sync.RWMutex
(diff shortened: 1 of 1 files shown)


weed/storage/volume.go
```diff
@@ -97,6 +98,8 @@ func (v *Volume) FileName(ext string) (fileName string) {
 }
 
 func (v *Volume) Version() needle.Version {
+	v.superBlockAccessLock.RLock()
```
chrislusf: should be just a Mutex.

kmlebedev: > should be just a Mutex.

@chrislusf  Done

Issue #3521, opened 2022-08-26
[master] panic if use hashicorp with -race detect


kmlebedev:
```
master branch
cd docker; make hashicorp_raft with -race
```

```
master2_1  | 
master2_1  | goroutine 1 [running]:
master2_1  | runtime.throw({0x334f6b4?, 0x26ed401?})
master2_1  |    /usr/local/go/src/runtime/panic.go:1047 +0x5d fp=0xc0008c6f88 sp=0xc0008c6f58 pc=0x47205d
master2_1  | runtime.checkptrAlignment(0x4aa339?, 0x4840a5?, 0x5865a28?)
master2_1  |    /usr/local/go/src/runtime/checkptr.go:26 +0x6c fp=0xc0008c6fa8 sp=0xc0008c6f88 pc=0x43e7ac
master2_1  | github.com/boltdb/bolt.(*Bucket).write(0xc0008c7140)
master2_1  |    /go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:626 +0x211 fp=0xc0008c7038 sp=0xc0008c6fa8 pc=0x26f13f1
master2_1  | github.com/boltdb/bolt.(*Bucket).CreateBucket(0xc000872018, {0x5865a28, 0x4, 0x4})
master2_1  |    /go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:188 +0x3e9 fp=0xc0008c7228 sp=0xc0008c7038 pc=0x26ed409
master2_1  | github.com/boltdb/bolt.(*Bucket).CreateBucketIfNotExists(0x0?, {0x5865a28, 0x4, 0x4})
master2_1  |    /go/pkg/mod/github.com/boltdb/bolt@v1.3.1/bucket.go:206 +0x4f fp=0xc0008c7278 sp=0xc0008c7228 pc=0x26ed6cf
master2_1  | github.com/boltdb/bolt.(*Tx).CreateBucketIfNotExists(...)
master2_1  |    /go/pkg/mod/github.com/boltdb/bolt@v1.3.1/tx.go:115
master2_1  | github.com/hashicorp/raft-boltdb.(*BoltStore).initialize(0xc000012360)
master2_1  |    /go/pkg/mod/github.com/hashicorp/raft-boltdb@v0.0.0-20220329195025-15018e9b97e0/bolt_store.go:100 +0xeb fp=0xc0008c7320 sp=0xc0008c7278 pc=0x270f3eb
[...]

PR #3551, opened 2022-08-30
etcd raft

kmlebedev:

joint consensus raft
https://github.com/seaweedfs/seaweedfs/pull/3519#issuecomment-1229064136

use etcd raft
https://github.com/etcd-io/etcd/tree/main/raft
https://medium.com/@kjanshair_4991/using-etcd-raftexample-for-go-b469aad2eadc

# How is the PR tested?

# Checks


kmlebedev:
@chrislusf There is a problem with etcd Raft, etcd in the example does not offer a ready-made interface for kv storage, but only raft.MemoryStorage which is not acceptable.

At the same time, hashicorp raft uses raftboltdb, but it is no longer maintained and there are problems with compilation -race https://github.com/seaweedfs/seaweedfs/issues/3521

chrislusf:
we can start with raft.MemoryStorage, and change to a different implementation if necessary.

kmlebedev:
> we can start with raft.MemoryStorage, and change to a different implementation if necessary.

Found a solution https://github.com/seaweedfs/seaweedfs/pull/3554

Issue #3564, opened 2022-09-01
[filer] DATA RACE on signal_handling


kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/3507

```
==================
WARNING: DATA RACE
Read at 0x00c0001276d0 by goroutine 36:
  github.com/seaweedfs/seaweedfs/weed/util/grace.init.0.func1()
      /go/src/github.com/seaweedfs/seaweedfs/weed/util/grace/signal_handling.go:31 +0xac

Previous write at 0x00c0001276d0 by main goroutine:
  github.com/seaweedfs/seaweedfs/weed/util/grace.OnInterrupt()
      /go/src/github.com/seaweedfs/seaweedfs/weed/util/grace/signal_handling.go:46 +0xfb
  github.com/seaweedfs/seaweedfs/weed/server.NewFilerServer()
      /go/src/github.com/seaweedfs/seaweedfs/weed/server/filer_server.go:179 +0x1999
  github.com/seaweedfs/seaweedfs/weed/command.(*FilerOptions).startFiler()
      /go/src/github.com/seaweedfs/seaweedfs/weed/command/filer.go:227 +0xcf7
  github.com/seaweedfs/seaweedfs/weed/command.runFiler()
      /go/src/github.com/seaweedfs/seaweedfs/weed/command/filer.go:203 +0x892
  main.main()
      /go/src/github.com/seaweedfs/seaweedfs/weed/weed.go:81 +0x943

Goroutine 36 (running) created at:
  github.com/seaweedfs/seaweedfs/weed/util/grace.init.0()
      /go/src/github.com/seaweedfs/seaweedfs/weed/util/grace/signal_handling.go:29 +0x18a
==================
Found 3 data race(s)
```

PR #3572, opened 2022-09-01
avoid data race on grace.hooks

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/3564

https://github.com/seaweedfs/seaweedfs/issues/3564

# How is the PR tested?

# Checks

### DIFF
--- a/weed/util/grace/signal_handling.go
+++ b/weed/util/grace/signal_handling.go
@@ -12,7 +12,7 @@ import (
 
 var signalChan chan os.Signal
 var hooks = make([]func(), 0)
-var hookLock sync.Mutex
+var hookLock sync.RWMutex
 
 func init() {
 	signalChan = make(chan os.Signal, 1)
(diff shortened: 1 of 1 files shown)


Woellchen:
Hey, I just saw that the new e2e tests started failing with this commit and I checked why.
I found that the filer service in the workflow didn't come up (port did not open).
It seems that the new `RLock` in the goroutine is at fault, because all `OnInterrupt()` calls would now block forever, for example in `NewFilerServer()`. This leads to the filer server never opening up it's port.

kmlebedev:
fixed https://github.com/seaweedfs/seaweedfs/pull/3575

kmlebedev:
@chrislusf Let's temporarily turn on CGO for darwin, since the tests are not informative, they always fall because of this ?
```
    pre_command: export CGO_ENABLED=0 && export GODEBUG=http2client=0
```

You also need to "fix" the CodeQL test

Screenshot 2022-09-02 at 12 10 40

Issue #2578, opened 2022-01-12
fsck: Orphan entries while filer.copy running


KyleSanderson:
**Describe the bug**
When running `weed filer.copy` (per the recommendation) and then `volume.fsck` it's returning orphans for already split files as the file entry has not been committed yet. This results in data loss when using the `-reallyDeleteFromVolume` option.

**System Setup**
- 5.4
- `unknown command: version` | `version 30GB 2.85 19555385 linux amd64`
- ```[leveldb2]
enabled = true
dir = "/data/filerldb2"```

**Expected behavior**
No data loss.

**Additional context**
There doesn't appear to be a way to configure volumes or anything with the k8s operator, so I've written my own deployment. Filer fails to deploy properly because the index is locked by the other rs. Additionally it's pretty painful to use something like k8s with seaweedfs because you cannot specify which filer.toml to use.

chrislusf:
Is the `volume.fsck` run after the `weed filer.copy`, or at the same time?

KyleSanderson:
Same time.

chrislusf:
Not good to run it at the same time, with the reason as you described.

Maybe later an "age" setting can be added to skip volumes too close to current timestamp.

KyleSanderson:
From what I'm seeing (I'm only like 3 hours into code - so pretty fresh) it looks like this is never a safe operation if there are any clients connected. I agree with the expert tag on the flag, however the documentation seems to indicate that removing the entry is as good as a deletion (from an encryption perspective). Running fsck twice does not guarantee the same operation, and as such is never safe to run in its present state.

This is my completely naive understanding of it.

chrislusf:
This should have been fixed with the above 2 PRs.

PR #3581, opened 2022-09-04
ADHOC: add read needle meta grpc

lyang24:
Add a grpc call/ resp to volume server to extract needle metadata (crc, ttl, last modified..) at specific offset with given size.

added function to read needle at specific offset with provided size on volume servers.  ec shards not supported.

# How is the PR tested?
unit test added

# Checks


weed/storage/erasure_coding/ec_volume.go
```diff
@@ -210,10 +210,14 @@ func (ev *EcVolume) LocateEcShardNeedle(needleId types.NeedleId, version needle.
 		return types.Offset{}, 0, nil, fmt.Errorf("FindNeedleFromEcx: %v", err)
 	}
 
-	shard := ev.Shards[0]
+	intervals = ev.LocateEcShardNeedleInterval(version, offset.ToActualOffset(), types.Size(needle.GetActualSize(size, version)))
+	return
+}
 
+func (ev *EcVolume) LocateEcShardNeedleInterval(version needle.Version, offset int64, size types.Size) (intervals []Interval) {
+	shard := ev.Shards[0]
```
lyang24, line 218: I am not sure if i can hardcode on to pick first shard here

weed/storage/store_ec.go
```diff
@@ -162,6 +162,32 @@ func (s *Store) ReadEcShardNeedle(vid needle.VolumeId, n *needle.Needle, onReadS
 	return 0, fmt.Errorf("ec shard %d not found", vid)
 }
 
+func (s *Store) ReadEcShardNeedleAt(vid needle.VolumeId, n *needle.Needle, offset int64, size int32) (int, error) {
```
lyang24: is it possible for me to pass a reader option that have flags that read meta data only

weed/storage/erasure_coding/ec_volume.go
chrislusf: do not need to support EC volumes.

kmlebedev:
@lyang24 Thank you very much, I would also like to know more specifically, could this PR be useful?

lyang24:
> 

Hi we will use this on volume.fsck command to fiter out files that modifed under the threshold  to avoid inconsistecies such as file chuck created on volume server but no information on filer yet
this is the related issue https://github.com/seaweedfs/seaweedfs/issues/2578

kmlebedev:
> Hi we will use this on volume.fsck command to fiter out files that modifed under the threshold to avoid inconsistecies such as file chuck created on volume server but no information on filer yet this is the related issue #2578

Thanks, it seems this PR will also be useful for https://github.com/seaweedfs/seaweedfs/issues/3614

PR #3583, opened 2022-09-04
Added ability to change replication settings upon volume.tier.move

bddrey:
Often when moving from tiers the replication setting will need to be changed. 

Instead of manually searching for the transferred volumes on the new tiered storage, this will allow user to automatically perform the action with the volume.tier.move operation.

# How is the PR tested?
Regression test; new feature tests

# Checks
- [ N/A ] I have added unit tests if possible.

### DIFF
--- a/weed/shell/command_volume_tier_move.go
+++ b/weed/shell/command_volume_tier_move.go
@@ -12,7 +12,11 @@ import (
 	"path/filepath"
 	"sync"
 	"time"
+	"context"
+	"errors"
 
+	"github.com/seaweedfs/seaweedfs/weed/operation"
+        "github.com/seaweedfs/seaweedfs/weed/pb/volume_server_pb"
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 )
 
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_tier_move.go
```diff
@@ -59,6 +63,8 @@ func (c *commandVolumeTierMove) Do(args []string, commandEnv *CommandEnv, writer
 	parallelLimit := tierCommand.Int("parallelLimit", 0, "limit the number of parallel copying jobs")
 	applyChange := tierCommand.Bool("force", false, "actually apply the changes")
 	ioBytePerSecond := tierCommand.Int64("ioBytePerSecond", 0, "limit the speed of move")
+	replicationString := tierCommand.String("replication", "", "the source disk type");
```
chrislusf: Is it a "toReplication"?

bddrey: I debated that but decided on just replication instead. If that is your preferred I will change it.

chrislusf: It could sound like a criteria to select which volume to move.

bddrey:
Requested change for 'toReplication' pushed to same commit. Ready for review.

PR #3588, opened 2022-09-05
arangodb s3 bucket name compatibility  

gfxlabs:

arangodb collection name restrictions are not the same as s3 bucket names. this causes some bucket names (those using "." or starting with a number) to break.

this introduces special character sets 'xN--' which is prepended to the collection name if starts with number, and 'xP--'  used to replace all '.' characters. since bucket names must not contain upper case letters, this should be 100% compatible with s3 bucket name rules. this is only in the storage layer and should be transparent to any user. 

I'm not sure if this is an issue for buckets for other filers, but this hack could be used for them as well if it is.

# additional changes

this also adds a defer to properly close the context used to initialize the arangodb connection, and properly escapes collection names in the AQL queries


chrislusf:
Is it possible to just replace `.` with underscore?

gfxlabs:
that would cause bucket names `foo.bar` and `foo_bar` to point towards the same bucket, which makes it incompatible with s3

chrislusf:
For current code, what problem it will cause?

gfxlabs:
Ah sorry, I guess the person who is my seaweed fs instance is using underscores, when S3 actually does not permit it. 

I don't mind forcing them to rename. I can edit the PR to replace . -> _ tomorrow

PR #3615, opened 2022-09-07
feat: Remove volume download lock

famosss:
The download of online services is sometimes very very slow. I added some debugging logs and found that when there is a very slow upload, the download side needs to wait for a long time to get the lock.
Because the entire `volume` is an append operation, there is no update operation, I did not expect the function of this `lock`, so deleted it
wecom-temp-195a08c56813a114dc0df566b84add56
wecom-temp-2faef2712b503d3b306a95c9e6f229f9

filer maxMB:32M

### DIFF
--- a/weed/storage/volume.go
+++ b/weed/storage/volume.go
@@ -37,6 +37,7 @@ type Volume struct {
 
 	dataFileAccessLock    sync.RWMutex
 	superBlockAccessLock  sync.Mutex
+	compactAccessLock  sync.RWMutex
 	asyncRequestsChan     chan *needle.AsyncRequest
 	lastModifiedTsSeconds uint64 // unix time in seconds
 	lastAppendAtNs        uint64 // unix time in nanoseconds
--- a/weed/storage/volume_read.go
+++ b/weed/storage/volume_read.go
@@ -100,9 +100,6 @@ func (v *Volume) readNeedleMetaAt(n *needle.Needle, offset int64, size int32) (e
 
 // read fills in Needle content by looking up n.Id from NeedleMapper
 func (v *Volume) readNeedleDataInto(n *needle.Needle, readOption *ReadOption, writer io.Writer, offset int64, size int64) (err error) {
-	v.dataFileAccessLock.RLock()
-	defer v.dataFileAccessLock.RUnlock()
-
 	nv, ok := v.nm.Get(n.Id)
 	if !ok || nv.Offset.IsZero() {
 		return ErrorNotFound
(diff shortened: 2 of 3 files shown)


kmlebedev:
@famosss Here we need to think carefully and refactor the dataFileAccessLock lock, as a vacuum race may occur.
It might be worth lowering the priority for vacuum and replacing it with TryLock()/TryRLock()

famosss:
> @famosss Here we need to think carefully and refactor the dataFileAccessLock lock, as a vacuum race may occur. It might be worth lowering the priority for vacuum and replacing it with TryLock()/TryRLock()

Help consider a scenario where this read lock is useful。

chrislusf:
there would be compaction operation that needs to lock.

chrislusf:
We can use read lock for non compaction operations.

famosss:
> We can use read lock for non compaction operations.

Is it possible to simply add a `compactAccessLock` to solve this problem? And then there will be no write impact on read.

chrislusf:
> Is it possible to simply add a `compactAccessLock` to solve this problem? And then there will be no write impact on read.

I would like to learn how to add it simply.

famosss:
> > Is it possible to simply add a `compactAccessLock` to solve this problem? And then there will be no write impact on read.
> 
> I would like to learn how to add it simply.

Submitted the code before, please check it.

chrislusf:
it is just changing one lock to another. What is the benefit?

kmlebedev:
Need more information about our installation.  Maybe it's a matter of performance or vacuum settings or the specifics of the workload. 

I can only show my graph with server volume response times.

Screenshot 2022-09-08 at 14 01 31

famosss:
> Need more information about our installation. Maybe it's a matter of performance or vacuum settings or the specifics of the workload.
> 
> I can only show my graph with server volume response times.
> 
> Screenshot 2022-09-08 at 14 01 31

`start command`
image

My local environment is speed-limited, and downloading a needle is very slow
image

This is the volume monitoring when there was a problem online yesterday。So I want to reproduce locally。
image

famosss:
After adjusting the lock, reading and writing are normal.There is no interaction between read and write, and the acquisition time of the read lock is normal.

image

image

famosss:
I just tested it again, and it can reproduce accurately. Start a speed-limited client to initiate a get request, such as 1KB/s. At this time, all uploads are blocked.

famosss:
[commit](https://github.com/seaweedfs/seaweedfs/commit/d14d029c736b819ab550818852817e8452c05931)  reduce lock scope is better.

Issue #3620, opened 2022-09-08
[aligo] optimization topology struct


kmlebedev:
https://github.com/essentialkaos/aligo

```
go install github.com/essentialkaos/aligo
aligo check weed/topology
```

```
Struct Collection (collection.go:12) fields order can be optimized (40 → 33)

type Collection struct {
  Name                     string
  volumeSizeLimit          uint64
  storageType2VolumeLayout *util.ConcurrentReadMap
  replicationAsMin         bool
}

Struct NodeImpl (node.go:40) fields order can be optimized (112 → 108)

type NodeImpl struct {
  id          NodeId          // -
  parent      Node            // -
  nodeType    string          // -
  value       interface{}     // -
  diskUsages  *DiskUsages     // -
  children    map[NodeId]Node // -
  RWMutex     sync.RWMutex    // lock children
  maxVolumeId needle.VolumeId // -
}

Struct Topology (topology.go:26) fields order can be optimized (304 → 297)

type Topology struct {
  NodeImpl             NodeImpl
  Sequence             sequence.Sequencer
  RaftServer           raft.Server
  vacuumLockCounter    int64
  collectionMap        *util.ConcurrentReadMap
  ecShardMap           map[needle.VolumeId]*..EcShardL…
  pulse                int64
  volumeSizeLimit      uint64
  chanFullVolumes      chan storage.VolumeInfo
  chanCrowdedVolumes   chan storage.VolumeInfo
  Configuration        *Configuration
  HashicorpRaft        *hashicorpRaft.Raft
  UuidMap              map[string][]string
  ecShardMapLock       sync.RWMutex
  RaftServerAccessLock sync.RWMutex
  UuidAccessLock       sync.RWMutex
  replicationAsMin     bool
}

Struct VolumeLayout (volume_layout.go:106) fields order can be optimized (156 → 149)

type VolumeLayout struct {
  growRequestTime  time.Time                          // -
  writables        []needle.VolumeId                  // transient array of writable volume id
  diskType         storage/types.DiskType             // -
  rp               *super_block.ReplicaPlacement      // -
  ttl              *storage/needle.TTL                // -
  vid2location     map[storage/needle.VolumeId]*..…   // -
  crowded          map[storage/needle.VolumeId]str…   // -
  readonlyVolumes  *volumesBinaryState                // readonly volumes
  oversizedVolumes *volumesBinaryState                // oversized volumes
  volumeSizeLimit  uint64                             // -
[...]

PR #3622, opened 2022-09-08
aligo topology

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/3620

https://github.com/seaweedfs/seaweedfs/issues/3620

# How is the PR tested?

# Checks

### DIFF
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -24,30 +24,24 @@ import (
 )
 
 type Topology struct {
-	vacuumLockCounter int64
 	NodeImpl
-
-	collectionMap  *util.ConcurrentReadMap
-	ecShardMap     map[needle.VolumeId]*EcShardLocations
-	ecShardMapLock sync.RWMutex
-
-	pulse int64
-
-	volumeSizeLimit  uint64
-	replicationAsMin bool
-
-	Sequence sequence.Sequencer
-
+	Sequence           sequence.Sequencer
+	RaftServer         raft.Server
+	vacuumLockCounter  int64
+	collectionMap      *util.ConcurrentReadMap
+	ecShardMap         map[needle.VolumeId]*EcShardLocations
+	pulse              int64
+	volumeSizeLimit    uint64
 	chanFullVolumes    chan storage.VolumeInfo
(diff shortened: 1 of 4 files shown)


weed/topology/volume_layout.go
```diff
@@ -104,19 +104,19 @@ func (v *volumesBinaryState) copyState(list *VolumeLocationList) copyState {
...
-	diskType         types.DiskType
 	vid2location     map[needle.VolumeId]*VolumeLocationList
-	writables        []needle.VolumeId // transient array of writable volume id
 	crowded          map[needle.VolumeId]struct{}
 	readonlyVolumes  *volumesBinaryState // readonly volumes
 	oversizedVolumes *volumesBinaryState // oversized volumes
 	volumeSizeLimit  uint64
-	replicationAsMin bool
 	accessLock       sync.RWMutex
+	growRequestCount int32
```
chrislusf, line 118: moving it here may cause atomic operation to fail on 32bit OS.

weed/topology/topology.go
```diff
@@ -24,30 +24,24 @@ import (
...
 	chanFullVolumes    chan storage.VolumeInfo
 	chanCrowdedVolumes chan storage.VolumeInfo
+	Configuration      *Configuration
+	HashicorpRaft      *hashicorpRaft.Raft
+	UuidMap            map[string][]string
 
-	Configuration *Configuration
-
-	RaftServer           raft.Server
+	ecShardMapLock       sync.RWMutex
```
chrislusf, line 41: separating the lock from the to-be-protected map is not good.

weed/topology/topology.go
```diff
@@ -24,30 +24,24 @@ import (
...
-	pulse int64
-
-	volumeSizeLimit  uint64
-	replicationAsMin bool
-
-	Sequence sequence.Sequencer
-
+	Sequence           sequence.Sequencer
+	RaftServer         raft.Server
+	vacuumLockCounter  int64
```
chrislusf, line 30: moving it here may cause atomic operation to fail on 32bit OS.

There is only one instance of Topology in memory. No need to optimize for a few bytes.

Issue #3623, opened 2022-09-08
[aligo] optimization storage structs


kmlebedev:
```
aligo check .

-- {GOPATH} ----------------------------------------------------------------------------

Struct Store (store.go:43) fields order can be optimized (193 → 186)

type Store struct {
  Locations           []*DiskLocation                    // -
  MasterAddress       pb.ServerAddress                   // -
  grpcDialOption      grpc.DialOption                    // -
  Ip                  string                             // -
  PublicUrl           string                             // -
  dataCenter          string                             // optional informaton, overwriting master setting if exists
  rack                string                             // optional information, overwriting master setting if exists
  volumeSizeLimit     uint64                             // read from the master
  Port                int                                // -
  GrpcPort            int                                // -
  NeedleMapKind       NeedleMapKind                      // -
  NewVolumesChan      chan pb/master_pb.VolumeShortIn…   // -
  DeletedVolumesChan  chan master_pb.VolumeShortInfor…   // -
  NewEcShardsChan     chan master_pb.VolumeEcShardInf…   // -
  DeletedEcShardsChan chan master_pb.VolumeEcShardInf…   // -
  connected           bool                               // -
  isStopping          bool                               // -
}

Struct Volume (volume.go:21) fields order can be optimized (304 → 287)

type Volume struct {
  SuperBlock             super_block.SuperBlock          // -
  dir                    string                          // -
  dirIdx                 string                          // -
  Collection             string                          // -
  DataBackend            backend.BackendStorageFile      // -
  nm                     NeedleMapper                    // -
  tmpNm                  TempNeedleMapper                // -
  lastIoError            error                           // -
  needleMapKind          NeedleMapKind                   // -
  asyncRequestsChan      chan *needle.AsyncRequest       // -
  lastModifiedTsSeconds  uint64                          // unix time in seconds
  lastAppendAtNs         uint64                          // unix time in nanoseconds
  lastCompactIndexOffset uint64                          // -
  volumeInfo             *pb/volume_server_pb.VolumeInfo // -
  location               *DiskLocation                   // -
  noWriteLock            sync.RWMutex                    // -
  dataFileAccessLock     sync.RWMutex                    // -
  superBlockAccessLock   sync.Mutex                      // -
  Id                     needle.VolumeId                 // -
  MemoryMapMaxSizeMb     uint32                          // -
  lastCompactRevision    uint16                          // -
[...]

PR #3624, opened 2022-09-08
aligo storage

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/3623

https://github.com/seaweedfs/seaweedfs/issues/3623

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -41,23 +41,25 @@ type ReadOption struct {
  * A VolumeServer contains one Store
  */
 type Store struct {
-	MasterAddress       pb.ServerAddress
-	grpcDialOption      grpc.DialOption
-	volumeSizeLimit     uint64 // read from the master
-	Ip                  string
-	Port                int
-	GrpcPort            int
-	PublicUrl           string
-	Locations           []*DiskLocation
-	dataCenter          string // optional informaton, overwriting master setting if exists
-	rack                string // optional information, overwriting master setting if exists
-	connected           bool
+	Locations       []*DiskLocation
+	MasterAddress   pb.ServerAddress
+	grpcDialOption  grpc.DialOption
+	Ip              string
+	PublicUrl       string
+	dataCenter      string // optional informaton, overwriting master setting if exists
+	rack            string // optional information, overwriting master setting if exists
+	volumeSizeLimit uint64 // read from the master
+	Port            int
+	GrpcPort        int
+
 	NeedleMapKind       NeedleMapKind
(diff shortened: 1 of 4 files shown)


weed/storage/store.go
```diff
@@ -41,23 +41,25 @@ type ReadOption struct {
...
-	volumeSizeLimit     uint64 // read from the master
-	Ip                  string
-	Port                int
-	GrpcPort            int
-	PublicUrl           string
-	Locations           []*DiskLocation
-	dataCenter          string // optional informaton, overwriting master setting if exists
-	rack                string // optional information, overwriting master setting if exists
-	connected           bool
+	Locations       []*DiskLocation
```
chrislusf, line 44: No need. Only one instance in memory.

weed/storage/volume.go
```diff
@@ -19,38 +19,37 @@ import (
...
+	lastModifiedTsSeconds  uint64 // unix time in seconds
+	lastAppendAtNs         uint64 // unix time in nanoseconds
 	lastCompactIndexOffset uint64
-	lastCompactRevision    uint16
+	volumeInfo             *volume_server_pb.VolumeInfo
+	location               *DiskLocation
 
-	isCompacting       bool
-	isCommitCompacting bool
+	noWriteLock          sync.RWMutex
```
chrislusf, line 40: Lose readability.

weed/storage/volume_info.go
```diff
@@ -10,21 +10,21 @@ import (
 )
 
 type VolumeInfo struct {
-	Id                needle.VolumeId
+	DiskType          string
+	Collection        string
+	RemoteStorageName string
```
chrislusf, line 15: Lose readability.

weed/storage/volume_vacuum.go
```diff
@@ -353,13 +353,13 @@ func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldI
 }
 
 type VolumeFileScanner4Vacuum struct {
-	version        needle.Version
-	v              *Volume
 	dstBackend     backend.BackendStorageFile
+	v              *Volume
```
chrislusf, line 357: This is fine, but it does not persist in memory. Not worth while.

weed/storage/store.go
kmlebedev: > No need. Only one instance in memory.

How to see the top 10 structures in memory?

chrislusf: I think that is ordered in terms of bytes saved for each type, not total amount of memory for the server.

Issue #3629, opened 2022-09-08
[master] Game on masterclient.go:288 updateVidMap


kmlebedev:
**Describe the bug**

I see such abnormal messages:
```
I0908 09:27:00.597693 masterclient.go:288 updateVidMap(DefaultDataCenter) .master: :0 volume add: 577, del: 547, add ec: 0 del ec: 0
I0908 09:27:00.585748 master_grpc_server.go:196 master see deleted volume 236 from :0
I0908 09:27:00.599290 master_grpc_server.go:196 master see deleted volume 316 from :0
I0908 09:27:00.599335 master_grpc_server.go:196 master see deleted volume 1293 from :0
I0908 09:27:00.599381 master_grpc_server.go:196 master see deleted volume 1506 from :0
I0908 09:27:00.599416 master_grpc_server.go:196 master see deleted volume 1415 from :0
I0908 09:27:00.599438 master_grpc_server.go:196 master see deleted volume 2587 from :0
I0908 09:27:00.599458 master_grpc_server.go:196 master see deleted volume 1674 from :0
I0908 09:27:00.599478 master_grpc_server.go:196 master see deleted volume 1169 from :0
```

normal messages:
```
I0908 09:27:43.339900 masterclient.go:288 updateVidMap(stagedc) .master: fast-volume-2.s3-fast-volume.service.stagedc.consul:8080 volume add: 957, del: 0, add ec: 0 del ec: 0
```

**System Setup**
```
3.26 master branch
```

kmlebedev:
Possibly related to https://github.com/seaweedfs/seaweedfs/pull/3621

famosss:
Post more logs?

PR #3630, opened 2022-09-08
skip deltaBeat if dn is zero

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/3629

https://github.com/seaweedfs/seaweedfs/issues/3629

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -127,7 +127,7 @@ func (ms *MasterServer) SendHeartbeat(stream master_pb.Seaweed_SendHeartbeatServ
 		ms.Topo.Sequence.SetMax(heartbeat.MaxFileKey)
 		if dn == nil {
 			// skip deltaBeat
-			if heartbeat.HasNoVolumes {
+			if heartbeat.DataCenter == "" || heartbeat.Rack == "" {
 				continue
 			}
 			dcName, rackName := ms.Topo.Configuration.Locate(heartbeat.Ip, heartbeat.DataCenter, heartbeat.Rack)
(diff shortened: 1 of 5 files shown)


kmlebedev:
@chrislusf This is a critical bug and seems to have appeared in version 3.25 or 3.26 not sure why it used to work fine.

kmlebedev:
@chrislusf I still don’t like the logic where the master sends a message to all clients with the removal of all volumes when leaving the cluster.

It seems that this is superfluous, and so, when changing the leader, we reset the volume map on the client.
What do you think?

kmlebedev:
@chrislusf Ping

weed/server/master_grpc_server.go
```diff
@@ -126,6 +127,10 @@ func (ms *MasterServer) SendHeartbeat(stream master_pb.Seaweed_SendHeartbeatServ
 
 		ms.Topo.Sequence.SetMax(heartbeat.MaxFileKey)
 		if dn == nil {
+			// skip deltaBeat
+			if heartbeat.Ip == "" {
```
chrislusf, line 131: Add a warning message here. It should never happen after this PR.

chrislusf:
> the master sends a message to all clients with the removal of all volumes when leaving the cluster.

which part of code are you referring to?

kmlebedev:
> > the master sends a message to all clients with the removal of all volumes when leaving the cluster.
> 
> which part of code are you referring to?

https://github.com/seaweedfs/seaweedfs/blob/31d2f77ceb8aa0acd4ebf7a85ca24c9375de0a17/weed/server/master_grpc_server.go#L86-L91

After adding the cache(https://github.com/seaweedfs/seaweedfs/pull/3350), it no longer makes sense

PR #3635, opened 2022-09-09
Balance by free space

dimm0:
Balance selectes the nodes with most volumes,not the nodes with most space available

Sorting by free volumes

# How is the PR tested?
On PRP cluster, with this change it moved the most occupied volumes to the least occupied node

# Checks


dimm0:
Ignore plz

Issue #3636, opened 2022-09-10
S3 connection to Cloudflare R2 generates error when attempting to sync from SeaweedFS to R2


devilshdespot:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

example of a good issue report:
https://github.com/seaweedfs/seaweedfs/issues/1005
example of a bad issue report:
https://github.com/seaweedfs/seaweedfs/issues/1008

**Describe the bug**
S3 connection to Cloudflare R2 generates error when attempting to sync from SeaweedFS to R2 because of "Header 'x-amz-storage-class' with value '' not implemented"  R2 api does not support x-amz-storage-class as per https://developers.cloudflare.com/r2/platform/s3-compatibility/api/

**System Setup**
- Command to start is 'weed server -dir=/seaweed -s3 -volume.max=50'
- OS version: Debian 11 x64
- output of `weed version`  version 30GB 3.26 c07ab9c060eafe26cc0b4246d507d5b33f32f317 linux amd64
- if using filer, show the content of `filer.toml` Just the blank filer.toml, I have not customized this for the limited test.

**Expected behavior**
The ability to use the s3 api for r2 or to create a new option like we have for b2 for remote config for cloud storage.

**Screenshots**

**Additional context**
Example of error output from issuing filer sync

[...]

PR #3639, opened 2022-09-10
[sync] override amz storage class, None to delete

kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/3636

https://github.com/seaweedfs/seaweedfs/issues/3636

https://github.com/seaweedfs/seaweedfs/issues/3636

# How is the PR tested?

# Checks

### DIFF
--- a/weed/command/filer_remote_sync_dir.go
+++ b/weed/command/filer_remote_sync_dir.go
@@ -37,6 +37,14 @@ func followUpdatesAndUploadToRemote(option *RemoteSyncOptions, filerSource *sour
 
 	var lastLogTsNs = time.Now().UnixNano()
 	processEventFnWithOffset := pb.AddOffsetFunc(func(resp *filer_pb.SubscribeMetadataResponse) error {
+		if option.storageClass != nil {
+			storageClass := *option.storageClass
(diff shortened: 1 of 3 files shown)


weed/command/filer_remote_sync_dir.go
```diff
@@ -37,6 +37,14 @@ func followUpdatesAndUploadToRemote(option *RemoteSyncOptions, filerSource *sour
 
 	var lastLogTsNs = time.Now().UnixNano()
 	processEventFnWithOffset := pb.AddOffsetFunc(func(resp *filer_pb.SubscribeMetadataResponse) error {
+		if option.storageClass != nil {
```
chrislusf: Nil check is not necessary.

weed/command/filer_remote_sync_dir.go
```diff
@@ -37,6 +37,14 @@ func followUpdatesAndUploadToRemote(option *RemoteSyncOptions, filerSource *sour
...
+			storageClass := *option.storageClass
+			if storageClass == "None" {
```
chrislusf: How about just use empty string to delete, but default to current default value?

PR #3659, opened 2022-09-14
docs: `indicates` readability comment fix

ryanrussell:
Signed-off-by: Ryan Russell <git@ryanrussell.org>

```bash
grep -r indictes
wdclient/resource_pool/managed_handle.go:	// This indictes a user is done with the handle and releases the handle
wdclient/net2/managed_connection.go:	// This indictes a user is done with the connection and releases the
```

`indictes` -> `indicates`

# Notes for merger
Feel free to let me know if you prefer these small, specific PR's, or would rather me lump comment readability things into a more signficant PR.

Best,
Ryan


chrislusf:
Smaller PRs is better.

PR #3680, opened 2022-09-15
fix backwards compatibility after

kmlebedev:
https://github.com/seaweedfs/seaweedfs/commit/b9112747b57a45d5c1ef476897ec5940c76fc7e8

Screenshot 2022-09-15 at 10 19 33

After changing the id, the master and the server of different versions became incompatible

New Methods added to the end without changing the order and id fields

# How is the PR tested?

# Checks


kmlebedev:
conflicts fixed in https://github.com/seaweedfs/seaweedfs/pull/3681/files

PR #3681, opened 2022-09-15
avoid grpc protocol service critical change

kmlebedev:

fix conflicts 
https://github.com/seaweedfs/seaweedfs/pull/3680

# How is the PR tested?

# Checks


chrislusf:
this is not needed now?

kmlebedev:
> this is not needed now?

@chrislusf Need now. Please read this https://developers.google.com/protocol-buffers/docs/proto3#updating

> Don't change the field numbers for any existing fields.

chrislusf:
The diff is only moving the proto definition to the end of the file.

kmlebedev:
> The diff is only moving the proto definition to the end of the file.

Yes.

chrislusf:
The ordering in the proto file should not matter.

Issue #3660, opened 2022-09-14
[Bug]S3 API Multipart Uploads with 0 Bytes


jerebear12:
**Describe the bug**
Multipart Uploads using the S3 API fail at the completion step. I am using the JavaScript SDK V3 with Typescript. Parts are visible in the Filer UI as they are sent but as soon as the CompleteMultipartUpload command is sent, the parts in the .uploads folder disappear(I believe this is expected behavior) and the file looks like it is added to the bucket folder, but it has a value of 0 bytes. Files shown with byte values other than 0 were uploaded without using the Multipart Upload command.
![filer-ui](https://user-images.githubusercontent.com/72420925/189993475-3daa9f25-1ee5-4589-a4d8-83a7369567c0.JPG)

**System Setup**
Set up from Getting Started in the Wiki. 
Version 3.23 running on Ubuntu VM
1 Master 1 Filer(with S3 gateway) 1 Volume Server 8 Volumes

**Expected behavior**
The file parts should disappear from the temporary .uploads folder, and the file should be created with all data in the bucket folder. Length in bytes should accurately reflect file size, and the file should open and display its contents as expected.

**Steps to reproduce**
I made a small repo that has all the files needed to reproduce the error. You will need a an instance of seaweed running with the S3 gateway enabled as well.

https://github.com/jerebear12/s3-multipart-error

betwowt:
Same problem
---
use s3 browser program
upload 1G size file 
use fuse mount watch 
<img width="617" alt="image" src="https://user-images.githubusercontent.com/16661764/190292188-b2b924ad-9432-4162-8b1d-fddd1c2217aa.png">

chrislusf:
@jerebear12 can you use some docker to reproduce? I have no knowledge about nodejs, and here is the output when following the readme.

```
dev/s3-multipart-error () - (main) > npm install
npm WARN old lockfile
npm WARN old lockfile The package-lock.json file was created with an old version of npm,
npm WARN old lockfile so supplemental metadata must be fetched from the registry.
npm WARN old lockfile
npm WARN old lockfile This is a one-time fix-up, please be patient...
npm WARN old lockfile
npm notice
npm notice New minor version of npm available! 8.1.2 -> 8.19.2
npm notice Changelog: https://github.com/npm/cli/releases/tag/v8.19.2
npm notice Run npm install -g npm@8.19.2 to update!
npm notice
npm ERR! code ENOENT
npm ERR! syscall open
[...]

PR #3682, opened 2022-09-15
IsTruncated returned by the ListMultipartUploads method is set to false by default

famosss:
Using weedfs as the s3 storage of the docker registry will report an error, and the registry will call IsTruncated by default, which will report a nil exception

`registry/storage/driver/s3-aws/s3.go:648`

image

Added default return


chrislusf:
Is this related to https://github.com/seaweedfs/seaweedfs/issues/3660 ?

famosss:
> Is this related to #3660 ?
no

PR #3702, opened 2022-09-16
volume: add "readBufSize" option to customize read optimization

famosss:
After the read logic is optimized, many locks need to be acquired each time a file is read, especially when reading a large file, it takes a long time to acquire the lock, which affects the query performance.

Add `readBufSize` option,larger values can optimize query performance but will increase some memory usage.

For example, I set the size to `maxMB`, so that each query only needs to acquire a lock once

Add `readBufSize` option


weed/command/server.go
```diff
@@ -132,6 +132,7 @@ func init() {
 	serverOptions.v.idxFolder = cmdServer.Flag.String("volume.dir.idx", "", "directory to store .idx files")
 	serverOptions.v.inflightUploadDataTimeout = cmdServer.Flag.Duration("volume.inflightUploadDataTimeout", 60*time.Second, "inflight upload data wait timeout of volume servers")
 	serverOptions.v.hasSlowRead = cmdServer.Flag.Bool("volume.hasSlowRead", false, "if true, this prevents slow reads from blocking other requests, but large file read P99 latency will increase.")
+	serverOptions.v.readBufSize = cmdServer.Flag.Int("volume.readBufSize", 1024 * 1024, "larger values can optimize query performance but will increase some memory usage,Use with hasSlowRead normally")
```
chrislusf: change to readBufferSize

add `<experimental>`

famosss: We are testing that `clickhouse` is stored in `weedfs`. In the case of a large number of `put` and `delete`, this parameter is effective。

ddorian: @famosss you mean clickhouse s3-tables in seaweedfs?

famosss: > @famosss you mean clickhouse s3-tables in seaweedfs?

We are testing clickhouse `mergetree` engine, `s3_only` and `hot-warm(s3)` storage_policy.

chrislusf: this is worthy of a wiki page on how to adjust these parameters after this PR.

PR #3708, opened 2022-09-16
fix set master options for single server

kmlebedev:

master options move to `-master` prefix

# How is the PR tested?

# Checks


chrislusf:
need to resolve the conflicts.

PR #3710, opened 2022-09-18
volume server:set the default value of "hasSlowRead" to true

famosss:

experimental

I think it is more appropriate to make the read and write less mutual influence by default. The new parameter `hasSlowRead` makes the configuration more complicated. The `readBufferSizeMB` should be able to alleviate the P99 time-consuming effect caused by the reads lock scope.

In my usage scenario, the download from the network disk will limit the speed. Turning on a VPN is equivalent to speed limiting. Many reasons,many unknown factors
. In fact, many times, I don't know if it has `slowRead`. Just think the network is good, no `slowRead`

set the default value of "hasSlowRead" to true


chrislusf:
Most use cases are within a data center and slow network is rare.

PR #3731, opened 2022-09-20
Introduce `Account` and `BucketMetadata`

shichanglin5:

Introduce `Account` to ensure that the account structure is consistent with Amazon S3, so as to ensure that multiple IAM-Users under one user account execute results in S3 ACL as expected

In Amazon S3, each user corresponds to an Account, and each Account can be configured with multiple IAM-Users. In S3 Access Control Policy(ACP), Account is used as the Owner of bucket and object, not IAM -User (using IAM-User as Owner will cause different IAM-Users of the same Account to have problems with ACL authentication)

Account is used as the owner of buckets and objects, not IAM-User, so that the ACL verify results of multiple IAM-Users under one account are consistent

# How is the PR tested?

# Checks


weed/s3api/s3api_server.go
```diff
@@ -76,7 +81,7 @@ func NewS3ApiServer(router *mux.Router, option *S3ApiServerOption) (s3ApiServer
 
 	s3ApiServer.registerRouter(router)
 
-	go s3ApiServer.subscribeMetaEvents("s3", filer.DirectoryEtcRoot, time.Now().UnixNano())
+	go s3ApiServer.subscribeMetaEvents("s3", time.Now().UnixNano(), filer.DirectoryEtcRoot, option.BucketsPath)
```
chrislusf: This will listen to all the changes under buckets folder, which will be unnecessarily too much.

shichanglin5: Yes, you are right, I didn't notice this, I meant to only subscribe to the events of the bucket, not the events of all the files in the bucket

shichanglin5: Do you know of a proper way to do this ?

shichanglin5: If the filer's Notification Queue write interface is open to the s3 Api Server (through grpc), then these s3-related path subscriptions are very flexible. Now I need to subscribe to bucket-related operations (such as create, delete, update acl, etc.), or Subscriptions of other operations (such as bucket policy) do not need to modify the code of the filer.

How about this ? @chrislusf

chrislusf: added https://github.com/seaweedfs/seaweedfs/commit/31922b2bf232af8de55adde44138437aeefea1c8 to watch specific directories

weed/server/filer_grpc_server_sub_meta.go
```diff
@@ -226,6 +226,9 @@ func (fs *FilerServer) eachEventNotificationFn(req *filer_pb.SubscribeMetadataRe
 		} else if matchByDirectory(dirPath, req.Directories) {
 			// good
 		} else {
+			if req.PathPrefix == "" {
```
chrislusf: separate PR if this is needed at all.

shichanglin5: Got it

PR #3733, opened 2022-09-21
skip pathPrefix prefix processing if `pathPrefix` is empty

shichanglin5:

Not specifying pathPrefix (empty string) causes subscription to all events

skip pathPrefix prefix processing if `pathPrefix` is empty

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/filer_grpc_server_sub_meta.go
+++ b/weed/server/filer_grpc_server_sub_meta.go
@@ -226,6 +226,9 @@ func (fs *FilerServer) eachEventNotificationFn(req *filer_pb.SubscribeMetadataRe
 		} else if matchByDirectory(dirPath, req.Directories) {
 			// good
 		} else {
+			if req.PathPrefix == "" {
+				return nil
+			}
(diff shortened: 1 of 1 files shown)


weed/server/filer_grpc_server_sub_meta.go
```diff
@@ -226,6 +226,9 @@ func (fs *FilerServer) eachEventNotificationFn(req *filer_pb.SubscribeMetadataRe
 		} else if matchByDirectory(dirPath, req.Directories) {
 			// good
 		} else {
+			if req.PathPrefix == "" {
```
chrislusf, line 229: this reverts the logic when pathprefix is empty

shichanglin5: Yes, it is; but in seaweed, all events will have a directory. The first character is `/`, so you can subscribe to all events by specifying pathPrefix as `/` instead of `""`; I think changing to pointer type is a better way, but there will be more code changes

PR #3748, opened 2022-09-27
fix: compact_map get error mismatching cokie

binbinshi:

when use snowflak gen id ，The following error occurs :
write to local disk: mismatching cookie xxxxxxx
the file write failue

fix the bug of compact_map get error ; when the slot end small than the key return  nil false

# How is the PR tested?

add compact get test 
test idx  
[1655451851227025229-502m-single-1_388.idx.zip](https://github.com/seaweedfs/seaweedfs/files/9654785/1655451851227025229-502m-single-1_388.idx.zip)

# Checks

### DIFF
--- a/weed/storage/needle_map/compact_map.go
+++ b/weed/storage/needle_map/compact_map.go
@@ -145,7 +145,7 @@ func (cs *CompactSection) deleteOverflowEntry(key SectionalNeedleId) {
 // return old entry size
 func (cs *CompactSection) Delete(key NeedleId) Size {
 	cs.Lock()
-	defer cs.RUnlock()
+	defer cs.Unlock()
 	ret := Size(0)
 	if key > cs.end {
 		return ret
--- a/weed/storage/needle_map/compact_map.go
+++ b/weed/storage/needle_map/compact_map.go
@@ -144,9 +144,13 @@ func (cs *CompactSection) deleteOverflowEntry(key SectionalNeedleId) {
 
 // return old entry size
 func (cs *CompactSection) Delete(key NeedleId) Size {
-	skey := SectionalNeedleId(key - cs.start)
 	cs.Lock()
+	defer cs.RUnlock()
 	ret := Size(0)
+	if key > cs.end {
+		return ret
+	}
+	skey := SectionalNeedleId(key - cs.start)
 	if i := cs.binarySearchValues(skey); i >= 0 {
 		if cs.values[i].Size > 0 && cs.values[i].Size.IsValid() {
 			ret = cs.values[i].Size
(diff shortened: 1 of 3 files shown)


chrislusf:
tests failed.

weed/storage/needle_map/compact_map_perf_test.go
```diff
@@ -90,3 +90,34 @@ func PrintMemUsage(totalRowCount uint64) {
 func bToMb(b uint64) uint64 {
 	return b / 1024 / 1024
 }
+
+func TestCompactSection_Get(t *testing.T) {
```
chrislusf: move to compact_map_test.go

weed/storage/needle_map/compact_map_perf_test.go
```diff
@@ -90,3 +90,34 @@ func PrintMemUsage(totalRowCount uint64) {
...
+	var maps []*CompactMap
+	totalRowCount := uint64(0)
+	indexFile, ie := os.OpenFile("../../1655451851227025229-502m-single-1_388.idx",
```
chrislusf: where is this file? Seems this test does not work at all.

binbinshi: [1655451851227025229-502m-single-1_388.idx.zip](https://github.com/seaweedfs/seaweedfs/files/9654785/1655451851227025229-502m-single-1_388.idx.zip)

binbinshi: the file  is 38MB, so i didn't commit it

weed/storage/needle_map/compact_map.go
```diff
@@ -144,9 +144,13 @@ func (cs *CompactSection) deleteOverflowEntry(key SectionalNeedleId) {
 
 // return old entry size
 func (cs *CompactSection) Delete(key NeedleId) Size {
-	skey := SectionalNeedleId(key - cs.start)
 	cs.Lock()
+	defer cs.RUnlock()
```
chrislusf: this is wrong. The test did not cover this function?

weed/storage/needle_map/compact_map_perf_test.go
chrislusf: try to create a test case directly. reproduce first, then fix.

binbinshi: reproduce  pipline 
1. remove the compact_map.go Get function  line 169-171
2. run the test TestCompactSection_Get func ,  may test failed of  get error 
3. recovery the compact_map.go Get function  line 169-171  , run the TestCompactSection_Get test will pass

PR #3753, opened 2022-09-28
s3: add account

shichanglin5:

In order to introduce s3 Acl related functions, it is necessary to add an Account structure to manage IAM-User, and use AccountId as the owner of the bucket/object instead of IAM-User (Identity) as the Owner, to ensure that IAM-Users under the same account will not Limited access due to Owner inconsistency

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3api_account.go
+++ b/weed/s3api/s3api_account.go
@@ -1,0 +1,69 @@
+package s3api
+
+import (
+	"sync"
+)
+
+//Predefined Accounts
+var (
+	// AccountAdmin is used as the default account for IAM-Credentials access without Account configured
+	AccountAdmin = Account{
+		Name:         "admin",
+		EmailAddress: "admin@example.com",
+		Id:           "admin",
+	}
+
+	// AccountAnonymous is used to represent the account for anonymous access
+	AccountAnonymous = Account{
+		Name:         "anonymous",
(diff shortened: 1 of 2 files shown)


weed/s3api/s3api_account.go
```diff
@@ -0,0 +1,69 @@
...
+
+	//Id is used to identify an Account when granting cross-account access(ACLs) to buckets and objects
+	Id string
+}
+
+type AccountManager struct {
+	sync.Mutex
+	s3a *S3ApiServer
+
+	IdNameMapping  map[string]*string
```
chrislusf: Why the map value type needs to be a pointer?

shichanglin5: Is there any problem with this, I'm used to using pointers to reduce value copying

weed/s3api/s3api_account.go
```diff
@@ -0,0 +1,69 @@
+package s3api
```
shichanglin5, line 1: Is it recommended to create a separate directory `s3_account` for `s3_account.go`, because if the shell command refers to `AccountAnonymous` or `AccountAdmin`, there may be a circular reference (s3api->command->s3api, now s3api has a reference to the command directory ); and if the management function of Account is added, it can also be placed in this directory

PR #3757, opened 2022-09-29
s3: add acl headers

shichanglin5:

[AWS S3 Overview](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html)
s3: add acl headers

When creating a bucket or uploading/updating an object, ACL (access control list) can be specified. If CannedAcl is specified, the request header `X-Amz-Acl` will contain the corresponding predefined values (such as `PublicRead`, `PublicReadWrite`, etc. etc.), if not, then according to the actual custom Grant information, request headers like `X-Amz-Grant-xxxx` will be assigned;

Supplement: If both CannedAcl and Custom Acl are specified, then only the Custom ACL will take effect, because the priority is higher

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3_constants/header.go
+++ b/weed/s3api/s3_constants/header.go
@@ -37,12 +37,21 @@ const (
 	AmzObjectTaggingDirective = "X-Amz-Tagging-Directive"
 	AmzTagCount               = "x-amz-tagging-count"
 
+	// S3 ACL headers
+	AmzCannedAcl      = "X-Amz-Acl"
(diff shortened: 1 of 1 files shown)


chrislusf:
not a good way to split PRs.

Each individual PR should be logically stand alone. Just split by file.

shichanglin5:
> not a good way to split PRs.
> 
> Each individual PR should be logically stand alone. Just split by file.



> not a good way to split PRs.
> 
> Each individual PR should be logically stand alone. Just split by file.

PR #3758, opened 2022-09-29
s3: add ACP(access control policy) related constants

shichanglin5:

[AWS S3 Overview](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html)

define ACP(access control policy) related constants

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3_constants/extend_key.go
+++ b/weed/s3api/s3_constants/extend_key.go
@@ -1,0 +1,6 @@
+package s3_constants
+
+const (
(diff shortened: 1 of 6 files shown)


weed/s3api/s3_constants/extend_key.go
```diff
@@ -0,0 +1,6 @@
+package s3_constants
+
+const (
+	ExtAcpKey       = "Weed-X-Amz-Acp"
```
chrislusf: use "Seaweed-"

Issue #3760, opened 2022-09-29
It seems that the actual Java client (3.13) lacks calculating of `Content-MD5` at the client side when uploading files.


kvtb:
**Describe the bug**

It seems that the actual Java client (3.13) lacks calculating of `Content-MD5` at the client side when uploading files.

chrislusf:
What is your use case?

kvtb:
> What is your use case?

To ensure there is no bitrot during network transmission and retry on checksum mismatch

PR #3761, opened 2022-09-29
client-java: provide Content-MD5 when uploading chunks

kvtb:

Bitrot during network transfer

Provide `Content-MD5` when uploading chunks so volume server could reject corrupt data
The volume server already support that, and other clients do it.

# How is the PR tested?

The change is trivial.

# Checks

Fixes: https://github.com/seaweedfs/seaweedfs/issues/3760

### DIFF
--- a/other/java/client/src/main/java/seaweedfs/client/SeaweedWrite.java
+++ b/other/java/client/src/main/java/seaweedfs/client/SeaweedWrite.java
@@ -14,7 +14,9 @@ import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.security.SecureRandom;
+import java.security.MessageDigest;
 import java.util.List;
+import java.util.Base64;
 
 public class SeaweedWrite {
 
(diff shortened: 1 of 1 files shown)


kvtb:
Also, HTTP client should check for 200 status.
Currently, if the server returns non-200, java client fires `NullPointerException`
It is acceptable but not nice

other/java/client/src/main/java/seaweedfs/client/SeaweedWrite.java
```diff
@@ -123,13 +125,20 @@ private static String multipartUpload(String targetUrl,
                                           final byte[] bytes,
                                           final long bytesOffset, final long bytesLength,
                                           byte[] cipherKey) throws IOException {
+        MessageDigest md = null;
+        try {
+            md = MessageDigest.getInstance("MD5");
+        } catch (java.security.NoSuchAlgorithmException e) {
```
chrislusf, line 131: should fail this or check whether md is null when using it later.

vobloeb: Letting this fail would change the interface of the API function (`throws IOException` -> `throws IOException, NoSuchAlgorithmException`)

`md` will never be `null`, "MD5" is supported on all JVM

PR #3765, opened 2022-09-30
add ownership rest apis

shichanglin5:

[Controlling object ownership](https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html)

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -16,6 +16,8 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
 )
 
+var IdentityAnonymous *Identity
+
 type Action string
 
 type Iam interface {
@@ -32,8 +34,13 @@ type IdentityAccessManagement struct {
 
 type Identity struct {
 	Name        string
+	AccountId   string
 	Credentials []*Credential
 	Actions     []Action
+}
+
+func (i *Identity) isAnonymous() bool {
+	return i.Name == AccountAnonymous.Name
 }
 
 type Credential struct {
@@ -125,9 +132,23 @@ func (iam *IdentityAccessManagement) loadS3ApiConfiguration(config *iam_pb.S3Api
 	for _, ident := range config.Identities {
 		t := &Identity{
 			Name:        ident.Name,
+			AccountId:   AccountAdmin.Id,
 			Credentials: nil,
 			Actions:     nil,
 		}
+
+		if ident.Name == AccountAnonymous.Name {
+			if ident.AccountId != "" && ident.AccountId != AccountAnonymous.Id {
+				glog.Warningf("anonymous identity is associated with a non-anonymous account ID, the association is invalid")
+			}
+			t.AccountId = AccountAnonymous.Id
+			IdentityAnonymous = t
+		} else {
+			if len(ident.AccountId) > 0 {
+				t.AccountId = ident.AccountId
+			}
+		}
+
 		for _, action := range ident.Actions {
 			t.Actions = append(t.Actions, Action(action))
 		}
@@ -138,6 +159,13 @@ func (iam *IdentityAccessManagement) loadS3ApiConfiguration(config *iam_pb.S3Api
 			})
 		}
 		identities = append(identities, t)
+	}
+
+	if IdentityAnonymous == nil {
+		IdentityAnonymous = &Identity{
+			Name:      AccountAnonymous.Name,
+			AccountId: AccountAnonymous.Id,
+		}
 	}
 	iam.m.Lock()
 	// atomically switch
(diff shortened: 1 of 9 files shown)


weed/s3api/filer_util.go
```diff
@@ -91,6 +91,22 @@ func (s3a *S3ApiServer) getEntry(parentDirectoryPath, entryName string) (entry *
 	return filer_pb.GetEntry(s3a, fullPath)
 }
 
+func (s3a *S3ApiServer) updateEntry(parentDirectoryPath string, newEntry *filer_pb.Entry) (entry *filer_pb.Entry, err error) {
```
chrislusf: what is the first return value?

weed/s3api/s3api_acp.go
```diff
@@ -0,0 +1,27 @@
+package s3api
+
+import (
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3_constants"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
+	"net/http"
+)
+
+func ensureAccountId(r *http.Request) string {
```
chrislusf: rename to getAccountId()?

weed/s3api/s3api_acp.go
```diff
@@ -0,0 +1,27 @@
...
+	id := r.Header.Get(s3_constants.AmzAccountId)
+	if len(id) == 0 {
+		return AccountAnonymous.Id
+	} else {
+		return id
+	}
+}
+
+func (s3a *S3ApiServer) OwnershipAccess(r *http.Request, bucket string) s3err.ErrorCode {
```
chrislusf: rename to checkAccessByOwnership()?

weed/s3api/s3api_bucket_handlers.go
```diff
@@ -343,3 +345,161 @@ func (s3a *S3ApiServer) GetBucketLocationHandler(w http.ResponseWriter, r *http.
...
+		return
+	}
+
+	if r.Body == nil || r.Body == http.NoBody {
+		s3err.WriteErrorResponse(w, r, s3err.ErrInvalidRequest)
+		return
+	}
+
+	var v s3.OwnershipControls
+	defer func(Body io.ReadCloser) {
```
chrislusf: defer util.CloseRequest(r)

weed/s3api/s3api_bucket_handlers.go
```diff
@@ -343,3 +345,161 @@ func (s3a *S3ApiServer) GetBucketLocationHandler(w http.ResponseWriter, r *http.
...
+	if err != nil {
+		if err == filer_pb.ErrNotFound {
+			s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchBucket)
+			return
+		}
+		s3err.WriteErrorResponse(w, r, s3err.ErrInternalError)
+		return
+	}
+
+	oldOwnership, ok := bucketEntry.Extended[s3_constants.ExtOwnershipKey]
```
chrislusf, line 401: bucketEntry.Extended can be nil.

weed/s3api/auth_credentials.go
```diff
@@ -259,6 +286,7 @@ func (iam *IdentityAccessManagement) authRequest(r *http.Request, action Action)
 		return identity, s3err.ErrAccessDenied
 	}
 
+	r.Header.Set(s3_constants.AmzAccountId, identity.AccountId)
```
chrislusf: is this necessary for anonymous mode?

weed/s3api/s3api_bucket_handlers.go
shichanglin5: A nil map can be read directly, but writing will cause NPE, so it doesn't matter here

weed/s3api/auth_credentials.go
shichanglin5: Do you mean anonymous access? There is no need to set it for anonymous access. When the request header reads `X-Amz-Account` is empty, the id of `AccountAnonymous` will be used as accountId
```go
func ensureAccountId(r *http.Request) string {
	id := r.Header.Get(s3_constants.AmzAccountId)
	if len(id) == 0 {
		return AccountAnonymous.Id
	} else {
		return id
	}
}
```

weed/s3api/s3api_bucket_handlers.go
famosss: > bucketEntry.Extended can be nil.

Tested, no exception will be thrown here.

weed/s3api/auth_credentials.go
famosss: > is this necessary for anonymous mode?

There are still several `pr` operation objects that need to use this header. The anonymous user is the anonymous accountId, which should not matter.

chrislusf: how about only add this header when the user is not anonymous?

weed/s3api/auth_credentials.go
```diff
@@ -259,6 +287,9 @@ func (iam *IdentityAccessManagement) authRequest(r *http.Request, action Action)
 		return identity, s3err.ErrAccessDenied
 	}
 
+	if !identity.isAnonymous() {
+		r.Header.Set(s3_constants.AmzAccountId, identity.AccountId)
+	}
 	return identity, s3err.ErrNone
```
shichanglin5, line 293: > how about only add this header when the user is not anonymous?

is this ok @chrislusf

PR #3814, opened 2022-10-10
revert disabling FSync for non Mac

kmlebedev:
https://github.com/seaweedfs/seaweedfs/commit/52580743b9de1931b149e1df40e27048d99ca3bf

When the memory limit is reached in the virtualization environment (k8s), during the execution of vaccum, an OOM killer may come and unexpectedly kill the container with the volume and break the volum files

To minimize  OOM killer effect, do a forced fsync on every vaccum step

https://github.com/golang/go/issues/26650

# How is the PR tested?

# Checks


kmlebedev:
next step https://github.com/seaweedfs/seaweedfs/pull/3832

PR #3824, opened 2022-10-11
split `ExtAcpKey` to `ExtAmzOwnerKey` and `ExtAmzAclKey` to avoid unn…

shichanglin5:

Avoid unnecessary `json.Unmarshal()` calls, in some cases (such as PutObject) return `errCode` directly when `Owner` is inconsistent without verifying grants

split `ExtAcpKey` to `ExtAmzOwnerKey` and `ExtAmzAclKey`，when verifying Owner, directly convert bytes to string without calling `json.Unmarshal`

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/bucket_metadata.go
+++ b/weed/s3api/bucket_metadata.go
@@ -1,16 +1,12 @@
 package s3api
 
 import (
-	"bytes"
 	"encoding/json"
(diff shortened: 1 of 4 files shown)


shichanglin5:
@chrislusf Take the time to look? Or point it out if there is a problem

PR #3832, opened 2022-10-12
vacuum metrics and force sync dst files

kmlebedev:

I would like to see the regression of the vacuum if it is when updating seaweedfs
https://github.com/seaweedfs/seaweedfs/issues/3837

add vacuum error count and  seconds metrics
Screenshot 2022-10-12 at 17 33 25

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/volume_vacuum.go
+++ b/weed/storage/volume_vacuum.go
@@ -120,9 +120,6 @@ func (v *Volume) CommitCompact() error {
 		v.nm = nil
 	}
 	if v.DataBackend != nil {
-		if err := v.DataBackend.Sync(); err != nil {
-			glog.V(0).Infof("failed to sync volume %d", v.Id)
-		}
 		if err := v.DataBackend.Close(); err != nil {
 			glog.V(0).Infof("failed to close volume %d", v.Id)
 		}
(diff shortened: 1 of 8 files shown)


weed/storage/volume_vacuum.go
```diff
@@ -265,13 +268,20 @@ func (v *Volume) makeupDiff(newDatFileName, newIdxFileName, oldDatFileName, oldI
 		return fmt.Errorf("open dat file %s failed: %v", newDatFileName, err)
 	}
 	dstDatBackend := backend.NewDiskFile(dst)
-	defer dstDatBackend.Close()
+	defer func() {
+		dstDatBackend.Sync()
```
chrislusf: consider moving the sync() into close()

guol-fnst:
I also found that current vacuum speed is much slower than before when I running test.

chrislusf:
> I also found that current vacuum speed is much slower than before when I running test.

@guol-fnst You are comparing which two versions?

kmlebedev:
> I also found that current vacuum speed is much slower than before when I running test.

This is to be expected due to direct-to-disk writing being enabled.
If speed is more important to you than data integrity, then you can move this behind the vacuum_fsync option

guol-fnst:
> > I also found that current vacuum speed is much slower than before when I running test.
> 
> @guol-fnst You are comparing which two versions? What is your OS?

Actually I found this problem when I tested my laster PR (https://github.com/seaweedfs/seaweedfs/pull/3830)
Besides I found the same problem when I tested my another PR(https://github.com/seaweedfs/seaweedfs/issues/3443)
And a related PR was reverted.

guol-fnst:
> 

Yes, I totally undanstand your concerns and know why it gets slower.
I just share this information here.
And thanks for your explanation.

guol-fnst:
> > I also found that current vacuum speed is much slower than before when I running test.
> 
> @guol-fnst You are comparing which two versions? What is your OS?

Almost forgot. Centos-7.2

And I think it's necessary to add a CI/CD for performance  test for each PR expecially when you are sensitive to performance

PR #3833, opened 2022-10-12
filer: get directory metadata

zemul:
use GET method get directory metadata 

other systems to access

# How is the PR tested?

# Checks


chrislusf:
Please add wiki for this.

PR #3843, opened 2022-10-13
implement `GetObjectAclHandler`

shichanglin5:

implement `GetObjectAclHandler`  to allow the user to query the object access control policy

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3api_acp.go
+++ b/weed/s3api/s3api_acp.go
@@ -1,9 +1,14 @@
 package s3api
 
 import (
+	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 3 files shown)


chrislusf:
this and the following changes do not have any test.

is there any integration test for these?

shichanglin5:
> this and the following changes do not have any test.
> 
> is there any integration test for these?

I just tested it with the sdk before splitting and submitting the pr

shichanglin5:
> Please add integration tests. One possible way is using docker compose.

用 Ceph 的 s3_tests 怎么样，上面有 acl的用例，我用这个测下

PR #3861, opened 2022-10-14
[volume.fsck] hotfix apply purging and add option verifyNeedle #3860

kmlebedev:

fix error: findExtraChunksInVolumeServers: purging volume 2823: delete fileId monitoring-thanos:2823,2a7a380095160caa00000000: volume 0 not found on fast-volume-1:8080

option verifyNeedle: It is necessary to be able to verify the integrity of the dat files so that all blobs are recoverable.
To fix a blob in time from a replica or from a backup

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_volume_fsck.go
+++ b/weed/shell/command_volume_fsck.go
@@ -44,6 +44,7 @@ type commandVolumeFsck struct {
 	verbose                  *bool
 	forcePurging             *bool
 	findMissingChunksInFiler *bool
+	verifyNeedle             *bool
 }
 
 func (c *commandVolumeFsck) Name() string {
@@ -82,6 +83,7 @@ func (c *commandVolumeFsck) Do(args []string, commandEnv *CommandEnv, writer io.
 	purgeAbsent := fsckCommand.Bool("reallyDeleteFilerEntries", false, "<expert only!> delete missing file entries from filer if the corresponding volume is missing for any reason, please ensure all still existing/expected volumes are connected! used together with findMissingChunksInFiler")
 	tempPath := fsckCommand.String("tempPath", path.Join(os.TempDir()), "path for temporary idx files")
 	cutoffTimeAgo := fsckCommand.Duration("cutoffTimeAgo", 5*time.Minute, "only include entries  on volume servers before this cutoff time to check orphan chunks")
+	c.verifyNeedle = fsckCommand.Bool("verifyNeedles", false, "try get head needle blob from volume server")
 
 	if err = fsckCommand.Parse(args); err != nil {
 		return nil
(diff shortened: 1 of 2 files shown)


weed/shell/command_volume_fsck.go
```diff
@@ -541,19 +543,38 @@ func (c *commandVolumeFsck) oneVolumeFileIdsSubtractFilerFileIds(dataNodeId stri
...
+			if v, ok := db.Get(nId); ok && v.Size.IsValid() {
+				newSize := types.Size(16)
+				if v.Size > newSize {
+					v.Size = newSize
+				}
+				if _, err := readSourceNeedleBlob(c.env.option.GrpcDialOption, voluemAddr, volumeId, *v); err != nil {
+					fmt.Fprintf(c.writer, "failed to read file %s NeedleBlob %+v: %+v", itemPath, nId, err)
+					if *c.forcePurging {
+						return
+					}
```
chrislusf, line 564: not clear what is the logic here.

weed/shell/command_volume_fsck.go
```diff
@@ -541,19 +543,38 @@ func (c *commandVolumeFsck) oneVolumeFileIdsSubtractFilerFileIds(dataNodeId stri
 		return
 	}
 
+	voluemAddr := pb.NewServerAddressWithGrpcPort(dataNodeId, 0)
 	if err = c.readFilerFileIdFile(volumeId, func(nId types.NeedleId, itemPath util.FullPath) {
+		inUseCount++
+		if *c.verifyNeedle {
+			if v, ok := db.Get(nId); ok && v.Size.IsValid() {
+				newSize := types.Size(16)
```
chrislusf: need a constant here, and why it is 16.

kmlebedev: > need a constant here, and why it is 16.

@chrislusf I honestly don't know why 16 is random. It seems that here you need to leave the size original and add the `isProbe` parameter to the grpc call

weed/shell/command_volume_fsck.go
kmlebedev: > not clear what is the logic here.

@chrislusf Logic in removing chunks that if their blobs are not available. Then you can run volume.check.disk and if the blob is in the replica, fix

chrislusf: seems the logic should be opposite.

kmlebedev: > seems the logic should be opposite.

@chrislusf It's a little unclear here. Since the next function is `db.Delete(nId)`
It removes the needle's metadata from the database, then deletes what is left.

That's why
```
if *c.forcePurging {
return
}
```
Ensures needle meta data of Needle in the database and is tagged as orphan

kmlebedev:
@chrislusf Do not release without merge this PR (purging broken)

PR #3894, opened 2022-10-24
fix: set user metadata key to lowercase

famosss:

image

After we migrated from `ceph s3` to `SeaweedFS`, we found that the metadata key has changed, so we changed the returned key according to the `aws` standard

replace return result

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -415,17 +415,13 @@ func (s3a *S3ApiServer) proxyToFiler(w http.ResponseWriter, r *http.Request, des
 }
 
 func setUserMetadataKeyToLowercase(resp *http.Response) {
-	newHeader := make(map[string][]string)
-
 	for key, value := range resp.Header {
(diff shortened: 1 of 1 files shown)


weed/s3api/s3api_object_handlers.go
```diff
@@ -408,10 +408,26 @@ func (s3a *S3ApiServer) proxyToFiler(w http.ResponseWriter, r *http.Request, des
...
 	}
 
+	setUserMetadataKeyToLowercase(resp)
+
 	responseStatusCode := responseFn(resp, w)
 	s3err.PostLog(r, responseStatusCode, s3err.ErrNone)
 }
 
+func setUserMetadataKeyToLowercase(resp *http.Response) {
+	newHeader := make(map[string][]string)
```
chrislusf: how about updating the original header?

famosss:
@chrislusf Is it necessary to add a configuration item to control whether to convert to lowercase, so that it will not affect the original user

chrislusf:
should be fine. most clients should be using case insensitive.

PR #3907, opened 2022-10-25
fix: isTruncated returns problematic

famosss:

image

When the subdirectory traversal ends and maxKeys is satisfied, the returned `isTruncated` is wrong, causing the client to not be able to list all the data

The easiest case to reproduce, when `maxkeys` is set to 1, when the first subdirectory traversal ends, the return is an error

The judgment of `isTruncated` should also be added to the `eof` side.


kmlebedev:
The most surprising thing is how the s3 tests worked ...

famosss:
> The most surprising thing is how the s3 tests worked ...

We are switching all `ceph s3` traffic to `weedfs`, which involves a lot of business, their feedback. ^_^

PR #3879, opened 2022-10-18
[volume.fsck] add param to verify read data hydrates the needle

kmlebedev:
# What problem are we solving?

1. param cutoff time ago  superfluous since now we get all ids from the filler
2. param verify allows detect and deleting file id with read data hydrates error(CRC error! Data On Disk Corrupted, index out of range)

# How are we solving the problem?

# How is the PR tested?

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #3935, opened 2022-10-31
[volume.fsck] avoid breaking due to needle reading error

kmlebedev:

split https://github.com/seaweedfs/seaweedfs/pull/3879

Most likely, binary search is not appropriate here.
Also, with an error in the comparison logic, needleMeta.AppendAtNs in previous versions of algae did not harm, which means there is no benefit from it

I can do more comparison tests later.

# How is the PR tested?

# Checks


weed/shell/command_volume_fsck.go
```diff
@@ -582,10 +563,25 @@ func (c *commandVolumeFsck) oneVolumeFileIdsSubtractFilerFileIds(dataNodeId stri
 	}
 
 	var orphanFileCount uint64
-	if err = volumeFileIdDb.AscendingVisit(func(n needle_map.NeedleValue) error {
+	cutoffFrom := c.collectCutoffFrom.Add(-*c.cutoffTimeAgo)
+	cutoffFromAtNs := uint64(cutoffFrom.UnixNano())
+	doCutoffOfLastNeedle := true
+	if err = volumeFileIdDb.DescendingVisit(func(n needle_map.NeedleValue) error {
```
chrislusf, line 569: this is descending based on needle id. But
1. The needle ids are not always monotonically increasing, e.g. for snow flake needle id generation.
2. The needle could get updated or deleted. The largest needle id is not necessarily the last appended entry.

kmlebedev: > 1. The needle ids are not always monotonically increasing, e.g. for snow flake needle id generation.
> 2. The needle could get updated or deleted. The largest needle id is not necessarily the last appended entry.

In this case, both options are bad, since it is necessary to make a grpc request for all needles

chrislusf: the binary search will do this efficiently, since the AppendAtNs is monotonically increasing.

kmlebedev: > the binary search will do this efficiently, since the AppendAtNs is monotonically increasing.

In this case, the Binary search algorithm is not suitable because algorithm for searching for an object by a given attribute in a set of objects ordered by its own attribute

You yourself wrote that objects are not sorted by AppendAtNs or even by needle id

chrislusf: > You yourself wrote that objects are not sorted by AppendAtNs or even by needle id

objects are not sorted by modified time or needle id. but they are sorted by AppendAtNs.
AppendAtNs is for appending time to the .dat file. Modified time can be set from the client side.

kmlebedev: @chrislusf I have two ideas:
1.  Cut-off time to send to a remote call to bulk remove of needles. And on the side by the will of the volume server to check already AppendAtNs.
2.  Binary search can be used after getting missing needles (transfer the filtering function from map to reduce)

PR #3937, opened 2022-11-02
[filer.backup] use `replication.source.filer`

zemul:
Can easily configure more excluded paths .
`filer.backup` use `replication.source.filer`.

# How is the PR tested?

# Checks
- [V ] I will add related wiki document changes and link to this PR after merging.


chrislusf:
1. how about making it a `filer.backup` command line option?
2. can we use "retention_duration" instead? "rolling_incremental_update" is hard to understand.
3. split this PR for separate changes.

PR #3948, opened 2022-11-04
fix(sec): upgrade org.apache.hadoop:hadoop-common to 3.3.3

chncaption:
### What happened？
There are 1 security vulnerabilities found in org.apache.hadoop:hadoop-common 2.10.1
- [CVE-2022-25168](https://www.oscs1024.com/hd/CVE-2022-25168)

### What did I do？
Upgrade org.apache.hadoop:hadoop-common from 2.10.1 to 3.3.3 for vulnerability fix

### What did you expect to happen？
Ideally, no insecure libs should be used.

### The specification of the pull request
[PR Specification](https://www.oscs1024.com/docs/pr-specification/) from OSCS

### DIFF
--- a/other/java/examples/pom.xml
+++ b/other/java/examples/pom.xml
@@ -23,7 +23,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>2.10.1</version>
+            <version>3.3.3</version>
             <scope>compile</scope>
         </dependency>
     </dependencies>


other/java/examples/pom.xml
```diff
@@ -23,7 +23,7 @@
         <dependency>
             <groupId>org.apache.hadoop</groupId>
             <artifactId>hadoop-common</artifactId>
-            <version>2.10.1</version>
+            <version>3.3.3</version>
```
chrislusf, line 26: this is to work with hadoop2. I am not sure whether this can still work moving from 2.x to 3.x

PR #3963, opened 2022-11-10
feat:add a uuid after uploadid

famosss:

When the same file is uploaded multiple times, and the multipart upload is concurrent, it is easy to cause problems with the final merged file.
When used as the underlying s3 storage of a docker registry, since blob files can be reused, the first concurrent build will result in concurrent multipart uploads

image

add a uuid after uploadid


kmlebedev:
Does it agree? https://github.com/seaweedfs/seaweedfs/pull/3968

PR #3983, opened 2022-11-14
enable faster vacuuming with concurrency

theRealWardo:
Due to the single threaded nature of the vacuum implementation, it is possible for high write volume clusters with low TTLs to be unable to actually evict the cache contents at a rate the operator desires.

Utilize all existing volume vacuum logic but enable it to run on multiple volumes at the same time via a go routine worker pool. The pool size can be grown via a flag.

# How is the PR tested?
Set the flag to >1 and both logs and disk usage metrics will show the system is able to vacuum multiple volumes at a time such that the garbage collection and compaction can occur much quicker.

# Checks
- [ ] ~I have added unit tests if possible.~ - **This code relies on existing unit tests as it wraps an existing implementation.**

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -2,12 +2,13 @@ package command
 
 import (
 	"fmt"
-	hashicorpRaft "github.com/hashicorp/raft"
 	"net/http"
 	"os"
 	"path"
 	"strings"
 	"time"
+
+	hashicorpRaft "github.com/hashicorp/raft"
 
 	"golang.org/x/exp/slices"
 
@@ -45,6 +46,7 @@ type MasterOptions struct {
 	// pulseSeconds       *int
 	defaultReplication *string
 	garbageThreshold   *float64
+	vacuumConcurrency  *int
 	whiteList          *string
 	disableHttp        *bool
 	metricsAddress     *string
@@ -70,6 +72,7 @@ func init() {
 	// m.pulseSeconds = cmdMaster.Flag.Int("pulseSeconds", 5, "number of seconds between heartbeats")
 	m.defaultReplication = cmdMaster.Flag.String("defaultReplication", "", "Default replication type if not specified.")
 	m.garbageThreshold = cmdMaster.Flag.Float64("garbageThreshold", 0.3, "threshold to vacuum and reclaim spaces")
+	m.vacuumConcurrency = cmdMaster.Flag.Int("vacuumConcurrency", 1, "number of volumes to concurrently vacuum")
 	m.whiteList = cmdMaster.Flag.String("whiteList", "", "comma separated Ip addresses having write permission. No limit if empty.")
 	m.disableHttp = cmdMaster.Flag.Bool("disableHttp", false, "disable http requests, only gRPC operations are allowed.")
 	m.metricsAddress = cmdMaster.Flag.String("metrics.address", "", "Prometheus gateway address <host>:<port>")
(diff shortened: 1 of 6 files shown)


chrislusf:
How to prevent 2 or more compactions running on the same volume server?

theRealWardo:
Why is that an issue? I've seen that multiple compactions run on the same volume server without any issue when using SSDs. If the cluster is running on HDDs, I suspect users will want to leave the default safe concurrency limit of 1.

Ideally, this code could be improved to look at volume server topology and try to target 1 compaction per volume server which would be an improvement over the existing implementation of 1 compaction per cluster. However, in practice a single volume server seems to be capable of doing multiple compactions at the cost of increased CPU + I/O.

Am I missing something that makes running two `VacuumVolumeCompact` RPCs at the same time unsafe?

chrislusf:
This is much better than before.

But if more than one GC is happening on one server, the latency will increase. Need a way to prevent it automatically instead of let users figuring out via try-and-errors.

theRealWardo:
Yes I noticed the increase in latency at times during the vacuum, but I'd take the trade off of a bit more latency vs. full disks and erroring writes!  😄 

Do you have any recommendations for what to do to improve this change such that it could be accepted and useful to others as well?

chrislusf:
When scheduling vacuum jobs, how about putting vacuumLocationList into a hashmap, and if there are duplication, skip to the next volume id? 

A little bit more work but could be easily unit tested.

chrislusf:
also need to fix the build errors.

theRealWardo:
looks like I had a merge go wrong. sorry about that.

I think I hear what you are saying but when we vacuum, I assume we want it to run on all volumes. I think a real skip and having to run the vacuum again to get all the volumes vacuumed is not ideal. It sounds like there's actually 2 concurrency settings here that we're talking about tuning for higher performance vacuuming:

1- volume servers to vacuum concurrently
2- volumes per server to vacuum concurrently

As implemented, my `-master.vacuumConcurrency` flag conflates these. This is a mediocre design at best, but it does solve some uses such as the case of a single node server.  😉 

If we want to allow operators more control of concurrent vacuuming, I think the simplest option is to:
- Provide `-master.vacuumWorkers [number]` to tune the concurrency on the master. The logic can skip or requeue volume servers that reject the request to be vacuumed. This allows operators to tune 1 above.
- Volume servers can choose to allow concurrency via `-volume.concurrentCompacts [number]` which simply is an atomic counter on the `VolumeServer` struct that limits concurrency on that instance. This allows operators to enable concurrency per-server as desired, tuning 2 above.

To reduce the risk of running hot and still ensure we try to vacuum all volumes as quickly as possible, we can alternate volume servers as much as possible when handing volume ids to the workers and run the volume ids that failed due to concurrency issues sequently after the concurrent pass. This is probably easily unit testable.

While this sounds like a lot, it does significantly improve the speed of vacuuming large clusters. Please let me know what you think and I'll give it a go! Thanks!

chrislusf:
A more detailed design:
1. collect all the volumes that should be vacuumed into a queue.
2. while the queue is not empty, pop the queue, if the volume replica locations are already busy, put it back to the queue. If the volume server is scheduled a vacuum job, put it to a hashmap, and remove it from the hashmap when the vacuum job is not running.

Also, need to avoid tight loop when there are no volume servers to run vacuum jobs.

PR #3990, opened 2022-11-18
add iscompressable from 128 byte

liuguofaa:
add IsCompressable like upload_content.go

# How is the PR tested?

# Checks


chrislusf:
this is on the server side. the CPU cost is noticeable many tests so these logic were removed before.

PR #4008, opened 2022-11-24
feat: Support for obtaining gzip files without passing accept-encoding header

famosss:
Get the gzip file, if no `accept-encoding` header  is passed, it will return `404`.

Because golang made this restriction. But ceph or other s3 storage will return 200 normally.Not passing `accept-encoding` header is an incorrect way to use it, but other storage does support it

`/net/http/transport.go:2196`
image

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -376,6 +376,13 @@ func (s3a *S3ApiServer) proxyToFiler(w http.ResponseWriter, r *http.Request, des
 	s3a.maybeAddFilerJwtAuthorization(proxyReq, isWrite)
 	resp, postErr := s3a.client.Do(proxyReq)
 
+	if resp.Uncompressed && r.Header.Get("Accept-Encoding") == "" {
+		r.Header.Set("Accept-Encoding","gzip")
+		util.CloseResponse(resp)
+		s3a.proxyToFiler(w, r, destUrl, false, passThroughResponse)
+		return
+	}
+
 	if postErr != nil {
 		glog.Errorf("post to filer: %v", postErr)
 		s3err.WriteErrorResponse(w, r, s3err.ErrInternalError)


weed/s3api/s3api_object_handlers.go
```diff
@@ -376,6 +376,13 @@ func (s3a *S3ApiServer) proxyToFiler(w http.ResponseWriter, r *http.Request, des
 	s3a.maybeAddFilerJwtAuthorization(proxyReq, isWrite)
 	resp, postErr := s3a.client.Do(proxyReq)
 
+	if resp.Uncompressed && r.Header.Get("Accept-Encoding") == "" {
+		r.Header.Set("Accept-Encoding","gzip")
+		util.CloseResponse(resp)
```
chrislusf, line 381: this line seems wrong

famosss: We tested and found no problems, it is to prevent infinite loop.
`resp.Uncompressed` knows that there is a problem with the data requested for the first time
`r.Header.Get("Accept-Encoding") == ""` is to prevent an infinite loop and retry at most once
After &&, this if can only be satisfied once

famosss: util.CloseResponse(resp) is to release the resources of the previous request

famosss: Use `content-encoding` to upload a file for gzip, and then request without `accept-encoding`,can reproduce it

chrislusf: what is the reason not to always set the "accept-encoding" to "gzip"?

famosss: > what is the reason not to always set the "accept-encoding" to "gzip"?

There are too many buckets migrated from ceph, and there are many, many businesses. I am worried that a decompression that does not support `gzip`. They have not passed `accept-encoding` before, and now i cannot be pushed them to change. But like `cos`, `ceph`, etc., it is true that you can return directly without passing. The `content-length` of weed returns -1 and status  `404` because of the limitation of golang. The code in the content

famosss:
@chrislusf 
Is there any other recommended solution? In addition to passing `Accept-Encoding`.

chrislusf (approved):

weed/s3api/s3api_object_handlers.go
chrislusf: Ack. But this fix is not a clean way to solve this problem, just for backward compatibility.

famosss: Yes, but I really didn't think of a more friendly solution. This change really helped us solve the problem of `ceph` and `cos`. If we can't merge into the `master`, we will have a little trouble using `seaweedfs` later.

famosss:
then?

chrislusf:
I do not know a good fix for this, but it should not send 2 proxy requests for client request just for compatibility.

Issue #3987, opened 2022-11-16
[S3API] Empty Etag returned when using Azure Storage as remote


aronneagu:
**Describe the bug**
Fetching the etag using the Minio Python SDK of a file in Azure Remote, using the SeaweedFS S3 API, is returning empty result

**System Setup**
```
> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 18.04.1 LTS
Release:        18.04
Codename:       bionic
> weed version
version 30GB 3.33  linux amd64
python3 --version
Python 3.8.0
```

**Expected behavior**
Fetching the Etag for the file to return the etag as defined in the remote

**Steps to reproduce**
Save the following as a bash script and run it. Export the variable ACCOUNT_KEY (was provided in private, starts with rwe...) before running the script: export ACCOUNT_KEY=rwe...
```
#!/bin/bash
# Set env vars for ACCOUNT_KEY
export ACCOUNT_NAME="seaweedsa"
# Access Key of the server.
pip3 install minio
TMPDIR=$(mktemp -d)
mkdir $TMPDIR/sync
echo Log files are saved in": $TMPDIR"
/usr/local/bin/weed -logdir=${TMPDIR} server -dir=${TMPDIR} -s3 2>/dev/null &
server_pid=$!
echo "remote.configure -name=azure -type=azure -azure.account_name=$ACCOUNT_NAME -azure.account_key=$ACCOUNT_KEY" | weed shell
echo "remote.mount -remote=azure/seaweed -dir=/buckets/seaweed" | weed shell
/usr/local/bin/weed -logdir=${TMPDIR}/sync filer.remote.sync -dir=/buckets/seaweed 2>&1 &
sync_pid=$!
cat <<EOF >min.py
from minio import Minio
import io, os
client = Minio("localhost:8333",secure=False)
result = client.put_object(
    "seaweed", "test-object", io.BytesIO(b"hello"), 5,
    metadata={"Uploaded-by": "aneagu"},
    content_type="application/parquet",
)
result = client.stat_object("seaweed","test-object")
print("The ETAG is: ")
result.etag
EOF
python3 min.py
echo "Waiting for sync to complete" && sleep 2
kill $sync_pid
kill $server_pid
```

**Screenshots**
Seaweed is aware of the etag in the remote (Azure Storage in this case), which can be seen in the key `remoteEtag`
```
# weed shell
> meta.cat /buckets/seaweed/test-object
{
[...]

PR #4025, opened 2022-11-30
Return ETag from remote when file doesn't exist on Filer

aronneagu:
https://github.com/seaweedfs/seaweedfs/issues/3987
Currently, the ETag for a file that exists ONLY in remote is always `d41d8cd98f00b204e9800998ecf8427e-0` which is MD5 of empty string.

Doing a check if file is not saved in Volume server, then return the ETag as set by the remote

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/filechunks.go
+++ b/weed/filer/filechunks.go
@@ -42,7 +42,7 @@ func ETag(entry *filer_pb.Entry) (etag string) {
 }
 
 func ETagEntry(entry *Entry) (etag string) {
-        if entry.Remote != nil && len(entry.Chunks) == 0 {
+        if entry.IsInRemoteOnly() {
 		return entry.Remote.RemoteETag
 	}
 	if entry.Attr.Md5 == nil {
--- a/weed/filer/filechunks.go
+++ b/weed/filer/filechunks.go
@@ -42,6 +42,9 @@ func ETag(entry *filer_pb.Entry) (etag string) {
 }
 
 func ETagEntry(entry *Entry) (etag string) {
+        if entry.Remote != nil && len(entry.Chunks) == 0 {
+		return entry.Remote.RemoteETag
+	}
 	if entry.Attr.Md5 == nil {
 		return ETagChunks(entry.GetChunks())
 	}


aronneagu:
I've discovered that when files are fetched from remote and saved to Filer, the files have empty `ETag` key for chunks. eg:
```
> meta.cat old_data_period.pq
{
  "name": "old_data_period.pq",
  "isDirectory": false,
  "chunks": [
    {
      "fileId": "17,7577ea1377",
      "offset": "0",
      "size": "2604226",
      "modifiedTsNs": "1669829772",
      "eTag": "",                        <- THIS IS EMPTY
      "sourceFileId": "",
      "fid": {
        "volumeId": 17,
        "fileKey": "117",
        "cookie": 2011829111
      },
      "sourceFid": null,
      "cipherKey": "",
      "isCompressed": false,
      "isChunkManifest": false
    }
  ],
  "attributes": {
    "fileSize": "2604226",
    "mtime": "1669741451",
    "fileMode": 420,
    "uid": 0,
    "gid": 0,
    "crtime": "0",
    "mime": "",
    "ttlSec": 0,
    "userName": "",
    "groupName": [],
    "symlinkTarget": "",
    "md5": "",
    "rdev": 0,
    "inode": "0"
  },
  "extended": {},
  "hardLinkId": "",
  "hardLinkCounter": 0,
  "content": "",
  "remoteEntry": {
    "storageName": "azure",
    "lastLocalSyncTsNs": "1669829772428494400",
    "remoteETag": "0x8DAD22BBCD7370A",
    "remoteMtime": "1669741451",
    "remoteSize": "2604226"
  },
  "quota": "0"
}chunks 1 meta size: 131 gzip:159   <- CHUNKS=1 means file is saved on Filer
```
 
I've found that the missing ETag is most likely coming from here https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/filer_grpc_server_remote.go#L153

But I am not sure exactly how to set the its value, its probably some kind of md5sum. If someone would be kind enough to point me to the right function here, I'd appreciate it

PR #4027, opened 2022-12-01
Update Makefile - use binary in current directory

aronneagu:
1. Makefile references `weed` which is found on the PATH
the build command creates the binary in the local directory, so we should be refencing `./weed` instead
2. debug_s3 and debug_webdav require Filer, so changed the command so it starts that as well

# How is the PR tested?

# Checks


chrislusf:
- One change at a time.
- Debug s3 and WebDAV are for debugging. Less components is easier.

PR #4040, opened 2022-12-08
don't set destination when copy object

shichanglin5:
Signed-off-by: changlin.shi <changlin.shi@ly.com>

When testing the s3 interface, many buckets will be created. When the CopyObject interface is called, because the custom destination header is overwritten, each bucket will create some volumes, which will cause a large amount of disk space to be occupied.

In addition, I see that when the filer is processed, if the destination header is not set, the collection will be selected according to the url, so this change will not cause other effects

When S3 CopyObject calls the filer, do not pass the destination parameter, which will overwrite the custom destination header

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3api_object_copy_handlers.go
+++ b/weed/s3api/s3api_object_copy_handlers.go
@@ -98,8 +98,7 @@ func (s3a *S3ApiServer) CopyObjectHandler(w http.ResponseWriter, r *http.Request
 		return
 	}
 	glog.V(2).Infof("copy from %s to %s", srcUrl, dstUrl)
-	destination := fmt.Sprintf("%s/%s%s", s3a.option.BucketsPath, dstBucket, dstObject)
-	etag, errCode := s3a.putToFiler(r, dstUrl, resp.Body, destination)
+	etag, errCode := s3a.putToFiler(r, dstUrl, resp.Body, "")
 
 	if errCode != s3err.ErrNone {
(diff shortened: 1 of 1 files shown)


shichanglin5:
> Is this just preventing creating the destination bucket during test?
> This would cause orphan garbage if the destination bucket is deleted, or cause missing data if the source bucket is deleted.

Yes, by specifying the destination header during the test, all bucket data is in a collection (a set of volumes) to avoid creating a large number of volumes (the test will create a thousand buckets). After the test is completed, manually Just delete the collection corresponding to the destination;

For non-test scenarios, no custom destination header will be set, and the collection selection logic of the filter is still the same as before

chrislusf:
but this means the test is not testing the actual behavior

shichanglin5:
> No, adding the destination header is just to avoid creating a lot of unnecessary volumes during the test process, which does not actually affect the functional test

For acl-related APIs, there are about 150 basic test sets. According to different ownership, bucket acl, and object acl, about 4*5*7*150=2100 test sets can be derived. There are many test cases that will affect each other. As a result, it cannot be executed concurrently, so each test process uses a different bucket to isolate different test processes

For example, the following figure is a part of the test suites, each branch represents a test suite, and will be executed in a separate bucket
image

chrislusf:
the pr will change the file content storage location, breaking existing data placement strategy.

can you not using different buckets?
or you can drop the bucket after each test?

PR #4042, opened 2022-12-08
Fix hardlink counting

wusongANKANG:
Signed-off-by: wusong <wangwusong@virtaitech.com>

HardLinkCounter incorrect during move operation 

keep HardLinkCounter unchanged

# How is the PR tested?
do test with commands( touch a, ln a b, mv b c, rm c) step by step and observe the value of HardLinkCounter.

# Checks


chrislusf:
Nice Fix! Thanks!

PR #4045, opened 2022-12-09
[filer]: Fix http rename

wusongANKANG:
Http api does not prevent parent directory from moving to its subdirectory which can result in filer corruption.

Add parent-child directory relationship judgment.

# How is the PR tested?
curl -i -X POST 'http://localhost:8888/a/b/?mv.from=/a/' should be failed.

# Checks

### DIFF
--- a/weed/filer/filer_rename.go
+++ b/weed/filer/filer_rename.go
@@ -2,11 +2,17 @@ package filer
 
 import (
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/util"
 	"strings"
+
+	"github.com/seaweedfs/seaweedfs/weed/util"
 )
 
-func (f *Filer) CanRename(source, target util.FullPath) error {
+func (f *Filer) CanRename(source, target util.FullPath, oldName string) error {
+	sourcePath := source.Child(oldName)
+	if strings.HasPrefix(string(target), string(sourcePath)) {
+		return fmt.Errorf("mv: can not move directory to a subdirectory of itself")
+	}
+
 	sourceBucket := f.DetectBucket(source)
 	targetBucket := f.DetectBucket(target)
 	if sourceBucket != targetBucket {
--- a/weed/server/filer_grpc_server_rename.go
+++ b/weed/server/filer_grpc_server_rename.go
@@ -19,7 +19,7 @@ func (fs *FilerServer) AtomicRenameEntry(ctx context.Context, req *filer_pb.Atom
 	oldParent := util.FullPath(filepath.ToSlash(req.OldDirectory))
 	newParent := util.FullPath(filepath.ToSlash(req.NewDirectory))
 
-	if err := fs.filer.CanRename(oldParent, newParent); err != nil {
+	if err := fs.filer.CanRename(oldParent, newParent, req.OldName); err != nil {
 		return nil, err
 	}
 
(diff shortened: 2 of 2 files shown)


weed/filer/filer_rename.go
```diff
@@ -2,11 +2,37 @@ package filer
...
+func (f *Filer) CanRename(source, target util.FullPath, oldName string) error {
+	sourcePath := strings.Split(string(source), "/")
+	if sourcePath[len(sourcePath)-1] == "" && len(sourcePath) > 0 {
+		sourcePath = sourcePath[:len(sourcePath)-1]
+	}
+	targetPath := strings.Split(string(target), "/")
+	sourceLen := len(sourcePath)
+	targetLen := len(targetPath)
+
+	if targetLen > sourceLen { // target is subdirectory
```
chrislusf: just check strings.HasPrefix(target, source)?

weed/server/filer_grpc_server_rename.go
```diff
@@ -34,6 +34,20 @@ func (fs *FilerServer) AtomicRenameEntry(ctx context.Context, req *filer_pb.Atom
 		return nil, fmt.Errorf("%s/%s not found: %v", req.OldDirectory, req.OldName, err)
 	}
 
+	if oldEntry.IsDirectory() {
```
chrislusf, line 37: these logic is different from what the PR described?

weed/filer/filer_rename.go
wusongANKANG: Thanks for your suggestion, I have modified it

weed/server/filer_grpc_server_rename.go
wusongANKANG: I'm sorry that I forgot to describe it in the PR.
This logic is the same as the move operation of mount, which is used to limit the directory move to non-empty directories and files.

weed/server/filer_grpc_server_rename.go
```diff
@@ -34,6 +34,20 @@ func (fs *FilerServer) AtomicRenameEntry(ctx context.Context, req *filer_pb.Atom
...
+		targetDir := newParent.Child(req.NewName)
+		newEntry, err := fs.filer.FindEntry(ctx, targetDir)
+		if err == nil {
+			if !newEntry.IsDirectory() {
+				fs.filer.RollbackTransaction(ctx)
+				return nil, fmt.Errorf("%s is not directory", targetDir)
+			}
+			if entries, _, _ := fs.filer.ListDirectoryEntries(context.Background(), targetDir, "", false, 1, "", "", ""); len(entries) > 0 {
+				return nil, fmt.Errorf("%s is not empty", targetDir)
```
chrislusf, line 46: should it also roll back the transaction?

wusongANKANG: I don't think it needs transaction rollback, it is only used as a condition to judge whether the move operation can be done.

chrislusf: line 42 has the rollback. why that is needed?

wusongANKANG: I refer to the logic of your previous StreamRenameEntry function. The FindEntry function has database operations, so transactions are used here.

wusongANKANG: At the same time, I have a question: There are also database operations in the ListDirectoryEntries function, so should a rollback operation be added to the StreamRenameEntry function and other functions that call ListDirectoryEntries?

chrislusf: you may want to send a separate PR for this change.

wusongANKANG: I'm considering to do it. Do you think there is anything else that needs to be changed in this PR?

Issue #4058, opened 2022-12-13
[S3API] Missing etag per chunks when fetching file from remote


aronneagu:
**Describe the bug**
Mount remote Azure Storage and expose via the S3API. Fetch a file from remote. The chunks in Filer have empty Etag.

**System Setup**
```
weed server -s3
export ACCOUNT_NAME="seaweedsa"
export ACCOUNT_KEY="rwex..."
echo "remote.configure -name=azure -type=azure -azure.account_name=$ACCOUNT_NAME -azure.account_key=$ACCOUNT_KEY" | weed shell
echo "remote.mount -remote=azure/seaweed -dir=/buckets/seaweed" | weed shell
```
```
> lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 20.04.4 LTS
Release:        20.04
Codename:       focal
```

**Expected behavior**

**Screenshots**
```
> cd /buckets/seaweed
> meta.cat kubelete.log
{
  "name": "kubelete.log",
  "isDirectory": false,
  "chunks": [
    {
      "fileId": "1,01199704485a",
      "offset": "12582912",
      "size": "512200",
      "modifiedTsNs": "1670947620",
      "eTag": "",                          <- EMPTY
      "sourceFileId": "",
      "fid": {
        "volumeId": 1,
        "fileKey": "281",
        "cookie": 2533640282
      },
      "sourceFid": null,
      "cipherKey": "",
      "isCompressed": false,
      "isChunkManifest": false
    },
    {
      "fileId": "1,011c791f6794",
      "offset": "8388608",
      "size": "4194304",
      "modifiedTsNs": "1670947621",
      "eTag": "",                          <- EMPTY
      "sourceFileId": "",
      "fid": {
        "volumeId": 1,
        "fileKey": "284",
        "cookie": 2032101268
      },
      "sourceFid": null,
      "cipherKey": "",
      "isCompressed": false,
      "isChunkManifest": false
    },
    {
      "fileId": "1,011b196e18f1",
      "offset": "4194304",
      "size": "4194304",
      "modifiedTsNs": "1670947621",
      "eTag": "",                          <- EMPTY
      "sourceFileId": "",
      "fid": {
        "volumeId": 1,
        "fileKey": "283",
        "cookie": 426645745
      },
      "sourceFid": null,
      "cipherKey": "",
      "isCompressed": false,
      "isChunkManifest": false
    },
    {
[...]

PR #4059, opened 2022-12-13
Issue 4058

aronneagu:
Chunks have no Etag when the file is fetched from remote (Azure Storage)

Create a remoteStorageClient and fetch the data for each chunk, then calculate the md5 per chunk, convert to base64 and save it in the Etag field 

# How is the PR tested?
See steps to reproduce in https://github.com/seaweedfs/seaweedfs/issues/4058

# Checks

### DIFF
--- a/weed/server/filer_grpc_server_remote.go
+++ b/weed/server/filer_grpc_server_remote.go
@@ -2,6 +2,7 @@ package weed_server
 
 import (
 	"context"
+	"crypto/md5"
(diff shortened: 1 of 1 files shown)


aronneagu:
Haven't noticed, but this was fixed in https://github.com/seaweedfs/seaweedfs/commit/e1ca6308cb66e25692d8b874715f61c940fa5b91

Issue #4074, opened 2022-12-21
S3 Gateway fails to start when installed via Helm


nadenf:
**Describe the bug**
Installing Seaweed on Kubernetes using Helm chart fails to start the S3 Gateway.
This is because the S3 gateway command is binding to localhost when it should be listening on all interfaces.

**System Setup**
Installing from latest master (at time of this issue) version of charts using a k3s cluster (v1.26.0+k3s1).

**Expected Behaviour**
S3 Gateway should start and be accessible from other pods either via Service or directly hitting the Pod.

chrislusf:
do you plan to send a PR?

PR #4075, opened 2022-12-21
Fixes 4074: S3 Gateway fails to start when installed via Helm

nadenf:
Fixes https://github.com/seaweedfs/seaweedfs/issues/4074.
S3 Gateway fails to start because it is listening only on localhost.

By binding to all interfaces.

# How is the PR tested?
Using k3s and AWS EKS.

# Checks

### DIFF
--- a/k8s/helm_charts2/templates/s3-deployment.yaml
+++ b/k8s/helm_charts2/templates/s3-deployment.yaml
@@ -79,6 +79,7 @@ spec:
               -v={{ .Values.global.loggingLevel }} \
               {{- end }}
               s3 \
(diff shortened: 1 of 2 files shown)


k8s/helm_charts2/templates/s3-deployment.yaml
```diff
@@ -79,6 +79,7 @@ spec:
               -v={{ .Values.global.loggingLevel }} \
               {{- end }}
               s3 \
+              -ip.bind=0.0.0.0 \
```
chrislusf: how about parameterizing this?

PR #4087, opened 2022-12-27
support enable/disable vacuum

guol-fnst:
Make vacuuming more flexible.
If the cluster is in heavy load or it's not the proper time to peform vacuuming.
We can used volume.suspendvacuum to suspend vacuuming and use volume.resumevacuum to continue vacuuming

# How is the PR tested?

# Checks


weed/shell/command_volume_resume_vacuum.go
```diff
@@ -0,0 +1,41 @@
...
+
+func init() {
+	Commands = append(Commands, &commandResumeVacuum{})
+}
+
+type commandResumeVacuum struct {
+}
+
+func (c *commandResumeVacuum) Name() string {
+	return "volume.resumevacuum"
```
chrislusf: This "resume" does not start from where the vacuum is paused. It should not be called "resume".

how about using `volume.vaccum.enable` and `volume.vaccum.disable`? 

And also change the file names and variable names accordingly.

guol-fnst: Done!

weed/topology/topology.go
```diff
@@ -35,6 +35,7 @@ type Topology struct {
 
 	volumeSizeLimit  uint64
 	replicationAsMin bool
+	VacuumStatus     bool
```
chrislusf: rename this

guol-fnst: How about renaming it  to "isDisableVacuum"

guol-fnst: Done

guol-fnst:
@chrislusf  Does this PR need more work?

Issue #3985, opened 2022-11-16
fuse mount exit unexpectly


SigmaStar:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

example of a good issue report:
https://github.com/seaweedfs/seaweedfs/issues/1005
example of a bad issue report:
https://github.com/seaweedfs/seaweedfs/issues/1008

**Describe the bug**
If there are more than one process reading and writing fuse mount point, sometimes the weed mount process exit with segmentation fault.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
- OS version
- output of `weed version`
- if using filer, show the content of `filer.toml`
./weed master
./weed volume -max=500 -mserver="localhost:9333" -port=8080
./weed filer -port=8888 -master="localhost:9333"
./weed mount -filer=localhost:8888 -dir=./mnt/

**Expected behavior**
Normal

**Screenshots**
![image](https://user-images.githubusercontent.com/22829962/202054527-797613f5-c91d-4ad9-8278-5b29074489ae.png)

**Additional context**
there are more than one processes reading and writing dir. Also while running mast there are some warnings which indicate that some file cannot be deleted but I have deleted them many ages ago.

zemul:
weed 3.34

[Trasnport endpoint is not connected ]
it also appears at my mount point, but there was no error or fatal message in the log.

zemul:
Please tell me how to reproduce this error. @sigmaStar

zemul:
Need some `recover` positioning issues. @chrislusf

SigmaStar:
I'm sorry I can't. Because I think it's not the problem of weed mount. My weed filesystem is corrupted, and maybe this exit is because failed operation submitted to the fuser. I think maybe we can add some corrupted file records to weed and then reproduce this error.

chrislusf:
the recent versions, 3.38, 3.39, have a lot of changes about mount. Please use the latest versions to test.

Super-six-java:
May I ask if you have resolved this issue? Did you solve it by upgrading to versions 3.38 and 3.39？thank you so much

Issue #3650, opened 2022-09-13
Follow-up VM disk Corruption and Errors


ProjectInitiative:
Opening a new issue to branching additional issues discovered from #3183.

VM disk checksums seem to working correctly with 3.27 when the VM is stopped. However, running VMs still don't seem to work. 

I have put together a testing method that can hopefully reproduce some of the errors I have been seeing, which are hard to track down.

The included zip file has everything to create a reproducible example, using Docker, Virtualbox and Vagrant. Please see the included `reproduce.md`

Please let me know if I can provide any additional information! 

[bug-report.zip](https://github.com/seaweedfs/seaweedfs/files/9552999/bug-report.zip)

chrislusf:
3.27 added ordered writes. Was the 3.27 making any difference?

ProjectInitiative:
For this specific problem, I didn't see any change. I included two docker files, one this spins up one master and one filer, and one that spins up 3 masters and 3 filers. The one with 3 masters and 3 filers throw some errors in the logs that may be unrelated, so I did testing with just a single filer and single master, using the latest commit.

paochiang:
I also encountered the same problem（`3.22` version, `defaultReplication=002`）:

<img width="1434" alt="image" src="https://user-images.githubusercontent.com/28077875/191252050-00d4f8df-9ca8-47fa-bb8d-290c333efecd.png">

Through investigation, we found that a network problem(we use RDMA network, and can find network error logs in `dmesg`) caused the data transmitted between the volume servers was modified, and  the volume server did not check the md5 before write data to disk. 

I suggest you add some logs to check if the data being transferred has been modified. 

My log example: https://github.com/seaweedfs/seaweedfs/compare/3.22...paochiang:seaweedfs:3.22-log
see diff files: <img width="1223" alt="image" src="https://user-images.githubusercontent.com/28077875/191254077-751594d7-9fe0-4a34-b06a-693269b472ad.png">

write file through `weed mount`, some inconsistent md5 error logs appear :
<img width="1411" alt="image" src="https://user-images.githubusercontent.com/28077875/191254423-8704d7c7-aac0-4d7a-b3d6-26e9f4232299.png">

Hope it helps!

ProjectInitiative:
@chrislusf any updates on potentially merging in @paochiang's network checksum logging?

chrislusf:
I think @paochiang means you can merge it to your own branch.

ProjectInitiative:
[...]

Issue #3183, opened 2022-06-15
Corrupted Files in Cluster


ProjectInitiative:
**Describe the bug**
Files copied to FUSE filer mount are being corrupted.

**System Setup**
- OS version **Debian GNU/Linux 11 (bullseye)**
- output of `weed version` **version 8000GB 3.11 d4ef06cdcf320f8b8b17279586e0738894869eff linux amd64**

```toml
[filer.options]
# with http DELETE, by default the filer would check whether a folder is empty.
# recursive_delete will delete all sub folders and files, similar to "rm -Rf"
recursive_delete = true

[leveldb3]
# similar to leveldb2.
# each bucket has its own meta store.
enabled = true
dir = "/opt/seaweedfs/REDACTED/filer/filerldb3"                    # directory to store level db files
```

**Expected behavior**
Files copied via FUSE mount should match md5sum like any other mount.

**Additional context**
fstab entry:

```ini
fuse /mnt/swfs/localshared fuse.weed filer='REDACTED:8889,REDACTED:8889,REDACTED:8889',filer.path=/,nofail 0 0
```

```bash
root:~# md5sum /var/lib/pve/local-btrfs/images/526/vm-526-disk-0/disk.raw 
fba3db35d8d631a8014bbf89dbb3e9ca  /var/lib/pve/local-btrfs/images/526/vm-526-disk-0/disk.raw
root:~# cp /var/lib/pve/local-btrfs/images/526/vm-526-disk-0/disk.raw /mnt/swfs/localshared/
root:~# md5sum /mnt/swfs/localshared/disk.raw 
f8555d7c8bc4e0081dde769255ddc815  /mnt/swfs/localshared/disk.raw
```

ProjectInitiative:
Currently looking to pull master, volume and filer logs, though they did not appear to report errors.

ProjectInitiative:
Testing with small text file yields expected behavior, larger files do not.
```bash
root:/mnt/swfs# cat test.txt 
this is a very important text file

root:/mnt/swfs# md5sum test.txt 
f473946c76d415891a380bb308a24086  test.txt

root:/mnt/swfs# cp test.txt localshared/
root:/mnt/swfs# cat localshared/test.txt 
this is a very important text file

root:/mnt/swfs# md5sum localshared/test.txt 
f473946c76d415891a380bb308a24086  localshared/test.txt

```

chrislusf:
how large is the large file?

[...]

PR #4089, opened 2022-12-27
more solid weed mount

chrislusf:

`weed mount` could not run virtual machines.

* https://github.com/seaweedfs/seaweedfs/issues/3650
* https://github.com/seaweedfs/seaweedfs/issues/3985
* https://github.com/seaweedfs/seaweedfs/issues/3183

* Add timestamp to each data slice written, and check timestamp during read time.
* Group chunks into sections, to speed up random writes on large files.
* Queue in each section uses linked list instead of arrays.
* Add lock during file handle release.

# How is the PR tested?

Following steps in https://github.com/seaweedfs/seaweedfs/issues/3650

# Checks

### DIFF
--- a/weed/filer/filechunk_group.go
+++ b/weed/filer/filechunk_group.go
@@ -35,9 +35,7 @@ func (group *ChunkGroup) AddChunk(chunk *filer_pb.FileChunk) error {
 	for si := sectionIndexStart; si < sectionIndexStop+1; si++ {
 		section, found := group.sections[si]
 		if !found {
-			section = &FileChunkSection{
-				sectionIndex: si,
-			}
+			section = NewFileChunkSection(si)
 			group.sections[si] = section
 		}
 		section.addChunk(chunk)
@@ -93,9 +91,7 @@ func (group *ChunkGroup) SetChunks(chunks []*filer_pb.FileChunk) error {
 		for si := sectionIndexStart; si < sectionIndexStop+1; si++ {
 			section, found := group.sections[si]
 			if !found {
-				section = &FileChunkSection{
-					sectionIndex: si,
-				}
+				section = NewFileChunkSection(si)
 				group.sections[si] = section
 			}
 			section.chunks = append(section.chunks, chunk)
--- a/weed/filer/filechunk_group.go
+++ b/weed/filer/filechunk_group.go
@@ -1,0 +1,152 @@
+package filer
+
+import (
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+	"github.com/seaweedfs/seaweedfs/weed/util/chunk_cache"
+	"github.com/seaweedfs/seaweedfs/weed/wdclient"
+	"sync"
+)
+
+type ChunkGroup struct {
+	lookupFn       wdclient.LookupFileIdFunctionType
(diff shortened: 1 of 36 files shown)


chrislusf:
@Woellchen I am having trouble running the End to End CI test. It is getting this error, which seems related to be over a memory limit. Not sure how to fix this.

```
fio: terminating on signal 15
Error: Process completed with exit code 137.
```

Woellchen:
> @Woellchen I am having trouble running the End to End CI test. It is getting this error, which seems related to be over a memory limit. Not sure how to fix this.
> 
> ```
> fio: terminating on signal 15
> Error: Process completed with exit code 137.
> ```

This could be because the FIO run exceeded 40 seconds runtime. If you check the e2e.yaml you will see the `fio` commands prepended with the `timeout` command.
I added these timeouts to basically catch locks that are not released anymore. 40 seconds should be more than enough to finish the FIO run. If it takes longer than that then probably something is locked or became really really slow.

ProjectInitiative:
So far the VM tests look promising. Was able to run a simple test VM with no problems on a 1 node system. Will begin larger cluster testing soon

ProjectInitiative:
Finished testing a 3 volume, 3 master, 3 filer, 1 filer store cluster and that seems to work well. Going to try testing Proxmox vs Virtualbox as the hypervisor. That will take more time though.

chrislusf:
> > @Woellchen I am having trouble running the End to End CI test. It is getting this error, which seems related to be over a memory limit. Not sure how to fix this.
> > ```
> > fio: terminating on signal 15
> > Error: Process completed with exit code 137.
> > ```
> 
> This could be because the FIO run exceeded 40 seconds runtime. If you check the e2e.yaml you will see the `fio` commands prepended with the `timeout` command. I added these timeouts to basically catch locks that are not released anymore. 40 seconds should be more than enough to finish the FIO run. If it takes longer than that then probably something is locked or became really really slow.

Currently the 8 FIO processes are started at the same time. How about changing them to run sequentially?

Woellchen:
No, they are running sequentially. See https://github.com/seaweedfs/seaweedfs/blob/master/.github/workflows/e2e.yml#L46

Woellchen:
You can see the full logs yourself. Download them here from the latest run: https://github.com/seaweedfs/seaweedfs/actions/runs/3885979780
It says `compose-mount-1   | I0110 18:05:41.825175 weedfs_file_sync.go:147 /fiotest set chunks: 4096`
The first FIO run uses `size=16M bs=4k`, that makes exactly the `4096` chunks, but since the rafactoring it doesn't batch the chunks together to 2M chunks, as is the default for `weed mount`.
While testing this behavior locally I noticed that the chunks are not even merged for bigger block sizes, it always maxes out at the FUSE limit (128k). This indicates chunks are not at all merged anymore, as was the case before.
This is probably the reason while the test takes longer now because it is busy uploading in the background.

chrislusf:
Thanks! I know the problem now. Let me try to fix it.

chrislusf:
the e2e test is passing again! thanks for the analysis!

Woellchen:
I can confirm it is working as expected now! Thank you!

PR #4093, opened 2023-01-01
filer.store.mysql: Use utf8mb4 instead of 3 byte UTF8, escape table columns in queries

CommanderRoot:
- Allow all UTF8 chars in files or directories instead of just 3 byte UTF8 ones.
- Escape columns names in SQL queries. For example `directory` is a [keyword](https://dev.mysql.com/doc/refman/8.0/en/keywords.html) so escaping makes sense.
- [Depreacted function](https://dev.mysql.com/doc/refman/8.0/en/insert-on-duplicate.html) usage in `upsertQuery`

- Switching mysql tables from utf8 to utf8mb4. I chose `utf8mb4_bin` as COLLATE instead of the [better](https://dba.stackexchange.com/questions/278010/what-is-the-difference-between-different-utf8mb4-binary-collations) `utf8mb4_0900_bin` because it has wider support across different mysql versions. `utf8mb4_0900_bin` would basically require mysql 8.0 and exclude mariadb and probably some more.
- Escaping columns makes sense as we already use `directory` with is a keyword.
- Using VALUES() on `ON DUPLICATE KEY` queries is deprecated so we switch to a non-deprecated way.

# How is the PR tested?
I tested these changes locally on a mysql 8.0 server and everything worked as expected

# Checks

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -41,13 +41,13 @@ enabled = false
 dbFile = "./filer.db"                # sqlite db file
 
 [mysql]  # or memsql, tidb
-# CREATE TABLE IF NOT EXISTS filemeta (
-#   dirhash     BIGINT               COMMENT 'first 64 bits of MD5 hash value of directory field',
-#   name        VARCHAR(1000) BINARY COMMENT 'directory or file name',
-#   directory   TEXT BINARY          COMMENT 'full path to parent directory',
-#   meta        LONGBLOB,
-#   PRIMARY KEY (dirhash, name)
-# ) DEFAULT CHARSET=utf8;
+# CREATE TABLE IF NOT EXISTS `filemeta` (
+#   `dirhash`   BIGINT NOT NULL       COMMENT 'first 64 bits of MD5 hash value of directory field',
+#   `name`      VARCHAR(766) NOT NULL COMMENT 'directory or file name',
+#   `directory` TEXT NOT NULL         COMMENT 'full path to parent directory',
+#   `meta`      LONGBLOB NOT NULL,
+#   PRIMARY KEY (`dirhash`, `name`)
+# ) DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
 
 enabled = false
 hostname = "localhost"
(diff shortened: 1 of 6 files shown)


weed/command/scaffold/filer.toml
```diff
@@ -41,13 +41,13 @@ enabled = false
...
-#   meta        LONGBLOB,
-#   PRIMARY KEY (dirhash, name)
-# ) DEFAULT CHARSET=utf8;
+# CREATE TABLE IF NOT EXISTS `filemeta` (
+#   `dirhash`   BIGINT NOT NULL       COMMENT 'first 64 bits of MD5 hash value of directory field',
+#   `name`      VARCHAR(766) NOT NULL COMMENT 'directory or file name',
+#   `directory` TEXT NOT NULL         COMMENT 'full path to parent directory',
+#   `meta`      LONGBLOB NOT NULL,
+#   PRIMARY KEY (`dirhash`, `name`)
+# ) DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin;
```
chrislusf, line 50: are these charset and collate changes backward compatible?

weed/command/scaffold/filer.toml
```diff
@@ -61,18 +61,18 @@ connection_max_lifetime_seconds = 0
...
 createTable = """
   CREATE TABLE IF NOT EXISTS `%s` (
-    dirhash BIGINT,
-    name VARCHAR(1000) BINARY,
-    directory TEXT BINARY,
-    meta LONGBLOB,
-    PRIMARY KEY (dirhash, name)
-  ) DEFAULT CHARSET=utf8;
+    `dirhash`   BIGINT NOT NULL,
+    `name`      VARCHAR(766) NOT NULL,
```
chrislusf, line 71: why need to reduce the length?

weed/command/scaffold/filer.toml
```diff
@@ -61,18 +61,18 @@ connection_max_lifetime_seconds = 0
...
+    `directory` TEXT NOT NULL,
+    `meta`      LONGBLOB NOT NULL,
```
chrislusf, line 73: this can be null

chrislusf:
there are several changes here. please separate into multiple dedicated PRs, for easier review and possible commit reverting.

weed/command/scaffold/filer.toml
CommanderRoot: The index key prefix length [limit](https://dev.mysql.com/doc/refman/8.0/en/innodb-limits.html) is 3072 bytes for InnoDB tables.
As we are now using the full utf8 space a char can have 4 bytes. So (3072 - 8 (for bigint)) / 4 = bytes left for `name`

weed/command/scaffold/filer.toml
CommanderRoot: Setting the charsets here only affects newly created tables. In that regard the changes are backward compatible. The only issue could arise if someone runs an "ancient" mysql version (prior to 5.5 which was released December 3, 2010) which doesn't have the utf8mb4 type. In that case the table creation would fail.
I haven't tested all different mysql protocol supporting servers (probably not really possible anyway) but if some of them support utf8 and not utf8mb4 it would also fail on them. My guess is however that this shouldn't be the case as utf8mb4 is around for ages. I already used `utf8mb4_bin` instead of `utf8mb4_0900_bin` to be as compatible as possible, even though sorting would be faster with `utf8mb4_0900_bin`

weed/command/scaffold/filer.toml
CommanderRoot: Fixed in new PR #4094

CommanderRoot:
I have split this PR into 3 parts (#4094, #4095 and #4096) so this PR here can be closed

PR #4097, opened 2023-01-02
Add useRaftHashicorp flag in helm chart

Programmeris:
Happy New Year!

In my case, it was necessary to use the implementation of the Raft algorithm from Hashicorp. Helm chart did not allow setting the required parameter for the master server

Added the useRaftHashicorp flag to the Helm chart, which is disabled by default

# How is the PR tested?
Enable useRaftHashicorp in your values.yaml

# Checks

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
+++ b/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
@@ -97,6 +97,12 @@ spec:
               value: {{ $value | quote }}
             {{- end }}
             {{- end }}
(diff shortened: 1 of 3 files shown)


chrislusf:
this option is not finalized yet.

cc @kmlebedev

kmlebedev:
> this option is not finalized yet.
finalized. For me, it solves the problem better, I can change the size of the cluster without stopping the service

chrislusf:
@kmlebedev can you add some instructions on how to change the master cluster size? You can put it in a wiki or just here. I can organize it later when `useRaftHashicorp` is default to true.

PR #4148, opened 2023-01-20
Improve S3 request signing performance

Woellchen:
S3 uploading is quite CPU intense.

This change is caching HMAC hashers for repeated use in subsequent requests and chunks, so they don't have to be initialized from scratch every time.
On my local computer this gives me ~5-6 times faster signature calculation and ~5-6.5% more throughput in S3 requests. The smaller the payload the better the throughput gets.

# How is the PR tested?
Existing unit tests + warp benchmarking tool.

# Checks

### DIFF
--- a/weed/s3api/auth_signature_v4.go
+++ b/weed/s3api/auth_signature_v4.go
@@ -32,6 +32,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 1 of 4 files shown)


weed/s3api/auth_signature_v4.go
```diff
@@ -458,6 +463,38 @@ func (iam *IdentityAccessManagement) doesPresignedSignatureMatch(hashedPayload s
...
+	date := t.Format(yyyymmdd)
+	hashID := "AWS4" + secretKey + "/" + date + "/" + region + "/" + service + "/" + "aws4_request"
+
+	iam.hashMu.RLock()
+	pool, ok := iam.hashes[hashID]
+	iam.hashMu.RUnlock()
+
+	if !ok {
+		iam.hashMu.Lock()
+		if pool, ok = iam.hashes[hashID]; !ok {
```
chrislusf: how to free the map entries?

Woellchen: You are right I totally forgot about it. I will add some cleanup logic as well

Woellchen: I have added logic to clean old entries from the map

weed/s3api/auth_signature_v4.go
```diff
@@ -458,6 +464,69 @@ func (iam *IdentityAccessManagement) doesPresignedSignatureMatch(hashedPayload s
...
+	iam.hashMu.RUnlock()
+
+	if !ok {
+		iam.hashMu.Lock()
+		defer iam.hashMu.Unlock()
+		pool, ok = iam.hashes[hashID]
+	}
+
+	if ok {
+		atomic.AddInt64(iam.hashCounters[hashID], 1)
```
Dimonyga: atomic.StoreInt64(iam.hashCounters[hashID], 1) ?
and maybe Int32 will be enough?

Woellchen: Thanks, makes sense, I adjusted it :)

Issue #3457, opened 2022-08-17
weedfs-3.22 S3  reports a directory as a file


tyronecai:
**Describe the bug**

we run weedfs to set up an S3 service,  used by Aapche Flink,  replace the former Minio.
when upgrade to weedfs-3.22, flink jobmanager start up fail, and report  s3://flink/recovery/flink/blob is a file !!!
But actually It is a directory.
![image](https://user-images.githubusercontent.com/3582423/185061458-661565ff-9b70-4df9-a762-06798b574bf8.png)

Roll back to weedfs-3.19, everything is OK

`Caused by: org.apache.hadoop.fs.FileAlreadyExistsException: Path is a file: s3://flink/recovery/flink/blob
        at org.apache.hadoop.fs.s3a.S3AFileSystem.innerMkdirs(S3AFileSystem.java:2042) ~[?:?]
        at org.apache.hadoop.fs.s3a.S3AFileSystem.mkdirs(S3AFileSystem.java:2007) ~[?:?]
        at org.apache.hadoop.fs.FileSystem.mkdirs(FileSystem.java:2275) ~[hadoop-common-3.3.4.jar:?]
        at org.apache.flink.fs.s3hadoop.common.HadoopFileSystem.mkdirs(HadoopFileSystem.java:183) ~[?:?]
        at org.apache.flink.core.fs.PluginFileSystemFactory$ClassLoaderFixingFileSystem.mkdirs(PluginFileSystemFactory.java:162) ~[flink-dist_2.11-1.12.4.jar:1.12.4]
        at org.apache.flink.runtime.blob.FileSystemBlobStore.<init>(FileSystemBlobStore.java:64) ~[flink-dist_2.11-1.12.4.jar:1.12.4]
        at org.apache.flink.runtime.blob.BlobUtils.createFileSystemBlobStore(BlobUtils.java:98) ~[flink-dist_2.11-1.12.4.jar:1.12.4]
        at org.apache.flink.runtime.blob.BlobUtils.createBlobStoreFromConfig(BlobUtils.java:76) ~[flink-dist_2.11-1.12.4.jar:1.12.4]
        at org.apache.flink.runtime.highavailability.HighAvailabilityServicesUtils.createHighAvailabilityServices(HighAvailabilityServicesUtils.java:115) ~[flink-dist_2.11-1.12.4.jar:1.12.4]
[...]

Issue #3086, opened 2022-05-21
[s3] put fakedir/ key object


kmlebedev:
required for integration with [NextCloud](https://docs.nextcloud.com/server/latest/admin_manual/configuration_files/primary_storage.html)
cmd put key

```
aws --debug --endpoint-url http://127.0.0.1:8333 s3api put-object --bucket test --key fakedir/
2022-05-21 13:01:18,267 - MainThread - botocore.endpoint - DEBUG - Sending http request: <AWSPreparedRequest stream_output=False, method=PUT, url=http://127.0.0.1:8333/test/fakedir/, headers={'User-Agent': b'aws-cli/2.5.1 Python/3.9.12 Darwin/20.6.0 source/x86_64 prompt/off command/s3api.put-object', 'Content-MD5': b'1B2M2Y8AsgTpgAmY7PhCfg==', 'X-Amz-Date': b'20220521T080118Z', 'X-Amz-Content-SHA256': b'e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855', 'Authorization': b'AWS4-HMAC-SHA256 Credential=some_access_key1/20220521/us-west-2/s3/aws4_request, SignedHeaders=content-md5;host;x-amz-content-sha256;x-amz-date, Signature=0e8e0ad443c9fad591a885de961a475f2fb6979bd22500e3e40940b7dedf5dfc', 'Content-Length': '0'}>
2022-05-21 13:01:18,268 - MainThread - urllib3.connectionpool - DEBUG - Starting new HTTP connection (1): 127.0.0.1:8333
2022-05-21 13:01:18,274 - MainThread - urllib3.connectionpool - DEBUG - http://127.0.0.1:8333 "PUT /test/fakedir/ HTTP/1.1" 200 0
2022-05-21 13:01:18,274 - MainThread - botocore.parsers - DEBUG - Response headers: {'Accept-Ranges': 'bytes', 'Content-Length': '0', 'Server': 'SeaweedFS S3', 'X-Amz-Request-Id': '1653120078293682028', 'Date': 'Sat, 21 May 2022 08:01:18 GMT'}
2022-05-21 13:01:18,274 - MainThread - botocore.parsers - DEBUG - Response body:
b''
[...]

PR #4169, opened 2023-01-31
 make directory lister trigger if mime type is httpd/unix-directory

elee1766:

when folders are created via the s3 gateway, the mime type `httpd/unix-directory` is attached to the folder. https://github.com/seaweedfs/seaweedfs/blob/master/weed/s3api/s3api_object_handlers.go#L102

however, when listing files via the filer server, the filer checks if the mime type is empty to determine whether or not to display the folder listing. https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/filer_server_handlers_read.go#L121

this causes folders uploaded via the s3 gateway to appear as blank white screen (404 error) when using the filer web ui

have filer server check if the mime type is either `httpd/unix-directory` or empty, instead of just empty, which will make it properly show directory listing on directories uploaded with s3

### DIFF
--- a/k8s/helm_charts2/Chart.yaml
+++ b/k8s/helm_charts2/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
 description: SeaweedFS
 name: seaweedfs
-appVersion: "3.40"
-version: "3.40"
+appVersion: "3.41"
+version: "3.41"
--- a/weed/util/constants.go
+++ b/weed/util/constants.go
@@ -5,7 +5,7 @@ import (
 )
 
 var (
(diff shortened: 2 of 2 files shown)


kmlebedev:
This was done on purpose and to show the folder as a file, not a directory.
Now the logic is broken https://github.com/seaweedfs/seaweedfs/issues/3086
This is an explicitly created directory, aka a file in s3 terms, it's an object that exists on its own. And black grouse on get or head request we get 404

kmlebedev:
And broken https://github.com/seaweedfs/seaweedfs/issues/3457

PR #4196, opened 2023-02-10
Modify the last modification time of the parent file

TDCBQ:
Modify the last modification time of the parent file

Last modification time of recursive modification of parent file

# How is the PR tested?

# Checks


chrislusf:
I prefer not to have this, which amplifies the write.

TDCBQ:
Yes, this will slow down the write speed
However, compared with other file systems, this part of the function is somewhat missing
So this is a suggestion. I respect your choice

PR #4208, opened 2023-02-15
compatibility patch for csi driver

washcycle:

Ability to use the seaweedfs-csi-driver with the seaweedfs helm chart.

adding namespace to volume `-ip` hostname.

image

# How is the PR tested?

Install csi driver and seaweedfs on seperate namespaces.

# Checks
- [] I will add related wiki document changes and link to this PR after merging.


washcycle:
I added namespace to all the places I could find.

chrislusf:
Thanks!

PR #4217, opened 2023-02-18
fix missing dereference 

washcycle:
fixes dereference in master statefulset
missed one, sorry
@chrislusf 

I'm working on a helm chart publish action for you that will catch these with helm lint and publish the chart to github pages.


chrislusf:
thanks! That'll be great!

chrislusf:
seems there are several places missing the `$` sign.

washcycle:
> seems there are several places missing the `$` sign.

it depended on the $ it was only needed in a few places if it was nested in a certain operator, it required the `$`

PR #4247, opened 2023-02-25
add -disk to filer command

lfhy:
When I start the volume with ```-disk=nvme```,and use filer to upload files.The filer will prompt no free volumes left and upload failed.
```
E0225 11:45:13.938780 filer_server_handlers_write.go:48 failing to assign a file id: rpc error: code = Unknown desc = no free volumes left for {"replication":{},"ttl":{"Count":0,"Unit":0}}
E0225 11:45:13.939355 filer_server_handlers_write_upload.go:209 upload error: rpc error: code = Unknown desc = no free volumes left for {"replication":{},"ttl":{"Count":0,"Unit":0}}
I0225 11:45:13.939424 common.go:70 response method:POST URL:/ with httpStatus:500 and JSON:{"error":"rpc error: code = Unknown desc = no free volumes left for {\"replication\":{},\"ttl\":{\"Count\":0,\"Unit\":0}}"}
```
I added - disk so that filer can write volume with same disk tag.

# How is the PR tested?
Start volume server with ```-disk=nvme```.
Start filer server with ```-disk=nvme```.
Upload the file to filer server.

# Checks

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -60,6 +60,7 @@ type FilerOptions struct {
 	localSocket             *string
 	showUIDirectoryDelete   *bool
 	downloadMaxMBps         *int
+	diskType                *string
 }
 
 func init() {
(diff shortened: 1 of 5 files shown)


chrislusf:
Need the same fix in "weed server" also.

lfhy:
> Need the same fix in "weed server" also.

OK, I pushed the new code and added ```filer.disk``` to ```weed server```. At the same time, if the filer service is selected start and ```filer.disk``` is empty but ```volume.disk``` is not empty, ```filer.disk``` will also be set to the value of ```volume.disk```.

PR #4267, opened 2023-03-03
adjust chunkSizeLimitMB default value

zemul:
-

Consistent with other defaults.

# How is the PR tested?

# Checks


jerebear12:
Wait, why change this? Won't a default size of 2Mb make the metadata storage [fill up quicker](https://github.com/seaweedfs/seaweedfs/wiki/Filer-Server-API)?

> When appending to a file, each append will create one chunk and added to the file metadata. If there are too many small appends, there could be too many chunks. So try to keep each append size reasonably big.

PR #4274, opened 2023-03-06
Avoid OOM when trying to consider all in some cases millions msg in queue

kmlebedev:

If replication has not worked for some time, then a large number of messages can accumulate in the queue, which are trying to be subtracted at one time, which leads to a memory load on the rabbitry server and the replicator may run out of memory

Add a prefetchCount parameter to notification.gocdk_pub_sub which limits the number of messages received at a time

# How is the PR tested?

# Checks


kmlebedev:
prefetchCount on the client side is not configured, since this must be done when creating a channel that is created inside the pub sub library
Therefore, prefetchCount will be configured from the server side as defaults

Issue #4271, opened 2023-03-05
cannot bind metrics port to ipv6


nivekuil:
**Describe the bug**
metrics bind ipv6 address is formatted wrong:
`2023/03/05 09:43:28 listen tcp: address ffff:ffff:ffff:ffff:ffff:ffff:687b:83af:9090: too many colons in address`
I believe this bug was introduced in https://github.com/seaweedfs/seaweedfs/commit/fe670a6241c752536de82aea7d358ecf03252099#diff-014d1186206a3991c7649ef2b7490a5fff7395500f438eaa5a4f0526967e3b81R290/fe670a6241c752536de82aea7d358ecf03252099
I think it needs to use `joinHostPort`

chrislusf:
Send a PR?

nivekuil:
It would involve either duplicating code or refactoring packages, because `util.JoinHostPort` can't be used from `stats` without creating a circular import.

chrislusf:
It's ok to duplicate a small piece of code.

rpodgorny:
maybe this can be closed now?

PR #4286, opened 2023-03-08
Fix binding metrics to ipv6

nivekuil:
https://github.com/seaweedfs/seaweedfs/issues/4271

copying JoinHostPort

# How is the PR tested?
deployed and manually checked

# Checks

### DIFF
--- a/weed/stats/metrics.go
+++ b/weed/stats/metrics.go
@@ -1,7 +1,6 @@
 package stats
 
 import (
-	"fmt"
 	"log"
 	"net"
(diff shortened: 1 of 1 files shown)


weed/stats/metrics.go
```diff
@@ -282,12 +281,22 @@ func LoopPushingMetric(name, instance, addr string, intervalSeconds int) {
 	}
 }
 
+// Copied from util/network.go
```
chrislusf:
```suggestion
```

PR #4306, opened 2023-03-14
Opengen master clients

kmlebedev:

Quickly support more clients for blob storage

Generating clients through openapi documentation using openapi-generator

# How is the PR tested?

# Checks


chrislusf:
This can be a separate repo.

t3hmrman:
How does this sound?

- Generate OpenAPI spec programmatically from this repo (so it stays in-sync with the code, as effortlessly as possible), checked in as part of the repo on releases
- Build automation to trigger a second repo (ex. `seaweedfs/clients`) that runs the openapi machinery to generate clients.

I think another thing worth thinking about is the usage of gRPC -- is having a public gRPC API of any interest? It seems like the current gRPC APIs are not for public consumption

chrislusf:
Automated repo would be nice, or maybe just one Makefile in a folder but git ignore generated files.

Public facing gRPC APIs would be nice too.

t3hmrman:
Those both sound like things I'd be excited to try and implement (I haven't written Golang in a while but the best part about Go is that's not a problem! :)

I know this sounds like far too much bureaucracy for an agile project like this one, but do you have an RFC/enhancement proposal process? Because I'd love to have a place to write out my approach for either of these things, get comments/collaboration, then go off to try and implement.

I can make a PR with a minimal process (i.e. `docs/`, `docs/sep/0-name-of-seaweed-enhancement-proposal/README.md`, `docs/sep/template.md`), along with an idea of how to implement one of those two things for comments

chrislusf:
Yes, please write down some designs before investing on a lot of code. PR description is one way to do this. A specific folder may seem good at first, but may have too many out of sync contents. Let's start with PR first.

Issue #4307, opened 2023-03-14
Single master server takes a long time to start up


stewartmiles:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

example of a good issue report:
https://github.com/seaweedfs/seaweedfs/issues/1005
example of a bad issue report:
https://github.com/seaweedfs/seaweedfs/issues/1008

**Describe the bug**

I would like to use seaweedfs for testing so the current startup speed is an issue for me. Starting the master server on a single node takes quite a long time (~20s). For example:

```shell
$ mkdir tmp
$ weed master -mdir $(readlink -f tmp) -ip localhost -port 9333
I0314 11:41:10.886451 file_util.go:23 Folder /home/stewart/tmp Permission: -rwxr-xr-x
I0314 11:41:10.887853 master.go:268 current: localhost:9333 peers:
I0314 11:41:10.887920 master_server.go:127 Volume Size Limit is 30000 MB
I0314 11:41:10.888133 master.go:149 Start Seaweed Master 30GB 3.43 3227e4175e2bf8df2ac8aeeff8cf73a819abc5a7 at localhost:9333
I0314 11:41:10.888407 raft_server.go:119 Starting RaftServer with localhost:9333
I0314 11:41:10.903295 raft_server.go:168 current cluster leader:
I0314 11:41:28.904858 master_server.go:215 [localhost:9333]  - is the leader.
W0314 11:41:28.905130 tls.go:39 pemfile.NewProvider({   5h0m0s}) grpc.master failed: pemfile: at least one credential file needs to be specified
I0314 11:41:28.907455 master.go:200 Start Seaweed Master 30GB 3.43 3227e4175e2bf8df2ac8aeeff8cf73a819abc5a7 grpc server at localhost:19333
I0314 11:41:30.408553 masterclient.go:155 No existing leader found!
I0314 11:41:30.408613 raft_server.go:190 Initializing new cluster
I0314 11:41:30.408659 master_server.go:174 leader change event:  => localhost:9333
I0314 11:41:30.408689 master_server.go:177 [localhost:9333] localhost:9333 becomes leader.
I0314 11:41:33.911888 master_grpc_server.go:349 + client .master@localhost:9333
```

[...]

PR #4309, opened 2023-03-15
Fix -raftHashicorp and -raftBootstrap flag propagation.

stewartmiles:
`weed server` was not correctly propagating
`-master.raftHashicorp` and `-master.raftBootstrap` flags when starting the master server.

Related to #4307

Propagating the following flags to the master server from the server command:
* -raftHashicorp
* -raftBootstrap

* Using the `masterOption` argument in `startMaster()` rather than global options.
* Exposing `-master.raftBootstrap` as a server command flag.

# How is the PR tested?

* Ran all unit tests and verified they passed.
* Ran master command to verify no change in behavior.
* Ran server command and verified the Hashicorp Raft server is used when the
  `-master.raftBootstrap` flag is specified.

# Checks

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -2,12 +2,13 @@ package command
 
 import (
 	"fmt"
-	hashicorpRaft "github.com/hashicorp/raft"
 	"net/http"
 	"os"
 	"path"
 	"strings"
 	"time"
+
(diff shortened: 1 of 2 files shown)


stewartmiles:
Thanks for the quick review @chrislusf !

PR #4313, opened 2023-03-16
Use exponential backoff to query leader.

stewartmiles:
`topology.Leader()` was using a backoff that typically resulted in at least a 5s delay when initially starting a master and raft server. This changes the backoff algorithm to use exponential backoff starting with 100ms and waiting up to 20s for leader selection.

Related to #4307

Trying to reduce latency when querying the leader master server.

I've changed the backoff algorithm to use exponential backoff starting with 100ms and waiting up to 20s for leader selection.

# How is the PR tested?

* Verified all unit tests pass
* Verified `weed master` and `weed server` commands work as expected.

# Checks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -10,6 +10,7 @@ require (
 	github.com/Azure/azure-storage-blob-go v0.15.0
 	github.com/Shopify/sarama v1.38.1
 	github.com/aws/aws-sdk-go v1.44.219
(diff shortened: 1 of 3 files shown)


stewartmiles:
I just noticed that this will only retry if an error is returned which is different to the original behavior where it would retry until the leader is not an empty string. Easy enough to fix, I'll send a PR in the morning

PR #4331, opened 2023-03-21
Make filer compatible with s3 when handling /buckets/../.. paths (an entry can be both a file and a directory)

shichanglin5:

Due to filer's mandatory compatibility with the path specification of posix, many files (both object files and logical directories in ceph s3) cannot be migrated from ceph to seaweed. 

When migrating these files, a response of 409 (prompt 'existing xxx is a directory' or 'existing xxx is a file') will be generated, which is very frustrating

Make filer compatible with s3 when handling /buckets/../.. paths (an entry can be both a file and a directory)

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/entry.go
+++ b/weed/filer/entry.go
@@ -140,3 +140,16 @@ func maxUint64(x, y uint64) uint64 {
 	}
 	return y
 }
+
+// When Entry.IsDirectory() is true, pass the Mime! = "" to determine whether it is an object file,
+// so if the mime was empty before, the default value is set to "application/octet-stream"
+func (e *Entry) ensureMimeForS3Entry(mime string) {
+	if e.Mime != "" {
+		return
+	}
+	if mime != "" {
+		e.Mime = mime
+	} else {
+		e.Mime = "application/octet-stream"
+	}
+}
--- a/weed/s3api/s3api_object_copy_handlers.go
+++ b/weed/s3api/s3api_object_copy_handlers.go
@@ -3,6 +3,7 @@ package s3api
 import (
 	"fmt"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3_constants"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
 	"modernc.org/strutil"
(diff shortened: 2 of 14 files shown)


chrislusf:
* Prefer not to introduce one more option. This option is only for object storage. It can be auto turned on if the file is in a bucket, where the prefix has `/buckets/.../`. 
* This approach is using one entry for both a directory and a file. Listing directory and deleting files, are tricky.

shichanglin5:
@chrislusf

shichanglin5:
> at `weed/pb/filer_pb/filer_pb_helper.go:20`
> 
> ```go
> func (entry *Entry) IsDirectoryKeyObject() bool {
> 	return entry.IsDirectory && entry.Attributes != nil && entry.Attributes.Mime != ""
> }
> ```
> 
> 我被这段代码误导了，所以现在的逻辑是当 `entry.IsDirectory==true` 时通过 `Mime!=""` 判断是否同时是一个 s3文件，但这是有问题的，因为通过 `S3 PubObject` 上传的文件可能不包含 ContentType，导致Mime 为空字符串，这种情况下，如果对应的 entry 已存在且是个目录的话，无法通过 Mime 判断这个entry是否是 s3 文件；
> 
> 为了能正确的识别一个directory entry 是否是 s3文件，是否可以加个 S3专用的 fs.FileType
> 
> ```go
> var FileModeS3 fs.FileMode = 1 << (32 - 14) // 偏移的offset设置为14是为了和 fs 的其他 FileMode 区分开
> ```
> 
> 当一个 entry 已经是 filer directory的时候，通过 `entry.Mode |= FileTypeS3` 来标识该 entry 是个 s3 文件

当一个 entry 已经存在，且 entry.IsDirectory() 为true 的话，那么当该 entry 关联的 url 上传文件后会将 entry的mime 属性更新，默认为application/octor-stream

weed/filer/entry.go
```diff
@@ -140,3 +140,16 @@ func maxUint64(x, y uint64) uint64 {
...
+// When Entry.IsDirectory() is true, pass the Mime! = "" to determine whether it is an object file,
+// so if the mime was empty before, the default value is set to "application/octet-stream"
+func (e *Entry) ensureMimeForS3Entry(mime string) {
+	if e.Mime != "" {
+		return
+	}
+	if mime != "" {
+		e.Mime = mime
+	} else {
+		e.Mime = "application/octet-stream"
```
chrislusf, line 153: this incurs unnecessary storage cost

weed/s3api/s3api_object_copy_handlers.go
```diff
@@ -309,3 +310,7 @@ func processMetadataBytes(reqHeader http.Header, existing map[string][]byte, rep
 
 	return
 }
+
+func IsS3FileEntry(entry *filer_pb.Entry) bool {
+	return !entry.IsDirectory || (entry.Attributes != nil && entry.Attributes.Mime != "")
```
chrislusf, line 315: s3 directories has mime as "httpd/unix-directory"

shichanglin5: Why is mime of s3 directory set to "httpd/unix-directory"? directory is only a logical concept in s3 api design
Here is to consider how to correctly handle s3 files when entry is both s3 files and directories, because directories do not have mime, and s3 files are generally mime (if not, they need to be set to the default application/octorstream). In this way, when an entry is a directory, we can know whether the entry is a file at the same time through mime.

chrislusf: @kmlebedev any comments about setting the mime of s3 directory to "httpd/unix-directory"?

Issue #4346, opened 2023-03-28
mount deadlock


neo-v:
**Describe the bug**
Mount may hung by deadlock.
Here is the reason:

![image](https://user-images.githubusercontent.com/10511493/228212087-537e4243-f464-42dc-af6c-bc9c3e3e04b1.png)

stacks:
```
goroutine 1248229 [semacquire, 12 minutes]:
sync.runtime_SemacquireMutex(0xc0000ff418?, 0x60?, 0xc0004261d0?)
        /usr/local/go/src/runtime/sema.go:71 +0x25
sync.(*Mutex).lockSlow(0xc000a7cc98)
        /usr/local/go/src/sync/mutex.go:162 +0x21d
sync.(*Mutex).Lock(0xc000a7cc98)
        /usr/local/go/src/sync/mutex.go:81 +0x65
github.com/seaweedfs/seaweedfs/weed/mount.(*FileHandle).AddChunks(0xc000a7cc80, {0xc00797bb90, 0x1, 0x1})
        /home/weed_source/seaweedfs/weed/mount/filehandle.go:95 +0x6b
github.com/seaweedfs/seaweedfs/weed/mount.(*ChunkedDirtyPages).saveChunkedFileIntervalToStorage(0xc0022d82a0, {0x495d780, 0xc00687c8a0}, 0x40000, 0x20000, 0xc00687c8a0?, 0x34a4bc0)
        /home/weed_source/seaweedfs/weed/mount/dirty_pages_chunked.go:79 +0x3b7
github.com/seaweedfs/seaweedfs/weed/mount/page_writer.(*MemChunk).SaveContent(0xc001521270, 0xc003eb9310)
        /home/weed_source/seaweedfs/weed/mount/page_writer/page_chunk_mem.go:105 +0x246
github.com/seaweedfs/seaweedfs/weed/mount/page_writer.(*UploadPipeline).moveToSealed.func1()
        /home/weed_source/seaweedfs/weed/mount/page_writer/upload_pipeline.go:184 +0xaa
github.com/seaweedfs/seaweedfs/weed/util.(*LimitedConcurrentExecutor).Execute.func1()
        /home/weed_source/seaweedfs/weed/util/limited_executor.go:38 +0x9f
created by github.com/seaweedfs/seaweedfs/weed/util.(*LimitedConcurrentExecutor).Execute
        /home/weed_source/seaweedfs/weed/util/limited_executor.go:33 +0x11b


[...]

PR #4352, opened 2023-03-29
fix:mount deadlock

zemul:
#4346 

# How is the PR tested?

# Checks


neo-v:
how about erase fh.entryLock.Lock()

zemul:
> how about erase fh.entryLock.Lock()
no problem.

zemul:
@chrislusf

chrislusf:
Is this still needed? If so, please write down some pseudo code describing current relationships between the two locks, to show why this PR can help and why it won't cause problems.

PR #4377, opened 2023-04-05
adjusted regex to be from 1 to 255 for the value

washcycle:
adjust regex for filer ttl validation


washcycle:
@chrislusf I updated this for 1 to 255 if you want to replace the previous one that I think only does [0 to 99]. I didn't do this one before bed I swear.

PR #4398, opened 2023-04-12
fix(chart): use filer data PVC correctly

samcday:
Follow-up to #4384

... Not sure if there's other related lurking bugs related to this ongoing `enablePVC -> data.*.type` transition. What's the plan on that? Maybe better to carry it out sooner rather than later ;) (I'm offering to do the work!)

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
+++ b/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
@@ -175,7 +175,7 @@ spec:
             - mountPath: /etc/sw
               name: config-users
               readOnly: true
-            {{- if (or .Values.filer.enablePVC (eq .Values.filer.data.type "hostPath")) }}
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks! Please help since you have more expertise here!

samcday:
> Thanks! Please help since you have more expertise here!

I'm not sure about that, but I'll help if I can :)

The Helm chart is being versioned in-line with the main SeaweedFS project. Personally, I'm a big fan of this!

Unfortunately though, it means if we want to actually remove the enablePVC option, we should be doing that as a major breaking change. (i.e a `4.x` version).

If you're not too worried about strictly observing semantic versioning guarantees on the Helm chart, we could instead add a small patch to NOTES.txt that causes the chart to fail installation if `filer.enablePVC` is set, with a message asking the user to instead switch to `filer.data`.

Please let me know what you think and I'll open an appropriate follow-up PR :)

samcday:
@chrislusf ping! I can open another issue to continue the discussion if you'd prefer.

chrislusf:
the latter approach should be easier for users to migrate.

Issue #4405, opened 2023-04-16
FUSE deadlock (in weed_copy_file_range?)


werdnum:
I've had a recurring issue where an individual file will get into a bad state where all attempts to read or write hang forever.

I've managed to pull a goroutine dump using dlv, and found this stack trace hanging forever:

```

Goroutine 1327 - User: /usr/local/go/src/runtime/sema.go:77 sync.runtime_Sem
acquireMutex (0x78f68) [timer goroutine (idle) 2444010484138]
         0  0x000000000004a1d0 in runtime.gopark
             at /usr/local/go/src/runtime/proc.go:382
         1  0x000000000005b79c in runtime.goparkunlock
             at /usr/local/go/src/runtime/proc.go:387
         2  0x000000000005b79c in runtime.semacquire1
             at /usr/local/go/src/runtime/sema.go:160
         3  0x0000000000078f68 in sync.runtime_SemacquireMutex
             at /usr/local/go/src/runtime/sema.go:77
         4  0x0000000000087584 in sync.(*Mutex).lockSlow
             at /usr/local/go/src/sync/mutex.go:171
         5  0x000000000169f7c0 in sync.(*Mutex).Lock
             at /usr/local/go/src/sync/mutex.go:90
         6  0x000000000169f7c0 in github.com/seaweedfs/seaweedfs/weed/mount.(*
FileHandle).readFromChunks
             at /github/workspace/weed/mount/filehandle_read.go:26
         7  0x00000000016a9b90 in github.com/seaweedfs/seaweedfs/weed/mount.re
adDataByFileHandle
             at /github/workspace/weed/mount/weedfs_file_read.go:80
         8  0x00000000016a7c20 in github.com/seaweedfs/seaweedfs/weed/mount.(*
WFS).CopyFileRange
             at /github/workspace/weed/mount/weedfs_file_copy_range.go:77
         9  0x000000000167d890 in github.com/hanwen/go-fuse/v2/fuse.doCopyFile
Range
             at /go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.2.0/fuse/opcode.go
:483
        10  0x0000000001685fd8 in github.com/hanwen/go-fuse/v2/fuse.(*Server).
handleRequest
             at /go/pkg/mod/github.com/hanwen/go-fuse/v2@v2.2.0/fuse/server.go
:514
        (truncated)
```

[...]

PR #4407, opened 2023-04-17
[mount]Fix copy file range

zemul:
#4405 

Remove duplicate locks.

# How is the PR tested?
local call copy_file_range.

# Checks


werdnum:
Glad to see this getting attention, but FWIW you can still get a deadlock if the same goroutine acquires a read-lock twice, if some other process calls `Lock()` in between. See https://pkg.go.dev/sync#RWMutex:

> If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.
>
> In the terminology of the Go memory model, the n'th call to Unlock “synchronizes before” the m'th call to Lock for any n < m, just as for Mutex. For any call to RLock, there exists an n such that the n'th call to Unlock “synchronizes before” that call to RLock, and the corresponding call to RUnlock “synchronizes before” the n+1'th call to Lock.

zemul:
> Glad to see this getting attention, but FWIW you can still get a deadlock if the same goroutine acquires a read-lock twice, if some other process calls `Lock()` in between. See https://pkg.go.dev/sync#RWMutex:
> 
> > If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.
> > In the terminology of the Go memory model, the n'th call to Unlock “synchronizes before” the m'th call to Lock for any n < m, just as for Mutex. For any call to RLock, there exists an n such that the n'th call to Unlock “synchronizes before” that call to RLock, and the corresponding call to RUnlock “synchronizes before” the n+1'th call to Lock.

This merge request has deleted the second lock on entryLock in the same goroutine

werdnum:
> > Glad to see this getting attention, but FWIW you can still get a deadlock if the same goroutine acquires a read-lock twice, if some other process calls `Lock()` in between. See https://pkg.go.dev/sync#RWMutex:
> > > If a goroutine holds a RWMutex for reading and another goroutine might call Lock, no goroutine should expect to be able to acquire a read lock until the initial read lock is released. In particular, this prohibits recursive read locking. This is to ensure that the lock eventually becomes available; a blocked Lock call excludes new readers from acquiring the lock.
> > > In the terminology of the Go memory model, the n'th call to Unlock “synchronizes before” the m'th call to Lock for any n < m, just as for Mutex. For any call to RLock, there exists an n such that the n'th call to Unlock “synchronizes before” that call to RLock, and the corresponding call to RUnlock “synchronizes before” the n+1'th call to Lock.
> 
> This merge request has deleted the second lock on entryLock in the same goroutine

Ah, my mistake – I see that the one left is the lock on the file itself, whereas `filehandle_read` locks the `entryLock`.

Issue #4381, opened 2023-04-05
Weed Mount Unable to Flush Large Files


ProjectInitiative:
**Describe the bug**
When copying large files into a fresh cluster, they start to fail the copy. Rsync and Rclone both fail because the are unable to close the file. the output of `weed mount` is shown below

```bash
I0405 22:18:01.447833 dirty_pages_chunked.go:75 /my/large/file.bin saveToStorage [49541021696,49543118848): upload data: unmarshalled error http://REDACTED:8888/?proxyChunkId=13,03453f8cf79849: failed to write to replicas for volume 13: [seaweedfs-volume-1.seaweedfs-volume.swfs:8080]: unmarshalled error http://seaweedfs-volume-1.seaweedfs-volume.swfs:8080/13,03453f8cf79849?ts=1680733080&ttl=&type=replicate: failed to write to local disk: volume 13 is read only
E0405 22:23:05.533213 weedfs_file_sync.go:103 /my/large/file.bin doFlush: flush data: upload data: unmarshalled error http://REDACTED:8888/?proxyChunkId=13,03453f8cf79849: failed to write to replicas for volume 13: [seaweedfs-volume-1.seaweedfs-volume.swfs:8080]: unmarshalled error http://seaweedfs-volume-1.seaweedfs-volume.swfs:8080/13,03453f8cf79849?ts=1680733080&ttl=&type=replicate: failed to write to local disk: volume 13 is read only
```

**System Setup**
- Kubernetes v1.25.7+k3s1

- Deployed with helm chart in repo, all defaults except what is listed below
- tried with `volumeSizeLimitMB: 30000` and `volumeSizeLimitMB: 1000`
- master: replica 3
- volume: replica 3
- filer: replica 1
- output of `weed version` version 30GB 3.45 422bfae linux amd64

**Expected behavior**
File operations should be able to close and flush properly.

**Additional Context** 

There are several TB of free space, and hundreds of free volumes available, not sure why the last volume the filer/mount tries to write to kills the entire copy.

Possibly related issues from the past: #3628 and  #3345

Issue #3628, opened 2022-09-08
Failed to write to local disk / flush data failed: volume XXX is read only


PUSHRCDN:
**Describe the bug**
When a volume becomes read only, some files being uploaded will end up being corrupted and will display a size of 0 bytes.

**System Setup**
- weed master -defaultReplication=001 -volumeSizeLimitMB=10240
- Ubuntu 22.04.1 LTS
- a cluster of 4 servers, 40 disk drives, weed mount on server 1. FTP server running on the same server to allow uploads to the mount
- weed version 30GB 3.23  linux amd64

- master.toml:

```
[master.maintenance]
# periodically run these scripts are the same as running them from 'weed shell'
scripts = """
  lock
  ec.encode -fullPercent=95 -quietFor=1h
  ec.rebuild -force
  ec.balance -force
  volume.deleteEmpty -quietFor=24h -force
  volume.balance -force
  volume.fix.replication
  s3.clean.uploads -timeAgo=24h
  unlock
"""
sleep_minutes = 17          # sleep minutes between each script execution
```

- filer.toml:

```
[redis3] # beta
enabled = true
address = "localhost:[redacted]"
password = "[redacted]"
database = 0
```

**Expected behavior**
This issue is observed on a setup where files are being uploaded to a weed FUSE mount via FTP. Expected behaviour would be for the file to be stored correctly. However, the resulting file will be corrupted, displaying a size of 0 bytes. This issue only happens when a volume is in read only state. 

**Screenshots**
Upload via FTP:
`Sep  8 02:03:45 SFS-FSN1-DC1-0001 pure-ftpd: (2220@[redacted]) [NOTICE] /home/sfs/2220/[redacted].mp4 uploaded  (7672027944 bytes, 9162.76KB/sec)`

Weed warning log:

`E0908 02:03:45.668145 weedfs_file_sync.go:106 /2220/[redacted].mp4 doFlush: flush data: upload data: unmarshalled error http://[redacted]:8082/438,72c1e1bd517e8b: failed to write to local disk: volume 438 is read only`

Another file:

`W0908 02:03:33.442864 upload_content.go:164 uploading 2 to http://[redacted]:8082/438,72c1e61248936c: unmarshalled error http://[redacted]:8082/438,72c1e61248936c: failed to write to local disk: volume 438 is read only`

Running stat on the file:

```
root@SFS-FSN1-DC1-0001 ~ # stat /home/sfs/2220/[redacted].mp4
[...]

Issue #3345, opened 2022-07-21
Failed to write to replicas for volume, volume is read only


ningfdx:
I write 240 big file （80GB） with 1 replication, and got 10 file flush failed.

Seems like the situation mentions in #2757, I didn't find the code bug, so what could be the reason? The master shell of volume.balance ?

Every failed weed mount program log is like
```
I0720 21:37:05 35209 slot_pool.go:60] --> 2
I0720 21:37:05 35209 filehandle.go:96] /test-80GB-copy182 existing 8199 chunks adds 1 more
I0720 21:37:05 35209 dirty_pages_chunked.go:89] /test-80GB-copy182 saveToStorage 38,013b8255d99798 [85993717760,85996367299)
I0720 21:37:05 35209 upload_pipeline.go:166]  uploaderCount 2 --> 1
I0720 21:37:05 35209 upload_pipeline.go:39] Free sealed chunk:  finished uploading chunk 8201
I0720 21:37:05 35209 slot_pool.go:60] --> 1
I0720 21:37:05 35209 filehandle.go:96] /test-80GB-copy182 existing 8200 chunks adds 2 more
I0720 21:37:05 35209 dirty_pages_chunked.go:89] /test-80GB-copy182 saveToStorage 23,013b80b8ae364c [85983232000,85993717760)
I0720 21:37:05 35209 upload_pipeline.go:166]  uploaderCount 1 --> 0
I0720 21:37:05 35209 upload_pipeline.go:39] Free sealed chunk:  finished uploading chunk 8200
I0720 21:37:05 35209 slot_pool.go:60] --> 0
E0720 21:37:05 35209 weedfs_file_sync.go:106] /test-80GB-copy182 doFlush: flush data: upload data: unmarshalled error http://192.168.114.13:8089/30,0139f379fb826d: failed to write to replicas for volume 30: [192.168.114.13:8081]: unmarshalled error http://192.168.114.13:8081/30,0139f379fb826d?ts=1658324217&ttl=&type=replicate: failed to write to local disk: volume 30 is read only
2022/07/20 21:37:05 tx 5249928:     5=input/output error
2022/07/20 21:37:05 rx 5249930: FLUSH n16901169560427606540 {Fh 6}
I0720 21:37:05 35209 weedfs_file_sync.go:102] doFlush /test-80GB-copy182 fh 0
[...]

PR #4416, opened 2023-04-21
mount: add retry for read only case

wusongANKANG:

When the disk space corresponding to a volume server is less than "minfreespace" (default 1%), "isDiskSpaceLow" will be set to true, and all volumes corresponding to the node are read only.

However, it takes 5 seconds for the volume to become read only until the master perceives "read only" through the heartbeat information. During this period, if the master allocates fid to the volume on this node, the write will fail.

According to my analysis, #4381 ， #3628 , #3345 failed to write because of this reason.

We add retry to mount for this situation, so that the master can assign fids to other nodes to avoid write failures.

# How is the PR tested?

env: more than 4 volume server nodes, ReplicaPlacement=002

When writing data through mount, use large files to fill up the disk corresponding to a volume server. At this time, all volumes on the node are read only. At this time, the mount will retry, and the write will succeed after the master assigns the fid to other nodes.

# Checks

### DIFF
--- a/weed/operation/upload_content.go
+++ b/weed/operation/upload_content.go
@@ -121,7 +121,8 @@ func UploadWithRetry(filerClient filer_pb.FilerClient, assignRequest *filer_pb.A
 			return true
 		})
 	} else {
-		err = util.UploadRetry("uploadWithRetry", doUploadFunc)
+		uploadErrList := []string{"transport", "is read only"}
+		err = util.MultiRetry("uploadWithRetry", uploadErrList, doUploadFunc)
 	}
 
 	return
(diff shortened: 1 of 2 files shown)


chrislusf:
Nice finding!

For this fix, the application logic is leaked into the util function. Is there any less hacky way to fix this?

wusongANKANG:
@chrislusf At present it seems that the only way to solve this problem is by retrying.

In order to avoid affecting the original util function logic, would it be better to write a new util retry function for retrying mount upload data?

chrislusf:
you can create a retry function with a list of errors that should be retried.

wusongANKANG:
@chrislusf added

chrislusf:
the retry function can take a list of errors as input parameter. If these errors are encountered, the operation should retry.

the recent change hard coded the errors, which is not so nice.

wusongANKANG:
@chrislusf  error list added

PR #4429, opened 2023-04-25
fix_replication commands support custom 'replication'

shichanglin5:
The collection replication is set on volume when each volume is created. If the initial replication setting is 000 and the subsequent change is 010, the existing volume will not take effect. At this time, we cannot add copies to improve the disaster preparedness capability.

In the fix_replication, increase the `replication` option. If a parameter value is specified, the fix_replication uses the parameter value to adjust the copy instead of using the original replication property of volume

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_volume_fix_replication.go
+++ b/weed/shell/command_volume_fix_replication.go
@@ -60,6 +60,7 @@ func (c *commandVolumeFixReplication) Do(args []string, commandEnv *CommandEnv, 
 	noDelete := volFixReplicationCommand.Bool("noDelete", false, "Do not delete over-replicated volumes, only fix under-replication")
 	retryCount := volFixReplicationCommand.Int("retry", 5, "how many times to retry")
 	volumesPerStep := volFixReplicationCommand.Int("volumesPerStep", 0, "how many volumes to fix in one cycle")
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_fix_replication.go
```diff
@@ -60,6 +60,7 @@ func (c *commandVolumeFixReplication) Do(args []string, commandEnv *CommandEnv,
 	noDelete := volFixReplicationCommand.Bool("noDelete", false, "Do not delete over-replicated volumes, only fix under-replication")
 	retryCount := volFixReplicationCommand.Int("retry", 5, "how many times to retry")
 	volumesPerStep := volFixReplicationCommand.Int("volumesPerStep", 0, "how many volumes to fix in one cycle")
+	replication := volFixReplicationCommand.String("replication", "", "Customize the number of replicas instead of following the setting when the volume was created, specify this parameter")
```
chrislusf, line 63: After a different replication setting, the volume would be considered wrongly replicated.

when this is needed? 

The volume should change the replication first then fix the replication?

shichanglin5: got it，I should have used "volume.configure.replication", then "volume.fix.replication".

PR #4431, opened 2023-04-26
[filer]fix: limit the directory move to non-empty directory and file

wusongANKANG:
limit the directory move to non-empty directories and files.

refer to #4045 

refer to StreamRenameEntry function

# How is the PR tested?

# Checks


chrislusf:
why is this needed?

wusongANKANG:
In order to maintain the same behavior as mount.
Or is it by design that this restriction is not needed?

chrislusf:
Could you please show the related code in mount?

wusongANKANG:
Maybe it wasn't clear.
 After the mount executes the rename operation, the filer side uses the StreamRenameEntry function for processing. This restriction is done in the StreamRenameEntry function.

The AtomicRenameEntry function handles http rename, so I also added this restriction to AtomicRenameEntry.

chrislusf:
there should not be this limitation. Moving a directory to an non-empty directory should be fine.

PR #4434, opened 2023-04-28
[Filer] post add param:saveInside

zemul:
The s3 configuration file uploaded using the filer http port cannot be recognized

# How is the PR tested?

# Checks
- [ yes] I will add related wiki document changes and link to this PR after merging.


zemul:
The s3 configuration file uploaded using filerUi is also unrecognizable. Is it more appropriate to store all the files uploaded to the /etc directory in the metadata?

Issue #4452, opened 2023-05-08
FUSE deadlock in page_chunk_swapfile (inconsistent ordering)


werdnum:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

example of a good issue report:
https://github.com/seaweedfs/seaweedfs/issues/1005
example of a bad issue report:
https://github.com/seaweedfs/seaweedfs/issues/1008

**Describe the bug**

FUSE deadlock - see go-deadlock output: https://gist.github.com/werdnum/9a1ce4ccf5e6bf0747a8b1c90596b21e

**System Setup**
Custom build based on 3.49 59f55c1a661516033e05f6cc306cdb92e885117c with go-deadlock added

zemul:
```
func (i *FileHandleToInode) ReleaseByInode(inode uint64) {
	i.Lock()
	defer i.Unlock()
	fh, found := i.inode2fh[inode]
	if found {
		fh.counter--
		if fh.counter <= 0 {
			delete(i.inode2fh, inode)
			delete(i.fh2inode, fh.fh)
			fh.ReleaseHandle()
		}
	}
[}](url)
```
`ReleaseHandle` is protected by a counter, deadlocks do not occur.

PR #4462, opened 2023-05-11
Fix deadlock

zemul:
#4452

Func AppendChunks  has been locked.

# How is the PR tested?

# Checks


zemul (commented):
This method is safe

PR #4480, opened 2023-05-18
support swift

mervynzhang:

swift doesn't support Object tagging. 
https://github.com/openstack/swift/blob/master/swift/common/middleware/s3api/s3request.py#L820

provider a flag s3.support_tagging to support swift without tagging. 

# How is the PR tested?
build a docker image and run locally.

# Checks


weed/remote_storage/s3/s3_storage_client.go
```diff
@@ -170,13 +170,27 @@ func (s *s3RemoteStorageClient) WriteFile(loc *remote_pb.RemoteStorageLocation,
...
-	_, err = uploader.Upload(&s3manager.UploadInput{
-		Bucket:       aws.String(loc.Bucket),
-		Key:          aws.String(loc.Path[1:]),
-		Body:         reader,
-		Tagging:      aws.String(tags),
-		StorageClass: aws.String(s.conf.S3StorageClass),
-	})
+	var s3_uploadInput s3manager.UploadInput
+	
+	if s.conf.S3SupportTagging {
```
chrislusf: how about conditionally setting Tagging to nil, without repeating the UploadInput function?

mervynzhang: Update code to conditionally set Tagging to nil, thank you.

weed/remote_storage/s3/s3_storage_client.go
```diff
@@ -169,12 +169,18 @@ func (s *s3RemoteStorageClient) WriteFile(loc *remote_pb.RemoteStorageLocation,
 		}
 	}
 
+	awsTags := aws.String(tags)
```
chrislusf: Nit: reverse this to avoid extra call to `aws.String()` function when not needed.

mervynzhang: updated. I'm new to golang, not sure if it's the best way to do this

PR #4484, opened 2023-05-19
Feat: etcd filer store keys should have customizable prefix

mhameed:
An etcd cluster is not necessarily only dedicated to seaweedfs. This security enhancement adds a customizable key_prefix option to the etcd filer store. This will allow an etcd cluster administrator to limit the seaweedfs etcd user to only read/write a subset of keys under the key_prefix, instead of all keys on the etcd cluster.

Now that the seaweedfs etcd user can be authenticated by the etcd cluster, the seaweedfs etcd user should not have access to all keys on etcd cluster.

By adding a customizable key_prefix.
This means that the etcd cluster administrator can restrict the seaweedfs etcd user to read/write operations for keys under this key_prefix.
 

# How is the PR tested?

By specifying the key_prefix in the filer.toml, and restricting the seaweedfs etcd user to the "seaweedfs." prefix, and seaweedfs were able to read/write the required keys/metadata as normal.

# Checks

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -264,6 +264,7 @@ enabled = false
 servers = "localhost:2379"
 username = ""
 password = ""
+key_prefix = "seaweedfs."
 timeout = "3s"
 
 [mongodb]
--- a/weed/filer/etcd/etcd_store.go
+++ b/weed/filer/etcd/etcd_store.go
@@ -25,6 +25,7 @@ func init() {
 
 type EtcdStore struct {
 	client *clientv3.Client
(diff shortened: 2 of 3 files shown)


n9:
@mhameed Are you using your change?

I need to change it the following way: https://github.com/n9/seaweedfs/commit/242d5e86a60fc73ca139f14c71b48f362deeb8c4

1. Add the end to the range, otherwise ETCD was failing on authorization.
2. Remove the prefix check, as it will not match since the returned keys have the `store.etcdKeyPrefix`. (Moreover, the check will be now redundant? As it should be now handled by the ETCD itself (1)?

PR #4508, opened 2023-05-26
Batch and pipeline multiple Redis calls together

Woellchen:
When using Redis store there are a lot of single commands sent to Redis sequentially. This causes many round trips and network IO wait.
For example for directories this produces a GET command for each child in the directory. Imagine a folder with a million files, it would sequentially execute 1 million GET commands.

Using the client-side pipelining feature to send all commands and get all their results in a single roundtrip.
This is implemented for `redis2` stores. If you like the idea I can go ahead and change the other redis stores as well.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/redis2/universal_redis_store.go
+++ b/weed/filer/redis2/universal_redis_store.go
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	DIR_LIST_MARKER = "\x00"
+	DirListMarker = "\x00"
 )
 
 type UniversalRedis2Store struct {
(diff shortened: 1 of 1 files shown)


agipson33:
As a Redis v7 user with SeaweedFS configured to use `redis_cluster3`, I would love to see these improvements and have many directories with thousands of files to test performance improvements against.

weed/filer/redis2/universal_redis_store.go
```diff
@@ -187,38 +251,65 @@ func (store *UniversalRedis2Store) ListDirectoryEntries(ctx context.Context, dir
...
+
+	entries, err := store.FindEntries(ctx, paths)
+	if err != nil {
+		glog.V(0).Infof("list %s : %v", dirPath, err)
+		return lastFileName, err
+	}
+
+	// Clean expired keys
+	var expiredEntries []*filer.Entry
+	for _, entry := range entries {
```
chrislusf, line 273: what if the entries in the first batch are all expired?

PR #4525, opened 2023-06-02
fix deadlock and full capacity volumes check

### LINKED ISSUE #4522: Strange behaviour / deadlock with seaweedfs 3.51
Every day we have task with huge amount of writes to seaweedfs cluster. Two days ago we upgraded to 3.51 and since them we've experienced troubles with write. Seaweedfs fuse hangs on writes after some time. This happend two times. Restarting seaweedfs itself helped. For now we've reverted to 3.50 and we experiance no problems again.

I'm writing load test to investigate this problem and will add more info very soon.

wusongANKANG:

fix #4522   and  optimize #4491 

In our test env, it is found that #4491 cannot completely solve the problem(no more writables volumes while disk free space is sufficient), because we set the volume size to 0 in the vacuum commit, but we may receive the old size(heartbeat) from the volume server before CollectDeadNodeAndFullVolumes, which will also cause misjudgment. And deadlock #4522 may occur due to locking on datanode.

We record the time of volume vacuum when vacuum commit, and check it in CollectDeadNodeAndFullVolumes with delay.

If a volume has been vacuumed in the past 20 seconds, we do not check whether it has reached full capacity.
After 20s(grpc timeout), theoretically all the heartbeats of the volume server have reached the master, the volume size should be correct, not the size before the vacuum.

In this way, the volume that should be writable will not be misjudged as unwritable. 

# How is the PR tested?

# Checks


weed/topology/volume_layout.go
```diff
@@ -112,6 +113,7 @@ type VolumeLayout struct {
 	vid2location     map[needle.VolumeId]*VolumeLocationList
 	writables        []needle.VolumeId // transient array of writable volume id
 	crowded          map[needle.VolumeId]struct{}
+	vacuumedVolumes  map[needle.VolumeId]int64
```
chrislusf, line 116: use `map[needle.VolumeId]time.Time`

chrislusf:
Also, can you split this PR into two for easier understanding?

PR #4529, opened 2023-06-02
shell meta load add concurrency

kmlebedev:

# How is the PR tested?

# Checks


chrislusf:
This can cause problems where file metadata is created first before the directory. To fix this, we can iterate through the folders as a single thread, and then concurrently to process all the files.

kmlebedev:
> This can cause problems where file metadata is created first before the directory. To fix this, we can iterate through meta dump twice: the first time to process the folders as a single thread, and the second time concurrently process all the file metadata.


@chrislusf I'm not sure if this is a problem at the time of loading.
The base scheme allows you to load in any order, and the display may suffer at the moment, but this is expected
```
create table seaweedfs.filemeta
(
    dirhash   bigint        not null comment 'first 64 bits of MD5 hash value of directory field',
    name      varchar(1000) not null comment 'directory or file name',
    directory text          null comment 'full path to parent directory',
    meta      longblob      null,
    primary key (dirhash, name)
)
    charset = utf8;
```

chrislusf:
for example, this would not work well for Redis.

kmlebedev:
@chrislusf I left one pass before creating folders in one thread, since only folders always go at the beginning of the file.

PR #4536, opened 2023-06-05
fix 404 not found

### LINKED ISSUE #4535: chunk 404 not found
**Describe the bug**

Through the volume.fsck tool scanning, it is found that there is a file lost chunk. And get input/output error when reading a file,  mount reports 404 not found

**System Setup**
1 master, 4 volume server, 1 filer

**Screenshots**
image

**Additional context**

This is the operation of pushing the image, and the intermediate files should be deleted. Mount has an error log (remove failed).

It seems to be that the chunks have been deleted but the metadata is still there.

wusongANKANG:

fix #4535 

First delete the metadata and then delete the chunks.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/filer_delete_entry.go
+++ b/weed/filer/filer_delete_entry.go
@@ -3,6 +3,7 @@ package filer
 import (
 	"context"
 	"fmt"
+
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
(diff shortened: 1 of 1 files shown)


chrislusf:
This PR did not cover when `shouldDeleteChunks && !isDeleteCollection` inside `doBatchDeleteFolderMetaAndData`.

It'll help to create some unit test for this.

wusongANKANG:
@chrislusf  It is difficult to create a unit test for the deleteEntry process. This problem was caused by network instability and other reasons, and we did not have this problem in the same test environment.

This pr only modifies the order in which metadata and chunks data delete, and does not affect other processes.

kmlebedev:
It seems that this is a replacement for an awl for soap. In any case, there will be inconsistency and it is better to focus on minimizing it.

As an option, use an external quorum queue when tasks for asynchronous operations such as deletions will arrive, where you can queue the operation in case of temporary problems within the application

wusongANKANG:
Inconsistencies will definitely exist, but I think we should give priority to ensuring that the metadata we see is correct.  If the metadata exists but the chunks are deleted, what we see on the mount side is  input/output error(data loss).  

Even though this may cause some orphan data, we can clear these orphan data through fsck.

PR #4540, opened 2023-06-05
revert to avoid potential deadlock

### LINKED ISSUE #4522: Strange behaviour / deadlock with seaweedfs 3.51

wusongANKANG:

fix  #4522 

fallback version (delete code that may introduce deadlock)

# How is the PR tested?

# Checks

### DIFF
--- a/weed/topology/node.go
+++ b/weed/topology/node.go
@@ -247,7 +247,6 @@ func (n *NodeImpl) CollectDeadNodeAndFullVolumes(freshThreshHold int64, volumeSi
 	if n.IsRack() {
 		for _, c := range n.Children() {
 			dn := c.(*DataNode) //can not cast n to DataNode
-			dn.RLock()
 			for _, v := range dn.GetVolumes() {
 				if v.Size >= volumeSizeLimit {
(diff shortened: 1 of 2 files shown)


chrislusf:
seems not needed any more after https://github.com/seaweedfs/seaweedfs/pull/4548

wusongANKANG:
@chrislusf  This pr is to roll back the code introduced by #4491, because #4491 may introduce a new deadlock (deadlock also appeared in our test environment).

#4548 is a new solution to fix the problem: no more writables volumes while disk free space is sufficient 
And #4491 does not completely solve this problem and may cause deadlock. So the solution of #4491 should be deprecated, it needs to be reverted

chrislusf:
Direct reverting has some conflicts. Please send a PR to revert it.

wusongANKANG:
@chrislusf  done in #4550

Issue #4532, opened 2023-06-04
Update to using Go-Redis v9


agipson33:
Go-Redis has updated to v9 and is now an official Redis client. The import statements need to be updated to take advantage of the new release and future ones as well.

Old v8: `import "github.com/go-redis/redis/v8"`
New v9: `import "github.com/redis/go-redis/v9"`

The blog post announcing the change can be found at https://redis.com/blog/go-redis-official-redis-client/.

chrislusf:
send a PR?

jessebot:
Could someone attempt to do this again? I unfortunately don't have time right now, but it looks like the last PR was reverted.

jessebot:
Was this done?

PR #4572, opened 2023-06-12
Initial patch for redis version 7 support

agipson33:
https://github.com/seaweedfs/seaweedfs/issues/4532

Adding support for Redis version 7

Updating package imports

# How is the PR tested?

# Checks


chrislusf:
Need to revert due to failing tests.

PR #4402, opened 2023-04-13
Rclone storage backend

dalbani:
# What problem are we solving?

SeaweedFS currently only supports S3 as storage backend for the Cloud tier functionality: https://github.com/seaweedfs/seaweedfs/wiki/Cloud-Tier.  
As a user of SeaweedFS, I'd like to be able to have more choice of storage protocols and providers. 

# How are we solving the problem?

This PR adds support for Rclone, well known for its versatility, as storage backend for Cloud Tier.

# Status / feedback

First, I have very little experience with programming in Go, so feedback is welcome.
Second, there are a couple of areas that probably need work and further improvement:
1. ~~Implement progress for `{download,upload}ViaRclone` functions.~~ ✅
2. ~~Add logging for Rclone related operations~~ ✅
    - Setting the `RCLONE_LOG_LEVEL=DEBUG` environment variable makes the Rclone library be more verbose. 

That being said, I've been using this functionality for weeks and have uploaded several terabytes of data via this Rclone backend (to OneDrive) without any error.

# How is the PR tested?

Manually (so far):
1. Configured a remote in Rclone, e.g. here called `onedrive-remote`.
   ```
   [onedrive-remote]
   type = onedrive
   drive_id = 123456789
   drive_type = personal
   ...
   ```
2. Configured SeaweedFS to use this Rclone remote:
   ```
   [storage.backend]
   [storage.backend.rclone.default]
   enabled = true
   remote_name = "onedrive-remote"
   ```
3. Filled SeaweedFS volumes with data (via the S3 API).
4. Triggered tiering in `weed shell` with:
   ```
   volume.tier.upload -dest rclone
   ```
5. Checked that the files were present in the storage provider (OneDrive in this case) configured in Rclone.
6. Retrieved some files via the S3 API and confirmed that they were consistent with what I originally put in.

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #4605, opened 2023-06-24
Support providing Rclone path instead of remote name

dalbani:

Although the Rclone backend implemented in https://github.com/seaweedfs/seaweedfs/pull/4402 works fine, the current implementation has the limitation of storing all files at the *root* of a Rclone remote.

By exposing a (new) configuration property specifying a Rclone *path* instead of a *remote name*.

# How is the PR tested?

Manually, by configuring such a backend:
```
[storage.backend]
[storage.backend.rclone.default]
enabled = true
rclone_path = "onedrive:/SeaweedFS"
```

And checking that SeaweedFS files were put under `SeadweedFS` in OneDrive.

# Checks

# Other considerations

I have no idea how much this Rclone backend is used in the wild (I have just mentioned in the documentation), so this PR takes the assumption that such a breaking change (due to the new property name) is acceptable.

### DIFF
--- a/weed/storage/backend/rclone_backend/rclone_backend.go
+++ b/weed/storage/backend/rclone_backend/rclone_backend.go
@@ -39,32 +39,31 @@ func (factory *RcloneBackendFactory) BuildStorage(configuration backend.StringPr
 
 type RcloneBackendStorage struct {
 	id         string
-	remoteName string
+	rclonePath string
 	fs         fs.Fs
 }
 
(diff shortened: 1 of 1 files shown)


dalbani:
Given the current drawbacks of this PR (1. introducing a non backward compatible property, and, more importantly, 2. not solving the issue of all the files being put in the _same_ folder), I'm planning to close this PR in favor of https://github.com/seaweedfs/seaweedfs/pull/4606.

PR #4606, opened 2023-06-24
Support templating name of files stored via Rclone backend

dalbani:

Although the Rclone backend implemented in https://github.com/seaweedfs/seaweedfs/pull/4402 works fine, the current implementation has the limitation of storing all files at the *root* of a Rclone remote, next to each other in the same "folder".
In a SeaweedFS deployment like mine with, hmm, tens of thousands of volume files, I'd like to have a way to split all those files into subfolders.

By introducing a new configuration property that lets the user provide a template generating the filename from the volume UUID.

# How is the PR tested?

Manually, by configuring such a backend:
```
[storage.backend]
[storage.backend.rclone.default]
enabled = true
remote_name = "onedrive"
key_template = "SeaweedFS/{{ slice . 0 2 }}/{{ slice . 2 4 }}/{{ slice . 4 }}"
```

And checking that volume files were uploaded in OneDrvive in such a structure:
```
SeaweedFS/
  5f/
    e5/
      03d8-0aba-4e42-b1c8-6116bf862f71
```

Which matches the output of `volume.list`:
```
...
volume id:8  
  size:106774344
  collection:"bucket-1"
  ...
  read_only:true
  version:3
  modified_at_second:1687612319
  remote_storage_name:"rclone.default"
  remote_storage_key:"SeaweedFS/5f/e5/03d8-0aba-4e42-b1c8-6116bf862f71" 
...
```

# Checks

# Other considerations

If this template functionality is deemed useful for the Rclone backend, I propose to implement it in a generic fashion, so that the S3 backend can benefit from it as well.

### DIFF
--- a/weed/storage/backend/rclone_backend/rclone_backend.go
+++ b/weed/storage/backend/rclone_backend/rclone_backend.go
@@ -1,12 +1,14 @@
 package rclone_backend
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"github.com/rclone/rclone/fs/config/configfile"
 	"github.com/seaweedfs/seaweedfs/weed/util"
 	"io"
 	"os"
+	"text/template"
 	"time"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


dalbani:
Thanks for this quick reaction.

FYI, I have just documented this new functionality in https://github.com/seaweedfs/seaweedfs/wiki/Cloud-Tier#rclone.

PR #4622, opened 2023-06-28
add S3ChunkSize 

mervynzhang:
It takes 5 minutes to upload a 137 mb file to backend swift with default 8 mb part size
image

With a bigger part size like 500mb, it take only 4 or 5 secs. 
image

Add a option to allow a bigger part size

# How is the PR tested?
Build locally and run upload to verify


chrislusf:
this is a work around but not a good fix.

mervynzhang:
it's not bad to provide an option to use different part size, 64 MB is used in s3_upload anyway. 
https://github.com/seaweedfs/seaweedfs/blob/a315490f7d4d2aa8b5226ea4029cb112edb88137/weed/storage/backend/s3_backend/s3_upload.go#L30C33-L30C33

PR #4651, opened 2023-07-10
avoid lock being preempted midway

wusongANKANG:
avoid lock being preempted midway

# How is the PR tested?

# Checks


chrislusf:
What problem does this PR solve?

wusongANKANG:
@chrislusf  no related PR.  Just saw this code and thought the lock should only be added once.

PR #4664, opened 2023-07-11
Fix data loss: add lock for metacache

### LINKED ISSUE #4648: Append write data loss
**Describe the bug**

When appending to a file, sometimes the written data may be deleted by mistake.

It's very easy to reproduce.

1: Append  write file

```
import os

os.system("rm a.txt")
file = 'a.txt'

for i in range(5000):
    with open(file, 'a+') as f:
        f.write("a\n")
```

2: do "ls" 

```
watch -n 1 ls
```

From the filer side, we can see that some data has been deleted.

image

On the Mount side, we can see that the data is deleted because the offset of the file obtained by two  append operations is  same.
image

**System Setup**

seaweedfs 3.53

wusongANKANG:

Fix #4648 

Add lock on meta cache

# How is the PR tested?

# Checks

### DIFF
--- a/weed/mount/meta_cache/meta_cache.go
+++ b/weed/mount/meta_cache/meta_cache.go
@@ -3,6 +3,7 @@ package meta_cache
 import (
 	"context"
 	"os"
+	"sync"
 
 	"github.com/seaweedfs/seaweedfs/weed/filer"
 	"github.com/seaweedfs/seaweedfs/weed/filer/leveldb"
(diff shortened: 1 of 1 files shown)


chrislusf:
What was the reason the lock was commented out?

wusongANKANG:
@chrislusf  When solving the problem, I found that the entry obtained by “ls” was the old entry when the data was deleted by mistake.

So , I add lock on metacache, and no data will be deleted by mistake after locks added.


And I see that you commented out this lock at the beginning. I wonder if there will be other problems with locking? Or is it based on performance considerations so no locks are added?

PR #4675, opened 2023-07-17
retry all errors

Numblgw:

The leader switch process may cause request failures.

Retry all errors.
The reason for the issue is that not all types of errors were retried when accessing the master node, only network issues were retried. After the election ends and a new leader appears, the topology information is not yet fully constructed, so the leader node is still in an unavailable state. At this time, the returned error should be retried.

# How is the PR tested?
Use S3 stress testing tool to conduct stress testing and stop the Leader process during the testing process to observe whether there are any request errors during the leader switch.

# Checks

### DIFF
--- a/weed/util/retry.go
+++ b/weed/util/retry.go
@@ -24,8 +24,6 @@ func Retry(name string, job func() error) (err error) {
 		if strings.Contains(err.Error(), "transport") {
 			hasErr = true
 			glog.V(0).Infof("retry %s: err: %v", name, err)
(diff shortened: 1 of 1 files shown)


Numblgw:
Fix the legacy issues of this commit.
https://github.com/seaweedfs/seaweedfs/commit/f7e0a65e752f08566c829ce67c4aa8ed1d6dd4f6

chrislusf:
Not all errors are retry-able.

Issue #3352, opened 2022-07-23
Filer cannot read content if using JWT auth


creeew:
**Describe the bug**
Using Filer HTTP cannot get the filer content when configured "jwt.signing.read" and "jwt.filer_signing.read" in security.toml.

We can get the filer's file list if we know the jwt.filer_signing.read key, but cannot get the file's content because we don't generate jwt token when we read file content by fid in volume server.

In this get file by fid code, we do not generate jwt token event we pass the filer server's jwt token authentication.
https://github.com/chrislusf/seaweedfs/blob/2c8818351f418e3584a6c5410c396f747aebd725/weed/util/http_util.go#L290
But in read s3 file code, it can generate jwt token by filer_signing.read key
https://github.com/chrislusf/seaweedfs/blob/2c8818351f418e3584a6c5410c396f747aebd725/weed/util/http_util.go#L372

I found a note in security.toml jwt.signing.read comment: 
NOTE: jwt for read is only supported with master+volume setup. Filer does not support this mode.
I don't know it's for more security or something. I thought we can get the file content if we pass the filer server's token authentication.

Further more, weed mount cannot work if we configured the "jwt.signing.read" even mount can get the read key. Maybe in mount code, we can read the security.toml and generate jwt token for every chunk? Maybe casue the performance issue if a file contain many chunks.

**System Setup**
- weed version 2.88

**Expected behavior**
We can get the file if we passed the filer's authentication.

zemul:
The volume service should not verify read sign

zemul:
It's also mentioned here.
https://github.com/seaweedfs/seaweedfs/wiki/Security-Overview

hencoappel:
I've created a pull request for this. Any chance it could be merged?

hencoappel:
I'm happy for this to now be closed. @creeew this still an issue?

PR #4694, opened 2023-07-19
Remove the auth check for volume server reads

hencoappel:
According to the documentation this shouldn't be authorized. It was breaking reads from the filer.

Issue [#3352
](https://github.com/seaweedfs/seaweedfs/issues/3352)
The auth check broke the filer being able to read files from the volume server if jwt.signing.key was set in security.toml.

Removing the auth check on reads.

### DIFF
--- a/weed/server/volume_server_handlers_read.go
+++ b/weed/server/volume_server_handlers_read.go
@@ -3,7 +3,6 @@ package weed_server
 import (
 	"bytes"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"mime"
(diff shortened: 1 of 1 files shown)


n9:
I read in the documentation:
> The **volume server can also check JWT for reads**. This mode does not work with weed filer. But this could be useful if the volume server is exposed to public and you do not want anyone to access it with a URL, e.g., paid content.

(source: https://github.com/seaweedfs/seaweedfs/wiki/Security-Overview#jwt-for-read-access-control)

n9:
A solution could be to compute JWT for reads in filer.

PR #4778, opened 2023-08-22
Fix s3 object list 

zemul:
Only the directories under the bucket can be listed, and the subdirectories cannot be listed

Path ending in "/" are considered subpath rather than prefix.

# How is the PR tested?

screenshot:
Level 1 sub directory of the bucket (Consistent result)
image
image

Level 2 sub directory of the bucket (Inconsistent result)
image

image

# Checks

### DIFF
--- a/weed/s3api/s3api_objects_list_handlers.go
+++ b/weed/s3api/s3api_objects_list_handlers.go
@@ -139,8 +139,8 @@ func (s3a *S3ApiServer) listFilerEntries(bucket string, originalPrefix string, m
 	var doErr error
 	var nextMarker string
 	cursor := &ListingCursor{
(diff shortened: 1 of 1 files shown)


zemul:
I don't understand the purpose of `prefixEndsOnDelimiter`, but it affect the normal listing of directory.

chrislusf:
@zemul need to add new test and fix the test.

lfhy:
image
我之前发现二级目录查看不了是因为部分filer store没有实现```ListDirectoryPrefixedEntries```这个方法导致的，像默认leveldb store就实现了所以可以查看到
image

zemul:
resolved #4834

PR #4834, opened 2023-09-12
Fix s3 api object list with params max-keys and prefix

annluda:

redis as filer store, suppose bucket have objects:

```text
demo_v1
demo_v10
demo_v11
demo_v12
demo_v2
demo_v20
```

when request s3api list objects with params max-keys and prefix, like `http://localhost:8333/test?max-keys=2&prefix=demo_v2`, we got 0 result. Because the item's rank in redis is bigger than the max-keys.

drop `isLastItemHasPrefix`, compare the prefix and lastFileName directly.

# How is the PR tested?

# Checks


chrislusf:
Thanks for the fix! This part of code is tricky. Do you want to add some tests for this?

PR #4855, opened 2023-09-19
Wrong default value for volume.ressource

armandleopold:

https://github.com/seaweedfs/seaweedfs/blob/0bb97709d41b1be4c74f01dcc65aac6d5f88bd16/k8s/charts/seaweedfs/values.yaml#L283

Getting  : 

aleopold@aleopold:~/desktop/helm/charts/qalita$ helm lint -f values.yaml .
coalesce.go:220: warning: cannot overwrite table with non table for qalita.seaweedfs.volume.resources (map[])
coalesce.go:220: warning: cannot overwrite table with non table for qalita.seaweedfs.volume.resources (map[])
coalesce.go:220: warning: cannot overwrite table with non table for qalita.seaweedfs.volume.resources (map[])
coalesce.go:220: warning: cannot overwrite table with non table for qalita.seaweedfs.volume.resources (map[])

when values are : 

```yaml
seaweedfs:
  volume:
    resources: |
      requests:
        memory: "100Mi"
        cpu: "50m"
...
```

Please replace to : 

```yaml
resources: null
```

# How is the PR tested?

Github Actions and helm lint

# Checks

### DIFF
--- a/k8s/charts/seaweedfs/Chart.yaml
+++ b/k8s/charts/seaweedfs/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 description: SeaweedFS
 name: seaweedfs
 appVersion: "3.56"
-version: "3.56"
+version: "3.57"
--- a/k8s/charts/seaweedfs/values.yaml
+++ b/k8s/charts/seaweedfs/values.yaml
@@ -280,7 +280,7 @@ volume:
   # should map directly to the value of the resources field for a PodSpec,
   # formatted as a multi-line string. By default no direct resource request
   # is made.
-  resources: {}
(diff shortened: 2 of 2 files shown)


k8s/charts/seaweedfs/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 description: SeaweedFS
 name: seaweedfs
 appVersion: "3.56"
-version: "3.56"
+version: "3.57"
```
chrislusf, line 5:
```suggestion
version: "3.56"
```
Need to keep it in sync with the binary release version.

Issue #4835, opened 2023-09-12
Support https/tls for weed filer/mount


isjo-org:
Hi there!

We are already using seaweedfs with all combinations(5 volume server, 3 master, 3 filer, mount/webdav/s3) in our dev.env, so far, so happy!

Now we want to go to production with master/volume/filer/s3/mount without any private network between those servers, so only public ip + self signed cert

**Is there any progess for supporting TLS?**
Im wondering, if we are the only ones with this problem, as for now i didnt see any tickets for this problem?

for sure, i could use JWT to trust client/server, but content is still unencrypted on read/write :/

```
# volume server https options
# Note: work in progress!
#     this does not work with other clients, e.g., "weed filer|mount" etc, yet.
[https.client]
enabled = true
[https.volume]
cert = ""
key = ""
```

TIA!

br

kmlebedev:
Hello, yes, encrypting traffic on volumes seems redundant, so there are few such configurations.

NiceRath:
Greetings!

We would like to use SeaweedFS on distributed servers that communicate over public WAN.
So enrypting any kind of traffic is necessary!

Greetings
- Rath

kmlebedev:
I just launched volume in tls mode, no problems with that.

NiceRath:
Greetings!

My college @isjo-org may be able to provide the exact error messages next week, but the problems ocurr when actually using it. (and blocking any traffic to non-tls ports)
We currently are trying to get it working by replacing all the hardcoded references to 'http://' with a config-switch.
(to see references - just run `cd  seaweedfs/weed && grep -r '\"http://'`)

NiceRath:
@kmlebedev Greetings!

May I ask if you have found the same issues as we have, while testing?

We have also sniffed the traffic using `tcpdump` and have found plaintext HTTP calls, even if TLS is enabled by config.

In regards to our 'PR' - we just wanted to show how we would approach the issue. But it is far from being clean, as we do not have an overview of the codebase..

oivindoh:
> I just launched volume in tls mode, no problems with that.

Yes, it will listen with https, but e.g. weed mount is hardcoded to use http, which will error out since it'll be trying to make plaintext calls to a https service. I'm not sure @chrislusf is aware of this state at the moment - we might try to submit a new PR on this, but so far not familiar enough with the code base to make an honest attempt

chrislusf:
[...]

PR #4874, opened 2023-09-26
Support https/tls for weed filer/mount

isjo-org:

We trie to implement https/tls for weed filer/mount

See issue here #https://github.com/seaweedfs/seaweedfs/issues/4835

For quicktesting -> single file (few MB) seems to work with write and reads. (no leftover http/1 connections)
If we test some bigger files (few 100MB), it fails with error, seems like an chunk problem?

`fatal error: concurrent map iteration and map write`  (there is also a very long stack trace if needed)

As additional info -> PR is quick and dirty, and not prepared for master, only for testing purpose.


chrislusf:
this PR is superseded by https://github.com/seaweedfs/seaweedfs/pull/5766

Issue #4647, opened 2023-07-07
[shell] volume 1415 replication 100, but over replicated +3 (critical)


kmlebedev:
Delete extra replicas, taking into account the replication factor, so that the replicas remain in different DCs.

```
I0707 05:13:12.448756 master_grpc_server.go:199 master see new volume 1415 from fast-volume-0.s3-fast-volume.service.dcvcs.consul:8080
volume 1415 replication 100, but over replicated +3
deleting volume 1415 from fast-volume-4.s3-fast-volume.service.dcix.consul:8080 ...
volume 1415 replication 100 is not well placed [0xc001bed030 0xc001beed00]
deleting volume 1415 from fast-volume-2.s3-fast-volume.service.dcvcs.consul:8080 ...
```

It is also necessary to handle such a case and, if the number of replicas is temporarily exceeded, write files to all.

PR #4875, opened 2023-09-26
fix: avoid deleting one replica without sync

kmlebedev:

With a larger number of actual replicas, the write occurs randomly
https://github.com/seaweedfs/seaweedfs/issues/4647

so it is necessary to synchronize with all replicas before deleting.

# How is the PR tested?

# Checks
- [yes] I have added unit tests if possible.
- [yes] I will add related wiki document changes and link to this PR after merging.

https://github.com/seaweedfs/seaweedfs/wiki/Replication/_compare/33814e2a4d53200170dcefb2d8634d568833ac42...7fefeb6d08bd7253355c59ee8ea13a582999694a


weed/shell/command_volume_fix_replication.go
```diff
@@ -58,6 +60,7 @@ func (c *commandVolumeFixReplication) Do(args []string, commandEnv *CommandEnv,
 	c.collectionPattern = volFixReplicationCommand.String("collectionPattern", "", "match with wildcard characters '*' and '?'")
 	skipChange := volFixReplicationCommand.Bool("n", false, "skip the changes")
 	noDelete := volFixReplicationCommand.Bool("noDelete", false, "Do not delete over-replicated volumes, only fix under-replication")
```
chrislusf: let's revert this existing option to positive, and then fix other places referencing this variable
```suggestion
	doDelete := volFixReplicationCommand.Bool("doDelete", true, "Also delete over-replicated volumes besides fixing under-replication")
```

kmlebedev:
@bddrey Hi, in this PR, backward compatibility with the `noDelete` option will be broken, it will need to be replaced with `doDelete=false` after release

PR #4881, opened 2023-10-02
 fix: disallow file name too long when writing a file

kmlebedev:

Long file names can cause a lot of problems in different places, so we disable them by default. But in the case where it is justified, it can be resolved using path-based co-ordination

Add LongerFileName option to fs.configure

# How is the PR tested?

# Checks


chrislusf:
The definition of "long" is vague. How about file name limit default to 1024, and can be customized to any desired value for any path?

kmlebedev:
> The definition of "long" is vague. How about file name limit default to 1024, and can be customized to any desired value for any path?

ok

kmlebedev:
A Google search reveals:
> Object key names must be up to 1024 characters long. The prefix / of an object is interpreted as a directory, and such directory segments are limited to 255 characters. See AWS S3 object name limitations.

The problem here is that 1024 characters do not equal 1024 bytes. See https://www.golangprograms.com/golang-get-number-of-bytes-and-runes-in-a-string.html

> In Go Strings are UTF-8 encoded, this means each charcter called rune can be of 1 to 4 bytes long. Here,the charcter ♥ is taking 3 bytes, hence the total length of string is 7.

@chrislusf  Accordingly, this is not a limitation on the length of the s3 key in runes, but rather the length of the file name in bytes is no more than 255 bytes for compatibility with Unix file systems. Which could be the equivalent of 63 to 255 characters.

kmlebedev:
@chrislusf Hi, what else can I do in this PR ?

chrislusf:
> A Google search reveals:
> 
> > Object key names must be up to 1024 characters long. The prefix / of an object is interpreted as a directory, and such directory segments are limited to 255 characters. See AWS S3 object name limitations.
> 
> The problem here is that 1024 characters do not equal 1024 bytes. See https://www.golangprograms.com/golang-get-number-of-bytes-and-runes-in-a-string.html
> 
> > In Go Strings are UTF-8 encoded, this means each charcter called rune can be of 1 to 4 bytes long. Here,the charcter ♥ is taking 3 bytes, hence the total length of string is 7.
> 
> @chrislusf Accordingly, this is not a limitation on the length of the s3 key in runes, but rather the length of the file name in bytes is no more than 255 bytes for compatibility with Unix file systems. Which could be the equivalent of 63 to 255 characters.

So we need to make the file length limit configurable, right? Just "long" is not enough. We can adjust our default limit to 255.

PR #4907, opened 2023-10-10
add fs.compact command into shell

yeganemehr:
Half-full volumes can easily cause volume-servers ran out of volumes although they may have plenty of free disk space.

As it already suggested in [this discussion](https://github.com/seaweedfs/seaweedfs/discussions/2647) if We could go throughout all files in filer and move all chunks to another volume and update they `Fid` in the filer store we can empty those volumes.

My solution heavily inspired by `fs.meta.changeVolumeId`.
We traverse on all files & directories and move they chunks to another volume.
In this action, chunks keep their `FileKey` and `Cookie` and just `VolumeId` would update.

# How is the PR tested?
I tested locally and on my own cluster.

# Checks

### DIFF
--- a/weed/shell/command_fs_compact.go
+++ b/weed/shell/command_fs_compact.go
@@ -1,0 +1,270 @@
+package shell
+
+import (
+	"context"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+
+	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
+	"github.com/seaweedfs/seaweedfs/weed/wdclient"
+
+	"github.com/seaweedfs/seaweedfs/weed/operation"
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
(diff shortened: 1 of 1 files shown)


weed/shell/command_fs_compact.go
```diff
@@ -0,0 +1,270 @@
...
+	client = &http.Client{}
+	Commands = append(Commands, &commandFsCompact{})
+}
+
+type commandFsCompact struct {
+	volumes map[needle.VolumeId]*master_pb.VolumeInformationMessage
+}
+
+func (c *commandFsCompact) Name() string {
+	return "fs.compact"
```
chrislusf: The word "compact" is already used to compact volumes when doing garbage collection.

how about "fs.mergeVolumes"?

yeganemehr: Ok I did rename it

kmlebedev:
@yeganemehr Thank you very much, the need for this has been long overdue, but I haven’t gotten around to implementing it.

yeganemehr:
@kmlebedev
Thanks to you and @chrislusf for you wonderful work in this project.

PR #4922, opened 2023-10-18
avoid error: no writable volumes available

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/3886

Avoid empty response for grpc Assign

# How is the PR tested?

# Checks

### DIFF
--- a/weed/topology/volume_layout.go
+++ b/weed/topology/volume_layout.go
@@ -314,7 +314,7 @@ func (vl *VolumeLayout) PickForWrite(count uint64, option *VolumeGrowOption) (*n
 				continue
 			}
 			counter++
-			if rand.Intn(counter) < 1 {
+			if rand.Intn(counter) < 1 || locationList == nil {
 				vid, locationList = v, volumeLocationList.Copy()
 			}
 		}


chrislusf:
the if rand.Intn(counter) < 1 is reservoir sampling, which will randomly pick at least one volume id.

Removing it would only pick the last one.

kmlebedev:
> the if rand.Intn(counter) < 1 is reservoir sampling, which will randomly pick at least one volume id.
> 
> Removing it would only pick the last one.

@chrislusf Look carefully, I didn't delete the random one.
I figured it would be a good option to leave the sample random and avoid returning an empty value.

chrislusf:
when n == 1, the assignment will always happen.

see https://pkg.go.dev/math/rand#Intn

kmlebedev:
> when n == 1, the assignment will always happen.
> 
> see https://pkg.go.dev/math/rand#Intn

Here's something I'm not sure about
See
https://go.dev/play/p/cunhkO1Nh_k
image
image

chrislusf:
`rand.Intn(1) < 1` should always be true

PR #4930, opened 2023-10-20
see_dat add print name 

binbinshi:
see_dat print  file  name 

add name in visitneedle 

# How is the PR tested?
see_dat that has name file 

# Checks


chrislusf (changes requested):
Print the name at the tail since the name length varies a lot.

PR #4945, opened 2023-10-24
avoid delete collection on fs.mv

kmlebedev:

```
> fs.mv /buckets/test /test
move: /buckets/test => /test
error: rpc error: code = Unknown desc = /buckets/test move error: fail to move /buckets/test => /test: delete file /buckets/test: filer store delete: delete /buckets/test: Error 1213 (40001): Deadlock found when trying to get lock; try restarting transaction
```

skip delete collection

# How is the PR tested?

localy

# Checks

### DIFF
--- a/weed/filer/filer_delete_entry.go
+++ b/weed/filer/filer_delete_entry.go
@@ -51,7 +51,9 @@ func (f *Filer) DeleteEntryMetaAndData(ctx context.Context, p util.FullPath, isR
 
 	if shouldDeleteChunks {
 		if isDeleteCollection {
-			f.doDeleteCollection(entry.Name())
+			if err := f.doDeleteCollection(entry.Name()); err != nil {
+				glog.Infof("failed delete collection %s: %v", entry.Name(), err)
+			}
 		} else {
 			f.DirectDeleteChunks(entry.GetChunks())
 		}
--- a/weed/filer/filer_delete_entry.go
+++ b/weed/filer/filer_delete_entry.go
@@ -49,13 +49,12 @@ func (f *Filer) DeleteEntryMetaAndData(ctx context.Context, p util.FullPath, isR
 		return fmt.Errorf("delete file %s: %v", p, err)
 	}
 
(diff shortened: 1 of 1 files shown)


weed/filer/filer_delete_entry.go
```diff
@@ -49,13 +49,12 @@ func (f *Filer) DeleteEntryMetaAndData(ctx context.Context, p util.FullPath, isR
...
-	}
-
-	if isDeleteCollection {
-		collectionName := entry.Name()
-		f.doDeleteCollection(collectionName)
+	if shouldDeleteChunks {
+		if isDeleteCollection {
+			f.doDeleteCollection(entry.Name())
+		} else {
+			f.DirectDeleteChunks(entry.GetChunks())
```
chrislusf, line 56: this would be deleting chunks for files deep under the bucket sub-directories, defeating the purpose of fast deletion when removing a collection.

weed/filer/filer_delete_entry.go
```diff
@@ -49,13 +49,12 @@ func (f *Filer) DeleteEntryMetaAndData(ctx context.Context, p util.FullPath, isR
...
 
-	if shouldDeleteChunks && !isDeleteCollection {
-		f.DirectDeleteChunks(entry.GetChunks())
-	}
-
-	if isDeleteCollection {
-		collectionName := entry.Name()
-		f.doDeleteCollection(collectionName)
+	if shouldDeleteChunks {
+		if isDeleteCollection {
```
chrislusf, line 53: This piece of logic is tricky. A unit test would be really nice.

kmlebedev: > This piece of logic is tricky. A unit test would be really nice.

I saw that in order to do a unit test you need to initialize the Filer structure, which is not a completely trivial task.

Here the logic is simple: shouldDeleteChunks is a synonym for fs.mv(move), so chunks do not need to be deleted, including the entire collection.

PR #4956, opened 2023-10-26
Improve the performance of prefix list by add a lower limit

lakeland1990:

after remove compare in pr #4924 , the loop will fetch some objects based on the limit parameter 
When the limit parameter is set too small and the number of objects is too large, it can cause performance issues.

The limit parameter only focuses on the number of results and should not affect the number of entries obtained each time. Modify the List operation of prefixFilterEntries to get a fixed number (500) of entries each time if the limit < 500 to improve performance. 

# How is the PR tested?

all test passed

# Checks

### DIFF
--- a/weed/filer/filerstore_wrapper.go
+++ b/weed/filer/filerstore_wrapper.go
@@ -281,8 +281,16 @@ func (fsw *FilerStoreWrapper) prefixFilterEntries(ctx context.Context, dirPath u
 		return actualStore.ListDirectoryEntries(ctx, dirPath, startFileName, includeStartFile, limit, eachEntryFunc)
 	}
 
+	// after remove compare in pr #4924 , the loop will fetch some objects based on the limit parameter
+	// When the limit parameter is set too small and the number of objects is too large, it can cause performance issues.
+	// so it will at least fetch 500 objects a time if limit < 500
+	fetchLimit := int64(500)
+	if limit > fetchLimit {
+		fetchLimit = limit
+	}
+
 	var notPrefixed []*Entry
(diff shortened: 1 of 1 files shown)


weed/filer/filerstore_wrapper.go
```diff
@@ -281,8 +281,16 @@ func (fsw *FilerStoreWrapper) prefixFilterEntries(ctx context.Context, dirPath u
...
+	// When the limit parameter is set too small and the number of objects is too large, it can cause performance issues.
+	// so it will at least fetch 500 objects a time if limit < 500
+	fetchLimit := int64(500)
+	if limit > fetchLimit {
+		fetchLimit = limit
+	}
+
 	var notPrefixed []*Entry
-	lastFileName, err = actualStore.ListDirectoryEntries(ctx, dirPath, startFileName, includeStartFile, limit, func(entry *Entry) bool {
+	lastFileName, err = actualStore.ListDirectoryEntries(ctx, dirPath, startFileName, includeStartFile, fetchLimit, func(entry *Entry) bool {
```
chrislusf, line 293: This will slow down the queries with small limit values.

lakeland1990: > This will slow down the queries with small limit values.

First of all, this is a fall back mechanism that does not affect the filerstore that has already implemented the prefixlist. If there is 10000 object in a directory and the limit is set to 1, there will be 10,000 times of queries in the original case. Each query has a limit of 1, and loop 10,000 times. In this case, if we change to fetch limit 500, it will only take 20 times.

PR #4960, opened 2023-10-30
Add ErrorGetNotFound and ErrorGetInternal to volume server metrics

SmsS4:
Add not found and internal errors as metrics to `SeaweedFS_volumeServer_request_total`  so we can find out if something is wrong with volume server (data loss for example)

I used these two functions in every internal and 404 responses
```go
func NotFound(w http.ResponseWriter) {
	stats.VolumeServerRequestCounter.WithLabelValues(stats.ErrorGetNotFound).Inc()
}

func InternalError(w http.ResponseWriter) {
	stats.VolumeServerRequestCounter.WithLabelValues(stats.ErrorGetInternal).Inc()
	w.WriteHeader(http.StatusInternalServerError)
}
```
for example:
```go
	if err != nil || count < 0 {
		glog.V(3).Infof("read %s isNormalVolume %v error: %v", r.URL.Path, hasVolume, err)
		if err == storage.ErrorNotFound || err == storage.ErrorDeleted {
			NotFound(w)
		} else {
			InternalError(w)
		}
		return
	}
```
Is there a better way to handle this?
I tried to add status code as label but I couldn't do it

# How is the PR tested?
I tested on my local but I couldn't coverage all return statements (for example when cookie is wrong)

# Checks

### DIFF
--- a/weed/server/volume_server_handlers_read.go
+++ b/weed/server/volume_server_handlers_read.go
@@ -21,12 +21,23 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/images"
 	"github.com/seaweedfs/seaweedfs/weed/operation"
+	"github.com/seaweedfs/seaweedfs/weed/stats"
 	"github.com/seaweedfs/seaweedfs/weed/storage"
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 	"github.com/seaweedfs/seaweedfs/weed/util"
 )
 
 var fileNameEscaper = strings.NewReplacer(`\`, `\\`, `"`, `\"`)
+
+func NotFound(w http.ResponseWriter) {
+	stats.VolumeServerRequestCounter.WithLabelValues(stats.ErrorGetNotFound).Inc()
+	w.WriteHeader(http.StatusNotFound)
+}
+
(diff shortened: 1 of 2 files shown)


weed/server/volume_server_handlers_read.go
```diff
@@ -147,15 +158,16 @@ func (vs *VolumeServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request)
...
 		} else {
-			w.WriteHeader(http.StatusInternalServerError)
+			InternalError(w)
 		}
 		return
 	}
 	if n.Cookie != cookie {
 		glog.V(0).Infof("request %s with cookie:%x expected:%x from %s agent %s", r.URL.Path, cookie, n.Cookie, r.RemoteAddr, r.UserAgent())
-		w.WriteHeader(http.StatusNotFound)
+		stats.VolumeServerRequestCounter.WithLabelValues(stats.ErrorGetNotFound).Inc()
```
kmlebedev, line 169: Look like twice increase metric counter

SmsS4: Thanks
I fixed it here https://github.com/seaweedfs/seaweedfs/pull/4961

LuuucasChen:
Is this Java client for filer effective ? (filer client communicates with Seaweed services via gRPC)

Issue #4944, opened 2023-10-24
Superblock and ".vif" files are not in sync


zehweh:
As discovered in #4934 :

It looks like in order to determine the replication setting,  sometimes the superblock is used and sometimes the `.vif` 

For example, when a new volume file gets created, the replication setting and the TTL is stored in the superblock only.
The `.vif` file looks like this:
```
{
  "files": [],
  "version": 3,
  "replication": "",
  "BytesOffset": 4
}
```

When I manually change the replication setting using `volume.configure.replication -collectionPattern * -replication 010`, the `.vif` file is updated but the superblock is not. The volume server now uses the `.vif`:

```
{
  "files": [],
  "version": 3,
  "replication": "010",
  "BytesOffset": 4
}
```

If I delete the `.vif` and restart the volume server, the `.vif` gets recreated without the `replication` setting and the volume server uses the outdated setting that it gets from the superblock.

**System Setup**
- Ubuntu 22.04
- output of `weed version`: version 30GB 3.57 0f8168c0c928bba3d2f48b0680d3bdce9c617559 linux amd64

**Expected behavior**
`volume.configure.replication` changes both superblock and `.vif` files

ljluestc:
​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​`​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​`​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​`​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​​
#!/bin/bash

# Script to update replication settings in SeaweedFS for both .vif and superblock
# Requirements: SeaweedFS v3.57 installed, weed command available
# Usage: ./update_replication.sh <collection_pattern> <new_replication>

# Configuration
MASTER_ADDRESS="localhost:9333"  # Replace with your master server address
VOLUME_SERVER_DIR="/path/to/volume/data"  # Replace with your volume server data directory
WEED_SHELL="weed shell"  # Path to weed shell command
NEW_REPLICATION="$2"  # e.g., "010"
COLLECTION_PATTERN="$1"  # e.g., "*" for all collections

# Check arguments
if [ -z "$COLLECTION_PATTERN" ] || [ -z "$NEW_REPLICATION" ]; then
[...]

PR #4975, opened 2023-11-02
Update superblock when changing replication

zehweh:
.vif file and superblock are not in sync, see issue #4944

With this PR, the superblock gets updated when the replication is changed using `weed shell`

It is basically the code from `seaweedfs/unmaintained/change_superblock/change_superblock.go`

I'm not sure if ConfigureVolume() is the best place to do this or if VolumeConfigure() in `volume_grpc_admin.go` if better.
So feel free to make changes :)

### DIFF
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -3,6 +3,7 @@ package storage
 import (
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"strings"
 	"sync"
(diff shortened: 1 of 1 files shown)


weed/storage/store.go
```diff
@@ -557,6 +559,25 @@ func (s *Store) ConfigureVolume(i needle.VolumeId, replication string) error {
 		if err != nil {
 			return fmt.Errorf("volume %d failed to save vif: %v", i, err)
 		}
+                // update superblock
```
chrislusf, line 562: please refactor this to a function, and just print a warning instead of failing in the middle.

zehweh: Ok, refactoring this into a function makes sense.
But I don't see why printing a warning here makes sense. If a problem occurs while loading or saving the ".vif", it fails with an error, so updating the superblock should be done the same way?

chrislusf: there should be only one source of truth, which is the `.vif` file.

PR #4992, opened 2023-11-09
chore(tests): update S3 compat tests

t3hmrman:

While present, the S3 compatibility tests were broken (at the very least when running on Linux) due to a few issues:

- `ubuntu:latest` moving ahead of Python 3.6
- Docker networking not working as expected (host.docker.internal)

This commit fixes the s3 compatibility tests, ensuring they run properly on linux, and updates the repository to contain some results.

Updating the machinery around running the tests

# How is the PR tested?

Pull the repository and attempt to run the tests. The previous instructions *should* work almost unchanged, on MacOS.

# Checks


t3hmrman:
Ah, I just realized that the s3 tests have actually moved to be [completely dockerized w/ compose](https://github.com/seaweedfs/seaweedfs/blob/master/docker/Makefile#L87). I can't tell which test is

That script doesn't work for me (`./bootstrap` is missing from more recent versions of `ceph/s3-tests`), but that code *looks* more recent than the code that was in `test/s3/compatibility`... Should I hollow out this directory and point at the `make` target @ `docker/Makefile`?

I can try to fix that test run as well (at least, pin it to a version of `ceph/s3-tests` that works).

PR #4999, opened 2023-11-13
fs.mergeVolumes: Make a plan based on volumes size

yeganemehr:
I had 500 half-full volumes on my cluster, this means 15TB of wasted disk, It was hard and time-consuming to merge volumes individually.

So i wrote a function to create a map (named `plan`) based on current status of volumes and decide how volumes must merge.
This function respect replication policy, volume TTL, collection, read-only volumes and max volumes size limit.

Now it's possible to
+ Merge volumes of a single collection
+ Merge only with specified destination volume (`-toVolumeId`)
+ Merge only with specified source volume (`-fromVolumeId`)

# How is the PR tested?
I run this locally and on my cluster.

### DIFF
--- a/weed/shell/command_fs_merge_volumes.go
+++ b/weed/shell/command_fs_merge_volumes.go
@@ -7,10 +7,13 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"sort"
 	"strings"
 
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 	"github.com/seaweedfs/seaweedfs/weed/wdclient"
+	"golang.org/x/exp/maps"
+	"golang.org/x/exp/slices"
 
 	"github.com/seaweedfs/seaweedfs/weed/operation"
 	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
@@ -28,7 +31,8 @@ func init() {
 }
 
 type commandFsMergeVolumes struct {
-	volumes map[needle.VolumeId]*master_pb.VolumeInformationMessage
+	volumes         map[needle.VolumeId]*master_pb.VolumeInformationMessage
+	volumeSizeLimit uint64
 }
 
 func (c *commandFsMergeVolumes) Name() string {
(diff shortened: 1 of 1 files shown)


weed/shell/command_fs_merge_volumes.go
```diff
@@ -40,7 +44,7 @@ func (c *commandFsMergeVolumes) Help() string {
 	
 	This would help clear half-full volumes and let vacuum system to delete them later.
 
-	fs.mergeVolumes -toVolumeId=y [-fromVolumeId=x] [-apply] /dir/
+	fs.mergeVolumes [-toVolumeId=y] [-fromVolumeId=x] [-collection="*"] [-apply] [/dir/]
```
chrislusf, line 47: How about changing the directory into a named parameter also, in a separate PR?

weed/shell/command_fs_merge_volumes.go
```diff
@@ -192,6 +195,94 @@ func (c *commandFsMergeVolumes) reloadVolumesInfo(masterClient *wdclient.MasterC
...
+}
+
+func (c *commandFsMergeVolumes) getVolumeSizeById(vid needle.VolumeId) uint64 {
+	return c.getVolumeSize(c.volumes[vid])
+}
+
+func (c *commandFsMergeVolumes) printPlan(plan map[needle.VolumeId]needle.VolumeId) {
+	fmt.Printf("max volume size: %d MB\n", c.volumeSizeLimit/1024/1024)
+	reversePlan := make(map[needle.VolumeId][]needle.VolumeId)
+	for src, dist := range plan {
```
chrislusf: dist => dest?

weed/shell/command_fs_merge_volumes.go
yeganemehr: OK, after you merged this I'll send another PR. Is it okay with you?

weed/shell/command_fs_merge_volumes.go
yeganemehr: done in 5091260

Issue #5014, opened 2023-11-16
[filer.backup] avoid backup truncated files


kmlebedev:
**Describe the bug**
Files from a month ago were synchronized, some of which had already been deleted.

**System Setup**
`weed filer.backup -filerPath "/"`
 `3.57`

replication.toml:
```
[sink.s3]
enabled = true
aws_access_key_id = "xxxx"
aws_secret_access_key = "yyyy"
region = "us-east-2"
bucket = "backups"
directory = "/docs"
endpoint = "https://storage"
is_incremental = true
```

**Expected behavior**
Skip backup there is a delete_chunks:true flag

**Additional context**
```
E1110 20:37:02.241428 filer_pb_tail.go:89 process directory:"/documents/8/7/2"  event_notification:{
new_entry:{name:"87279e8754b8d728c696c47445ebaf10-thumb-bak"  chunks:{file_id:"8,679a6278ced870"  size:14490  modified_ts_ns:1692373608704630899  e_tag:"zT9IpGyOtAT5/jgfmOD+eg=="  fid:{volume_id:8  file_key:6789730  cookie:2026821744}}  attributes:{file_size:14490  mtime:1678822702  file_mode:4
20  uid:990  gid:990  crtime:1692373608  mime:"image/jpeg"  inode:9740092430445530816}}  delete_chunks:true  new_parent_path:"/documents/8/7/2"  signatures:1520837357  signatures:-1416986718}  ts_ns:169237360870954990
1: create entry1 : RequestError: send request failed
caused by: Put "https://storage/backups/docs/2023-03-14/documents/8/7/2/87279e8754b8d728c696c47445ebaf10-thumb-bak": http://192.168.0.1:8080/8,679a6278ced870?readDeleted=true: 404 Not Found
```

chrislusf:
it seems the `https://storage/backups/docs/2023-03-14/documents/8/7/2/87279e8754b8d728c696c47445ebaf10-thumb-bak` file was updated later. So it should be fine to skip current error.

kmlebedev:
Here is the full response:
```
E1110 20:37:29.028503 filer_pb_tail.go:89 process directory:"/documents/8/7/2"  
event_notification: { 
 old_entry: {
 	name:".872809481ab4c04988b491bd70b73fc4-thumb-bak.vVA5U6"
 	chunks: {
[...]

PR #5015, opened 2023-11-16
fix: avoid backup deleted files

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/5014

# How is the PR tested?

# Checks

### DIFF
--- a/weed/command/filer_sync.go
+++ b/weed/command/filer_sync.go
@@ -427,6 +427,11 @@ func genProcessFunction(sourcePath string, targetPath string, excludePaths []str
 			return nil
 		}
 
+		// handle truncate entries
+		if filer_pb.IsUpdate(resp) && resp.EventNotification.DeleteChunks && dataSink.IsIncremental() {
+			return nil
+		}
+
 		// handle updates
 		if strings.HasPrefix(string(sourceOldKey), sourcePath) {
 			// old key is in the watched directory


weed/command/filer_sync.go
```diff
@@ -427,6 +427,11 @@ func genProcessFunction(sourcePath string, targetPath string, excludePaths []str
 			return nil
 		}
 
+		// handle truncate entries
+		if filer_pb.IsUpdate(resp) && resp.EventNotification.DeleteChunks && dataSink.IsIncremental() {
```
chrislusf, line 431: `resp.EventNotification.DeleteChunks` is only used in sink to another filer.

Not sure what does this mean. It sounds like "for incremental mode, do not handle updates".

kmlebedev: > Not sure what does this mean. It sounds like "for incremental mode, do not handle updates".

May need more time to reproduce. 
I found the following calls:
image

Do not fit where signatures is nill and newEntry or oldEntry is nill and deleteChunks is false.
And that remains:
1. https://github.com/seaweedfs/seaweedfs/blob/17710e1ecbd36409fa5b14861c025aacf943b6fd/weed/filer/filer.go#L245
2. https://github.com/seaweedfs/seaweedfs/blob/17710e1ecbd36409fa5b14861c025aacf943b6fd/weed/server/filer_grpc_server.go#L195

kmlebedev:
> Not sure what does this mean. It sounds like "for incremental mode, do not handle updates".

PR #5017, opened 2023-11-17
fix: guard whiteList for file id handle on master

kmlebedev:

With whitelist restrictions configured, one handle with a file ID is available for requests from other IPs
```
I1117 04:35:54.641257 common.go:75 response method:GET URL:/sitemap.xml with httpStatus:404 and JSON:{"error":"volume id sitemap.xml not found: Unknown volume id sitemap.xml"}
```

add grant whiteList for `{fileId}` handler

# How is the PR tested?

# Checks


kmlebedev:
> With the cookie, `fileId` is unlikely guessable. This should not be necessary.

@chrislusf  This is more a desire to protect yourself from security scanners that flood the logs with errors.

PR #5018, opened 2023-11-17
fix: add doDeleteFile option for filer backup

kmlebedev:

The `weed filer.backup` command is responsible for backup, but it also deletes files, which is not obvious.

Added an option responsible for deleting files and by default it is disabled.

# How is the PR tested?

# Checks


chrislusf:
Does it duplicate the incremental mode?

kmlebedev:
> Does it duplicate the incremental mode?

No. So the path to the files will be original

Issue #4953, opened 2023-10-26
Fuse mount deadlock on write


kvaster:
We're experiancing deadlock on write using fuse mount. It was max 20 files write in parallel.

weed command for seaweedfs-csi was built from commit: 40888b561226004765ce158e922df6aca5dcfe70

[deadlock-stracktrace.txt](https://github.com/seaweedfs/seaweedfs/files/13174960/deadlock-stracktrace.txt)

chrislusf:
Possibly to mitigate this, try to increase this `weed mount -concurrentWriters=32` to a high number.

kvaster:
I think will improve situation but will not close the root of the problem. I think that the problem is here:

https://github.com/seaweedfs/seaweedfs/blob/bf839651ecdcb400d7707349b1c0a2016a570782/weed/mount/page_writer/upload_pipeline.go#L157

Do we really need to hold the lock while waiting for writes to complete ? It's really straight way for dead lock. The deadlock may occur even if we don't hit maximum number of concurrent writers.

kvaster:
Also 32 is a default value for concurrentWriters :)

kvaster:
@chrislusf, also I think that waiting for 'no jobs' is not really good idea. End of the flush should be predictable and we should wait only for finishing writes which were in progress only when FlushAll was issued.

PR #5019, opened 2023-11-17
Waiting for writes to complete with acquired lock may lead to deadlock

kvaster:
Waiting for writes to complete with acquired lock may lead to deadlock, fixes #4952


kvaster:
Fixes #4953

Issue #4991, opened 2023-11-08
Weed fix leads to data loss


SmsS4:
**Describe the bug**
Looks like volume server sorts needles in needle map with offset but after `weed fix` needles are sorted by key so when volume server starts, truncate `dat` files

**Reproduce**
Reproduce with volume server:
```shell
weed volume -ip.bind  -dir vol 0.0.0.0
weed master -volumeSizeLimitMB 1000  -ip.bind 0.0.0.0
```
reprod.py:
```python
import requests

MASTER = "http://127.0.0.1:9333/dir"
VOLUME = "http://127.0.0.1:8080"
NUMBER_OF_UPLOADS = 1000
FILE_SIZE = 100
FILE_PATH = "/tmp/file_to_upload"


with open(FILE_PATH, "w") as f:
    for i in range(FILE_SIZE):
        f.write(str(i % 10))
needles = [
    requests.get(f"{MASTER}/assign").json()["fid"] for _ in range(NUMBER_OF_UPLOADS)
]
needles.sort(reverse=True)
for needle_id in needles:
    print(needle_id)
    with open(FILE_PATH, "rb") as f:
        print(
            requests.post(
                f"{VOLUME}/{needle_id}",
                files={"file": f},
            ).text,
        )
```
```shell
python reprod.py 
# stop volume server
weed fix vol
weed volume -dir vol -ip.bind 0.0.0.0
W1108 11:46:52.806182 volume_checking.go:118 Truncate vol/1.dat from 22328 bytes to 5192 bytes!
W1108 11:46:52.806211 volume_checking.go:118 Truncate vol/7.dat from 22040 bytes to 5480 bytes!
```

Reproduce with s3 (This doesn't always leads to dataloss):
```shell
weed volume -ip.bind  -dir vol 0.0.0.0
weed master -volumeSizeLimitMB 1000  -ip.bind 0.0.0.0
weed filer -s3
warp put --insecure --host localhost:8333 --noclear  --obj.size 1KiB --concurrent 300 --noprefix --duration 10s
# stop volume server
weed fix vol
weed volume -ip.bind  -dir vol 0.0.0.0
> W1108 11:58:15.023936 volume_checking.go:118 Truncate vol/warp-benchmark-bucket_2.dat from 13389696 bytes to 13386408 bytes!
echo "fs.verify -concurrency 30" | weed shell -filer localhost:8888
```
```
[...]

PR #5020, opened 2023-11-17
Fix weed fix

SmsS4:
https://github.com/seaweedfs/seaweedfs/issues/4991

Sorting needles by offset when mapdb wants to save idx to file

# How is the PR tested?
I tested using `reprod.py` in issue. Also tested vacuum manually because vacuum use `SaveToIdx`

# Checks


chrislusf:
reverting. let's discuss in @kmlebedev 's thread.

PR #5022, opened 2023-11-18
chore(filer): add system log dir for filer

duanhongyi:

# How is the PR tested?

# Checks


chrislusf:
similar to `/buckets` folder, this is not meant to be customizable.

PR #5042, opened 2023-11-21
consul filer store

unRob:

I run Hashicorp [Consul](https://www.consul.io/) at home and would like to use it as the kv store for filer. I'm new to the project and will be running my fork to try it out for the first time. My end goal is to use seaweed through the CSI driver on my nomad cluster as a simple way to provide persistent storage.

Copied the flow of the etcd store (thanks, previous contributors!) and used consul's official golang sdk to implement the `filer.FilerStore` interface.

# How is the PR tested?

Based test on existing etcd test, enabled it and made sure it ran against a dockerized Consul in dev mode. Also ran `make test_consul`, again, trying to emulate the existing etcd implementation but not sure how to measure success here, as apparently I missed a step where a bucket is created for the registry. 

As said above, will be running a fork at home to fully test out the implementation (as well as getting to know the project). Will mark as ready for review once I review it running for a while.

# Checks


chrislusf:
plan changed?

PR #4946, opened 2023-10-25
support tls in master-ingress

Programmeris:
# What problem are we solving?

It is necessary to add the ability to configure TLS in the master Ingress

# How are we solving the problem?

Added optional tls field to values.yaml similar to filer ingress 

# How is the PR tested?

helm lint/helm template on local machine, CI jobs

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #5048, opened 2023-11-24
Helm: add post install hook to support creating default buckets

cloudymax:

Enhancements:
  - Enable administrators to create buckets using the helm chart in a declarative way
  
  - Provide the option for created buckets to be publicly accessible

Housekeeping: 

- Changes  nodeSelector tag from `beta.kubernetes.io/arch: amd64` to `kubernetes.io/arch: amd64` because it was deprecated in version 1.14 and removed in 1.18

- bump chart from `3.59.2` to `3.59.3`

This PR enables the creation of new buckets via the use of a [Post-install helm Hook](https://helm.sh/docs/topics/charts_hooks/#the-available-hooks) on a Kubernetes Job, which executes after all other resources are loaded into Kubernetes.

The Kubernetes job invokes `weed shell` to create a list of buckets.

# How is the PR tested?

1. Create an updated values file containing the new bucket options.

    Example filer section:

    ```yaml
    filer:
      s3:
        enabled: true
        enableAuth: true
        skipAuthSecretCreation: false
        # You may specify buckets and users to be created during the install process
        createBuckets:
          - name: bucket-a
            anonymousRead: true
          - name: bucket-b
            anonymousRead: false
    ```

2. install the helm chart from the local directory:

    ```console
    $ helm install --values=test-values.yaml seaweedfs . --wait
    LAST DEPLOYED: Thu Nov 23 13:37:55 2023
    NAMESPACE: default
    STATUS: deployed
    REVISION: 1
    TEST SUITE: None
    ```

3. add some data to the bucket and expose the s3-endpoint service

    

4. Share the public endpoint with friends

    A view of a web browser opened to a public URL that shows the sprout image that was uploaded to bucket-a

# Checks


k8s/charts/seaweedfs/values.yaml
```diff
@@ -30,7 +30,7 @@ global:
   # Z number of replica in other servers in the same rack
   replicationPlacment: "001"
   extraEnvironmentVars:
-    WEED_CLUSTER_DEFAULT: "sw"
+    WEED_CLUSTER_DEFAULT: "config"
```
chrislusf: What no need to change the default value here?

k8s/charts/seaweedfs/values.yaml
```diff
@@ -570,12 +570,19 @@ filer:
     # Suffix of the host name, {bucket}.{domainName}
     domainName: ""
     # enable user & permission to s3 (need to inject to all services)
-    enableAuth: false
+    enableAuth: true
     skipAuthSecretCreation: false
     # set to the name of an existing kubernetes Secret with the s3 json config file
     # should have a secret key called seaweedfs_s3_config with an inline json configure
-    existingConfigSecret: ""
+    existingConfigSecret: false
```
chrislusf: This doesn't sound like a Boolean.

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
```diff
@@ -0,0 +1,100 @@
...
+        {{- if .Values.master.readinessProbe.enabled }}
+    {{- $hostpath_exists := include "master.hostpath_exists" . -}}
+    {{- $existing_claims := include "master.existing_claims" . -}}
+    {{- if .Values.filer.s3.enableAuth }}
+      volumes:
+        - name: config-users
+          secret:
+            defaultMode: 420
+            {{- if .Values.filer.s3.existingConfigSecret }}
+            secretName: {{ .Values.filer.s3.existingConfigSecret }}
```
chrislusf, line 92: Used as string here.

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
```diff
@@ -0,0 +1,100 @@
...
+            #name: swfs-master-grpc
+        {{- if .Values.master.readinessProbe.enabled }}
+    {{- $hostpath_exists := include "master.hostpath_exists" . -}}
+    {{- $existing_claims := include "master.existing_claims" . -}}
+    {{- if .Values.filer.s3.enableAuth }}
+      volumes:
+        - name: config-users
+          secret:
+            defaultMode: 420
+            {{- if .Values.filer.s3.existingConfigSecret }}
```
chrislusf: Used as bool here.

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
```diff
@@ -0,0 +1,100 @@
...
+        app.kubernetes.io/managed-by: {{ .Release.Service | quote }}
+        app.kubernetes.io/instance: {{ .Release.Name | quote }}
+    spec:
+      restartPolicy: Never
+      containers:
+      - name: post-install-job
+        image: {{ template "master.image" . }}
+        env:
+          - name: WEED_CLUSTER_DEFAULT
+            value: "sw"
```
chrislusf, line 30: Used "sw" for default value.

k8s/charts/seaweedfs/values.yaml
cloudymax: This is an artifact left over form debugging, my mistake.

cloudymax: reverted to previous default in https://github.com/seaweedfs/seaweedfs/pull/5048/commits/601661c9e42feded4f73b9f320ec7dae185f92cb

We were debugging the interaction between `enableAuth`, `skipAuthSecretCreation`, and `existingConfigSecret` at the time and I mistakenly believed the secret mount was related to the `sw` default value when it was actually something else.

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
cloudymax: reverted the change to the default value based on previous feedback 👍

k8s/charts/seaweedfs/values.yaml
cloudymax: You are correct, it should be a string. I will alter my approach accordingly 👍

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
cloudymax: should now be consistently treated as a string by using 

```go
{{- if not (empty .Values.filer.s3.existingConfigSecret) }}
```
changed in https://github.com/seaweedfs/seaweedfs/pull/5048/commits/ecdc3217cb88d7f580ff53765fd5b56a2c0403ef

k8s/charts/seaweedfs/templates/post-install-bucket-hook.yaml
cloudymax: should now be consistently treated as a string by using:

```go
{{- if not (empty .Values.filer.s3.existingConfigSecret) }}
```
changed in https://github.com/seaweedfs/seaweedfs/pull/5048/commits/ecdc3217cb88d7f580ff53765fd5b56a2c0403ef

cloudymax:
@chrislusf Thank you so much for all the feedback! 

I think all points have been addressed now 👍

k8s/charts/seaweedfs/Chart.yaml
```diff
@@ -2,4 +2,4 @@ apiVersion: v1
 description: SeaweedFS
 name: seaweedfs
 appVersion: "3.59"
-version: 3.59.2
+version: 3.59.3
```
jessebot, line 5: Just a note for anyone curious. This PR also includes releasing this PR https://github.com/seaweedfs/seaweedfs/pull/4946

PR #5060, opened 2023-11-30
chore: add dsn for connection to mysql

kmlebedev:

it is necessary to pass arbitrary connection param to the msql driver like [tls](https://github.com/go-sql-driver/mysql#tls)

add dsn to filer config

# How is the PR tested?

localy

# Checks

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -60,6 +60,7 @@ connection_max_idle = 2
 connection_max_open = 100
 connection_max_lifetime_seconds = 0
 interpolateParams = false
+dsn = "root:password@tcp(localhost:3306)/seaweedfs?collation=utf8mb4_bin"
 # if insert/upsert failing, you can disable upsert or update query syntax to match your RDBMS syntax:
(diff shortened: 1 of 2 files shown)


weed/command/scaffold/filer.toml
```diff
@@ -60,6 +60,8 @@ connection_max_idle = 2
 connection_max_open = 100
 connection_max_lifetime_seconds = 0
 interpolateParams = false
+#     [username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
+dsn = "root@tcp(localhost:3306)/seaweedfs?collation=utf8mb4_bin"
```
chrislusf: move the new section to after line 53, and add comments that this dsn will take priority over "hostname, port, username, password, database".

kmlebedev:
@chrislusf done

chrislusf:
To adapt with password, the password needs to be set first? If so, need to mention it in the comment.

kmlebedev:
> password needs to be

make it so you don't need to set a password

PR #5081, opened 2023-12-06
Always include volumes in statefulsets

dzsibi:

When both `master.data.type` and `master.logs.type` is `persistentVolumeClaim`, deployment of the Helm chart fails:

```
create Pod seaweedfs-master-0 in StatefulSet seaweedfs-master failed error: Pod "seaweedfs-master-0" is invalid: spec.containers[0].volumeMounts[1].name: Not found: "master-config"
```

This is due to a condition that omits the entire volumes section in this case:

https://github.com/seaweedfs/seaweedfs/blob/d5d9fbb8aa5e86f2b866e25f58bf36a7a989478d/k8s/charts/seaweedfs/templates/master-statefulset.yaml#L217C20-L217C20

The issue was introduced by #4797, as it did not take into account this condition.

I removed the conditions entirely, including the ones for volume-statefulset. They also interfere with a bunch of other options (`extraVolumes`, `enableSecurity`).

# How is the PR tested?

For the default configuration (using hostPath volumes), deployment works before and after my changes. For a configuration that uses PVCs instead, deployments fail before and work after. 

An example values.yaml to trigger the bug:

```
seaweedfs:
  master:
    data:
      type: persistentVolumeClaim
    logs:
      type: persistentVolumeClaim
```

# Checks

I am unsure if these are relevant for this PR.

### DIFF
--- a/k8s/charts/seaweedfs/templates/_helpers.tpl
+++ b/k8s/charts/seaweedfs/templates/_helpers.tpl
@@ -122,19 +122,6 @@ Inject extra environment vars in the format key:value, if populated
 {{- end -}}
 {{- end -}}
 
-{{/* check if any Volume HostPath exists */}}
-{{- define "volume.hostpath_exists" -}}
-{{- if or (or (eq .Values.volume.data.type "hostPath") (and (eq .Values.volume.idx.type "hostPath") .Values.volume.dir_idx )) (eq .Values.volume.logs.type "hostPath") -}}
-{{- printf "true" -}}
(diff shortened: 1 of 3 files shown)


chrislusf:
Thanks! Merging.

cc @aviau let me know if you have any concern.

dzsibi:
Hmm, I didn't know it's going to auto-publish immediately, I think it may have overwritten the previous version since I did not include a version bump in the PR.

PR #5102, opened 2023-12-13
Remote gateway options bug fix

itsfarbod:
I encountered the issue when i was trying to have a `filer.remote.gateway` instance for async backup.
The problem is that when you set filer address in `filer.remote.gateway` `-filer` flag, It won't use the filer address and encounters error while trying to write to the filer. Actually it uses RemoteSyncOptions instead of RemoteGatewayOptions. So because RemoteSyncOptions have default values it will use localhost:8888 as filer address and will ignore the actual filer address.

By changing `filer_remote_gateway_buckets.go` file to use `RemoteGatewayOptions` instead of `RemoteSyncOptions` 

# How is the PR tested?
Locally on docker compose

# Checks
No need for editing wiki


chrislusf:
added commit which should be a little more cleaner

PR #5109, opened 2023-12-14
Set allowed origins in config

jerebear12:

When making a request to a server at a different domain from a browser, the browser will impose a set of restrictions known as CORS. When making a request with credentials from a browser to a server on a different domain, the server with a different domain cannot accept the credentials in the usual way. The usual way is to use the header "Access-Control-Allow-Credentials" with value true. The browser fails the request when the "Access-Control-Allow-Origin" header has a value of *(wildcard). If the value of "Access-Control-Allow-Origin" is set to a specific domain, the browser will let the request through.

This issue appears when making a request to the Filer or S3 Gateway from a web client with credentials included. This happens when the client needs to send HTTP-only cookies to SeaweedFS, a use case addressed in this pull: #5077. 

I added a way to specify allowed domains(plural) in the security.toml file and through a command line flag `-allowedDomains`. The default value is still "*". The value set in the configuration file or command line flag adjust the value of the "Access-Control-Allow-Origin" header.

The domains are split by comma in the code and added as a list of type string. When the Filer or S3 Gateway receive a request, we check for the "Domain" header the browser will include. If the header is not empty, we compare the value to the domains added to the config. If there is a match, we return the matched value to give the browser the "go ahead" and respond to the request. If the domain is not found(no match), the request will fail with a status code of 503(Resource Forbidden). If no configuration is changed and the command line flag is not set, the default value is used, and requests that do not include credentials will go through as usual. 

**These changes apply to the Filer and S3 gateway**.

# How is the PR tested?

Tested on my local machine with a web client. I tested with the default value, with the localhost domain in the security.conf, and using the command line flag

# Checks

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -258,7 +258,7 @@ func (fo *FilerOptions) startFiler() {
 		ShowUIDirectoryDelete: *fo.showUIDirectoryDelete,
 		DownloadMaxBytesPs:    int64(*fo.downloadMaxMBps) * 1024 * 1024,
 		DiskType:              *fo.diskType,
-		AllowedOrigins:        *fo.allowedOrigins,
+		AllowedOrigins:        strings.Split(*fo.allowedOrigins, ","),
 	})
 	if nfs_err != nil {
 		glog.Fatalf("Filer startup error: %v", nfs_err)
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -61,6 +61,7 @@ type FilerOptions struct {
 	showUIDirectoryDelete   *bool
 	downloadMaxMBps         *int
 	diskType                *string
+	allowedOrigins          *string
 }
 
 func init() {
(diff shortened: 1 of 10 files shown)


weed/command/filer.go
```diff
@@ -91,6 +92,7 @@ func init() {
 	f.showUIDirectoryDelete = cmdFiler.Flag.Bool("ui.deleteDir", true, "enable filer UI show delete directory button")
 	f.downloadMaxMBps = cmdFiler.Flag.Int("downloadMaxMBps", 0, "download max speed for each download request, in MB per second")
 	f.diskType = cmdFiler.Flag.String("disk", "", "[hdd|ssd|<tag>] hard drive or solid state drive or any tag")
+	f.allowedOrigins = cmdFiler.Flag.String("allowedOrigins", "*", "comma separated list of allowed origins")
```
chrislusf, line 95: how about defaulting `allowedOrigins` to an empty string?

jerebear12: The filer and S3 gateway will need a value of an address or the wildcard on startup in order to work properly. With a blank value the requests from the browser will fail unless they are from the same origin I believe. Requests from a non web could would most likely continue to work.

PR #5123, opened 2023-12-20
Using filer.remote.sync concurrency in filer.remote.gateway

itsfarbod:
- Chnaged ProcessEvenFn to be concurrent just like filer.remote.sync

So the problem appeared in a comparison between `filer.remote.sync` and `filer.remote.gateway`
In action `filer.remote.sync` was much more faster, and I measured the difference and it was something between 80x to 100x performance difference.

So I checked out the code and it was a simple procedure used in `filer.remote.sync` which I think is used for concurrency.
Here is the part of code I changed in `weed/command/filer_remote_gateway_buckets.go`:
```
// read filer remote storage mount mappings
	if detectErr := option.collectRemoteStorageConf(); detectErr != nil {
		return fmt.Errorf("read mount info: %v", detectErr)
	}

	eachEntryFunc, err := option.makeBucketedEventProcessor(filerSource)
	if err != nil {
		return err
	}

	processor := NewMetadataProcessor(eachEntryFunc, 128)

	var lastLogTsNs = time.Now().UnixNano()
	processEventFnWithOffset := pb.AddOffsetFunc(func(resp *filer_pb.SubscribeMetadataResponse) error {
		processor.AddSyncJob(resp)
		return nil
	}, 3*time.Second, func(counter int64, lastTsNs int64) error {
		if processor.processedTsWatermark == 0 {
			return nil
		}
		now := time.Now().UnixNano()
		glog.V(0).Infof("remote sync %s progressed to %v %0.2f/sec", *option.filerAddress, time.Unix(0, processor.processedTsWatermark), float64(counter)/(float64(now-lastLogTsNs)/1e9))
		lastLogTsNs = now
		return remote_storage.SetSyncOffset(option.grpcDialOption, pb.ServerAddress(*option.filerAddress), option.bucketsDir, processor.processedTsWatermark)
	})
```
When I tested `filer.remote.gateway` after this change, Its performance was exactly like `filer.remote.sync`
# How is the PR tested?
Locally via docker compose.
On a k8s cluster.

# Checks
It's a Bug Fix. No need to change the wiki.

### DIFF
--- a/weed/command/filer_remote_gateway_buckets.go
+++ b/weed/command/filer_remote_gateway_buckets.go
@@ -30,10 +30,20 @@ func (option *RemoteGatewayOptions) followBucketUpdatesAndUploadToRemote(filerSo
 		return err
 	}
 
-	processEventFnWithOffset := pb.AddOffsetFunc(eachEntryFunc, 3*time.Second, func(counter int64, lastTsNs int64) error {
-		lastTime := time.Unix(0, lastTsNs)
-		glog.V(0).Infof("remote sync %s progressed to %v %0.2f/sec", *option.filerAddress, lastTime, float64(counter)/float64(3))
-		return remote_storage.SetSyncOffset(option.grpcDialOption, pb.ServerAddress(*option.filerAddress), option.bucketsDir, lastTsNs)
+	processor := NewMetadataProcessor(eachEntryFunc, 128)
+
+	var lastLogTsNs = time.Now().UnixNano()
+	processEventFnWithOffset := pb.AddOffsetFunc(func(resp *filer_pb.SubscribeMetadataResponse) error {
+		processor.AddSyncJob(resp)
+		return nil
+	}, 3*time.Second, func(counter int64, lastTsNs int64) error {
+		if processor.processedTsWatermark == 0 {
+			return nil
+		}
+		now := time.Now().UnixNano()
+		glog.V(0).Infof("remote sync %s progressed to %v %0.2f/sec", *option.filerAddress, time.Unix(0, processor.processedTsWatermark), float64(counter)/(float64(now-lastLogTsNs)/1e9))
+		lastLogTsNs = now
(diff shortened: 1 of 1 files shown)


weed/command/filer_remote_gateway_buckets.go
```diff
@@ -30,10 +30,20 @@ func (option *RemoteGatewayOptions) followBucketUpdatesAndUploadToRemote(filerSo
 		return err
 	}
 
-	processEventFnWithOffset := pb.AddOffsetFunc(eachEntryFunc, 3*time.Second, func(counter int64, lastTsNs int64) error {
-		lastTime := time.Unix(0, lastTsNs)
-		glog.V(0).Infof("remote sync %s progressed to %v %0.2f/sec", *option.filerAddress, lastTime, float64(counter)/float64(3))
-		return remote_storage.SetSyncOffset(option.grpcDialOption, pb.ServerAddress(*option.filerAddress), option.bucketsDir, lastTsNs)
+	processor := NewMetadataProcessor(eachEntryFunc, 128)
```
kmlebedev, line 33: 128 - Let's put it in a static variable and call it so that it is clear what it is responsible for. 
Also, maybe it’s worth connecting to the number core via GOMAXPROCS?

itsfarbod: I agree, but i actually tried to use the same logic from `filer.remote.sync`
But how is it possible to use GOMAXPROCS‌ for this code?

itsfarbod: Also I think merging this PR before the release will be helpful and I'll change the logic behind the concurrency value in future. Because it needs a change in both `filer.remote.gateway` and `filer.remote.sync`. 
In other hands I think founding the actual proper value for max concurrency parameter needs a separate discussion for setting it to a appropriate value.
So I will be very thankful if this PR gets merged.
Thanks.

PR #5134, opened 2023-12-23
Update network.go since To16 converts the IP address ip up to a 16-byte representation. If ip is not an IP address (it is the wrong length), To16 returns nil.

vivekkoya:
Optimizing Network code performance

To16 also works for ipv4 addresses as referenced in the go documentation: https://pkg.go.dev/net. The referenced change removes a variable and an accompanying check for IPV4/6, improving performance.

# How is the PR tested?
The code should pass the existing network tests.

# Checks

### DIFF
--- a/weed/util/network.go
+++ b/weed/util/network.go
@@ -15,18 +15,13 @@ func DetectedHostAddress() string {
 		return ""
 	}
 
-	if v4Address := selectIpV4(netInterfaces, true); v4Address != "" {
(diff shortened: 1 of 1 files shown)


vivekkoya:
Can you please provide more details regarding the OS, docker, runtime environment, and if docker ipv6 support is enabled under Linux: https://docs.docker.com/config/daemon/ipv6/. ? Thank you

PR #5166, opened 2024-01-05
chore: filer sync add doDeleteFiles option for create only mode

kmlebedev:

I would like to have a backup cluster where nothing is deleted instantly.

Add an option to skip synchronization of deleted and updated files.

Trying to create an existing file will result in an error that can be ignored.

# How is the PR tested?

```
make test
```

# Checks

### DIFF
--- a/weed/command/filer_sync.go
+++ b/weed/command/filer_sync.go
@@ -47,7 +47,8 @@ type SyncOptions struct {
 	metricsHttpIp   *string
 	metricsHttpPort *int
 	concurrency     *int
-	doDeleteFiles   *bool
+	aDoDeleteFiles  *bool
+	bDoDeleteFiles  *bool
 	clientId        int32
 	clientEpoch     int32
 }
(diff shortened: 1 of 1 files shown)


weed/command/filer_sync.go
```diff
@@ -90,6 +91,7 @@ func init() {
 	syncMemProfile = cmdFilerSynchronize.Flag.String("memprofile", "", "memory profile output file")
 	syncOptions.metricsHttpIp = cmdFilerSynchronize.Flag.String("metricsIp", "", "metrics listen ip")
 	syncOptions.metricsHttpPort = cmdFilerSynchronize.Flag.Int("metricsPort", 0, "metrics listen port")
+	syncOptions.doDeleteFiles = cmdFilerSynchronize.Flag.Bool("doDeleteFiles", true, "delete and update files when synchronizing")
```
chrislusf: The user may need to set for `a` or `b` sync direction.

kmlebedev: > The user may need to set for `a` or `b` sync direction.

done

PR #5180, opened 2024-01-08
Removed problematic if statement

jerebear12:

The removed if statement was causing the value of option.AllowedOrigins to always be equal to "*". 

Removing the if statement that was causing the config values to only be set when the option.AllowedOrigins property is nil or empty. 

Now the values in the config file will be used when present. This allows for people who don't need this feature to not update their security.toml files(same as before).

# How is the PR tested?

I tested on my local machine using a web client that makes requests that fail when SeaweedFS has the wildcard config. Then I adjusted the config values and retried the same requests. The AccessControlAllowOrigin header was successfully changed according to the config in my tests.

# Checks


jerebear12:
Thanks @chrislusf!

Issue #5182, opened 2024-01-09
[Critical] volume write more than MaxPossibleVolumeSize bytes, case offset truncate


kungf:
**Describe the bug**
volume write more than MaxPossibleVolumeSize bytes, case offset truncate

**System Setup**
-  "weed master", "weed volume"
- OS version: Ubuntu 20.04.6 LTS
- weed version: 3.53

**Expected behavior**

**Screenshots**
I add some log in weed fix command, and found that, the last chunk offset after thransfer to  smaller size exceeding uint32, so it became a truncate value from 34360015656 to 277288, this chunk can not be read!
![image](https://github.com/seaweedfs/seaweedfs/assets/24663329/b6005160-b30c-4277-befa-f3742a7a6028)

![image](https://github.com/seaweedfs/seaweedfs/assets/24663329/4bfad833-d8b7-4bf0-b61e-93107065c1b8)

**Additional context**

PR #5190, opened 2024-01-11
fix write volume over size MaxPossibleVolumeSize

kungf:
volume over size MaxPossibleVolumeSize
https://github.com/seaweedfs/seaweedfs/issues/5182

1. when write, n.Size is not init yet before prepareWriteBuf, so n.Size.IsValid()  alaways false!
2. v.nm.ContentSize() only have write data use size not include delete use size, so if volume has many delete ops(n) before full, more than n*32 bytes can also write to volume, even volume true offset is large than MaxPossibleVolumeSize

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/needle/needle_write.go
+++ b/weed/storage/needle/needle_write.go
@@ -127,7 +127,7 @@ func (n *Needle) Append(w backend.BackendStorageFile, version Version) (offset u
 		err = fmt.Errorf("Cannot Read Current Volume Position: %v", e)
 		return
 	}
-	if offset >= MaxPossibleVolumeSize && n.Size.IsValid() {
+	if offset >= MaxPossibleVolumeSize && len(n.Data) != 0 {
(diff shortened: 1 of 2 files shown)


chrislusf:
* How is this PR tested?
* Deletion can still work if over the size limit?

kungf:
hi @chrislusf，I am glad to be able to do something for SeaweedFS！

> * How is this PR tested?
I will test it by these steps：
1. write some files
2. delete these files
3. repeated 1,2 until volume full
I write a tool to check whether need set over flow。


> * Deletion can still work if over the size limit?
Yes，If we disable deletion after volume full，it will unable to reclaim space。

> * This PR is functionally the same as before. What is the difference this can make related to [[Critical] volume write more than MaxPossibleVolumeSize bytes, case offset truncate #5182](https://github.com/seaweedfs/seaweedfs/issues/5182) ?
This can fix size mismatch err when vacuuming。

Issue #4758, opened 2023-08-15
filer.sync with default concurrency(32) leaves lot of orphans


ents-hqx:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

- master,volume,filer and s3 are different processes

weed master -disableHttp -volumeSizeLimitMB=10240  -volumePreallocate=false  -mdir=/seaweedfs/mdir
weed volume -dataCenter=dc -dir=/seaweedfs/volume -dir.idx=/seaweedfs/index -max=0 
weed filer -defaultStoreDir=/seaweedfs/filer -ui.deleteDir=false 
weed s3 -ip.bind=${BIND_IP} -port=9000 -domainName=${S3_DOMAINS}

weed -logdir /var/log/seaweedfs filer.sync -a filer1:8888 -b filer2:8888

Typical file sizes in filer are 200-600MB

- debian 11.7, kernel 5.10.0-23-amd64
- weed version 30GB 3.55 7eafa3420b8e5ae83c8873cddd03ded90a0fc921 linux amd64
-  filer.toml content 

[filer.options]

recursive_delete = false

[leveldb3]
enabled = true
dir = "/seaweedfs/filer/filerldb3"                    # directory to store level db files

*******************

When running filer.sync(bootstrapping) with default concurrency(32), it makes lot of orphans

# volume.fsck -cutoffTimeAgo=1s
total 16 directories, 240 files
dataNode:filer2:8080       volume:18       entries:3847    orphan:1428     37.12%  3794475429B
dataNode:filer2:8080       volume:7        entries:4426    orphan:759      17.15%  1967575264B
dataNode:filer2:8080       volume:8        entries:1       orphan:1        100.00% 448640B
dataNode:filer2:8080       volume:1        entries:4407    orphan:756      17.15%  2003626853B
dataNode:filer2:8080       volume:20       entries:3753    orphan:1373     36.58%  3567856595B
dataNode:filer2:8080       volume:15       entries:3938    orphan:1439     36.54%  3802318932B
dataNode:filer2:8080       volume:16       entries:3879    orphan:1449     37.35%  3826251075B
dataNode:filer2:8080       volume:3        entries:4328    orphan:724      16.73%  1844336028B
[...]

PR #5216, opened 2024-01-18
fix: skip s3 .uploads

bernardx:
#4758 [filer]filer.sync with default concurrency(32) leaves lot of orphans 
- `.uploads` folder is used during s3 multi-part uploads and will be deleted after a successful upload
- During filer.sync, as the real data is referenced by two files (the final file and temp file in .uploads), it will be synced twice

skip .uploads folder during sync
@chrislusf  may be it's a better idea not to include `.uploads` folder in system log?

# How is the PR tested?
- add `-s3` param to `local-sync-mount-compose.yml` 
- using the following python code to upload a large(100MB) file to node-1
```
from minio import Minio
from minio.error import S3Error

def main():
    # Create a client with the MinIO server playground, its access key
    # and secret key.
    client = Minio("192.168.207.2:8333",
        access_key="",
        secret_key="",
        secure=False

    )

    # The file to upload, change this path if needed
    source_file = "test.mp4"

    # The destination bucket and filename on the MinIO server
    bucket_name = "test"
    destination_file = "test.mp4"

    # Make the bucket if it doesn't exist.
    found = client.bucket_exists(bucket_name)
    if not found:
        client.make_bucket(bucket_name)
        print("Created bucket", bucket_name)
    else:
        print("Bucket", bucket_name, "already exists")

    # Upload the file, renaming it in the process
    client.fput_object(
        bucket_name, destination_file, source_file,
    )
    print(
        source_file, "successfully uploaded as object",
        destination_file, "to bucket", bucket_name,
    )

if __name__ == "__main__":
    try:
        main()
    except S3Error as exc:
        print("error occurred.", exc)
```
- Verify that the file has been replicated to node-2 successfully and there is no `.uploads` folder created

# Checks

### DIFF
--- a/weed/command/filer_sync.go
+++ b/weed/command/filer_sync.go
@@ -11,6 +11,7 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/replication/sink"
 	"github.com/seaweedfs/seaweedfs/weed/replication/sink/filersink"
 	"github.com/seaweedfs/seaweedfs/weed/replication/source"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3_constants"
 	"github.com/seaweedfs/seaweedfs/weed/security"
 	statsCollect "github.com/seaweedfs/seaweedfs/weed/stats"
 	"github.com/seaweedfs/seaweedfs/weed/util"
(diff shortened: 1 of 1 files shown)


chrislusf:
thanks for the PR. 

The metadata is needed to replicate updates to other filers. So the special folder can not be skipped.

kmlebedev:
> thanks for the PR.
> 
> The metadata is needed to replicate updates to other filers. So the special folder can not be skipped.

@chrislusf 
> **can not be skipped.** ?

PR #5241, opened 2024-01-25
weed/shell: Cleanup volume balance logic

BenoitKnecht:

Reading through the `volume.balance` code, I noticed a few things that made the code more complex than it needed to be:

- The case where `*collection == "ALL_COLLECTIONS"` was treated separately, while it's actually taken care of by the default case in the if/else conditional;
- The `balanceVolumeServers*()` functions required a `volumeSizeLimit` argument, but never used it;
- The loop over a map used the `for key, _ := range x` syntax, instead of the equivalent `for key := range x` (which makes the linter happy).

See descriptions of the individual commits for details.

# How is the PR tested?

I ran the tests, deployed the new code on my test cluster and used the `volume.balance` command successfully.

# Checks


BenoitKnecht:
@chrislusf I'm not sure why the fuse mount check failed, but I don't think it's related to those changes. I don't know how to rerun the check though.

PR #5245, opened 2024-01-26
weed/shell: Cluster check other disk types

BenoitKnecht:

The `cluster.check` command only took the empty (`""`) and `hdd` disk types into consideration, but a cluster with only `ssd` or `nvme` disk types would be equally valid.

This commit simply checks that _any_ disk type is defined, and that some volumes are available for it.

# How is the PR tested?

Spawned a cluster and ran `cluster.check` on it.

# Checks

### DIFF
--- a/weed/shell/command_cluster_check.go
+++ b/weed/shell/command_cluster_check.go
@@ -90,9 +90,7 @@ func (c *commandClusterCheck) Do(args []string, commandEnv *CommandEnv, writer i
 
 	// collect all masters
 	var masters []pb.ServerAddress
-	for _, master := range commandEnv.MasterClient.GetMasters() {
-		masters = append(masters, master)
-	}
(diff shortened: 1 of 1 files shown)


chrislusf:
The original intention is to ensure there is a hdd disk type defined, which would be used by the filer metadata logs.

BenoitKnecht:
> The original intention is to ensure there is a hdd disk type defined, which would be used by the filer metadata logs.

I see. In that case, perhaps it would make sense to check the empty and `hdd` disk types when `len(filers) > 0`, so that we don't error out when filers aren't even deployed?

But in principle, I think it would make sense for the metadata logs to prefer `hdd` disk type but fall back to `ssd` or `nvme` if those are the only available volume types. I haven't looked at that part of the code yet though.

BenoitKnecht:
@chrislusf Is this better?

PR #5267, opened 2024-01-31
helm using external master address

cuisongliu:

1. using external master address

# How is the PR tested?

# Checks


cuisongliu:
image
volume standalone deployment

cuisongliu:
image

In this way, the volume can be expanded at will when deployed later.

But it depends on pr https://github.com/seaweedfs/seaweedfs/pull/5268

Issue #5215, opened 2024-01-18
[volume] error reading newly written files after vacuum


kmlebedev:
**Describe the bug**

error reading newly written files after vacuum:
```
Jan 17, 2024 @ 17:36:38.738	I0117 12:36:09.943514 needle_map_memory.go:111 loading idx from offset 0 for file: /data/bucket_306.cpx
Jan 17, 2024 @ 17:36:38.738	I0117 12:36:23.346201 volume_loading.go:139 updating memory compact index /data/bucket_306.idx 
Jan 17, 2024 @ 17:36:38.738	W0117 12:36:23.066922 volume_checking.go:121 Truncate /data/bucket_306.dat from 38649746072 bytes to 34362613416 bytes!
Jan 17, 2024 @ 17:36:43.709	W0117 12:36:23.066922 volume_checking.go:121 Truncate /data/bucket_306.dat from 38649746072 bytes to 34362613416 bytes!
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:03.562706 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 11776 offset 34843571792 fileSize 34562495176: EOF
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:07.184218 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 472 offset 36492488624 fileSize 34562495176: EOF
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:05.063442 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 11776 offset 34843571792 fileSize 34562495176: EOF
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:06.182941 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 472 offset 36492488624 fileSize 34562495176: EOF
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:02.562017 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 11776 offset 34843571792 fileSize 34562495144: EOF
Jan 17, 2024 @ 19:35:10.120	E0117 14:35:06.182941 needle_read.go:45 /data/bucket_306.dat read 0 dataSize 472 offset 36492488624 fileSize 34562495176: EOF
[...]

PR #5272, opened 2024-01-31
avoid unexpected compact size

kmlebedev:

avoid unexpected compact size after https://github.com/seaweedfs/seaweedfs/issues/5215

return an error if the data size is smaller than the expected size

# How is the PR tested?

localy
```
1. make server
2. warp mixed --insecure --host 127.0.0.1:8000 --noclear  --obj.size 3KiB --concurrent 300 --noprefix --duration 20s --access-key some_access_key1 --secret-key some_secret_key1
3. echo -e "lock\nvolume.vacuum -garbageThreshold 0.01nvolume.list\n" | weed shell
```

# Checks

### DIFF
--- a/weed/storage/volume_vacuum.go
+++ b/weed/storage/volume_vacuum.go
@@ -423,7 +423,10 @@ func (v *Volume) copyDataBasedOnIndexFile(srcDatName, srcIdxName, dstDatName, da
 	if dstDatBackend, err = backend.CreateVolumeFile(dstDatName, preallocate, 0); err != nil {
 		return err
 	}
-	defer dstDatBackend.Close()
+	defer func() {
+		dstDatBackend.Sync()
+		dstDatBackend.Close()
+	}()
 
 	oldNm := needle_map.NewMemDb()
 	defer oldNm.Close()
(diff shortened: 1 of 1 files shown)


weed/storage/volume_vacuum.go
```diff
@@ -484,7 +487,20 @@ func (v *Volume) copyDataBasedOnIndexFile(srcDatName, srcIdxName, dstDatName, da
 	if err != nil {
 		return err
 	}
-
+	dstDatSize, _, err := dstDatBackend.GetStat()
+	if err != nil {
+		return err
+	}
+	if v.nm.ContentSize() > v.nm.DeletedSize() {
+		expectedContentSize := v.nm.ContentSize() - v.nm.DeletedSize()
```
chrislusf, line 495: when using leveldb, `needleMapMetricFromIndexFile()` function used a bloomfilter to estimate the deleted size. So this expectedContentSize may not be accurate.

kmlebedev: I looked at the inaccuracy with the size of the deleted bytes; it may be smaller, which will not harm the logic.

We need to observe on real data; if there are false errors, I will fix them as problems arise.

PR #5297, opened 2024-02-12
Disable filer UI in configuration

jerebear12:

There is currently no way to disable the Filer returning directory metadata when the Filer is exposed to the public. The previous best option was to require the JWT for reads and set the UI to readonly mode(which isn't always possible depending on the use case). This is good, but still exposes the paths and names of files. If the Filer is not set to readonly mode, then a client can request file paths via the Filer, view them, and a valid JWT, delete them. 

This pull adds a configuration property that allows you to hide the Filer UI by setting the filer.expose_directory_metadata.enabled property to `false`. This will make it harder for malicious users to access the files in SeaweedFS. This way the only way the paths to images will be exposed is if your application exposes their routes. 

# How is the PR tested?

This was tested on my local machine with the configuration property set to `true` and `false`. 

# Checks


weed/command/filer.go
```diff
@@ -93,6 +94,7 @@ func init() {
 	f.downloadMaxMBps = cmdFiler.Flag.Int("downloadMaxMBps", 0, "download max speed for each download request, in MB per second")
 	f.diskType = cmdFiler.Flag.String("disk", "", "[hdd|ssd|<tag>] hard drive or solid state drive or any tag")
 	f.allowedOrigins = cmdFiler.Flag.String("allowedOrigins", "*", "comma separated list of allowed origins")
+	f.exposeDirectoryData = cmdFiler.Flag.Bool("exposeDirectoryData", true, "whether to return directory metadata and content in Filer UI")
```
chrislusf, line 97: usually server.go file also needs to add a corresponding option.

weed/server/filer_server.go
```diff
@@ -115,6 +116,10 @@ func NewFilerServer(defaultMux, readonlyMux *http.ServeMux, option *FilerOption)
 	domains := strings.Split(allowedOrigins, ",")
 	option.AllowedOrigins = domains
 
+	v.SetDefault("filer.expose_directory_metadata.enabled", true)
```
jerebear12, line 119: @chrislusf Is this the option you are referring to in your comment above?

weed/command/filer.go
jerebear12: See my comment below. I believe the server.go file referenced is the filer_server.go file?

weed/server/filer_server.go
chrislusf: this file https://github.com/seaweedfs/seaweedfs/blob/e88467bb36b3618723c663d6b0e96541f654371c/weed/command/server.go#L109

jerebear12: Thank you, updated it.

hencoappel:
@jerebear12 What is the intention here? I see exposeDirectoryData option has been added here but it's never passed to NewFilerServer() in filer.go?
Even if it was, it looks like it's always overwritten by the value from 'filer.expose_directory_metadata.enabled'?
Is this expected?

PR #5301, opened 2024-02-14
filer: reduce delete entry ErrNotFound logs

sberthier:

Too many logs when deleting non existing object in filer: "filer: no entry is found in filer store"

Make these "error" silent: only log error when not ErrNotFound

### DIFF
--- a/weed/server/filer_server_handlers_write.go
+++ b/weed/server/filer_server_handlers_write.go
@@ -213,14 +213,12 @@ func (fs *FilerServer) DeleteHandler(w http.ResponseWriter, r *http.Request) {
 
 	err := fs.filer.DeleteEntryMetaAndData(context.Background(), util.FullPath(objectPath), isRecursive, ignoreRecursiveError, !skipChunkDeletion, false, nil)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


weed/server/filer_server_handlers_write.go
```diff
@@ -213,13 +213,13 @@ func (fs *FilerServer) DeleteHandler(w http.ResponseWriter, r *http.Request) {
 
 	err := fs.filer.DeleteEntryMetaAndData(context.Background(), util.FullPath(objectPath), isRecursive, ignoreRecursiveError, !skipChunkDeletion, false, nil)
 	if err != nil {
-		glog.V(1).Infoln("deleting", objectPath, ":", err.Error())
-		httpStatus := http.StatusInternalServerError
 		if err == filer_pb.ErrNotFound {
-			httpStatus = http.StatusNoContent
```
kmlebedev, line 219: Hello, here you don’t have to declare a new variable, but simply pass the status to the function below

sberthier: Yes, better like that

PR #5302, opened 2024-02-14
Fix/s3 config secret

spastorclovr:

This PR fixes couple of issues with s3 and its config as a secret :

1. the filer statefulset would try to mount the secret even if s3 is not enabled
2. the s3 deployment tests `existingConfigSecret` wrongly from `.Values.filer.s3`
3. the secret could be created enven if s3 is not enabled or enableAuth is not on

1. use the secret only when `Values.filer.s3.enabled` is on in `filer-statefulset.yaml`
2. fix wrong condition to test if an existingConfigSecret is used in s3-deployment (.Values.filer.s3.existingConfigSecret vs .Values.s3.existingConfigSecret)
3. rewrite the condtion to create the secret to make sure we check for both s3 and filer.s3 if s3 is enabled and .enableAuth is on + removed the use of `skipAuthSecretCreation` which is redondant with `existingConfigSecret`

# How is the PR tested?

Manually

# Checks

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
+++ b/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
@@ -286,7 +286,7 @@ spec:
         - name: db-schema-config-volume
           configMap:
             name: seaweedfs-db-init-config
-        {{- if .Values.filer.s3.enableAuth }}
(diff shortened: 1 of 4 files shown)


spastorclovr:
Thx for merging @chrislusf when do you think we could have it released into the helm Chart? I understand that the helm Chart is now triggered on new tag only (which is great to avoid breaking changes) :+1:

PR #5403, opened 2024-03-20
Add TLS support, initial etcd connection check, fix key_prefix handling for etcd backend meta storage

sample:

This PR introduces several enhancements to the EtcdStore. The specific issues addressed include:

**1. Lack of TLS Handshake for Secure Communication**: Previously, communication with Etcd did not leverage TLS handshakes, leaving data transmissions potentially vulnerable to eavesdropping or tampering.

**2. Absence of Client Certificate Authentication**: The system did not support client certificate authentication with Etcd, limiting the ability to enforce stricter security policies.

**3. Initial Etcd Connection Check Missing**: The system proceeded without verifying the initial connection to Etcd, which could lead to failures in subsequent operations if the Etcd server was unreachable or misconfigured.

**4. Inaccurate key_prefix Handling in Etcd**: We're fixing a bug where file listing doesn't work correctly with the etcd filer storage if key_prefix is used

**1. Lack of TLS Handshake for Secure Communication:** Previously, communication with Etcd did not leverage TLS handshakes

**2. Absence of Client Certificate Authentication:** The system did not support client certificate authentication with Etcd, limiting the ability to enforce stricter security policies.

**3. Initial Etcd Connection Check Missing**: The system proceeded without verifying the initial connection to Etcd, which could lead to failures in subsequent operations if the Etcd server was unreachable or misconfigured.

**4. Bug file listing doesn't work correctly with the etcd filer storage if key_prefix is used**: The solution is to modify the `ListDirectoryEntries` function to optimize the search process by accurately using Etcd range queries, ensuring that only relevant directory entries are retrieved. Also, redundant bytes.HasPrefix checks have been removed

Additionally, the `WEED_ETCD_KEY_PREFIX` environment variable is added to the `test-etcd-filer.yml` Docker Compose configuration to facilitate testing and ensure that the prefix is correctly applied in test environments.

# How is the PR tested?

Manual testing is carried out in a local Docker environment, specifically by setting different values for `WEED_ETCD_KEY_PREFIX` in `test-etcd-filer.yml` and observing the behavior of the `ListDirectoryEntries` function to ensure it behaves as expected across different scenarios.

# Checks

### DIFF
--- a/weed/filer/etcd/etcd_store.go
+++ b/weed/filer/etcd/etcd_store.go
@@ -1,9 +1,10 @@
 package etcd
 
 import (
-	"bytes"
 	"context"
+	"crypto/tls"
 	"fmt"
+	"go.etcd.io/etcd/client/pkg/v3/transport"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


weed/filer/etcd/etcd_store.go
```diff
@@ -164,8 +164,10 @@ func (store *EtcdStore) ListDirectoryEntries(ctx context.Context, dirPath weed_u
 		return lastFileName, fmt.Errorf("list %s : %v", dirPath, err)
 	}
 
+	etcdKeyWithDirectoryPrefix := append([]byte(store.etcdKeyPrefix), directoryPrefix...)
```
chrislusf: how about adding a `if/else` check to avoid the append operation if `store.etcdKeyPrefix` is empty.

n9:
Hi @sample, in case you are using ETCD, I have added some improvements (mTLS) to ETCD in my fork https://github.com/seaweedfs/seaweedfs/compare/master...n9:seaweedfs:master. Feel free to include them in your PR if they would be useful for you.

n9:
PS: This commit may solve the same problem https://github.com/seaweedfs/seaweedfs/commit/242d5e86a60fc73ca139f14c71b48f362deeb8c4?

sample:
> Hi @sample, in case you are using ETCD, I have added some improvements (mTLS) to ETCD in my fork [master...n9:seaweedfs:master](https://github.com/seaweedfs/seaweedfs/compare/master...n9:seaweedfs:master). Feel free to include them in your PR if they would be useful for you.

this was next task in my todo-list. I will to some research and add this code into pr, if you don't mind

>PS: This commit may solve the same problem https://github.com/seaweedfs/seaweedfs/commit/242d5e86a60fc73ca139f14c71b48f362deeb8c4? yes, it is. do you know if there is any difference in speed between clientv3.WithFromKey and clientv3.WithRange?

n9:
> I will do some research and add this code into pr, if you don't mind

Sure!

> do you know if there is any difference in speed between clientv3.WithFromKey and clientv3.WithRange?

I haven't measured that.

I am using Seaweedfs as a simple storage for K8S cluster(s) that would survive a node degradation (e.g. during cluster upgrades).

3×ETCD + 3×master (STS) + volumes (STS) + filer(s) + CSIs.

(Speed is not the most important metric in my case.)

sample:
1. Added TLS configuration
2. Added initial connection check
3. clientv3.WithFromKey() replaced with clientv3.WithRange(clientv3.GetPrefixRangeEnd(store.etcdKeyPrefix+string(directoryPrefix))) following the example of n9's code

PR #5409, opened 2024-03-22
[mount] fix GetAttr blocks count

### LINKED ISSUE #5408: du -sh file don't meet expectation
**Describe the bug**
Model parsing failed during AI training. At the same time, du -sh model file and found that the size did not meet expectations.

It can be reproduced by du -sh  while reading the file.

**Screenshots**

du
md5

wusongANKANG:

fix #5408 

# How is the PR tested?

# Checks


chrislusf:
Nice fix! Thanks a lot!

PR #5410, opened 2024-03-22
add basic support for sentry error reporting

doc-sheet:
Defaults are chose to be as less impactful as possible.

To enable reporting some environment variables must be set
* SENTRY_DSN (from project settings)
* SENTRY_ENVIRONMENT (production, prod, test or whatever is used in organization)
* SENTRY_RELEASE (`weed@0.27.0`)

All defaults to empty string if not set.

I guess it's better to define Release option at compile time, but I personally prefer to pass it as environment variable.

Lack of monitoring in small environment.

Added support for sentry reporting. May be used with saas Sentry (free or paid) or self-hosted version.

# How is the PR tested?
Hope for gh actions

# Checks

### DIFF
--- a/weed/weed.go
+++ b/weed/weed.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 	"sync"
 	"text/template"
+	"time"
 	"unicode"
 	"unicode/utf8"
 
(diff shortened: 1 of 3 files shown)


weed/weed.go
```diff
@@ -49,6 +51,19 @@ func main() {
 	glog.MaxFileCount = 5
 	flag.Usage = usage
 
+	err := sentry.Init(sentry.ClientOptions{
```
kmlebedev, line 54: Let's initialize if the corresponding environment variables are declared?

doc-sheet: ok. I'll add it in next PR with tweaks to report rates.
Basically sentry sdk will not initialize itself when DSN is empty https://github.com/getsentry/sentry-go/blob/424b37d3354557bab8f4e7eb08c3bb6518bbd489/client.go#L328

But some linters expect to handle returned errors anyway.

PR #5427, opened 2024-03-26
[tests] add action s3 tests

kmlebedev:

avoid commits that break s3 tests

add action for s3 tests

# How is the PR tested?

# Checks

### DIFF
--- a/.github/workflows/s3tests.yml
+++ b/.github/workflows/s3tests.yml
@@ -1,0 +1,181 @@
+name: "Ceph S3 tests"
+
+on:
+  push:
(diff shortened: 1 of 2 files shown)


chrislusf:
Added test is failing. Is this expected?

kmlebedev:
> Added test is failing. Is this expected?

The tests should pass, but now there is clearly a problem with the settings; yesterday there was not enough time to find the error in the configuration.

kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/5431

PR #5458, opened 2024-04-02
[volume] Reduce the number of buffers for uploading one chunk

kmlebedev:

For one uploading file of 64 MB in size, 3 different buffers of 64 MB are allocated

profile023

Help GC and force memory cleanup
https://betterprogramming.pub/memory-optimization-and-garbage-collector-management-in-go-71da4612a960

# How is the PR tested?

# Checks


kmlebedev:
empty server 
profile001

kmlebedev:
after benchmark wrap
```
warp mixed --host=127.0.0.1:8000 --access-key=some_access_key1 --secret-key=some_secret_key1 --autoterm
```
image

chrislusf:
To compare, any profiling before the fix?

kmlebedev:
> To compare, any profiling before the fix?

profile004

chrislusf:
There seem no difference before and after. Also the fix seems not related to the techniques mentioned in the linked doc https://betterprogramming.pub/memory-optimization-and-garbage-collector-management-in-go-71da4612a960.

kmlebedev:
> There seem no difference before and after.
Yes, Now the idea is to convert all buffers to a common pool of buffers, so that reuse would be higher

chrislusf:
What conclusion from the last "only volume" profiling?

Specifically for the changes, if it is useful at all, I suspect just setting the values to nil would also work and no need for the defer clauses.

kmlebedev:
It seems better this way
profile012

kmlebedev:
It seems you can still read range queries in the same pool to allocate memory more efficiently
image

kmlebedev:
with reset and put 
image

image

chrislusf:
> with reset and put

Could not understand why it makes a difference.

kmlebedev:
https://github.com/seaweedfs/seaweedfs/issues/5461

kmlebedev:
> Could not understand why it made a difference.

@chrislusf If after each filling of the buffer you do a reset, then the cache hit drops and the buffer is allocated again.
In fact, we allocate a buffer three times for the same file and due to the cache we can reduce memory consumption by three times.

chrislusf:
let's move the bytes buffer pool to a common package, maybe under `util`? So that everywhere else could use the same pool.

kmlebedev:
after concurrent 20
```
./warp put --insecure --tls --host=192.168.0.119:8443 --access-key=SUPERSECRET --secret-key=SUPERSECRET --autoterm --obj.size 64MiB --concurrent 20 --duration 30s
```
```
/usr/bin/weed -v 0 -logtostderr=true volume -pprof -ip.bind 192.168.0.119 -ip 192.168.0.119 -port 8080 -publicUrl http://192.168.0.119:8080 -max 256 -dir /opt/seaweedfs_volume_01 -mserver 192.168.0.119:9333 -dataCenter DefaultDataCenter -rack DefaultRack -metricsPort 9326 -compactionMBps 64 -concurrentDownloadLimitMB 256 -concurrentUploadLimitMB 0 -preStopSeconds 5 -minFreeSpace 1GiB
```

image

kmlebedev:
before
```
/warp put --insecure --tls --host=192.168.0.119:8443 --access-key=SUPERSECRET --secret-key=SUPERSECRET --autoterm --obj.size 64MiB --concurrent 20 --duration 30s
warp: Benchmark data written to "warp-put-2024-04-04[095639]-lxQ1.csv.zst"

----------------------------------------
Operation: PUT. Concurrency: 20
* Average:115.34 MiB/s, 1.80 obj/s

Throughput, split into 9 x 1s:
 * Fastest:115.3MiB/s, 1.80 obj/s
 * 50% Median:115.3MiB/s, 1.80 obj/s
 * Slowest:115.3MiB/s, 1.80 obj/s
```
profile033

chrislusf:
the "after concurrent 20" does not have throughput info.

kmlebedev:
> let's move the bytes buffer pool to a common package, maybe under util? So that everywhere else could use the same pool.

Done

weed/topology/store_replicate.go
```diff
@@ -87,7 +89,10 @@ func ReplicatedWrite(masterFn operation.GetMasterFn, grpcDialOption grpc.DialOpt
 					pairMap[needle.PairNamePrefix+k] = v
 				}
 			}
-
+			bytesBuffer := buffer_pool.SyncPool.Get().(*bytes.Buffer)
```
chrislusf: Better simplify this line into one function call.

kmlebedev: @chrislusf Done

weed/util/buffer_pool/sync_pool.go
```diff
@@ -0,0 +1,12 @@
+package buffer_pool
+
+import (
+	"bytes"
+	"sync"
+)
+
+var SyncPool = sync.Pool{
```
chrislusf: I mean only expose GetBuffer, PutBuffer functions here.

kmlebedev: @chrislusf done

Issue #5066, opened 2023-12-04
Failed to commit new volume file while performing garbage collection on Windows


JavaNullPointer:
**Describe the bug**
When the system runs for a while, the volume becomes unavailable, and when recreated, it shows that the. bat file is being occupied by other processes.
Causing historical files to be unreadable and writable, and being able to recover after restarting.
Tested other versions on Windows that also have the same issue, such as 3.51 and 3.59.

**System Setup**
Windows 11
version 30GB 3.58 windows amd64
Start with weed -logdir=D:\ims-data\seaweedfs\logs server -s3 -ip=0.0.0.0 -dir=D:\ims-data\seaweedfs\datanode -s3.port=8333 -s3.config=C:\ims-depends\seaweedfs\script\s3_auth_config.json -filer.port=8888 -master.port=9333 -master.volumeSizeLimitMB=30000 -volume.max=1400 -volume.port=9011
Upload a file and delete
Trigger the vacuum

**Expected behavior**
Vacuum completes with no errors.

**Screenshots**

**Additional context**
I1130 14:43:18.806449 volume_layout.go:383 Volume 4 becomes unwritable
I1130 14:43:18.806449 topology_vacuum.go:74 0 Start vacuuming 4 on 0.0.0.0:9011
I1130 14:43:18.822075 topology_vacuum.go:93 0 vacuum 4 on 0.0.0.0:9011 processed 292051360 bytes, loadAvg 0.00%
I1130 14:43:18.868939 needle_map_memory.go:111 loading idx from offset 0 for file: D:\ims-data\seaweedfs\datanode/test-bucket_4.cpx
I1130 14:43:18.931785 topology_vacuum.go:102 Complete vacuuming 4 on 0.0.0.0:9011
I1130 14:43:18.931785 topology_vacuum.go:128 Start Committing vacuum 4 on 0.0.0.0:9011
I1130 14:43:18.931785 volume_vacuum.go:106 Committing volume 4 vacuuming...
E1130 14:43:18.947110 volume_grpc_vacuum.go:89 failed commit volume 4: remove D:\ims-data\seaweedfs\datanode/test-bucket_4.dat: The process cannot access the file because it is being used by another process.
[...]

Issue #4530, opened 2023-06-02
Failed to commit new volume file when running vacuum on Windows


KodamaSakuno:
**Describe the bug**
When the server is committing new volume file, the old one is still opening and can't be deleted, so the vacuum will fail.

**System Setup**
- Windows 11
- version 30GB 3.51 4310e1fac4e59a4391159e6c349f45a8e1cd0601 windows amd64
- Start with `weed server`
- Upload a file and delete
- Trigger the vacuum

**Expected behavior**
Vacuum completes with no errors.

**Additional context**
```
I0603 03:12:21.420840 volume_vacuum.go:106 Committing volume 5 vacuuming...
E0603 03:12:21.421865 volume_grpc_vacuum.go:87 failed commit volume 5: remove C:\Users\***\AppData\Local\Temp/5.dat: The process cannot access the file because it is being used by another process.
```

chrislusf:
Added a possible fix. No windows machine to test it. Please help to check the fix.

KodamaSakuno:
Unfortunately it doesn't work. But I've attached debugger and captured some process activities to that volume file with Process Monitor:

Start server and wait until ready:
| Time of Day      | Operation                       | Result          | Detail                                                                                                                                                                                       |
| ---------------- | ------------------------------- | --------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 13:21:25.4759388 | CreateFile                      | SUCCESS         | Desired Access: Read Attributes, Disposition: Open, Options: Open Reparse Point, Attributes: n/a, ShareMode: Read, Write, Delete, AllocationSize: n/a, OpenResult: Opened                    |
| 13:21:25.4760499 | QueryNetworkOpenInformationFile | SUCCESS         | ***                                                                                                                                                                                          |
| 13:21:25.4760547 | CloseFile                       | SUCCESS         |                                                                                                                                                                                              |
| 13:21:25.4763634 | CreateFile                      | SUCCESS         | Desired Access: Generic Read/Write, Disposition: OpenIf, Options: Synchronous IO Non-Alert, Non-Directory File, Attributes: N, ShareMode: Read, Write, AllocationSize: 0, OpenResult: Opened |
| 13:21:25.4763857 | QueryInformationVolume          | SUCCESS         | ***                                                                                                                                                                                          |
| 13:21:25.4763912 | QueryAllInformationFile         | BUFFER OVERFLOW | ***                                                                                                                                                                                          |
[...]

PR #5459, opened 2024-04-02
fix: always close volume file (#4530)

hencoappel:
If sync fails then close is never called. We should always be calling close on the file.

The issue described in #4530 and #5066. When vacuuming the file is not closed properly so the file could not be deleted.

Always make sure to close the file, even if error occurs during sync.

# How is the PR tested?
By running the seaweed, storing data, clearing it and then letting the vacuum process run.

# Checks

### DIFF
--- a/weed/storage/backend/disk_file.go
+++ b/weed/storage/backend/disk_file.go
@@ -81,13 +81,20 @@ func (df *DiskFile) Close() error {
 	if df.File == nil {
 		return nil
 	}
-	if err := df.Sync(); err != nil {
-		return err
+	err := df.Sync()
+	var err1 error
+	if df.File != nil {
+		// always try to close
+		err1 = df.File.Close()
 	}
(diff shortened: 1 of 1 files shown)


chrislusf:
Actually previously there were no explicit sync operations. Only close operations were used. Supposedly sync should happen inside close?

hencoappel:
The sync was added as part of this change: https://github.com/seaweedfs/seaweedfs/pull/3832

chrislusf:
How was the PR tested on windows? Is there any error logged during the sync?

chrislusf:
It does not hurt to have this. But is it fixing some reproducible cases?

hencoappel:
Yes, it fixes the issue of files handles not being released properly on Windows when vacuuming meaning the .dat files can't be deleted and eventually running out of writable volumes.

PR #5460, opened 2024-04-03
fix completed multiupload lost data

kungf:
If there are putObjectPart requests with the same uploadId during completeMultiPart, it can result in data loss. putObjectPart requests might be due to timeout retries.

S3 completeMultiUpload makes data lost.
If there are putObjectPart requests with the same uploadId during completeMultiPart, it can result in data loss. putObjectPart requests might be due to timeout retries.

Add an uuid id to the part object name to avoid covering, if this partId has retry requests.

# How is the PR tested?
1. I use aws go client SDK to simulate multi-upload, when doing completedMultiUpload to loop putObjectPart one part simultaneously.
2. read the object with readDeleted=false to check whether it has data lost.

# Checks

### DIFF
--- a/weed/s3api/filer_multipart.go
+++ b/weed/s3api/filer_multipart.go
@@ -4,15 +4,16 @@ import (
 	"encoding/hex"
 	"encoding/xml"
 	"fmt"
-	"github.com/google/uuid"
-	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
-	"golang.org/x/exp/slices"
 	"math"
 	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
 	"time"
+
+	"github.com/google/uuid"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
+	"golang.org/x/exp/slices"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/s3"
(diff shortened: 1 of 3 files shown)


kmlebedev:
Who will fix the failed test now?

chrislusf:
@kungf please fix the s3 tests

kungf:
@chrislusf ok

chrislusf:
what is the history of the `test_multipart_resend_first_finishes_last`?

kmlebedev:
```
def test_multipart_resend_first_finishes_last():
    bucket_name = get_new_bucket()
    client = get_client()
    key_name = "mymultipart"

    response = client.create_multipart_upload(Bucket=bucket_name, Key=key_name)
    upload_id = response['UploadId']

    #file_size = 8*1024*1024
    file_size = 8

    counter = Counter(0)
    # upload_part might read multiple times from the object
    # first time when it calculates md5, second time when it writes data
    # out. We want to interject only on the last time, but we can't be
    # sure how many times it's going to read, so let's have a test run
    # and count the number of reads

    fp_dry_run = FakeWriteFile(file_size, 'C',
        lambda: counter.inc()
        )

    parts = []

    response = client.upload_part(UploadId=upload_id, Bucket=bucket_name, Key=key_name, PartNumber=1, Body=fp_dry_run)

    parts.append({'ETag': response['ETag'].strip('"'), 'PartNumber': 1})
    client.complete_multipart_upload(Bucket=bucket_name, Key=key_name, UploadId=upload_id, MultipartUpload={'Parts': parts})

    client.delete_object(Bucket=bucket_name, Key=key_name)

    # clear parts
    parts[:] = []

    # ok, now for the actual test
    fp_b = FakeWriteFile(file_size, 'B')
    def upload_fp_b():
        response = client.upload_part(UploadId=upload_id, Bucket=bucket_name, Key=key_name, Body=fp_b, PartNumber=1)
        parts.append({'ETag': response['ETag'].strip('"'), 'PartNumber': 1})

    action = ActionOnCount(counter.val, lambda: upload_fp_b())

    response = client.create_multipart_upload(Bucket=bucket_name, Key=key_name)
    upload_id = response['UploadId']

    fp_a = FakeWriteFile(file_size, 'A',
        lambda: action.trigger()
        )

    response = client.upload_part(UploadId=upload_id, Bucket=bucket_name, Key=key_name, PartNumber=1, Body=fp_a)

    parts.append({'ETag': response['ETag'].strip('"'), 'PartNumber': 1})
    client.complete_multipart_upload(Bucket=bucket_name, Key=key_name, UploadId=upload_id, MultipartUpload={'Parts': parts})

    _verify_atomic_key_data(bucket_name, key_name, file_size, 'A')

```

kmlebedev:
@kungf You have fixed a special case of the test, and not the general case when the last loaded part number ends up in the final file.

kungf:
hi, @kmlebedev I think the test_multipart_resend_first_finishes_last test case is unreasonable, for uncertain input(two part 1) from the client, an error should be returned。 We should not speculate on which option to use, as our guess may not necessarily be what the user wants. We should return an error and let the user confirm their input.

so, how about deleting this test case, and adding a PR to return ErrInvalidPart in scenarios involving the use of the same part?

kmlebedev:
Let's leave the test, since s3 clients are different and why give an error if you can return the expected result.

 https://github.com/seaweedfs/seaweedfs/pull/5471

kungf:
hi, @kmlebedev which version of s3test seaweedfs use?

PR #5466, opened 2024-04-04
[wip] fix s3test failed test

kungf:

# How is the PR tested?

# Checks


kmlebedev:
@kungf This hack is superfluous if the Etag check conditions work.
image

PR #5479, opened 2024-04-08
fix(volume.fix.replication): adjust volume count, not free volume count

CodeRusher:

In the keepDataNodesSorted function, the calculation of the free volume count uses MaxVolumeCount - VolumeCount, thus it's necessary to adjust VolumeCount, rather than FreeVolumeCount.

# How is the PR tested?

# Checks


CodeRusher:
> @CodeRusher You described a solution to a certain problem. But what practical problem are you solving? What does it express itself in?

I have modified it!

PR #5490, opened 2024-04-10
Unification Extended-key format in Java FilerClient

stillmoon:
When using Filer Http to process files, setting Extended-key must carry the "Seaweed-" prefix, and the prefix will be persisted to the file. 
When using Java FilerClient to process files, setting Extended-key does not force the "Seaweed-" prefix. 
Therefore, the format of Extended-key for the files processed in the two ways is not uniform. 
Therefore, when dealing with Entry in Client, it is necessary to distinguish between the two formats of Extended-key, but it is obviously unrealistic to require users to pay extra attention to the two formats of Extended-key when using FilerClient, and it is necessary to unify the format of Extended-key in Client.

Refer to the implementation of S3, add a certain prefix to the user-defined metadata, and erase the prefix when querying. 
Add a utility class to add and remove "Seaweed-" prefixes for Extended-key. 
The "Seaweed-" prefix is automatically appended when the Entry is submitted, and the "Seaweed-" prefix is automatically removed when the Entry is pulled. 
Achieve the unification of Extended-key format, that is, the Extended-key of Entry will carry the "Seaweed-" prefix, but the user will not display the "Seaweed-" prefix when obtaining Entry (such as lookUpEntry). 
Therefore, when using Client, users are completely unaware of the "Seaweed-" prefix, but the format of Extended is unified.

# How is the PR tested?
add Unit test in SeaweedFilerTest

# Checks
- [ √] I have added unit tests if possible.
- [ √] I will add related wiki document changes and link to this PR after merging.

### DIFF
--- a/other/java/client/src/main/java/seaweedfs/client/ExtendedFormatUtil.java
+++ b/other/java/client/src/main/java/seaweedfs/client/ExtendedFormatUtil.java
@@ -1,0 +1,50 @@
+package seaweedfs.client;
+
+import com.google.protobuf.ByteString;
+import java.util.HashMap;
+import java.util.Map;
+
+/**
+ * A tool class for uniformly formatting the keys of Entry's ExtendedMap <br>@date 2024/4/9 10:44<br>
+ * Process "Seaweed-" prefix, consistent with http upload file situation
+ * <br>
+ *
+ * <b>Premise:</b>
+ * <br>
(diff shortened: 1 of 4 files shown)


chrislusf:
But with this, a user would not be able to set the S3 specific headers with Java FilerClient.

PR #5512, opened 2024-04-20
add sentry support to filer

doc-sheet:

Reporing http errors to sentry collector

Capture http and some common errors and send it to sentry.
Sentry DSN if provided now required to be valid.

# How is the PR tested?

With hope for gh actions.

# Checks

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -10,8 +10,10 @@ import (
 	"strings"
 	"time"
 
+	sentryhttp "github.com/getsentry/sentry-go/http"
 	"google.golang.org/grpc/reflection"
 
+	"github.com/getsentry/sentry-go"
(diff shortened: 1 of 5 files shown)


chrislusf:
There is already Prometheus metrics which we can use for error tracking.

doc-sheet:
Hm. I don't see error-related metrics for weed server.
Sentry also could catch and aggregate different types of errors.
In this PR I added reporting to filer because it is good target to test things, small chance to break anything.

I have plans however to add reporting to master/volume too.

Looks like basic sentry.Init() https://github.com/seaweedfs/seaweedfs/pull/5410 is not enough.


It would be much simpler to add hook to glog, but there is no builtin support in sentry-go. Only with 3rd party package https://github.com/yext/glog-contrib

PR #5530, opened 2024-04-24
fix evacuate volume to different disk types

kungf:
fix evacuate volume to different disk types

set different  types disk free count to -1
 

# How is the PR tested?
volumeServer.evacuate

# Checks


chrislusf:
What is the problem?

kungf:
hi @chrislusf , this  pr stop evacuate volume from hdd to SSD or SSD to HDD, what do you think, does it need to prevent this behaver，if not， I will close this pr.

PR #5537, opened 2024-04-26
fix issue: sometimes volume.fsck report 'volume not found'

ehooi:

sometimes volume.fsck report 'volume not found' when a volume server has multiple disk types

reuse volumeId to VInfo map when is exists.

# How is the PR tested?

I have tested on my cluster.

# Checks

### DIFF
--- a/weed/shell/command_volume_fsck.go
+++ b/weed/shell/command_volume_fsck.go
@@ -653,7 +653,10 @@ func (c *commandVolumeFsck) collectVolumeIds() (volumeIdToServer map[string]map[
 	eachDataNode(topologyInfo, func(dc string, rack RackId, t *master_pb.DataNodeInfo) {
 		for _, diskInfo := range t.DiskInfos {
 			dataNodeId := t.GetId()
-			volumeIdToServer[dataNodeId] = make(map[uint32]VInfo)
+			if _, ok := volumeIdToServer[dataNodeId]; !ok {
+				volumeIdToServer[dataNodeId] = make(map[uint32]VInfo)
+			}
+			volumeCount := 0
 			for _, vi := range diskInfo.VolumeInfos {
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_fsck.go
```diff
@@ -654,9 +658,10 @@ func (c *commandVolumeFsck) collectVolumeIds() (volumeIdToServer map[string]map[
 					isEcVolume: true,
 					isReadOnly: true,
 				}
+				volumeCount += 1
```
chrislusf: Shard count is different from volume count.

ehooi: Sorry, I don't have much information of seaweedfs.
I thought the count here is volume count. normal volume + ec volume (not shard)
Moreover, the reported value is not changed after code changes. (I have ec volumes too in my cluster)

Please let me know how can I fix this.

chrislusf: you can have separate counts for normal volumes and for ec shards.

ehooi:
Thank you for merging.

However, I am confused.
Suppose a node has below files:
```
bucket_680.ec01
bucket_680.ec05
bucket_680.ec09
bucket_680.ec13
```
I thought the node has 4 shards of an EC volume. (volume id=680)

The original code (before I modified it) reported a count of 1, not 4.
So I initially named the counter as `volumeCount`.
normal volume count + ec volume (though it's a fraction) count

The code you modified still reports 1 under the name `ec shards`.
I believe it should be 4 under the name `ec shards`.

If I'm missing something or if I'm mistaken, please let me know.

I found the following code to count ec shards:
```
count += erasure_coding.ShardBits(ecShardInfo.EcIndexBits).ShardIdCount()
```

chrislusf:
the added volume count and ec shard count are only for verbose logging and do not have any actual effect.

PR #5539, opened 2024-04-26
feat: upload files directly without multipart/form-data

KevinBLT:
Hello!

This is my first pull request in golang ever, please review.
It worked locally, although I have the feeling parts of the code between the
two parse functions should be shared by using a separate function.
After extracting the file of a multipart upload, it would be used
by the same function as the direct upload file.

In some cases, the user should be able to submit data into SeaweedFS without creating a `multipart/form-data` (boundary etc). The user would just POST with the corresponding `content-type` header and be done (see JS snippet below as an example).

When an upload  happens, the headers are checked for the `content-type` header,
if a content-type is explicitly set, parsing of `multipart/form-data` is omitted
and the data is uploaded as is, using the `content-type` and `content-disposition` header.

# How is the PR tested?

I never used go before but this feature would be really convenient for me.

So far I have tested with JS fetch function

```javascript
import { readFile } from 'node:fs/promises';

const file = 'test.png', fileId = '7,071075d841';

const r = await fetch('http://127.0.0.1:8080/' + fileId, {
  method: 'POST',
  body: await readFile(file),
  duplex: 'half',
  headers: { 'content-type': 'image/png' }
});

console.log(await r.text());
```

It worked properly and makes POSTing a file much easier as
any code to create a valid `multipart/form-data` request ist not needed.
Just send the file as is since the request itself has the relevant headers.

# Checks

_________________
Discussion: https://github.com/seaweedfs/seaweedfs/discussions/5531

### DIFF
--- a/weed/storage/needle/needle_parse_upload.go
+++ b/weed/storage/needle/needle_parse_upload.go
@@ -44,7 +44,15 @@ func ParseUpload(r *http.Request, sizeLimit int64, bytesBuffer *bytes.Buffer) (p
 	}
 
 	if r.Method == "POST" {
-		e = parseMultipart(r, sizeLimit, pu)
+		contentType := r.Header.Get("Content-Type")
+
+		// If content-type is explicitly set, upload the file without parsing form-data
+		if contentType != "" && !strings.Contains(contentType, "form-data") {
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks! Merging. We can refactor the code in a later PR.

PR #5540, opened 2024-04-27
refactor ifs

kmlebedev:

small refactor https://github.com/seaweedfs/seaweedfs/pull/5539

- if to switch
- string metrod to const

improve readability

# How is the PR tested?

# Checks

### DIFF
--- a/weed/storage/needle/needle_parse_upload.go
+++ b/weed/storage/needle/needle_parse_upload.go
@@ -34,7 +34,7 @@ type ParsedUpload struct {
 }
 
 func isMultiPartFormData(contentType string) bool {
-	return strings.Contains(contentType, "form-data")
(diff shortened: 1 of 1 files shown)


chrislusf:
Go test failed.

KevinBLT:
@kmlebedev

I would assume test fails because at least curl for example sends `Content-Type: multipart/form-data; boundary="----xyz"`, so I think it's not possible to simply test with equality operator here.  

Additionally, I am not sure if the `content-type` is always set to `multipart/form-data` by existing clients, so if it's empty I think it should be treated as `multipart/form-data` for backwards compatibility. 🙂

kmlebedev:
https://github.com/seaweedfs/seaweedfs/pull/5726

PR #5543, opened 2024-04-28
helm: Fix annotations parameters to be standard maps

### LINKED ISSUE #5542: helm s3 ingress templating throwing `wrong type for value; expected string; got map[string]interface {}`
### Describe the bug
https://github.com/seaweedfs/seaweedfs/pull/5463 has introduced a bug for the S3 ingress annotation. After the latest version of the helm chart (`3.65.0`) I get this in the logs: 

Failure Log example:

> Failed to load target state: failed to generate manifest for source 1 of 1: rpc error: code = Unknown desc = Manifest generation error (cached): `helm template . --name-template nextcloud-seaweedfs --namespace nextcloud --kube-version 1.29 --values /tmp/14146e97-0d68-491a-bdbf-864746d34c6a <api versions removed> --include-crds` failed exit status 1: coalesce.go:289: warning: destination for seaweedfs.s3.ingress.annotations is a table. Ignoring non-table value () Error: template: seaweedfs/templates/s3-ingress.yaml:15:19: executing "seaweedfs/templates/s3-ingress.yaml" at <.Values.s3.ingress.annotations>: wrong type for value; expected string; got map[string]interface {} Use --debug flag to render out invalid YAML

#### To Reproduce

Clone the current repo on the master branch and run:
```
# change to the correct directory
cd k8s/charts/seaweedfs
```
Then, try changing the values.yaml to have a standard annotation for `s3.ingress.annotations` (don't forget to set `s3.ingress.enabled` to `true`):
```yaml
s3:
  ingress:
    enabled: true
    annotations:
      nginx.ingress.kubernetes.io/proxy-body-size: 1G
```
Then run `helm template .`:
```console
helm template .
Error: template: seaweedfs/templates/s3-ingress.yaml:15:19: executing "seaweedfs/templates/s3-ingress.yaml" at <.Values.s3.ingress.annotations>: wrong type for value; expected string; got map[string]interface {}

Use --debug flag to render out invalid YAML
```

### System Setup
This is for the helm chart specifically and to do with ingress. I'm using the latest verison released here:
https://github.com/seaweedfs/seaweedfs/releases/tag/3.65

## Expected behavior
The issue is that the correct type is not list `[]` or string `""`.
The correct type for annotations is a map `{}`.

You can verify this by creating a new helm chart with:
```bash
helm create test-chart
```

Then inspect the default `annotations` in values.yaml and then the deployment template:
```console
$ grep annotations test/values.yaml
  annotations: {}
  annotations: {}
  
$ grep annotations -A2 -B1 test-chart/templates/deployment.yaml
      {{- with .Values.podAnnotations }}
      annotations:
        {{- toYaml . | nindent 8 }}
      {{- end }}
```

You can see another example at the nextcloud helm chart [here](https://github.com/nextcloud/helm/blob/30c69c16298c189dd78c930660a17a2b5b774773/charts/nextcloud/values.yaml#L28)

This is the standard so that users can then provide a map in the values.yaml like so:

```yaml
annotations:
  cert-manager.io/cluster-issuer: 'letsencrypt-staging'
```

I'll submit a PR to fix the helm chart.

jessebot:

- fixes #5542 

Please read the above issue to learn more about this PR.

I've converted all annotations back to map types. The reason is that this is the default behavior of helm charts to have annotations be maps. I've also bumped the helm chart version to release the chart.

**Background**: The two PRs that originally changed these were because they said it allowed them to override annotations (https://github.com/seaweedfs/seaweedfs/pull/4799) but that can still be done by setting any of the annotation fields to `{}` if you'd like to not have any annotations, or by just using the standard map to supply your own in your local values.yaml.

The second PR (https://github.com/seaweedfs/seaweedfs/pull/5463) was to make all the annotations behave the same way, which is non-standard.

# How is the PR tested?
Clone the repo, and then run:
```
# change to the correct directory
cd k8s/charts/seaweedfs

# test that templating renders fine, which will render every template
helm template .

# lint the chart
helm lint
```

Then, try changing the values.yaml to have a standard annotation for `s3.ingress.annotations` (don't forget to set `s3.ingress.enabled` to `true`):
```yaml
s3:
  ingress:
    enabled: true
    annotations:
      nginx.ingress.kubernetes.io/proxy-body-size: 1G
```

Now run the `helm template .` again, and you will get all the manifests rendered, but the most important one is this:

<details>
  <summary><code>seaweedfs/templates/s3-ingress.yaml</code></summary>
  
```yaml
---
# Source: seaweedfs/templates/s3-ingress.yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: ingress-seaweedfs-s3
  namespace: argocd
  annotations:
    nginx.ingress.kubernetes.io/proxy-body-size: 1G
  labels:
    app.kubernetes.io/name: seaweedfs
    helm.sh/chart: seaweedfs-3.65.1
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/instance: release-name
    app.kubernetes.io/component: s3
spec:
  ingressClassName: "nginx"
  tls:

      []
  rules:
  - http:
      paths:
      - path: /
        pathType: ImplementationSpecific
        backend:
          service:
            name: seaweedfs-s3
            port:
              number: 8333
              #name:
    host: seaweedfs.cluster.local
```

</details>

# Checks
- [x] I have added unit tests if possible. - not applicable for helm at this time

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer-ingress.yaml
+++ b/k8s/charts/seaweedfs/templates/filer-ingress.yaml
@@ -11,9 +11,9 @@ kind: Ingress
 metadata:
   name: ingress-{{ template "seaweedfs.name" . }}-filer
   namespace: {{ .Release.Namespace }}
-  {{- if .Values.filer.ingress.annotations }}
+  {{- with .Values.filer.ingress.annotations }}
   annotations:
-    {{ tpl .Values.filer.ingress.annotations . | nindent 4 | trim }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
   labels:
     app.kubernetes.io/name: {{ template "seaweedfs.name" . }}
--- a/k8s/charts/seaweedfs/Chart.yaml
+++ b/k8s/charts/seaweedfs/Chart.yaml
@@ -2,4 +2,4 @@ apiVersion: v1
 description: SeaweedFS
 name: seaweedfs
 appVersion: "3.65"
-version: 3.65.0
+version: 3.65.1
--- a/k8s/charts/seaweedfs/templates/master-ingress.yaml
+++ b/k8s/charts/seaweedfs/templates/master-ingress.yaml
@@ -11,9 +11,9 @@ kind: Ingress
 metadata:
   name: ingress-{{ template "seaweedfs.name" . }}-master
   namespace: {{ .Release.Namespace }}
-  {{- if .Values.master.ingress.annotations }}
+  {{- with .Values.master.ingress.annotations }}
   annotations:
-    {{ tpl .Values.master.ingress.annotations . | nindent 4 | trim }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
   labels:
     app.kubernetes.io/name: {{ template "seaweedfs.name" . }}
--- a/k8s/charts/seaweedfs/templates/s3-ingress.yaml
+++ b/k8s/charts/seaweedfs/templates/s3-ingress.yaml
@@ -10,9 +10,9 @@ kind: Ingress
 metadata:
   name: ingress-{{ template "seaweedfs.name" . }}-s3
   namespace: {{ .Release.Namespace }}
-  {{- if .Values.s3.ingress.annotations }}
+  {{- with .Values.s3.ingress.annotations }}
   annotations:
-    {{- tpl .Values.s3.ingress.annotations . | nindent 4 }}
+    {{- toYaml . | nindent 4 }}
   {{- end }}
   labels:
     app.kubernetes.io/name: {{ template "seaweedfs.name" . }}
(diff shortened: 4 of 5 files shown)


k8s/charts/seaweedfs/templates/filer-ingress.yaml
```diff
@@ -11,9 +11,9 @@ kind: Ingress
 metadata:
   name: ingress-{{ template "seaweedfs.name" . }}-filer
   namespace: {{ .Release.Namespace }}
-  {{- if .Values.filer.ingress.annotations }}
+  {{- with .Values.filer.ingress.annotations }}
```
jessebot, line 14: this is changed back to the `with` keyword, as it has an implicit `if` assumed

PR #5566, opened 2024-05-07
Fix deadlock in lock table locks

Woellchen:
This fixes a race condition in which multiple Goroutines release and acquire locks for the same file. Because the lock entries in the lock table and the counters in the lock entries themselves are independent and not synchronized, it can happen that two Goroutines acquire a lock and a third Goroutine releases the same lock in between them, leading to two lock entries being created. Those two lock entries would now wait forever on entry.cond.Wait() and can't proceed.

The fix is to track the locks that are in flight in the lock table itself, so that it knows when it is safe to release a lock before creating a second instance.

# How is the PR tested?
The `FUSE Mount` step in the e2e workflow is failing. This PR should fix it and make it succeed again.

# Checks

### DIFF
--- a/weed/util/lock_table.go
+++ b/weed/util/lock_table.go
@@ -2,17 +2,19 @@ package util
 
 import (
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"sync"
 	"sync/atomic"
+
+	"github.com/seaweedfs/seaweedfs/weed/glog"
 )
 
 // LockTable is a table of locks that can be acquired.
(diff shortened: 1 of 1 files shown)


eliphatfs:
I have a stupid question: does this fix #5380?

Woellchen:
It sounds like it could be a possible fix for it, yes, because this bug led to unfinished IO requests and was blocking the kernel forever. However, I didn't encounter uninterruptible sleep during my tests, so I'm not 100% sure this is the same issue, but could be.

PR #5569, opened 2024-05-08
feat: add log when postPolicyForm check or parse failed

7y-9:

[add log when postPolicyForm check or parse failed]

add log

# How is the PR tested?
no test

# Checks


chrislusf:
Print out the content also for debugging.

7y-9:
ok fine,mayby I lose sight of this.

PR #5581, opened 2024-05-13
Fixes races with tests.

cognusion:

Tests of the lock manager lazily read from concurrently-written variables without using read locks.  This can result in false failures of the tests when the testing system is under heavy load.

Using the provided read locks in the tests.

# How is the PR tested?

Existing test suite passes, and previously failing occurrences under contrived conditions are no longer failing. 

# Checks


chrislusf:
This seems overkill for tests.

cognusion:
I was reluctant to submit it as a PR, as I assumed this would be the response. I will continue to patch tests locally as they fail inconsistently otherwise, and withdraw the PR in the morning.

PR #5602, opened 2024-05-18
Feature/mongodb security

baursn:
added insecure_skip_verify option to mongodb filer store

# How is the PR tested?
tested on my local instance

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -286,6 +286,7 @@ ssl = false
 ssl_ca_file = ""
 ssl_cert_file = ""
 ssl_key_file = "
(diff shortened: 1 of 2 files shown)


kmlebedev:
@baursn  I would recommend doing automatic loading of certificates for https without restarting the process

baursn:
> @baursn I would recommend doing automatic reloading of certificates for https without restarting the process. #
> 
> https://github.com/seaweedfs/seaweedfs/blob/3fae87632f023ee971a869f89075faf4fb94dbe4/weed/command/s3.go#L293-L322
> 
> Maybe the driver itself can replace the certificate during reconnections?

Is there any examples in other connectors, then i can take a look?

PR #5628, opened 2024-05-31
Added context for the MasterClient's methods to avoid endless loops

vadimartynov:
When writing a client, there was a need to complete endless loops in the MasterClient's methods.
It is not possible to rewrite the client in another code, because MasterClient uses hidden variables in methods that are not accessible from the outside.

Added contexts to the MasterClient' methods.
And added a context closure check to the methods.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/shell_liner.go
+++ b/weed/shell/shell_liner.go
@@ -51,7 +51,7 @@ func RunShell(options ShellOptions) {
 
 	if commandEnv.option.FilerAddress == "" {
 		var filers []pb.ServerAddress
-		commandEnv.MasterClient.WithClient(context.Background(), false, func(client master_pb.SeaweedClient) error {
+		commandEnv.MasterClient.WithClient(false, func(client master_pb.SeaweedClient) error {
 			resp, err := client.ListClusterNodes(context.Background(), &master_pb.ListClusterNodesRequest{
 				ClientType: cluster.FilerType,
 				FilerGroup: *options.FilerGroup,
--- a/weed/shell/shell_liner.go
+++ b/weed/shell/shell_liner.go
@@ -46,8 +46,9 @@ func RunShell(options ShellOptions) {
 
 	commandEnv := NewCommandEnv(&options)
 
-	go commandEnv.MasterClient.KeepConnectedToMaster(context.Background())
-	commandEnv.MasterClient.WaitUntilConnected(context.Background())
+	ctx := context.Background()
+	go commandEnv.MasterClient.KeepConnectedToMaster(ctx)
+	commandEnv.MasterClient.WaitUntilConnected(ctx)
 
 	if commandEnv.option.FilerAddress == "" {
 		var filers []pb.ServerAddress
--- a/weed/shell/shell_liner.go
+++ b/weed/shell/shell_liner.go
@@ -46,12 +46,12 @@ func RunShell(options ShellOptions) {
 
 	commandEnv := NewCommandEnv(&options)
 
-	go commandEnv.MasterClient.KeepConnectedToMaster()
-	commandEnv.MasterClient.WaitUntilConnected()
+	go commandEnv.MasterClient.KeepConnectedToMaster(context.Background())
+	commandEnv.MasterClient.WaitUntilConnected(context.Background())
 
 	if commandEnv.option.FilerAddress == "" {
 		var filers []pb.ServerAddress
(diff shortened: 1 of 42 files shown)


weed/wdclient/masterclient.go
```diff
@@ -297,9 +318,9 @@ func (mc *MasterClient) updateVidMap(resp *master_pb.KeepConnectedResponse) {
 		len(resp.VolumeLocation.NewEcVids), len(resp.VolumeLocation.DeletedEcVids))
 }
 
-func (mc *MasterClient) WithClient(streamingMode bool, fn func(client master_pb.SeaweedClient) error) error {
+func (mc *MasterClient) WithClient(ctx context.Context, streamingMode bool, fn func(client master_pb.SeaweedClient) error) error {
```
vadimartynov: I did it incorrectly here. The context must be created inside the retry. I'll fix it now

weed/shell/shell_liner.go
```diff
@@ -46,8 +46,8 @@ func RunShell(options ShellOptions) {
 
 	commandEnv := NewCommandEnv(&options)
 
-	go commandEnv.MasterClient.KeepConnectedToMaster()
-	commandEnv.MasterClient.WaitUntilConnected()
+	go commandEnv.MasterClient.KeepConnectedToMaster(context.Background())
+	commandEnv.MasterClient.WaitUntilConnected(context.Background())
```
chrislusf: should these two calls share the same context?

weed/wdclient/masterclient.go
```diff
@@ -220,39 +235,45 @@ func (mc *MasterClient) tryConnectToMaster(master pb.ServerAddress) (nextHintedL
...
-				glog.V(0).Infof("%s.%s masterClient failed to receive from %s: %v", mc.FilerGroup, mc.clientType, master, err)
-				stats.MasterClientConnectCounter.WithLabelValues(stats.FailedToReceive).Inc()
-				return err
-			}
+			select {
+			case <-ctx.Done():
+				glog.V(0).Infof("Connection attempt to master stopped: %v", ctx.Err())
+				return ctx.Err()
+			default:
+				resp, err := stream.Recv()
```
chrislusf: If there are no messages, which would be normal case most of the time, the ctx.Done() would not be reached, and the context cancelling would not actually work.

weed/shell/shell_liner.go
vadimartynov: The Background context is used, so nothing should break. I can make a general one if context.WithCancel is needed in the future

vadimartynov: Done🙂

weed/wdclient/masterclient.go
vadimartynov: If there is a chance that a goroutine might go into an infinite loop, it is better to add a check. For the service, this won't be an issue, but for writing a client, it could become a problem🙂

(The function runs in the KeepConnectedToMaster goroutine)

chrislusf: "For the service, this won't be an issue" Why is it?

chrislusf:
> @chrislusf , hi! Tell me, please, is there anything else left to do for the pull request?

see https://github.com/seaweedfs/seaweedfs/pull/5628#discussion_r1625063235

weed/wdclient/masterclient.go
vadimartynov: There will be no problems for the service, since the `KeepConnectedToMaster` function runs in a single instance. After the application is completed, the `KeepConnectedToMaster` function running in a single instance will also be completed.

If used in the client, the client call can run a lot of `KeepConnectedToMaster` goroutines. If they don't stop after the client completes, then the only option to stop will be to turn off the application (client).

vadimartynov:
Thank you very much, I didn't notice🙂

weed/wdclient/masterclient.go
chrislusf: Line 243 the `stream.Recv()` is a blocking call. There would be no messages for a long time if there are no messages received, and line 239 `ctx.Done()` would not be reached.

vadimartynov: I changed it to another check. The context completion check is added at the end of an infinite loop.

https://github.com/seaweedfs/seaweedfs/pull/5628/files#diff-19d64e3838987026ca53e25e1814f47fd552bfa15ed43c1a5743a2916daf3618L172-L259

vadimartynov: @chrislusf , Done🙂

PR #5630, opened 2024-06-02
Support concurrent ec.decode

NyaMisty:

ec.decode actually is a CPU-intensive task. We should allow it using more computation power.

Adding a -concurrency option like other commands

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_ec_decode.go
+++ b/weed/shell/command_ec_decode.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/seaweedfs/seaweedfs/weed/pb"
 	"github.com/seaweedfs/seaweedfs/weed/storage/types"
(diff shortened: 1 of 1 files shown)


chrislusf:
need to have a way to ensure limit the number of concurrent ec.decode jobs per server.

NyaMisty:
does that matter? ec.decode dynamically selects the target server, and decoding a volume is actually a CPU-intensive job, I believe there's no need to ensure concurrency for each single server?

chrislusf:
> does that matter? ec.decode dynamically selects the target server, and decoding a volume is actually a CPU-intensive job, I believe there's no need to ensure concurrency for each single server?

it's possible to select the same target server with multiple volumes to decode.

PR #5631, opened 2024-06-02
Support concurrent volume.tier.upload

NyaMisty:

volume.tier.upload actually is a network-intensive task. We should allow it using more network bandwidth.

Adding a -concurrency option like other commands

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_volume_tier_upload.go
+++ b/weed/shell/command_volume_tier_upload.go
@@ -5,6 +5,7 @@ import (
 	"flag"
 	"fmt"
 	"github.com/seaweedfs/seaweedfs/weed/pb"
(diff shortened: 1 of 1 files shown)


chrislusf:
need to have a way to ensure limit the number of concurrent volume.tier.upload jobs per server.

NyaMisty:
I believe the uploading files is not a very fast operation even provided with 10Gbps network. Is it really needed to limit concurrency for each single server?

chrislusf:
> I believe the uploading files is not a very fast operation even provided with 10Gbps network. Is it really needed to limit concurrency for each single server?

The network resource could be limited.

PR #5637, opened 2024-06-02
Better volume.tier.move performance

NyaMisty:

volume.tier.move is slow, so a concurrency algorithm is needed.

We introduced a task queue for each volume server, and make as many volume servers busy at the same time.

# How is the PR tested?

Tested on my cluster. I believe unit test is not needed for this command.

# Checks

### DIFF
--- a/weed/shell/command_volume_tier_move.go
+++ b/weed/shell/command_volume_tier_move.go
@@ -64,6 +64,7 @@ func (c *commandVolumeTierMove) Do(args []string, commandEnv *CommandEnv, writer
 	applyChange := tierCommand.Bool("force", false, "actually apply the changes")
 	ioBytePerSecond := tierCommand.Int64("ioBytePerSecond", 0, "limit the speed of move")
 	replicationString := tierCommand.String("toReplication", "", "the new target replication setting")
+	disableSrcParallelLimit := tierCommand.Bool("disableSrcParallelLimit", false, "limit each src server to only one job")
 
 	if err = tierCommand.Parse(args); err != nil {
 		return nil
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_tier_move.go
```diff
@@ -64,6 +64,7 @@ func (c *commandVolumeTierMove) Do(args []string, commandEnv *CommandEnv, writer
 	applyChange := tierCommand.Bool("force", false, "actually apply the changes")
 	ioBytePerSecond := tierCommand.Int64("ioBytePerSecond", 0, "limit the speed of move")
 	replicationString := tierCommand.String("toReplication", "", "the new target replication setting")
+	disableSrcParallelLimit := tierCommand.Bool("disableSrcParallelLimit", false, "limit each src server to only one job")
```
chrislusf, line 67: how about change this option to the concurrent job count limit?

NyaMisty: Sorry for the late reply. The problem is there's actually two different bottleneck during volume.tier.move, so we actually need both read concurrency limit and write concurrency limit.
For example moving from ssd to hdd, we would like limiting the write concurrency to hdd, but not limiting read concurrency from ssd.

Probably we can change -disableSrcParallelLimit to -concurrency & -srcConcurrency, the -concurrency limits how many copy jobs can execute in parallel, and -srcConcurrency limits how many volumes can be read concurrently from each source volume server.

If you feel good for above change, I'll start to implement it :)

chrislusf: Sounds good!

PR #5647, opened 2024-06-04
Set the capacity of clientChan to 10000

CodeRusher:
Setting the capacity of clientChan to 100 is too conservative; it will block jobs like KeepConnected/SendHeartbeat/ProcessGrowRequest when a large number of events occur.

Set the capacity of clientChan to 10000

# How is the PR tested?

# Checks


chrislusf:
with https://github.com/seaweedfs/seaweedfs/pull/5639 this seems not necessary?

CodeRusher:
> with #5639 this seems not necessary?

This PR resolves the deadlock issue, but if the capacity of the channel is too small and the consum is too slow, wouldn't it block these jobs for a while?

PR #5678, opened 2024-06-17
[fs.verify] getmeta data from filer log

kmlebedev:

A full crawl creates a very excessive load on the filerStore if you need to check the recorded files for the last day, week, month

Get a list of files from the filler log

# How is the PR tested?

localy
```
make server
warp put --host=127.0.0.1:8000 --access-key=some_access_key1 --secret-key=some_secret_key1 --autoterm --duration 30s
fs.verify -v -metadataFromLog -modifyTimeAgo 2000s -concurrency 2
verified 1957 files, error 4178 files 
```

# Checks


chrislusf:
Nice approach!

PR #5698, opened 2024-06-21
merge upstream

chocotan:
create pr


chocotan:
Wrong pr by mistake, please ignore it.

chrislusf:
there are some good ideas in this. You may want to create separate PRs.

chocotan:
Okay, We have currently switched many business scenarios to SeaweedFS, and the switching process is still ongoing. After the system stabilizes, we will select some commits to create separate pull requests.

PR #5711, opened 2024-06-25
Adding tls for http

vadimartynov:

- TLS for the Filer HTTP handler
- Global client for communication (with a secret pool), with override HTTP/HTTPS schemes in requests
- Separate clients prepared for volume, filer, and master services.

# Next Steps:
Due to the use of redirects/proxies in the code, it is difficult to determine where each specific client (master/volume/filer) should be used. Therefore, a general client named global_client was created. In the future, I think it would be better to replace the global client with the appropriate client (master/volume/filer).

# How is the PR tested?
```bash
# I only checked the loading and unloading of data through the filler web interface
# with dev env:
cd docker; make dev_tls
```

# Checks
- [-] I will add related wiki document changes and link to this PR after merging.


kmlebedev:
@vadimartynov Please use the options with auto-renewal of the certificatehttps://github.com/seaweedfs/seaweedfs/blob/8211b29689b2f5a09d2ee244be91658268a687ea/weed/command/s3.go#L299

kmlebedev:
@vadimartynov for local debugging run`cd docker; make dev_tls`

vadimartynov:
@kmlebedev , I saw this comment in the config
https://github.com/seaweedfs/seaweedfs/blob/master/weed/command/scaffold/security.toml#L98-L100
```
# volume server https options
# Note: work in progress!
#     this does not work with other clients, e.g., "weed filer|mount" etc, yet.
```
Is there already work on "work in progress" somewhere?

vadimartynov:
So far, the http_unknown client has been added because I am not sure which functions the various services use. I have prepared HTTP clients for the master, volume, and filler services.

Example for `security.toml`:

```
[https.unknown]
cert = "/certs-for-server/server.crt"
key = "/certs-for-server/server.key"
ca = "/certs-global/rootCA.crt"

[https.volume]
cert = "/certs-for-server/server.crt"
key = "/certs-for-server/server.key"
ca = "/certs-global/rootCA.crt"

[https.master]
cert = "/certs-for-server/server.crt"
key = "/certs-for-server/server.key"
ca = "/certs-global/rootCA.crt"

[https.filer]
cert = "/certs-for-server/server.crt"
key = "/certs-for-server/server.key"
ca = "/certs-global/rootCA.crt"
```

weed/util/http/volume/http_util.go
```diff
@@ -0,0 +1,72 @@
+package volume
```
chrislusf: try to use different package name other than filer, master, volume, http. They are already used or too general.

also, maybe change http_unknown => http_general

vadimartynov: Chris, hello!

> try to use different package name other than filer, master, volume, http. They are already used or too general.

Got it, thanks!

> also, maybe change http_unknown => http_general

I temporarily added `http_unknown`. I was thinking of fixing `http_unknown` to `volume`/`filter`/`master` http clients (if they are used in the code) in the next stage (next commits to this branch).

Can you tell me if there is a way to determine which client functions from `weed/util/http_util.go` were used for?
And what should I pay attention to? For example, are there functions in the code that can simultaneously use `volume`/`filer`/`master` clients (besides `weed/util/http_util.go`)? If there are functions that are used by different clients, then perhaps my solution will be incorrect, and it will be necessary to add client initialization to the config objects of the microservices during initialization (and their transfer to subsequent functions).

chrislusf: Not possible to tell. Maybe add a parameter during execution time.

unmaintained/repeated_vacuum/repeated_vacuum.go
```diff
@@ -35,7 +35,7 @@ func main() {
 	go func() {
 		for {
 			println("vacuum threshold", *garbageThreshold)
-			_, _, err := http_unknown.Get(fmt.Sprintf("http://%s/vol/vacuum?garbageThreshold=%f", pb.ServerAddress(*master).ToHttpAddress(), *garbageThreshold))
+			_, _, err := util_http.Get(util_http.GetGlobalHttpClient(), fmt.Sprintf("http://%s/vol/vacuum?garbageThreshold=%f", pb.ServerAddress(*master).ToHttpAddress(), *garbageThreshold))
```
chrislusf, line 38: make this simpler?

```
util_http.Get(util_http.GetGlobalHttpClient(), "url")
```
into
```
util_http.Get("url")
```

vadimartynov: @chrislusf , hello!

If I replace the function with a method, it will no longer:
- perform JWT authorization by token;
- check the content for gzip (I suspect that gzip is not needed in this case since no data is being transferred);
- handle StatusCode > 500.

The StatusCode check can be done where the function is called. If JWT authorization is not needed, I can change it to a Get method. Or did I misunderstand you?

vadimartynov: Hi @chrislusf,

I guess I misunderstood you. Did you mean to remove the client argument from the `util_http` functions?

Did I understand correctly that I need to remove the client argument for all functions in the `weed/util/http/http_util.go` file (return the arguments to their original state):
- `Post`;
- `Get`;
- `GetAuthenticated`;
- `Head`;
- `Delete`;
- `DeleteProxied`;
- `GetBufferStream`;
- `GetUrlStream`;
- `DownloadFile`;
- `Do`;
- `NormalizeUrl`;
- `ReadUrl`;
- `ReadUrlAsStream`;
- and so on....

vadimartynov: Hi @chrislusf,

Please let me know if there's anything else I need to do to have the PR accepted🙂

vadimartynov:
Split:
- https://github.com/seaweedfs/seaweedfs/pull/5764
- https://github.com/seaweedfs/seaweedfs/pull/5766

PR #5721, opened 2024-06-27
Move cluster role to a separate template.

### LINKED ISSUE #5612: [Helm chart] seaweedfs service account usage is not clear in Filer
**Describe the bug**
We have a cluster in which permissions are restricted to specific namespace only, So we are trying to skip the ClusterRole, ClusterRoleBinding creation using flag `global.createClusterRole` (these will be created by cluster admins prior to the chart installation). 
But in doing so apart from skipping of ClusterRole and ClusterRoleBinding resource creation `serviceAccount` is also being removed from `filer` statefulset due to this condition https://github.com/seaweedfs/seaweedfs/blob/master/k8s/charts/seaweedfs/templates/filer-statefulset.yaml#L53.

So its not really clear if `serviceAccount` is required or not for `filer` to function normally.
If its needed for may be certain scenario(s) don't you think better approach will be to separate rbac resources creation and `serviceAccount` usage ?

**Expected behavior**
RBAC creation and serviceAccount usage is separated.

gtudan:

/closes #5612 
The template has a flag for disabling the cluster role. This has the side-effect that disables the service account,  which breaks the deployment.

I moved the cluster role to its own template to make sure the service account is not accidentally toggled out.

# How is the PR tested?

Installed this without cluster role into a separate namespace.

# Checks

I would love to see a wiki page explaining what the cluster role is for - the comments in the chart suggest that it is used for migration, but I don't really understand what is being migrated. Is that still required?

### DIFF
--- a/k8s/charts/seaweedfs/templates/cluster-role.yaml
+++ b/k8s/charts/seaweedfs/templates/cluster-role.yaml
@@ -1,0 +1,35 @@
+{{- if .Values.global.createClusterRole }}
+#hack for delete pod master after migration
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: {{ .Values.global.serviceAccountName }}-rw-cr
+  labels:
+    app.kubernetes.io/name: {{ template "seaweedfs.name" . }}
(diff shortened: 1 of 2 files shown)


rgaduput:
@gtudan thank you.
@gtudan @chrislusf  appreciate if you can add some info about the usage of the cluster role.

PR #5724, opened 2024-06-29
[Helm chart] Remove createClusterRole dependency from serviceAccount usage from filer statefulset

rgaduput:
Recently through PR https://github.com/seaweedfs/seaweedfs/pull/5721 cluster role creation and service account dependency is removed but there is still left over in filer startefulset where `serviceAccount` usage depends in `createClusterRole` condition which is unnecessary. 
Originally part of issue https://github.com/seaweedfs/seaweedfs/issues/5612
Also added option for automountServiceAccountToken for service account at global level

Remove the condition of `createClusterRole` condition usage from filer statefulset. 

# How is the PR tested?
Helm templated and installed.

# Checks


chrislusf:
@gtudan

gtudan (approved):
Good catch, thanks a lot

PR #5718, opened 2024-06-26
add s3test for sql

kmlebedev:
# What problem are we solving?

split https://github.com/seaweedfs/seaweedfs/pull/5580

# How are we solving the problem?

# How is the PR tested?

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #5736, opened 2024-07-02
fix(filer): unescape on delete

mdenushev:

Deleting/renaming files with spaces in name using S3 doesn't work after 3.66. Caused by adding url escaping on S3 side.

Unescape in filer hander

# How is the PR tested?

# Checks


chrislusf:
Thanks! Will this PR enable some s3 tests currently disabled?

kmlebedev:
@chrislusf Let's get this PR out of the way first?  https://github.com/seaweedfs/seaweedfs/pull/5718

kmlebedev:
@mdenushev  Why can't you separate url escaping logic in s3 from the filler?

kmlebedev:
@mdenushev In my understanding, the filer should not know anything about the encoding. Since this is only on the s3 side. Accordingly, I would like to solve the problem only on the s3 side

chrislusf:
Does this fix a S3 test? If no existing test, needs to add one.

kmlebedev:
> Does this fix a S3 test? If no existing test, needs to add one.

@chrislusf  This is an incorrect fix, since you need to [commit](https://github.com/seaweedfs/seaweedfs/commit/e6bfd061c71ef7ef06e559a6d1d041b997ea633f) to fix it

PR #5760, opened 2024-07-09
[s3] revert skip deletion error, since the error file was not found is already skipped on the side of the grpc function

kmlebedev:

return an error when deleting a file

There should not be a `filer_pb.ErrNotFound` error at this location
since it is already passed on the server side

https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/filer_grpc_server.go#L296

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/s3api_bucket_handlers.go
+++ b/weed/s3api/s3api_bucket_handlers.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


weed/s3api/s3api_bucket_handlers.go
```diff
@@ -346,7 +347,9 @@ func (s3a *S3ApiServer) GetBucketLifecycleConfigurationHandler(w http.ResponseWr
 			Expiration: Expiration{Days: days, set: true},
 		})
 	}
-
+	sort.Slice(response.Rules, func(i, j int) bool {
```
chrislusf: this change seems not related.

kmlebedev: > this change seems not related.

@chrislusf  Yes. I checked. Recently test_lifecycle_get started fluttering due to sorting.
reverted

PR #5764, opened 2024-07-11
Added tls for http handlers

vadimartynov:

- TLS for HTTP handlers

# How is the PR tested?
```bash
# I only checked the loading and unloading of data through the filler web interface
# with dev env:
cd docker; make dev_tls
```

# Checks
- [-] I will add related wiki document changes and link to this PR after merging.

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -1,6 +1,9 @@
 package command
 
 import (
+	"context"
+	"crypto/tls"
+	"crypto/x509"
 	"fmt"
 	"net"
 	"net/http"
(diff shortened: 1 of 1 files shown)


weed/command/filer.go
```diff
@@ -329,15 +341,62 @@ func (fo *FilerOptions) startFiler() {
...
+
+		httpS.TLSConfig = &tls.Config{
+			GetCertificate: fo.GetCertificateWithUpdate,
+			ClientAuth:     clientAuth,
+			ClientCAs:      caCertPool,
+		}
+
+		if filerLocalListener != nil {
+			go func() {
+				if err := httpS.ServeTLS(filerLocalListener, "", ""); err != nil {
```
chrislusf, line 382: TLS seems not necessary if working with localhost

vadimartynov: Hi Chris,

I added it similarly to s3. Example: https://github.com/seaweedfs/seaweedfs/blob/8211b29689b2f5a09d2ee244be91658268a687ea/weed/command/s3.go#L336-L345
It will be needed for another PR: https://github.com/seaweedfs/seaweedfs/pull/5766

PR #5766, opened 2024-07-11
Added tls for http clients

vadimartynov:

- TLS for HTTP clients

# How is the PR tested?
- For testing, changes from PR https://github.com/seaweedfs/seaweedfs/pull/5764 are required.
- Tested in a dev environment consisting of 3 weed server nodes. Config security.toml:
    ```toml
    # HTTPS
    # https client for master/volume/filer connection
    [https.client]
    enabled = true
    cert = "/certs-for-server/server.crt"
    key = "/certs-for-server/server.key"
    ca = "/certs-global/rootCA.crt"

    # volume server https options
    [https.volume]
    cert = "/certs-for-server/server.crt"
    key = "/certs-for-server/server.key"
    ca = "/certs-global/rootCA.crt"

    # master server https options
    [https.master]
    cert = "/certs-for-server/server.crt"
    key = "/certs-for-server/server.key"
    ca = "/certs-global/rootCA.crt"

    # filer server https options
    [https.filer]
    cert = "/certs-for-server/server.crt"
    key = "/certs-for-server/server.key"
    ca = "/certs-global/rootCA.crt"
    disable_tls_verify_client_cert=true
    ```
- As a test, performed data upload/download through the filer's web interface.

# Checks
- [+] I will add related wiki document changes and link to this PR after merging:
   > added info to scaffold


weed/util/http/client/http_client.go
```diff
@@ -0,0 +1,205 @@
...
+		Transport: httpClient.Transport,
+	}
+
+	for _, opt := range opts {
+		opt(&httpClient)
+	}
+	return &httpClient, nil
+}
+
+func GetStringOptionFromSecurityConfiguration(clientName ClientName, stringOptionName string) string {
```
chrislusf: May need to reduce `GetStringOptionFromSecurityConfiguration` visibility since it is not used elsewhere. Ditto for other similar new functions.

weed/util/http/client/http_client.go
```diff
@@ -0,0 +1,205 @@
...
+	onceLoadSecurityConfiguration.Do(func() {
+		util.LoadConfiguration("security", false)
```
chrislusf: `util.LoadConfiguration("security", false)` is also called in many other places. This call may be duplicated, either though it is just one more time.

vadimartynov:
Fixed conflicts via `git rebase master && git push --force`

weed/util/http/client/http_client.go
vadimartynov: Changed the visibility of functions🙂: https://github.com/seaweedfs/seaweedfs/pull/5766/commits/efd944c44e3bd0160d4a32af3f09c028430a456a

weed/util/http/client/http_client.go
vadimartynov: Moved the loading of the security config to a separate function: https://github.com/seaweedfs/seaweedfs/pull/5766/commits/9a04357a743324298df615fa96e4a6a679ed754d

chrislusf: https://github.com/seaweedfs/seaweedfs/commit/9a04357a743324298df615fa96e4a6a679ed754d is a refactoring without changing the execution. [My question](https://github.com/seaweedfs/seaweedfs/pull/5766#discussion_r1674259211) is still the same.

vadimartynov: @chrislusf, hello!

Should we ensure that the `sequrity` call is executed only once in this specific place, or do we need to replace the `sequrity` call throughout the entire code?

If we need to ensure that the `sequrity` call is executed only once in this specific place, then it is already happening as the `sequrity` call is wrapped in a `sync.Once`.

If we need to replace the `sequrity` call throughout the entire code with the `loadSecurityConfig` function, then I think this pull request would be overloaded with additional changes. It would be better to handle this in a separate PR.

chrislusf: > If we need to replace the `sequrity` call throughout the entire code with the `loadSecurityConfig` function, then I think this pull request would be overloaded with additional changes. It would be better to handle this in a separate PR.

This is better and please send a PR for it.

vadimartynov: @chrislusf , hi!
Added PR: https://github.com/seaweedfs/seaweedfs/pull/5792

vadimartynov:
Fixed conflicts via git rebase master && git push --force

weed/util/http/client/http_client.go
vadimartynov: @chrislusf, hi!

Added a fix: https://github.com/seaweedfs/seaweedfs/pull/5766/commits/55a578089773350dd708fb83522d4ddeef34881e

jerebear12:
We had issues after updating to the 3.71 Seaweed release containing this merge. The security.toml file has the https.client property set to true by default which caused our Seaweed cluster to refuse reads and writes. Changing the https.client property to false solved this problem for us.

chrislusf:
cc @vadimartynov 

@jerebear12 what is the `security.toml` file content before and after (removing the credentials of course) ? No change to the server settings?

vadimartynov:
Previously, the `https.client` parameter was not used anywhere in the code. Probably because it was always set to `true` in example (`scaffold`)
https://github.com/seaweedfs/seaweedfs/pull/5766/files#diff-99227e26a33ed46d7e5da4dfa073d750e6156c923d93d4084f693770cdfc410aL97-R104, this happened.

Now the parameter is active, and likely, the error occurred because the parameter was enabled.

jerebear12:
> Previously, the `https.client` parameter was not used anywhere in the code. Probably because it was always set to `true` in example (`scaffold`) https://github.com/seaweedfs/seaweedfs/pull/5766/files#diff-99227e26a33ed46d7e5da4dfa073d750e6156c923d93d4084f693770cdfc410aL97-R104, this happened.
> 
> Now the parameter is active, and likely, the error occurred because the parameter was enabled.

Ah, I understand. This merge does not adjust the value of the https.client parameter, only makes use of it. Yes, the parameter caused the problem. 

1. Should this be considered a breaking change? 
2. Why wasn't there any mention of the negative affects of leaving this unused parameter set to true in the release notes?
3. Should the parameter be set to false by default instead of true?

jerebear12:
> cc @vadimartynov
> 
> @jerebear12 what is the `security.toml` file content before and after (removing the credentials of course) ? No change to the server settings?

The file content is pretty standard accept it doesn't have the latest changes implemented in this merge as we didn't re-run the scaffold command to recreate the file. This means the end of the security.toml file looked like this:

```toml
# volume server https options
# Note: work in progress!
#         this does not work with other clients, e.g., "weed filer|mount" etc, yet.
[https.client]
enabled = true

[https.volume]
cert = '"
key = ""
ca = ""

[https.master]
cert = ""
key = ""
ca = ""
```

After I adjusted the parameter the end of the file of course looked like this:

```toml
# volume server https options
# Note: work in progress!
#         this does not work with other clients, e.g., "weed filer|mount" etc, yet.
[https.client]
enabled = false

[https.volume]
cert = '"
key = ""
ca = ""

[https.master]
cert = ""
key = ""
ca = ""
```

The error we were seeing when attempting to read a file from the Filer with this URL(https://10.100.1.13:8080/35,059c1933ed1232?readDeleted=true) looked like this: "read chunk: Get "https://10.100.1.13:8080/35,059c1933ed1232?readDeleted=true": http: server gave HTTP response to HTTPS client". 

Setting the URL to HTTP instead of HTTPS like this: http://10.100.1.13:8080/35,059c1933ed1232?readDeleted=true, returned the file as expected. This caused us to search for SSL and HTTPS-related commits leading us here.

chrislusf:
@jerebear12 seems the previous `security.toml` file is setting `[https.client] enabled = true` but without the certs. It should not have worked.

jerebear12:
> @jerebear12 seems the previous `security.toml` file is setting `[https.client] enabled = true` but without the certs. It should not have worked.

I agree, however, I think what @vadimartynov said above is that all security.toml files created before the 3.71 release have the `[https.client]` property set to true by default, meaning that everyone who upgrades to 3.71 with a security.toml file created before the most recent release will experience this issue. 

Please correct me if I'm wrong, but if this is the case I believe this problem should be addressed somehow.

aagrxyz:
+1 

I was using version 3.64 and upgraded to 3.72 and suddenly there was no ability to read or write using the fuse mounts. I do think, we should call this out as a note in the release.

vadimartynov:
Guys, I’m sorry, I genuinely didn’t mean to cause any problems. My goal was to add TLS support🙂

I’ve updated the default parameter to ensure no one else runs into this issue - https://github.com/seaweedfs/seaweedfs/pull/6320
In my opinion, Scaffold should be seen more as a guide rather than a ready-to-use configuration

PR #5772, opened 2024-07-12
fix: avoid timeout if datacenter does not exist in topology

kmlebedev:

If filer started earlier than volume or all volumes are temporarily unavailable in one DC. Then we wait 10 seconds for a timeout before a repeat request is sent on assign without specifying the data center.

Added a check that the data center is in the topology

# How is the PR tested?

# Checks

### DIFF
--- a/weed/stats/metrics.go
+++ b/weed/stats/metrics.go
@@ -76,7 +76,7 @@ var (
 			Subsystem: "master",
 			Name:      "volume_layout_total",
 			Help:      "Number of volumes in volume layouts",
-		}, []string{"collection", "replica", "type"})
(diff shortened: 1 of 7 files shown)


weed/stats/metrics.go
```diff
@@ -76,7 +76,7 @@ var (
 			Subsystem: "master",
 			Name:      "volume_layout_total",
 			Help:      "Number of volumes in volume layouts",
-		}, []string{"collection", "replica", "type"})
+		}, []string{"collection", "dataCenter", "type"})
```
chrislusf: this is a different change.

kmlebedev:
@chrislusf change of metrics reverted

PR #5775, opened 2024-07-12
metrics stats of volume layout depends on the data center

kmlebedev:

the metric is not informative when using several data centers

add  datacenter to lables

# How is the PR tested?

# Checks


chrislusf:
this changes the metric meaning. Not backward compatible with existing grafana panels.

kmlebedev:
> this changes the metric meaning. Not backward compatible with existing grafana panels.

It is unlikely that anyone seriously used this, I added as an experiment to observe the statistics total volume count is important to avoid overhead for operations.

chrislusf:
ok. May need to fix grafana after merging this.

Issue #5735, opened 2024-07-01
Optimizing chunk assign when one bucket occupies the entire cluster


kmlebedev:
When a cluster is allocated to one bucket with more than 20 disks, then the volume of date files is distributed randomly across the disks, which leads to distortions. Somewhere it’s empty, and somewhere it’s dense.
Therefore, I would like to have a distribution algorithm that would automatically cover all available disks and evenly distribute active data volume files across them.

<img width="1265" alt="image" src="https://github.com/seaweedfs/seaweedfs/assets/9497591/a0e9c217-cc9f-4138-922e-aeaf7e916c67">

It might be worth using an algorithm https://highscalability.com/consistent-hashing-algorithm/

kmlebedev:
[// randomize the writables](https://github.com/seaweedfs/seaweedfs/blob/master/weed/topology/volume_layout.go#L317)

kmlebedev:
https://github.com/seaweedfs/seaweedfs/blob/04f4b108842742a7b0205a02baed32d3bacbb73f/weed/topology/volume_growth.go#L145

kmlebedev:
@chrislusf Hi, do you remember why this condition?

https://github.com/seaweedfs/seaweedfs/blob/04f4b108842742a7b0205a02baed32d3bacbb73f/weed/shell/command_fs_configure.go#L104

chrislusf:
volumeGrowthCount is physical volume count.

kmlebedev:
@chrislusf I would like to use the volume_growth_count field in PathConf with a value of 0 or -1 as an indication that I need to automatically maintain one volume on all disks in the background.
It`s ok ?

chrislusf:
Maybe have more details on the design first. It sounds good, but maybe a dedicated field could be better.

kmlebedev:
I would like to start by prefactoring the growth of volumes, since at the moment they are allocated synchronously with PUT/POST requests to the filer and the client waits for new volumes to be allocated, this is ok if we are talking about 2-7 volumes, if you need to allocate more than 10 it can be slowly.

Also, according to the current logic, the growth of volumes occurs in a ladder, for example, 7 volumes are filled evenly and at one time become full and it is necessary to request the growth of another 7 volumes, but if at this moment something goes wrong, then recording will not be possible. Therefore, I would like to gradually growth the volumes without waiting for the first 7 to be completely filled

chrislusf:
The first change is worthy and may be good enough if done right.

[...]

PR #5781, opened 2024-07-15
[master] Do Automatic Volume Grow in background

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/5735

Grow volumes continuously in the background

# How is the PR tested?

# Checks


kmlebedev:
@chrislusf Duplicated the function of automatic volume growth, then only in the background. 
Next PR I will select weights, how many volumes to create in the background, and how many in assigning

Issue #5527, opened 2024-04-24
The files on the mounted point are not deleted after the TTL expires


liangyihan:
**Describe the bug**
Create a bucket named "test2" with a TTL (Time To Live) of 1 minute. Upload files into the bucket. After the TTL expires, verify that the files no longer exist inside the bucket. However, the files on the mounted point are still present.

upload file
curl -F file=@Readme.txt "http://10.xxx.xx.x:9283/home/hdpu/lyh/mount_dir/test2/20240424/Readme1.txt"
curl -F file=@Readme.txt "http://10.xxx.xx.x:9283/home/hdpu/lyh/mount_dir/test2/20240424/Readme2.txt"
curl -F file=@Readme.txt "http://10.xxx.xx.x:9283/home/hdpu/lyh/mount_dir/test2/20240424/Readme3.txt"

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".

./weed -logdir=/data/obs/hdd/log/master master -mdir=/data/obs/ssd/data/master -peers=10.xxx.xx.x:9666 -ip=10.xxx.xx.x -port=9666 -defaultReplication=000 -volumePreallocate -volumeSizeLimitMB=80

/weed -logdir=/data/obs/hdd/log/volume volume -index=leveldb -idleTimeout=30 -dir=/data/obs/ssd/data/volume,/data/obs/hdd/data/volume,/data/obs/warm/data/volume,/data/obs/cold/data/volume -disk=ssd,hdd,warm,cold -mserver=10.xxx.xx.x:9666 -dataCenter=dc -rack=rack1 -ip=10.xxx.xx.x -port=9189 -concurrentDownloadLimitMB=2048 -max=0

./weed -logdir=/data/obs/hdd/log/filer filer -master=10.xxx.xx.x:9666 -ip=10.xxx.xx.x -port=9283 -defaultReplicaPlacement=000 -maxMB=4 -s3 -s3.allowEmptyFolder -encryptVolumeData -s3.port=9383 -iam -iam.port=9483

./weed mount -cacheCapacityMB=0 -filer=10.xxx.xx.x:9283 -volumeServerAccess=filerProxy -dir=/home/hdpu/lyh/mount_dir/ -filer.path=/home/hdpu/lyh/mount_dir/

- OS version  centos7
- output of `weed version`  seaweedfs 3.15
- if using filer, show the content of `filer.toml`
[cassandra]
enabled = true
[...]

PR #5783, opened 2024-07-15
Fixed TTL expiration check to trigger file deletion on mounted points

andrea-agostini:
https://github.com/seaweedfs/seaweedfs/issues/5527

Changing delete method calls in filer.go so that they also trigger a NotifyUpdateEvent and have a filer.remote.sync daemon simultaneously delete files on remote mounted point.

# How is the PR tested?
1. Start seaweed master/filer/volume docker images with patched exe, e.g.:
```
services:
  master:
    image: chrislusf/seaweedfs:local
    ports:
      - "9333:9333"
      - "19333:19333"
    command: "master -ip=master -ip.bind=0.0.0.0"
  volume:
    image: chrislusf/seaweedfs:local
    ports:
      - "8089:8080"
      - "18089:18080"
    command: 'volume -mserver="master:9333" -ip.bind=0.0.0.0 -port=8080'
    depends_on:
      - master
  filer:
    image: chrislusf/seaweedfs:local
    ports:
      - "8888:8888"
      - "18888:18888"
    command: 'filer -master="master:9333" -ip.bind=0.0.0.0'
    tty: true
    stdin_open: true
    depends_on:
      - master
      - volume
```
2. From weed shell:
    -  configure S3 credentials and mount some remote S3 bucket:
    ```
    > remote.configure -name=ttltest -type=s3 -s3.access_key=xxxx -s3.secret_key=yyyy -s3.storage_class=STANDARD -s3.region=eu-west-1
    > remote.mount -dir=/buckets/mytest -remote=ttltest/ttl-local-test
   ```
   - configure TTL on said bucket
   ```
   > fs.configure -locationPrefix=/buckets/mytest -ttl=1m -apply
   ```
3. Start a filer.remote.sync to sync filer updates to the bucket:
 ```
./weed filer.remote.sync -filer=localhost:8888 -dir=/buckets/mytest
```
4. Upload a random file to the bucket:
```
curl -F "file=@blah.txt" "http://localhost:8888/buckets/mytest/blah.txt
```
5. Wait for configured TTL to expire and execute fs.ls command from weed shell:
```
>fs.ls /buckets/mytest
```
6. Verify that file has been removed from filer *AND ALSO* deleted on remote /ttl-local-test S3 bucket.

### DIFF
--- a/weed/filer/filer.go
+++ b/weed/filer/filer.go
@@ -334,7 +334,10 @@ func (f *Filer) FindEntry(ctx context.Context, p util.FullPath) (entry *Entry, e
 	entry, err = f.Store.FindEntry(ctx, p)
 	if entry != nil && entry.TtlSec > 0 {
 		if entry.Crtime.Add(time.Duration(entry.TtlSec) * time.Second).Before(time.Now()) {
-			f.Store.DeleteOneEntry(ctx, entry)
+			err = f.doDeleteEntryMetaAndData(ctx, entry, false, false, nil)
+			if err != nil {
+				return nil, err
+			}
 			return nil, filer_pb.ErrNotFound
 		}
 	}
@@ -350,7 +353,10 @@ func (f *Filer) doListDirectoryEntries(ctx context.Context, p util.FullPath, sta
 		default:
 			if entry.TtlSec > 0 {
 				if entry.Crtime.Add(time.Duration(entry.TtlSec) * time.Second).Before(time.Now()) {
-					f.Store.DeleteOneEntry(ctx, entry)
+					err = f.doDeleteEntryMetaAndData(ctx, entry, false, false, nil)
+					if err != nil {
+						return false
+					}
 					expiredCount++
 					return true
 				}

### CODE REVIEW

File: weed/filer/filer.go [hunk 1]
```diff
@@ -334,7 +334,10 @@ func (f *Filer) FindEntry(ctx context.Context, p util.FullPath) (entry *Entry, e
 	entry, err = f.Store.FindEntry(ctx, p)
 	if entry != nil && entry.TtlSec > 0 {
 		if entry.Crtime.Add(time.Duration(entry.TtlSec) * time.Second).Before(time.Now()) {
-			f.Store.DeleteOneEntry(ctx, entry)
+			err = f.doDeleteEntryMetaAndData(ctx, entry, false, false, nil)
```
togettoyou, line 337: In fact, the volume server calls the CollectHeartbeat method every 5 seconds, which checks for expired volumes and deletes them. 

https://github.com/seaweedfs/seaweedfs/blob/master/weed/storage/store.go#L272-L274

The TTL mechanism for the filer server should only manage metadata and not interfere with volumes, as this could trigger a large number of chunk deletion operations, potentially leading to performance issues.

andrea-agostini: @togettoyou I'm definitely a seaweed newbie, but I think what you're saying doesn't fix the problem liangyihan and I have? Please also see this issue: https://github.com/seaweedfs/seaweedfs/issues/5527
I mean, if I understood it right, what you linked above is an expiration check on entire volumes, *not* on single files. Or maybe I was the one misunderstanding from the beginning the -ttl option of the fs.configure command?
https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration

togettoyou: @andrea-agostini My understanding is that SeaweedFS has two TTL mechanisms: one for the metadata database of the `filer server` and another for the volumes on the `volume server`. The TTL set in `fs.configure` will assign a TTL to the volume when files are uploaded. After this, the volume server will trigger the deletion of the volume through a scheduled mechanism. However, the expiration deletion of the metadata on the filer server requires a `list` or `get` operation to be triggered.

togettoyou: It seems that the author of issue #5527 has already fixed it.

PR #5811, opened 2024-07-24
fix allocate reduplicated volumeId to different volume

kungf:
when using HashicorpRaft， it may allocate a reduplicated volume ID to new volume just after restarting masters， the volume ID had been allocated to others。

when master restarts, it does these jobs:
1. restore snap
2. select leader , after this step, if leader master gets a growing request, the nextMaxVolumeId may be the old one in the raft log 
3. sync raft log
4. apply log

when the master became the leader， using raft.Barrier() to wait for all applying logs to be done, then began to serve leader service.

# How is the PR tested?
grow volume after restarting the masters.

# Checks

### DIFF
--- a/weed/server/raft_hashicorp.go
+++ b/weed/server/raft_hashicorp.go
@@ -5,6 +5,14 @@ package weed_server
 
 import (
 	"fmt"
+	"math/rand"
+	"os"
+	"path"
+	"path/filepath"
+	"sort"
+	"strings"
+	"time"
+
 	transport "github.com/Jille/raft-grpc-transport"
 	"github.com/armon/go-metrics"
 	"github.com/armon/go-metrics/prometheus"
@@ -14,13 +22,6 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/pb"
 	"github.com/seaweedfs/seaweedfs/weed/stats"
 	"google.golang.org/grpc"
-	"math/rand"
-	"os"
-	"path"
-	"path/filepath"
-	"sort"
-	"strings"
-	"time"
 )
 
 const (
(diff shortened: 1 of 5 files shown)


kmlebedev:
@kungf Let's separate PR, separate fixed conditions and separate changes to update raft peers? Since this can lead to problems in some extreme cases.

weed/server/raft_hashicorp.go
```diff
@@ -56,46 +57,67 @@ func (s *RaftServer) AddPeersConfiguration() (cfg raft.Configuration) {
...
+					updatePeers = false
 				}
-				if _, found := s.peers[peerLeader]; !found {
-					glog.V(0).Infof("removing old leader peer: %s", peerLeader)
-					s.RaftHashicorp.RemoveServer(raft.ServerID(peerLeader), 0, 0)
+
+				glog.V(0).Infof("raft do barrier")
+				barrier := s.RaftHashicorp.Barrier(2 * time.Minute)
+				if err := barrier.Error(); err != nil {
+					glog.Fatalf("failed to wait for barrier, error %s", err)
```
kmlebedev: This is really not good at all

weed/server/raft_hashicorp.go
```diff
@@ -56,46 +57,67 @@ func (s *RaftServer) AddPeersConfiguration() (cfg raft.Configuration) {
...
+				if updatePeers {
+					s.updatePeers()
+					updatePeers = false
 				}
-				if _, found := s.peers[peerLeader]; !found {
-					glog.V(0).Infof("removing old leader peer: %s", peerLeader)
-					s.RaftHashicorp.RemoveServer(raft.ServerID(peerLeader), 0, 0)
+
+				glog.V(0).Infof("raft do barrier")
+				barrier := s.RaftHashicorp.Barrier(2 * time.Minute)
```
kmlebedev: It seems better to do this when trying to allocate a new ID after changing the leader.
I didn’t want to increase the time needed to change the leader.

kungf: It's better to put barrier in a common place if we have some other read op in the feature.

kungf:
> @kungf Let's separate PR, separate fixed conditions and separate changes to update raft peers? Since this can lead to problems in some extreme cases.

ok

weed/server/raft_hashicorp.go
kungf: this should not err if raft service work fine

weed/server/raft_hashicorp.go
kmlebedev: > It's better to put barrier in a common place if we have some other read op in the feature.

For us, quickly building a new topology of volumes is more critical than reading the log; perhaps we should just stupidly slow down the growth of volumes for example for 2 minutes.

weed/server/raft_hashicorp.go
kmlebedev: > this should not err if raft service work fine

I looked at the code for this function and the error can occur for several reasons. And you need to be able to correctly handle this error, since it affects only one number, maxVolumeID, and it can be double-checked. This is better than sending a message and calling to os.Exit(1)

kungf: yes, you are right. i modify it

kungf:
hi @kmlebedev  update.  
Because only one place can receive the leader change event, so I have to monitor the leader change event in the place update raft peer. 
> Since this can lead to problems in some extreme cases.
which problems can it have?

kungf:
hi @kmlebedev @chrislusf, I need some review, please.

PR #5819, opened 2024-07-26
Check `ShouldGrowVolumes` before returning error in assign.

eliphatfs:
The problem is described in #5818

The master server will still discard the grow request if not necessary in `ProcessGrowRequest`, but the assign request will fail before that.
By checking that the request is not necessary, the assign request and hence the write operation can succeed when there is a lot of volumes that are not crowded, but there are no free volumes to allocate. This usually happens when the cluster is full then cleaned up.

# How is the PR tested?
I tested the PR on our cluster of 8 volume servers + 1 master/filer server. Now the error will not be triggered.

# Checks
- [ ] I have added unit tests if possible. [Not Applicable?]
- [ ] I will add related wiki document changes and link to this PR after merging. [Not Applicable]


kmlebedev:
@chrislusf  This commit significantly slows down the work of this double check synchronously at the moment of writing the file if the datacenter is specified.

chrislusf:
reverted for now e2a07d11d5127edad63bdae12a8f911a29e729c1

PR #5821, opened 2024-07-26
[master] do sync grow request only if absolutely necessary

kmlebedev:

https://github.com/seaweedfs/seaweedfs/pull/5819

Since now requests for growth are made by a background process https://github.com/seaweedfs/seaweedfs/pull/5781
There is no need to perform requests for growth in synchronous mode when dn is available

# How is the PR tested?

# Checks


chrislusf:
Similar logic also exists in master_server_handlers.go

kmlebedev:
> Similar logic also exists in master_server_handlers.go

done

PR #5834, opened 2024-07-29
fix duplicate volumeClaimTemplates keys in Helm chart

kvaps:
Signed-off-by: Andrei Kvapil <kvapss@gmail.com>

When isntalling seaweedfs chart using flux, a kustomize post-processsor used, it shows error of duplicated keys:

```
 seaweedfs-system       4m52s   False   Helm upgrade failed for release tenant-root/seaweedfs-system with chart cozy-seaweedfs@1.10.1: error while running post render on files: map[string]int erface {}(nil): yaml: unmarshal errors:...
```

move standard `volumeClaimTemplates` block under `if not $pvc_exists` check, and add `else` condition

# How is the PR tested?

# Checks


kvaps:
wrong PR, I probably specified both `enablePVC` and `filer.data`

PR #5836, opened 2024-07-29
[filer] Added a version to the title of Filer UI

JiffsMaverick:
In the Filer UI, the version is not indicated in the title, unlike in the Master UI and Volume UI.

I simply added the version display similar to the master and volume. This is just an aesthetic change to appease perfectionism :)

# How is the PR tested?

# Checks


chrislusf:
> This is just an aesthetic change to appease perfectionism :)

It's always better to be perfect.

PR #5844, opened 2024-08-01
fix CVE-2016-2183

zuzuviewer:
CVE-2016-2183

add tls configuration with tls min version, tls max version and cipher suites

# How is the PR tested?

# Checks

### DIFF
--- a/weed/security/tls.go
+++ b/weed/security/tls.go
@@ -199,6 +199,8 @@ func FixTlsConfig(viper *util.ViperProxy, config *tls.Config) error {
 
 func TlsVersionByName(name string) (uint16, error) {
 	switch name {
+	case "":
+		return 0, nil
 	case "SSLv3":
 		return tls.VersionSSL30, nil
 	case "TLS 1.0":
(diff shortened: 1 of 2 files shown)


weed/security/tls.go
```diff
@@ -62,7 +63,22 @@ func LoadServerTLS(config *util.ViperProxy, component string) (grpc.ServerOption
 			RootProvider: serverRootProvider,
 		},
 		RequireClientCert: true,
-		VerificationType:             advancedtls.CertVerification,
+		VerificationType:  advancedtls.CertVerification,
+	}
+	options.MinTLSVersion, err = TlsVersionByName(config.GetString("tls.min_version"))
```
chrislusf, line 68: `tls.min_version` config must exist? How about making them optional?

zuzuviewer: there're all optional, if tls.min_version not set, TlsVersionByName will return default value 0

zuzuviewer: you're right, it's bug, and I fix it now, thank you.

Issue #5877, opened 2024-08-10
Can s3 api support these parameters ?mode=fit&width=200&height=100


rehe0x:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

example of a good issue report:
https://github.com/seaweedfs/seaweedfs/issues/1005
example of a bad issue report:
https://github.com/seaweedfs/seaweedfs/issues/1008

**Describe the bug**

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
- OS version
- output of `weed version`
- if using filer, show the content of `filer.toml`

**Expected behavior**

**Screenshots**

**Additional context**

rehe0x:
<img width="1057" alt="image" src="https://github.com/user-attachments/assets/4f576646-9977-4510-a437-1ddf34340bac">****

chrislusf:
this is not following S3 API and should not merge into master branch.

bwlfhu:
How can we use thumbnails in an S3 environment?

PR #5878, opened 2024-08-10
s3 api support these parameters ?mode=fit&width=200&height=100

rehe0x:

https://github.com/seaweedfs/seaweedfs/issues/5877

# How is the PR tested?
https://github.com/seaweedfs/seaweedfs/issues/5877

# Checks


chrislusf:
This feature does not follow S3 API.

PR #5893, opened 2024-08-15
[master] fix master panic

wusongANKANG:

master panic: fatal error: concurrent map iteration and map write
image

# How is the PR tested?

# Checks


chrislusf:
How about refactor the code to use the lock only inside disk.go file?

wusongANKANG:
> How about refactor the code to use the lock only inside disk.go file?

done

winwill2012:
Is the following issue also related？
image
Reproduction Steps： 
1. start weed shell process
2. run `weed move`
3. move and tail operator has already completed, but source volume server rpc occur exception while deleting source volume

thanks~

wusongANKANG:
@winwill2012  This issue doesn't seem to be related to your problem. This is master panic, while yours is volume panic.

Which version are you using?

PR #5934, opened 2024-08-24
filer.remote.gateway not filer.remote.sync for remote.mount.buckets help string.

sierra-alpha:

Incorrect (I think?) help strings for command documentation.

Correcting the help string.

# How is the PR tested?

untested, doc string change, are there tests for that?

# Checks
- [ ] I will add related wiki document changes and link to this PR after merging. -> I already updated the wiki when it caught me out.


sierra-alpha:
Also related to #5935

sierra-alpha:
I noticed when trying to do a `weed filer.remote.sync -createBucketAt=cloud1` and it complained `flag provided but not defined: -createBucketAt` so a bit of code searching lead me to think a bit of copy paste errors in the doc strings had happened.

PR #5958, opened 2024-09-02
Show volume server which failed to verify a chunk

evgmik:

If volume replication is large then 1 it is useful to know which volume server has problems.

This patch adds volume server info to the log message of `fs.verify` shell command.

# How is the PR tested?

Manually, by created 2 replicas and corrupting a file in one of them.

# Checks

### DIFF
--- a/weed/shell/command_fs_verify.go
+++ b/weed/shell/command_fs_verify.go
@@ -246,7 +246,7 @@ func (c *commandFsVerify) verifyEntry(path string, chunks []*filer_pb.FileChunk,
 					if err := c.verifyChunk(volumeServer, fChunk.Fid); err != nil {
 						if !(*c.metadataFromLog && strings.HasSuffix(err.Error(), "not found")) {
 							fmt.Fprintf(c.writer, "%s failed verify fileId %s: %+v, at volume server %v\n",
-								fileMsg, chunk.GetFileIdString(), err, volumeServer)
+								msg, fChunk.GetFileIdString(), err, volumeServer)
 						}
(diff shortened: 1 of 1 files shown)


weed/shell/command_fs_verify.go
```diff
@@ -245,8 +245,8 @@ func (c *commandFsVerify) verifyEntry(path string, chunks []*filer_pb.FileChunk,
 					defer wg.Done()
 					if err := c.verifyChunk(volumeServer, fChunk.Fid); err != nil {
 						if !(*c.metadataFromLog && strings.HasSuffix(err.Error(), "not found")) {
-							fmt.Fprintf(c.writer, "%s failed verify fileId %s: %+v\n",
-								msg, fChunk.GetFileIdString(), err)
+							fmt.Fprintf(c.writer, "%s failed verify fileId %s: %+v, at volume server %v\n",
+								fileMsg, chunk.GetFileIdString(), err, volumeServer)
```
chrislusf: No need to change to `fileMsg, chunk`

chrislusf (changes requested):

evgmik:
Fixed. Sorry for not catching it up myself. The code looked quite duplicate, so I just copy pasted it without giving it a second thought.

weed/shell/command_fs_verify.go
evgmik: Fixed. Sorry for not catching it up myself. The code looked quite duplicate, so I just copy pasted it without giving it a second thought.

Issue #5954, opened 2024-08-30
WORM (Write once read many) support 


27149chen:
Is there a plan or roadmap to support [WORM (Write once read many)](https://en.wikipedia.org/wiki/Write_once_read_many) ?
I find a related issue (https://github.com/seaweedfs/seaweedfs/issues/4268). But that issue discusses S3, while I am more interested in fuse mount.

chrislusf:
you can run `weed mount -readonly` for readonly mode.

27149chen:
@chrislusf readonly means I can not create any new files, but worm means I can open a new file, write something and close, after that, I can not write it again

27149chen:
@chrislusf  there are some issues in your pr
1. worm files should not be deleted (see https://en.wikipedia.org/wiki/Write_once_read_many)
2. modify time will change, so I can write&close a file again and again in 10 second
3. there are some other ways to modify a file within the mount point, like uploading a file directly thru filer

PR #5983, opened 2024-09-06
improve worm support

27149chen:
https://github.com/seaweedfs/seaweedfs/issues/5954

1. add a new field `Frozen` to filer `Entry`, it is a bool value, if true, the file is readonly and can not be deleted
2. if worm is enabled, set  `Frozen` to true when closing the file
3. when opening a file with write permission, if it is frezen, return EPERM
4. when deleting a file, if it is the last hard link, return EPERM
5. if worm is enabled in filer, write/delete an existing file will return "operation not permitted"

# How is the PR tested?

# Checks


weed/mount/weedfs_file_mkrm.go
```diff
@@ -110,6 +111,8 @@ func (wfs *WFS) Mknod(cancel <-chan struct{}, in *fuse.MknodIn, name string, out
 
 }
 
+var swapFileNameRegexp = regexp.MustCompile(`^\.([a-zA-Z0-9._-]+)\.swp$`)
```
chrislusf: Not generic.

chrislusf (changes requested):

weed/mount/weedfs_file_mkrm.go
27149chen: Do you have any suggestion?

weed/server/filer_server_handlers_write.go
```diff
@@ -197,6 +196,10 @@ func (fs *FilerServer) move(ctx context.Context, w http.ResponseWriter, r *http.
 // curl -X DELETE http://localhost:8888/path/to?recursive=true&ignoreRecursiveError=true
 // curl -X DELETE http://localhost:8888/path/to?recursive=true&skipChunkDeletion=true
 func (fs *FilerServer) DeleteHandler(w http.ResponseWriter, r *http.Request) {
+	if fs.option.WORM {
```
half-life666: We still need a way to delete a WORM file. Consider a file is half written, there is no way to delete it if WORM is enabled.

There are 2 possible options:
1. set up a retention, after the retention, the file is deleted automatically, or can be deleted by DELETE API here. 
2. Implement a new cmd, or force option (or privilieged option, which can be used for special user), wich careful control

Without above, the only thing we can do is to delete entry from filemeta table?

weed/mount/weedfs_file_mkrm.go
half-life666: Just document the consequences when worm is enabled? Then if we have a solution to delete the file (swap or whatever), provide that solution together in the document. 

If above are provided, we can delete the special hanling here?

weed/server/filer_server_handlers_write.go
27149chen: add `?force=true`

chrislusf:
Now I think a better design is to use https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration to configure the WORM. And It does not support specific files such as `.swp`.

27149chen:
> Now I think a better design is to use https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration to configure the WORM. And It does not support specific files such as `.swp`.

@chrislusf what do you mean by "It does not support specific files such as `.swp`", I can not create such files or they do not follow the rules?

weed/mount/weedfs_file_mkrm.go
27149chen: removed. It seams that vi will create temp files not only .swp, but also .swx (x = [a-z]), and other names.
We should document that vi (and other editiors which will create temp files) is not allowed in WORM mode

27149chen:
> Now I think a better design is to use https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration to configure the WORM. And It does not support specific files such as `.swp`.

@chrislusf move the worm flag from filter cmd to fs configure, please take a look

weed/mount/filehandle_read.go
```diff
@@ -47,6 +47,8 @@ func (fh *FileHandle) readFromChunks(buff []byte, offset int64) (int64, int64, e
 	if fileSize == 0 {
 		glog.V(1).Infof("empty fh %v", fileFullPath)
 		return 0, 0, io.EOF
+	} else if offset >= fileSize {
```
chrislusf: unrelated change, separate the PR.

weed/filer/filechunk_group.go
```diff
@@ -46,6 +47,9 @@ func (group *ChunkGroup) AddChunk(chunk *filer_pb.FileChunk) error {
 }
 
 func (group *ChunkGroup) ReadDataAt(fileSize int64, buff []byte, offset int64) (n int, tsNs int64, err error) {
+	if offset >= fileSize {
```
chrislusf: unrelated change, separate the PR.

weed/server/filer_server_handlers_write.go
```diff
@@ -197,7 +196,7 @@ func (fs *FilerServer) move(ctx context.Context, w http.ResponseWriter, r *http.
 // curl -X DELETE http://localhost:8888/path/to?recursive=true&ignoreRecursiveError=true
 // curl -X DELETE http://localhost:8888/path/to?recursive=true&skipChunkDeletion=true
 func (fs *FilerServer) DeleteHandler(w http.ResponseWriter, r *http.Request) {
-
+	force := r.URL.Query().Get("force") == "true"
```
chrislusf: this should not be needed since filer config can be changed

other/java/client/src/main/proto/filer.proto
```diff
@@ -125,6 +125,7 @@ message Entry {
 
     RemoteEntry remote_entry = 10;
     int64 quota = 11; // for bucket only. Positive/Negative means enabled/disabled.
+    bool frozen = 12; // if true, the file is readonly and cannot be unfrozen.
```
chrislusf: change to write_once_read_many here

other/java/client/src/main/proto/filer.proto
```diff
@@ -442,6 +443,7 @@ message FilerConf {
         string data_node = 11;
         uint32 max_file_name_length = 12;
         bool disable_chunk_deletion = 13;
+        bool worm = 14;
```
chrislusf, line 445: change to write_once_read_many here

weed/command/mount.go
```diff
@@ -33,7 +33,7 @@ type MountOptions struct {
 	localSocket        *string
 	disableXAttr       *bool
 	extraOptions       []string
-	writeOnceReadMany  *bool
+	worm               *bool
```
chrislusf: change to writeOnceReadMany here

weed/command/mount.go
```diff
@@ -71,7 +71,7 @@ func init() {
 	mountOptions.debugPort = cmdMount.Flag.Int("debug.port", 6061, "http port for debugging")
 	mountOptions.localSocket = cmdMount.Flag.String("localSocket", "", "default to /tmp/seaweedfs-mount-<mount_dir_hash>.sock")
 	mountOptions.disableXAttr = cmdMount.Flag.Bool("disableXAttr", false, "disable xattr")
-	mountOptions.writeOnceReadMany = cmdMount.Flag.Bool("writeOnceReadMany", false, "write once, read many times")
+	mountOptions.worm = cmdMount.Flag.Bool("worm", false, "wrom mode, if true, you can change a file only once, and then it is readonly and undeletable, see https://en.wikipedia.org/wiki/Write_once_read_many")
```
chrislusf: change to writeOnceReadMany here

weed/command/mount_std.go
```diff
@@ -247,7 +248,7 @@ func RunMount(option *MountOptions, umask os.FileMode) bool {
 		Cipher:             cipher,
 		UidGidMapper:       uidGidMapper,
 		DisableXAttr:       *option.disableXAttr,
-		WriteOnceReadMany:  *option.writeOnceReadMany,
+		WORM:               *option.worm,
```
chrislusf: change to writeOnceReadMany here

weed/filer/entry.go
```diff
@@ -43,6 +43,7 @@ type Entry struct {
 	Content         []byte
 	Remote          *filer_pb.RemoteEntry
 	Quota           int64
+	Frozen          bool
```
chrislusf: change to WriteOnceReadMany here

weed/mount/weedfs_file_mkrm.go
```diff
@@ -130,6 +130,11 @@ func (wfs *WFS) Unlink(cancel <-chan struct{}, header *fuse.InHeader, name strin
 		return code
 	}
 
+	// you can't delete a frozen file in worm mode
+	if wfs.option.WORM && entry != nil && entry.HardLinkCounter <= 1 && entry.Frozen {
```
chrislusf: need to prevent hard linking to a writeOnceReadMany file/dir

chrislusf: in here, wfs.option.WORM should not matter

weed/pb/filer.proto
```diff
@@ -125,6 +125,7 @@ message Entry {
 
     RemoteEntry remote_entry = 10;
     int64 quota = 11; // for bucket only. Positive/Negative means enabled/disabled.
+    bool frozen = 12; // if true, the file is readonly and cannot be unfrozen.
```
chrislusf: change to write_once_read_many

weed/pb/filer.proto
```diff
@@ -442,6 +443,7 @@ message FilerConf {
         string data_node = 11;
         uint32 max_file_name_length = 12;
         bool disable_chunk_deletion = 13;
+        bool worm = 14;
```
chrislusf, line 445: change to write_once_read_many

weed/mount/weedfs_file_mkrm.go
27149chen: there is an example - a worm filesystem which support hardlink: https://docs.oracle.com/cd/E22586_01/html/E22571/gkxss.html

27149chen:
@chrislusf I think worm is better than writeOnceReadMany, because worm is a proper noun, it is a specific file system mode, but writeOnceReadMany is just a feature and cannot indicate that files like this cannot be deleted. see an example here https://docs.oracle.com/cd/E22586_01/html/E22571/gkxss.html

weed/command/mount.go
chrislusf: fix spelling error "wrom mode"

chrislusf: "if true, you can change a file only once, and then it is readonly and undeletable"  simplify this.

weed/filer/entry.go
```diff
@@ -43,6 +43,7 @@ type Entry struct {
 	Content         []byte
 	Remote          *filer_pb.RemoteEntry
 	Quota           int64
+	WORM            bool
```
chrislusf: use "Worm"

chrislusf:
Still I prefer "WriteOnceReadMany". No need to add one more layer of barrier to understand the code, even though it is simple or common.

27149chen:
But how do we know a file can not been deleted if we set "WriteOnceReadMany"?

weed/command/mount.go
27149chen: > "if true, you can change a file only once, and then it is readonly and undeletable" simplify this.

I think it is simple enough

chrislusf:
The design needs a big change.

* If using path specific config, the WriteOnceReadMany flag is not needed.
* However, this would not protect against hardlinks crossing the path boundary. So the hardlinks crossing the path boundary should not be allowed.

27149chen:
ok, I'll remove hardlink support, we can implement it later
But why "If using path specific config, the WriteOnceReadMany flag is not needed."?

chrislusf:
> But why "If using path specific config, the WriteOnceReadMany flag is not needed."?

Only use the path-specific config to determine WriteOnceReadMany flag. When the path-specific config removes the flag, the files would become normal again and no need to "force" delete them.

27149chen:
> > But why "If using path specific config, the WriteOnceReadMany flag is not needed."?
> 
> Only use the path-specific config to determine WriteOnceReadMany flag. When the path-specific config removes the flag, the files would become normal again and no need to "force" delete them.

You means the worm flag in weed mount command?

chrislusf:
> You mean the worm flag in weed mount command?

No. https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration

27149chen:
> > You mean the worm flag in weed mount command?
> 
> No. https://github.com/seaweedfs/seaweedfs/wiki/Path-Specific-Configuration

the worm flag is not needed in path-specific config? I cannot understand, can you explain how to implement worm features using current path-specific options?

chrislusf:
the worm flag is not needed in file entry.

27149chen:
> the worm flag is not needed in file entry.

if worm flag is removed from file entry, how do I know if write is allowed when opening a file?

chrislusf:
> if worm flag is removed from file entry, how do I know if write is allowed when opening a file?

check the path-specific configuration

27149chen:
current flow:
1. first open: writable
2. first close(flush): set worm flag in file entry
3. second open: readonly (by checking the worm flag in file entry)

If worm flag is removed ,how do I know if it is the first open? Can you describe this process?

chrislusf:
If it is first open, it would be a new entry and there would not be existing entry.

27149chen:
> If it is first open, it would be a new entry and there would not be existing entry.

entry is created in Mknod，Open is to open an existing entry

27149chen:
> Check created time equals modified time.

Is the accuracy of these two times in seconds not precise enough?

chrislusf:
Should be enough. Also check 0 file size.

27149chen:
I think a dedicated flag is better than guessing the state by other attributes, a write may finish in one second, also zero size is not equal to the first write

27149chen:
echo "1" > /xxx
result:
```
"Mtime": "2024-09-12T12:44:23+08:00",
"Crtime": "2024-09-12T12:44:23+08:00",
```

chrislusf:
A flag would need more special handling when need to change back to normal file.

chrislusf:
> echo "1" > /xxx
> 
> result:
> 
> ```
> 
> "Mtime": "2024-09-12T12:44:23+08:00",
> 
> "Crtime": "2024-09-12T12:44:23+08:00",
> 
> ```

File size is not zero.

27149chen:
Is it possible to parallelly write a new file? In such case, opening a new file might cause non-zero file size

chrislusf:
> Is it possible to parallelly write a new file? In such case, opening a new file might cause non-zero file size

The purpose for WriteOnceReadMany is to protect against future modifications. This edge case, assuming it is possible, is out of scope.

27149chen:
changes:
1. remove worm flag in Entry, if fileSize != 0 and crtime != mtime, it is not a new file
2. load latest filer conf when weed mount starts, and subscribe new changes
3. check if a file is in a worm enbaled path by rules in filer conf

weed/server/filer_server_handlers_write_autochunk.go
```diff
@@ -121,6 +126,10 @@ func (fs *FilerServer) doPutAutoChunk(ctx context.Context, w http.ResponseWriter
 		contentType = ""
 	}
 
+	if err := fs.checkPermissions(ctx, r, ""); err != nil {
```
jerebear12, line 129: Is passing this empty string intentional?

27149chen: I think it is a mistake, pr: https://github.com/seaweedfs/seaweedfs/pull/6142

PR #5992, opened 2024-09-09
[shell] fix volume grow in shell

kmlebedev:

force volume grow via shell

add VolumeGrow grpc func 

# How is the PR tested?

localy
```
make server
> volume.grow -collection test -dataCenter DefaultDataCenter -rack DefaultRack -dataNode 192.168.3.43:8080 -count 1
```

# Checks


chrislusf:
why a new volume grow grpc API is needed?

kmlebedev:
> why a new volume grow grpc API is needed?

Because the Assign function is created for something else and it is limited to checking shouldGrow, and I want to have grpc logic similar to http volumeGrowHandler

weed/pb/master.proto
```diff
@@ -51,6 +51,8 @@ service Seaweed {
   }
   rpc RaftRemoveServer (RaftRemoveServerRequest) returns (RaftRemoveServerResponse) {
   }
+  rpc VolumeGrow (AssignRequest) returns (VolumeGrowResponse) {
```
chrislusf: create a VolumeGrowRequest and remove unnecessary fields

kmlebedev: done

PR #6007, opened 2024-09-11
prevent dead loop for followers of master node

dsd2077:
prevent dead loop for followers of master node

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/master_grpc_server_volume.go
+++ b/weed/server/master_grpc_server_volume.go
@@ -3,11 +3,12 @@ package weed_server
 import (
 	"context"
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/stats"
 	"math/rand/v2"
 	"strings"
 	"sync"
 	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/stats"
 
 	"github.com/seaweedfs/seaweedfs/weed/topology"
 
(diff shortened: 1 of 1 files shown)


kmlebedev:
@dsd2077  Thanks for noticing, let's just add it `time.Sleep(time.Second * 1)` inside `if !ms.Topo.IsLeader():` ?

dsd2077:
> @dsd2077 Thanks for noticing, let's just add it `time.Sleep(time.Second * 1)` inside `if !ms.Topo.IsLeader():` ?

Hi, kmlebedev.
Recently, I noticed that the sleep time in the auto grow volume logic is too long due to our high throughput capacity. 
https://github.com/seaweedfs/seaweedfs/blob/1cf433abf22a2d3faccd4eaef65678c2e5e01d9b/weed/server/master_grpc_server_volume.go#L43
I’m considering making the sleep time configurable to better adapt to different data scales. My question is: Does the current sleep time have any specific significance? Would it be okay to modify it?

kmlebedev:
> > @dsd2077 Thanks for noticing, let's just add it `time.Sleep(time.Second * 1)` inside `if !ms.Topo.IsLeader():` ?
> 
> Hi, @kmlebedev. Recently, I noticed that the sleep time in the auto grow volume logic is too long due to our high throughput capacity.
> 
> https://github.com/seaweedfs/seaweedfs/blob/1cf433abf22a2d3faccd4eaef65678c2e5e01d9b/weed/server/master_grpc_server_volume.go#L43
> 
> 
> I’m considering making the sleep time configurable to better adapt to different data scales. My question is: Does the current sleep time have any specific significance? Would it be okay to modify it?

Yes, it does matter, but not critically. Now it is 15 minutes and this means that at the moment the filling speed of two volumes should exceed this time. And in fact, you can regulate the volume size. For example, for a 30Gb volume, you need a speed of 66 (30000 * 2 / (60 * 15)) MB per second.  Which is clearly not enough in the case of SSD drives.

Issue #3745, opened 2022-09-26
No local caching of files/chunks on fuse mount


CoolOhm:
**Describe the bug**
When using "./weed mount", the system does not seem to use local caching of the files/chunks only the metadata is cached. Meaning no matter how many times I load/access the same file back-to-back it takes the same time as opposed to only having to pull it from the network once, and after that having fast access from some local on disk cache.

This happens with all files I have tried (the range was from 1kb all the way to 10GB), so I do not think it is only a file size issue.

Looking at the cacheDir, it looks like only metadata is being downloaded or kept as even when accessing several files over 1GB the files in cacheDir are only in the kilobytes.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
- OS version: Redhat Enterprise Linux 7.9
- output of `weed version`: version 30GB 3.29 1ffb1e696e1f651c6294fc34c86b9abb6db1985d linux amd64
- if using filer, show the content of `filer.toml`:  using Filer, but did not make a "filer.toml"
- security.toml: exactly as in wiki, certs made with certstrap
- '3 masters set up with: `./weed server -dir=/seaweedFS/ -master.defaultReplication=010 -filer -master.peers=s2.seaweed.swf.local:9333,s3.seaweed.swf.local:9333 -ip.bind=0.0.0.0 -ip=s1.seaweed.swf.local -rack=rack01 -dataCenter=dc1 -master.volumePreallocate &`
- one test mount set up with:
`sudo  ./weed  mount -filer=s1.seaweed.swf.local:8888,s2.seaweed.swf.local:8888,s3.seaweed.swf.local:8888 -dir=/seaweed-vol/ -filer.path=/  -cacheCapacityMB=25000 -cacheDir=/var/tmp &`

**Expected behavior**
Remote file that have been recently accessed should be cached for faster access, and subsequent access attempts should be near to the speed of local disk access.

**Screenshots**

**Additional context**
I am testing seaweedfs to see if it can be a good replacement for NFS but with added redundancy and security.
looking at my data

chrislusf:
I can not reproduce this.

If possible, can you reproduce it with a docker compose file similar to this? https://github.com/seaweedfs/seaweedfs/blob/master/docker/compose/local-mount-compose.yml

User-26:
Hi, Chris! I'm facing the same issue. Here are the steps to reproduce it:

1. git clone https://github.com/seaweedfs/seaweedfs
[...]

PR #6009, opened 2024-09-12
Feature limit caching to prescribed number of bytes per file

evgmik:

The patch addresses #3745.

In particular it protects against putting the same needle into the cache multiple time. This is done by implementing a new method which checks if fileId is already in the cache.

It also does not create `downloader` if fileId is already in the cache. This would not
be normally triggered prior to the patch since only the 1st needle of a file were subject to caching.

New feature: several needles of the large file are cached as long as they collectively less than `maxFilePartSizeInCache` (new parameter in the Chunk Cache).

Rationale for the above: prior to the parch only the first needle of a file was subject to caching (2MB by default). This is quite small size for a modern data analysis. With this patch,  the cached portion of a file limited to total permitted cache size divided by 4. Which allows a user to be in control of the caching strategy.

# How is the PR tested?

Manual tests seem to work. Major speed up is seen for reading files large then 2MB on consequent requests.

### DIFF
--- a/weed/util/chunk_cache/chunk_cache.go
+++ b/weed/util/chunk_cache/chunk_cache.go
@@ -14,6 +14,7 @@ type ChunkCache interface {
 	ReadChunkAt(data []byte, fileId string, offset uint64) (n int, err error)
 	SetChunk(fileId string, data []byte)
 	IsInCache(fileId string, lockNeeded bool) (answer bool)
+	GetMaxFilePartSizeInCache() (answer uint64)
 }
 
 // a global cache for recently accessed file chunks
@@ -24,6 +25,7 @@ type TieredChunkCache struct {
 	onDiskCacheSizeLimit0 uint64
 	onDiskCacheSizeLimit1 uint64
 	onDiskCacheSizeLimit2 uint64
+	maxFilePartSizeInCache uint64
 }
 
 var _ ChunkCache = &TieredChunkCache{}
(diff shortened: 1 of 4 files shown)


weed/util/chunk_cache/chunk_cache.go
```diff
@@ -39,10 +42,49 @@ func NewTieredChunkCache(maxEntries int64, dir string, diskSizeInUnit int64, uni
...
+func (c *TieredChunkCache) GetMaxFilePartSizeInCache() (answer uint64) {
+	return c.maxFilePartSizeInCache
+}
+
+func (c *TieredChunkCache) IsInCache(fileId string, lockNeeded bool) (answer bool) {
+	if c == nil {
+		return false
+	}
+
+	if lockNeeded {
```
chrislusf, line 59: when can a lock be skipped?

chrislusf (changes requested):

chrislusf:
Tests failed.

weed/util/chunk_cache/chunk_cache.go
evgmik: If it is acquired in a function which calls it, otherwise this will wait for the lock forever.

In particular ` DoSetChunk` function acquires the lock before calling the check.

evgmik:
Fixed the test cases. Sorry about it. Have not use go for a couple years, still rusty on the steps before PR.

PR #6046, opened 2024-09-20
add pprof for master

dsd2077:
In some situation, we need pprof to monitor the status of master server. So I add pprof for master.

Add a http sock for pprof.

# How is the PR tested?

# Checks


kungf:
weed/command/imports.go has import pprof, if you set -metricsPort when start master, you can get the pprof

chrislusf:
does this work with `weed server` also?

dsd2077:
> weed/command/imports.go has import pprof, if you set -metricsPort when start master, you can get the pprof

Thanks for your reply. I have a question "the help manual said metricsPort parameter is for Prometheus. How can I obtain the pprof file?"

kungf:
> > weed/command/imports.go has import pprof, if you set -metricsPort when start master, you can get the pprof
> 
> Thanks for your reply. I have a question "the help manual said metricsPort parameter is for Prometheus. How can I obtain the pprof file?"

curl http://{masterIp}:{metricsPort}/debug/pprof

kungf:
> does this work with `weed server` also?

yes, I tested

dsd2077:
> does this work with `weed server` also?

No. Because ```weed server``` already has it. So does ```weed filer``` and ```weed volume```.

dsd2077:
> > does this work with `weed server` also?
> 
> yes, I tested

I see. Thank you so much.

Issue #6040, opened 2024-09-18
persist readonly state stay after volume.move


SmoothDenis:
**Describe the bug**
PR https://github.com/seaweedfs/seaweedfs/pull/5977
adds the read_only state to .vif file of the server volume

```
{
  "files":  [],
  "version":  3,
  "replication":  "",
  "BytesOffset":  4,
  "datFileSize":  "0",
  "DestroyTime":  "0",
  "readOnly":  true
}
```

unfortunately, volume.move moves this file to a new server and as a result, the file remains in the read_only state despite a successful move

affects admin functions such as fix.replication, volume.move

**System Setup**
verison 3.73

**Expected behavior**
read_only should be saved only if it was assigned initially, and not as a result of temporarily setting the flag

**Additional context**
```
exec -it fast-master-0 -- sh -c "echo 'volume.list -volumeId 99321' | weed shell | grep 99321 -A 2"
Defaulted container "seaweedfs" out of: seaweedfs, vault-init (init)
I0918 09:25:54.027097 masterclient.go:228 master s3-fast-mserver.query.consul:9333 redirected to leader fast-master-2.s3-fast-master-direct.service.stagedc.consul:9333
.          volume id:99321  size:8  collection:"sms-commission-reports"  replica_placement:10  version:3  modified_at_second:1726643946
        Disk hdd total size:8 file_count:0
      DataNode fast-volume-3.s3-fast-volume.service.stagedc.consul:8080 total size:8 file_count:0
--
          volume id:99321  size:8  collection:"sms-commission-reports"  replica_placement:10  version:3  modified_at_second:1726643946
        Disk hdd total size:8 file_count:0
      DataNode fast-volume-0.s3-fast-volume.service.stagedc.consul:8080 total size:8 file_count:0


volume.move -source fast-volume-0.s3-fast-volume.service.stagedc.consul:8080 -target fast-volume-2.s3-fast-volume.service.stagedc.consul:8080 -volumeId 99321
2024/09/18 09:26:30 copying volume 99321 from fast-volume-0.s3-fast-volume.service.stagedc.consul:8080 to fast-volume-2.s3-fast-volume.service.stagedc.consul:8080
2024/09/18 09:26:30 tailing volume 99321 from fast-volume-0.s3-fast-volume.service.stagedc.consul:8080 to fast-volume-2.s3-fast-volume.service.stagedc.consul:8080
[...]

PR #6058, opened 2024-09-24
fix(volume): don't persist RO state in specific cases

mdenushev:
#6040

Add flag `persist` for store.MarkVolumeWritable. Persist flag is true only when marking disk with command volume.mark, volume.move/volume.fsck/volume.tier.move/volume.tier.upload does not persist in vif file.

# How is the PR tested?
Locally

1. Run cluster in docker 
2. Upload any file
3. Call `volume.move` on non-empty volume

# Checks
- [ ] ~~I have added unit tests if possible.~~ - No, but I would like to add some admin commands related tests, should discuss it.
- [ ] ~~I will add related wiki document changes and link to this PR after merging.~~


weed/storage/store.go
```diff
@@ -470,26 +470,30 @@ func (s *Store) HasVolume(i needle.VolumeId) bool {
...
-func (s *Store) MarkVolumeWritable(i needle.VolumeId) error {
+func (s *Store) MarkVolumeWritable(i needle.VolumeId, persist bool) error {
 	v := s.findVolume(i)
 	if v == nil {
 		return fmt.Errorf("volume %d not found", i)
 	}
 	v.noWriteLock.Lock()
 	v.noWriteOrDelete = false
-	v.PersistReadOnly(false)
+	if persist {
```
chrislusf: this is not needed

PR #6061, opened 2024-09-25
support dameng as filer store

Vegetable540:
support dameng as filer store

add dameng store

# How is the PR tested?
use test_suite.go。but I modified the testcase 

# Checks
- [y] I have added unit tests if possible.
- [y ] I will add related wiki document changes and link to this PR after merging.


chrislusf:
@Vegetable540 are you connected with dameng? The `gitee.com/chunanyong/dm` library does not have license specified.

Vegetable540:
> @Vegetable540 are you connected with dameng? The `gitee.com/chunanyong/dm` library does not have license specified.

I'm not connected with Dameng.The library is from the Dameng website, and there doesn't seem to be any license statement available there. https://eco.dameng.com/download/

chrislusf:
There is a 10% size increase to include this library:

```
weed$ ls -al `which weed`
-rwxr-xr-x@ 1 chrislu  staff  89842178 Oct 21 22:35 /Users/chrislu/go/bin/weed
weed$ go install
go: downloading gitee.com/chunanyong/dm v1.8.16
weed$ go install
weed$ ls -al `which weed`
-rwxr-xr-x@ 1 chrislu  staff  100483762 Oct 21 22:37 /Users/chrislu/go/bin/weed
```

We do not need to include this in the default version, but we can add build tags to include this in the full version, similar to `elastic,gocdk,rclone,sqlite,tikv,ydb`

Issue #6062, opened 2024-09-25
The `unclaimedSpaces` calculation logic in `MaybeAdjustVolumeMax()` is incorrect.


shichanglin5:
Currently, the volume server calculates the unclaimed space based on `freeSpace - unusedSpace`.

If preallocate is enabled, the disk space is pre-allocated, so freeSpace is already reduced. The file size obtained through statfs reflects the actual data written, excluding the pre-allocated size, resulting in an incorrect calculation.

Therefore, if preallocate is enabled, unusedSpace should not be subtracted. Currently, unusedSpace is calculated as `volumeSizeLimit - stat.Size()`.

<img width="848" alt="image" src="https://github.com/user-attachments/assets/58800c83-67d7-4262-baad-64f669a9edd9">

PR #6063, opened 2024-09-25
fix unclaimed spaces calculation when volumePreallocate is enabled

shichanglin5:
fix https://github.com/seaweedfs/seaweedfs/issues/6062

when calculating unclaimedSpaces in volume, it should determine whether to subtract unusedSpace based on the preallocate setting. 

Therefore, the volumeServer needs to add `preallocate` filed to store, and it should be initialized along with`volumeSizeLimit` when receiving a heartbeat response from the master.


weed/pb/master.proto
```diff
@@ -90,6 +90,7 @@ message HeartbeatResponse {
   uint32 metrics_interval_seconds = 4;
   repeated StorageBackend storage_backends = 5;
   repeated string duplicated_uuids = 6;
+  bool preallocate = 7;
```
chrislusf, line 93: how about just sending the preallocate size?

shichanglin5: 如果 preallocate 禁用，那么 preallocateSize 是 0，这种情况，volume server 就无法获取 volume size limit 了

PR #6068, opened 2024-09-27
feat: add col/mark master api to mark collection writable/readonly

zouyixiong:

A new master api `col/mark?collection=***&flag=writable|readonly` to support collection live control out of the service.

In my usage, I create collection daily, and write to it,  after the day, the collection can be mark as readonly.

(volume server config with max = 0)
I found that, the max available volumes will the affect by the existed writable but not full volumes. So, it's helpful for calc a more correct newMaxVolumeCount if I mark some out of day collection as readonly.

image

a new api

# How is the PR tested?
tested in my dev env.

# Checks


chrislusf:
In general, no more http APIs. Only add gRPC APIs.

zouyixiong:
OK, I will change it later.

zouyixiong:
@chrislusf  Updated to grpc api, please check it again.

add a small test here: test/grpc/master/master-grpc-client.go

chrislusf:
* there is already `VolumeMarkReadonly` API

PR #6092, opened 2024-10-03
s3api: Fix signature v4 with reverse proxy at sub-path

### LINKED ISSUE #6086: [S3] Signature V4 doesn't work with reverse proxy
**Describe the bug**

When you try to use S3 signature v4 with URL like `https://your.server/s3`, Seaweed throws error:
> ERROR: S3 error: 403 (SignatureDoesNotMatch): The request signature we calculated does not match the signature you provided. Check your key and signing method.

When try to access directly by port or by using v2 (`signature_v2 = True` in s3cmd), everything works fine.

**System Setup**
- output of `weed version`: was 3.72 during tests but seems like latest 3.74 still have this bug.

**Additional context**

`/s3` is reverse proxy to s3 port, for Caddy it's something like
```caddy
redir /s3 /s3/
handle_path /s3/* {
	reverse_proxy localhost:8333
}
```

er2off:

This PR closes #6086

Server checks if `X-Forwarded-Prefix` header was passed, in this case add additional URL signature check.
Additionally, `X-Forwarded-Host` is handled in same way as in f747767aa10da345c19a61a63687d52876efe6a5.

# How is the PR tested?

Tried to use s3cmd, aws-cli and rclone, everything worked fine.

# Checks

### DIFF
--- a/weed/s3api/auth_signature_v4.go
+++ b/weed/s3api/auth_signature_v4.go
@@ -148,15 +148,38 @@ func (iam *IdentityAccessManagement) doesSignatureMatch(hashedPayload string, r 
 		}
 	}
 
+	if forwardedPrefix := r.Header.Get("X-Forwarded-Prefix"); forwardedPrefix != "" {
+		// Handling usage of reverse proxy at prefix. Note that it's an undefined behavior for AWS S3 and not supported in MinIO.
+		// Trying with prefix before main path.
+
+		// Get canonical request.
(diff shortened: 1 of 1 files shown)


er2off:
For s3cmd it's required to set `X-Forwarded-Host` header because it can't properly subdivide host and I can't detect if it was used. Luckily, aws-cli and rclone works as intended.

PR #6094, opened 2024-10-06
feat: Add support for "custom" dataDirs type for volume server in kubernetes helm chart

TheHackmeister:
The helm chart doesn't seem to support running multiple volume servers with host paths for mounts. 

I have 5 drives mounted in one machine. With a replication of 000 I could mount all the drives into the volume server and it would work. But to support replication between drives I found that really needed multiple volume servers. When I tried to increase the number of replicas I found the mounts weren't generic enough and I could get 1 pod running but not any more. 

Ultimately I could use `extraVolumeMounts` and `subPathExpr: $(POD_NAME)` to generically mount in the pod, but  the dataDirs logic always created a volumeMount that conflicted with anything I put in extraVolumeMounts. 

Added a "custom" type for the dataDirs parameter to allow use of extraVolumes and extraVolumeMounts 

After looking over the code I realized there was just one place I needed to change to support use of extraVolumeMounts as a dataDirs option. It was to skip adding the dataDirs volume to the volumeMount if the type is custom. 

# How is the PR tested?
Needed for my cluster, so that's where I've tested it. You can find my implementation here: https://github.com/OwnYourIO/SpencersLab/blob/9be00356009ee30951da16e7fde31ffc9c30224c/services/infra/prod/values.yaml#L388-L398

Probably worth noting I wrote this code in the wee hours and haven't really deployed any data yet. That'll come this week. 

# Checks

# Other comments
- I put some documentation in the code itself. Would love some pointers on where else documentation might go. 
- This adds support for using extraVolumeMounts for dataDirs but another approach would be to add support for subPathExpr directly.

### DIFF
--- a/k8s/charts/seaweedfs/values.yaml
+++ b/k8s/charts/seaweedfs/values.yaml
@@ -301,9 +301,9 @@ volume:
   #   - name: data
   #     type: "emptyDir"
   #     maxVolumes: 0  # If set to zero on non-windows OS, the limit will be auto configured. (default "7")
-  # 
+  #
   # If these don't meet your needs, you can use "custom" here along with extraVolumes and extraVolumeMounts
-  # Particularly useful when using more than 1 for the volume server replicas. 
+  # Particularly useful when using more than 1 for the volume server replicas.
   #   - name: data
   #     type: "custom"
   #     maxVolumes: 0  # If set to zero on non-windows OS, the limit will be auto configured. (default "7")
@@ -394,7 +394,7 @@ volume:
   #     subPathExpr: $(POD_NAME)
   # extraVolumes: |
   #   - name: drive
-  #     hostPath: 
+  #     hostPath:
   #       path: /var/mnt/
   extraVolumes: ""
   extraVolumeMounts: ""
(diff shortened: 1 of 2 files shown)


k8s/charts/seaweedfs/values.yaml
```diff
@@ -381,6 +387,15 @@ volume:
...
+  # Example for use when using more than 1 volume server replica
+  # extraVolumeMounts: |
+  #   - name: drive
+  #     mountPath: /drive
+  #     subPathExpr: $(POD_NAME)
+  # extraVolumes: |
+  #   - name: drive
+  #     hostPath:
+  #       path: /var/mnt/
   extraVolumes: ""
```
chrislusf, line 399: `extraVolumes` and `extraVolumeMounts` should not be strings?

TheHackmeister: That was my first thought too. But when I tried to deploy I got this error: 

Error: template: infra/charts/seaweedfs/templates/volume-statefulset.yaml:213:26: executing "infra/charts/seaweedfs/templates/volume-statefulset.yaml" at <.Values.volume.extraVolumeMounts>: wrong type for value; expected string; got []interface {} Use --debug flag to render out invalid YAML


So then I changed it to a string and it worked. I wasn't super surprised as it's a common enough pattern in k8s, especially when extra variable replacement needs to be done (like the $(POD_NAME) here). 

Happy to pull more info if you're curious, that just was the relevant bit.

PR #6154, opened 2024-10-22
stop renaming worm files

27149chen:
renaming should not be allowed for worm files/dirs, since a file may be moved from a worm path to a non-worm path

check if it is a worm file/dir before renaming

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/filer_server_handlers_write.go
+++ b/weed/server/filer_server_handlers_write.go
@@ -160,6 +160,14 @@ func (fs *FilerServer) move(ctx context.Context, w http.ResponseWriter, r *http.
 		return
 	}
 
+	rule := fs.filer.FilerConf.MatchStorageRule(src)
+	if rule.Worm {
+		// you cannot move a worm file or directory
+		err = errors.New("operation not permitted")
+		writeJsonError(w, r, http.StatusForbidden, err)
+		return
+	}
+
 	oldDir, oldName := srcPath.DirAndName()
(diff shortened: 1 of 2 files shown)


weed/server/filer_server_handlers_write.go
```diff
@@ -160,6 +160,14 @@ func (fs *FilerServer) move(ctx context.Context, w http.ResponseWriter, r *http.
 		return
 	}
 
+	rule := fs.filer.FilerConf.MatchStorageRule(src)
+	if rule.Worm {
+		// you cannot move a worm file or directory
+		err = errors.New("operation not permitted")
```
chrislusf: be a bit more specific about the error reason

27149chen: updated

weed/server/filer_server_handlers_write.go
```diff
@@ -160,6 +160,14 @@ func (fs *FilerServer) move(ctx context.Context, w http.ResponseWriter, r *http.
 		return
 	}
 
+	rule := fs.filer.FilerConf.MatchStorageRule(src)
+	if rule.Worm {
+		// you cannot move a worm file or directory
+		err = fmt.Errorf("cannot move entry from '%s' to '%s': operation not permitted", src, dst)
```
chrislusf:
```suggestion
		err = fmt.Errorf("cannot move write-once entry from '%s' to '%s': operation not permitted", src, dst)
```

Issue #6146, opened 2024-10-20
gRPC Volume Server BLOB read/write API


ollelogdahl:
**Case**
The workflow described in the main README document is not out of the box achievable trough gRPC (see https://github.com/seaweedfs/seaweedfs/discussions/2297). This includes reading & writing a BLOB to a volume server, after it has been assigned trough the master. Currently, it seems like only the http api supports this trough the following:
```
curl -F file=@/home/chris/myphoto.jpg http://127.0.0.1:8080/3,01637037d6
```

In the previously linked discussion, it seems like running the filer server or manually implementing write is the solution. I was wondering if there are any reasons why gRPC read/write is not supported, or if the project would be open for a patch adding at least read & write.

**Why**
The change could improve interoperability between other distributed systems also using gRPC, by not requiring them to also bundle a http client with multipart support. This would also not require bundling the filer server just for uploads, making the system lighter.

**How**
My thought would be to simply copy the behaviour from the volume server http handler, and instead take parameters from a protobuf object `ReadRequest`/`WriteRequest`.

**Relevant code**
- [weed/pb/volume_server.proto:10](https://github.com/seaweedfs/seaweedfs/blob/d04859b01c2128389b9a81007e231a0b2db091d7/weed/pb/volume_server.proto#L10)
- [weed/server/volume_server_handlers_write.go:18](https://github.com/seaweedfs/seaweedfs/blob/d04859b01c2128389b9a81007e231a0b2db091d7/weed/server/volume_server_handlers_write.go#L18C23-L18C78) for current implementation of the http write API.

chrislusf:
This may be a good idea. A few points to keep in minde:
* do add some tests to cover this, since it is not used in normal code path.
* avoid scope creeping. Not sure how much code change it may incur.

ollelogdahl:
@chrislusf great to hear. Suprised this hasn't been commented on earlier. I'm thinking a first pass would be Write/Read, but I guess in the long run more API:s could be supported trough gRPC. I will submit a patch when I have the time. Future API:s can be reported in new issues and discussed further to avoid feature creep.

PR #6156, opened 2024-10-22
volume server: added write blob grpc api similar to http

ollelogdahl:
Currently, only the http API on the volume server can create new blobs and automatically handle replication. This patch
adds a gRPC API that can be used by other systems to write new blobs. Compared to the other gRPC procedures which are more _internal_, this API is intended specifically for external systems. See issue https://github.com/seaweedfs/seaweedfs/issues/6146

- Much of the logic depends on `http.Request`, so that has been broken out.
- A new gRPC procedure called `WriteBlob` which performs the same logic as [weed/server/volume_server_handlers_write.go:18](https://github.com/seaweedfs/seaweedfs/blob/d04859b01c2128389b9a81007e231a0b2db091d7/weed/server/volume_server_handlers_write.go#L18C23-L18C78).

# How is the PR tested?
Currently only by writing a separate client and connect to it.

# Todo
- Testing: I'm not fully sure on how to test this, but I'm working on it.
- Handling more options; jwt, mime, filename, etc. I'm not sure here if everything should be supported.

# Checks


codewithnithish:
Hi @chrislusf  
Are you planning to merge this code?
It seems that we may require feature in our workflow.

Issue #6182, opened 2024-10-31
SubscribeMetadata stuck processing new events in some situation


half-life666:
**Describe the bug**
Similar to https://github.com/seaweedfs/seaweedfs/issues/6181, we will subscribe events from filer, but here we use fuse mount, so the basic setup is like below:

- Start weed master, volume, filer, mount
- Write files to fuse mount using tools like dd

After some time, e.g., 5-6 days, the IO tool is stopped, like for 1 or 2 days, then we resume the IO stress again. And we found that after this point, we can no longer receive any more events. However, we can workaround this by reseting the subscriber, by setting the `SinceNs` to 0, so the subscriber will got all events from beginning.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
Same as 6181
- OS version
- output of `weed version`
 version 30GB 3.77 linux amd64
- if using filer, show the content of `filer.toml`

**Expected behavior**
Subscribe should continue receieve new events.

**Screenshots**

**Additional context**
When debugging https://github.com/seaweedfs/seaweedfs/issues/6181, after comment out the `MetaAggregator` part, I found that the problem can be easily recreated. 

For example, yesterday 10-30 I removed `MetaAggregator` logic and everything works fine for subscribing events, but today 10-31, no more events can be receieved any more. 

```go
	for {

		glog.V(4).Infof("read on disk %v aggregated subscribe %s from %+v", clientName, req.PathPrefix, lastReadTime)

		processedTsNs, isDone, readPersistedLogErr = fs.filer.ReadPersistedLogBuffer(lastReadTime, req.UntilNs, eachLogEntryFn)
		if readPersistedLogErr != nil {
			return fmt.Errorf("reading from persisted logs: %v", readPersistedLogErr)
		}
		if isDone {
			return nil
		}

		if processedTsNs != 0 {
			lastReadTime = log_buffer.NewMessagePosition(processedTsNs, -2)
		}
```

Together with above code, since the subscriber will remember the last update time as 10-30 xxx (the last entry of log), so `processedTsNs` is 0?  

Which means if the subscriber last time points to the very end of a day, and then IO stopped, and IO restarted next day, processedTsNs will be 0 since there is no more entries from yesterday, so lastReadTime will not be updated.

I need to add some log and do more verify later, but open this issue first.

half-life666:
See pr https://github.com/seaweedfs/seaweedfs/pull/6183

Issue #6181, opened 2024-10-31
SubscribeMetadata can miss some events when big files are written to filer via s3


half-life666:
**Describe the bug**
We use `SubscribeMetadata` API to subscribe events from filer, so we can capture all the file metadata, to build service on top of that. Since the filer metadata is independent from upper layer access protocols, like fuse or s3, no matter which protocol, you can always subscribe to filer to get the file metadata events.

Our setup is as below:
- Start weed master, volume, filer and enable s3 service
- Start weed mount as well for s3 filer
- Write a simple server to subscribe the filer events using `SubscribeMetadata`
- Use a tool (like rclone) to write files to weed s3 endpoint

When we write a batch of large files, like 50 files with 512MB size, we can only receive like 10+ or 20+ file create events from subscriber.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
```
/usr/bin/weed -logtostderr=true master -mdir=/data -volumePreallocate -volumeSizeLimitMB=1024 -ip=192.168.3.138 -volumePreallocate=false -volumeSizeLimitMB=4096 -defaultReplication=000
/usr/bin/weed -logtostderr=true volume -dir=/data -dir=/mnt/weed1 -mserver=localhost:9333 -max=0 -port=8080 -metricsPort=9325 -preStopSeconds=1
/usr/bin/weed -logtostderr=true volume -dir=/data -dir=/mnt/weed2 -mserver=localhost:9333 -max=0 -port=8081 -metricsPort=9326 -preStopSeconds=1
/usr/bin/weed -logdir=/var/log/udm filer -master=localhost:9333 -port 8888 -s3 -s3.config=/etc/seaweedfs/s3.json -s3.port=8333 -s3.allowEmptyFolder=false -s3.allowDeleteBucketNotEmpty=false
```
- OS version
- output of `weed version`
version 30GB 3.77  linux amd64

- if using filer, show the content of `filer.toml`

**Expected behavior**
subscriber should return file counts match to files written

**Screenshots**

**Additional context**
after some debugging, we found the problem is at the implementation of MetaAggregator, but we can't identify which part of it has problem, below is the code of `SubscribeMetadata`:

```go


	for {

		glog.V(4).Infof("read on disk %v aggregated subscribe %s from %+v", clientName, req.PathPrefix, lastReadTime)

[...]

PR #6169, opened 2024-10-28
remove dup code for section.reader access

half-life666:
# What problem are we solving?
https://github.com/seaweedfs/seaweedfs/issues/6168

# How are we solving the problem?
resolve the thread safty for function

# How is the PR tested?
not yet, will do long IO test

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #6183, opened 2024-10-31
Fix 6181/6182

half-life666:
https://github.com/seaweedfs/seaweedfs/issues/6181
https://github.com/seaweedfs/seaweedfs/issues/6182

For 6181:
- when no entry is found (`processedTsNs` is 0), try next day

For 6182:
- set larger buf size

# How is the PR tested?
Using weed s3 and transfer 100 * 512MB files, no events is lost

# Checks

### DIFF
--- a/weed/filer/filer_notify_read.go
+++ b/weed/filer/filer_notify_read.go
@@ -4,14 +4,15 @@ import (
 	"container/heap"
 	"context"
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
-	"github.com/seaweedfs/seaweedfs/weed/util/log_buffer"
-	"github.com/seaweedfs/seaweedfs/weed/wdclient"
-	"google.golang.org/protobuf/proto"
 	"io"
 	"math"
 	"strings"
 	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+	"github.com/seaweedfs/seaweedfs/weed/util/log_buffer"
+	"github.com/seaweedfs/seaweedfs/weed/wdclient"
+	"google.golang.org/protobuf/proto"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/util"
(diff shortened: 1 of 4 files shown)


half-life666:
The UT failed at command_ec_encode.go, which is fixed in pr 6169 https://github.com/seaweedfs/seaweedfs/pull/6169/commits/9f7588eda5abeefd71c4ccdca378ff2c3fde06f2

weed/filer/filer_notify_read.go
```diff
@@ -39,6 +40,19 @@ func (f *Filer) collectPersistedLogBuffer(startPosition log_buffer.MessagePositi
 
 }
 
+func (f *Filer) HasPersistedLogFiles(startPosition log_buffer.MessagePosition) (bool, error) {
+	startDate := fmt.Sprintf("%04d-%02d-%02d", startPosition.Year(), startPosition.Month(), startPosition.Day())
+	dayEntries, _, listDayErr := f.ListDirectoryEntries(context.Background(), SystemLogDir, startDate, true, math.MaxInt32, "", "", "")
```
chrislusf:
```suggestion
	dayEntries, _, listDayErr := f.ListDirectoryEntries(context.Background(), SystemLogDir, startDate, true, 1, "", "", "")
```

weed/server/filer_grpc_server_sub_meta.go
```diff
@@ -62,8 +63,19 @@ func (fs *FilerServer) SubscribeMetadata(req *filer_pb.SubscribeMetadataRequest,
 			return nil
 		}
 
+		glog.V(4).Infof("processed to %v: %v", clientName, processedTsNs)
 		if processedTsNs != 0 {
 			lastReadTime = log_buffer.NewMessagePosition(processedTsNs, -2)
+		} else {
+			nextDayTs := util.GetNextDayTsNano(lastReadTime.UnixNano())
+			position := log_buffer.NewMessagePosition(nextDayTs, -2)
+			found, err := fs.filer.HasPersistedLogFiles(position)
```
chrislusf, line 72: this only check whether there are logs in the next day.

What if there are multiple days without any logs?

chrislusf:
the problem happens only when the UTC date changes?

weed/server/filer_grpc_server_sub_meta.go
half-life666: > this only check whether there are logs in the next day.
> 
> What if there are multiple days without any logs?

Right, we also need to consider this.

half-life666:
> the problem happens only when the UTC date changes?

Yes.

weed/server/filer_grpc_server_sub_meta.go
half-life666: If there are multiple days without any logs, the for loop should loop every day, until today? I can verify this in our teststand.

half-life666: Tested in our env that, last weekend there is no IO for 2 days, today we write some files and new events can still be subscribed with no problems.

zemul:
For the same problem, the file uploaded through the filer can be displayed on mount, but the file uploaded through s3 cannot be displayed on mount (it can be seen after restart mount). This pr does not seem to have specific repair details.

zemul:
> For the same problem, the file uploaded through the filer can be displayed on mount, but the file uploaded through s3 cannot be displayed on mount (it can be seen after restart mount). This pr does not seem to have specific repair details.对于相同的问题，通过文件管理器上传的文件可以在挂载时显示，但通过 s3 上传的文件无法在挂载时显示（重新启动挂载后可以看到）。这个 pr 似乎没有具体的维修细节。


Can't be repeated, sometimes normal.(Small file,not related to part upload.)

PR #6184, opened 2024-10-31
fix deadlock hang when broadcast to clients

kungf:
when master thransfer leader, the old master will disconnect with all filers and volumeServers, if the cluster is a big one, the broadcast messages may be much more than the max len of the channel 100, and if the KeepConnect just not listen on the channel in disconnect, the broadcastToClients will hang. and the whole cluster will not serve!

Related to： https://github.com/seaweedfs/seaweedfs/commit/91e4eca1e98cb5195346d90a2cc1fb9c92557213

1.  broadcast to client with no block, and add metirc to monitor full err
2. volumeServer heartbeat no need to broadcast when transfer leader

# How is the PR tested?
hard code the len the message channel to 1, it's easy deadlock when transfer leader, after fix, it will not deadlock

# Checks

### DIFF
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -4,11 +4,12 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"github.com/google/uuid"
-	"github.com/seaweedfs/seaweedfs/weed/cluster"
 	"net"
 	"sort"
 	"time"
+
+	"github.com/google/uuid"
+	"github.com/seaweedfs/seaweedfs/weed/cluster"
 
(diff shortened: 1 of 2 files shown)


chrislusf:
Nice fix!

chrislusf:
s3 integration tests start to fail. Need to look into this.

kungf:
> s3 integration tests start to fail. Need to look into this.

hi， @chrislusf I ran the test locally， and all passed. 
image

The error about test_lifecycle_get does not look related to this pr.
image

kungf:
recheck

Issue #6177, opened 2024-10-30
When uploading a file named ".txt" in chunks, an error occurs when reading the uploaded file


liangyihan:
**Describe the bug**
When uploading a file named ".txt" in chunks, an error occurs when reading the uploaded file. However, if the file is uploaded without chunking, it can be read normally.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".

-v=4 master -mdir=E:/guangdian/weedfs/data/master -peers=127.0.0.1:9666 -ip=127.0.0.1 -port=9666 -defaultReplication=000 -volumePreallocate -volumeSizeLimitMB=10
-v=4 volume -index=leveldb -idleTimeout=30 -dir=E:/guangdian/weedfs/data/volume,E:/guangdian/weedfs/data/warm,E:/guangdian/weedfs/data/cold -disk=hdd,warm,cold -mserver=127.0.0.1:9666 -dataCenter=dc -rack=rack1 -ip=127.0.0.1 -port=9189 -concurrentDownloadLimitMB=2048 -max=100
-v=4 filer -master=127.0.0.1:9666 -ip=127.0.0.1 -port=9283 -defaultReplicaPlacement=000 -maxMB=5 -encryptVolumeData -s3 -s3.allowEmptyFolder -s3.port=9383 -iam -iam.port=9483

- OS version
windows10
- output of `weed version`
seaweedfs 3.77
- if using filer, show the content of `filer.toml`

**Expected behavior**
The content of the ".txt" file can be read normally when uploaded in chunks.

**Screenshots**
Uploading a txt file with a size of 6454122 bytes using a chunk size of 5242880 bytes.
![e2dde233c2b6a0b597f916e40876fab](https://github.com/user-attachments/assets/3966f6b4-f4f3-4b0a-bf1c-d943d9389b69)
The result failed to be read.
![c47d47c150ca3f0f443bf0051954102](https://github.com/user-attachments/assets/7e50dbd2-0a30-4087-9895-fcff6e6552c8)
Debugging the reading process
![c864661d4f7c15906789cdf4ac81b1e](https://github.com/user-attachments/assets/86dc4ee8-bdc5-4ac3-9f78-158d53c84b7c)
flsCompressed =  false
![5c3ef50d596af8ae0e52607fa63ebde](https://github.com/user-attachments/assets/c42d2d3d-c32d-42d3-9ef0-d6944ac9f6be)

Error log
[...]

PR #6013, opened 2024-09-13
Feature: compress before encryption

evgmik:
Addresses #5524

Prior to the patch, the encryption disabled compression.
Now we compress (if beneficial) ==> encrypt (if requested). As result, data takes less space on the volume servers.

Lucky bonus: the old volumes/meta is compatible with new method, since `isCompressed` flag was always `false` when encryption is on.

# How are we solving the problem?

Effectively by not disabling compression.

# How is the PR tested?

filler was run with `-encryptVolumeData`, test data was saved into the weed fuse mount, the data was read back and checksums were compared. No deviation was observed. Meanwhile the volume server storage was checked, and it shows that it takes less space than uncompressed file sizes.

A relevant part of `fs.meta on_a_file`

````
      "cipherKey": "HY/Kzc9W26EQIzBC8sX7LGQ7lpzBNiTkY+DvVj7KJtA=",
      "isCompressed": true,
````

PR #6188, opened 2024-11-01
Update filer_multipart.go

liangyihan:
fix the #6177 bug

this bug
https://github.com/seaweedfs/seaweedfs/issues/6177

# How is the PR tested?
Upload files using cipher
Can read the correct value of IsCompressed when reading files

# Checks


liangyihan:
The problem stems from the modification of PR https://github.com/seaweedfs/seaweedfs/pull/6013

PR #6191, opened 2024-11-02
[shell] volume.balance collect volume servers by dc rack node

kmlebedev:

alancing between racks and nodes

aad select dn options rack and nodes

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -147,11 +147,11 @@ func collectVolumeServersByDcOrRack(t *master_pb.TopologyInfo, selectedDataCente
 			continue
 		}
 		for _, r := range dc.RackInfos {
-			if selectedRacks != "" && strings.Contains(selectedRacks, r.Id) {
+			if selectedRacks != "" && !strings.Contains(selectedRacks, r.Id) {
 				continue
 			}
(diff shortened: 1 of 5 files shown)


weed/shell/command_volume_balance.go
```diff
@@ -73,23 +74,30 @@ func (c *commandVolumeBalance) Do(args []string, commandEnv *CommandEnv, writer
 	balanceCommand := flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 	collection := balanceCommand.String("collection", "ALL_COLLECTIONS", "collection name, or use \"ALL_COLLECTIONS\" across collections, \"EACH_COLLECTION\" for each collection")
 	dc := balanceCommand.String("dataCenter", "", "only apply the balancing for this dataCenter")
+	racks := balanceCommand.String("racks", "", "only apply the balancing for this racks")
+	nodes := balanceCommand.String("nodes", "", "only apply the balancing for this nodes")
+	noLock := balanceCommand.Bool("noLock", false, "do not lock the admin shell at one's own risk")
```
chrislusf: `no lock` is a separate change. Better move it out of this PR.

kmlebedev: ok

Issue #6001, opened 2024-09-10
[S3] Head dir return unexpected response


Jackson-Wang-7:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
Report issues here. Ask questions here https://stackoverflow.com/questions/tagged/seaweedfs
Please ask questions in https://github.com/seaweedfs/seaweedfs/discussions

**Describe the bug**
When we create a directory by putting "bucket/folder/" object with empty contentType, SeaweedFS S3 API will set it as "httpd/unix-directory".
code is here https://github.com/seaweedfs/seaweedfs/blob/3.67/weed/s3api/s3api_object_handlers_put.go#L79
Then when we try to head the directory, there are some different results:
1. If the contentType is "httpd/unix-directory", no matter if the request is "HEAD bucket/folder" or "HEAD bucket/folder/", the response will be 200 OK. It confuses our program because we think the "folder" and "folder/" both exist.
2. If the contentType is not "httpd/unix-directory", no matter if the request is "HEAD bucket/folder/" or "HEAD bucket/folder/", it will return 404 not found.

**System Setup**
Starts the S3 API and filer at the same time, configured with IAM identities.
SeaweedFs version is 3.67

**Expected behavior**
the expected response:
For S3 standard behavior, when the "bucket/folder/" is a directory, the request "HEAD bucket/folder" should return 404. the request "HEAD bucket/folder/" should return 200. it indicates that the object exists as a directory.
Assuming the "bucket/folder" is a file, the request "HEAD bucket/folder" should return 200. the request "HEAD bucket/folder/" should return 404.
I guess you want to go into this code part to handle this https://github.com/seaweedfs/seaweedfs/blob/3.67/weed/server/filer_server_handlers_read.go#L140
but I think it should be like
```
        if entry.IsDirectory() {
                .....
                if !isForDirectory {
		        w.WriteHeader(http.StatusNotFound)
		        return
                }
        }
        if isForDirectory && !entry.IsDirectory() {
		w.WriteHeader(http.StatusNotFound)
		return
	}

```
Absolutely, I know you want to support Filer API and S3 API at the same time, but it increases complexity and the implementation here is not compatible with all situations.

**Screenshots**
<img width="1355" alt="image" src="https://github.com/user-attachments/assets/fef0df58-680c-47fe-a0d1-8ee99feca88a">

**Additional context**
[...]

PR #6194, opened 2024-11-04
Fix a bug where Get/Head a s3 path returns unexpected results

YichuanSun:

https://github.com/seaweedfs/seaweedfs/issues/6001#:~:text=If%20the%20contentType%20is%20%22httpd/unix%2Ddirectory%22%2C%20no%20matter%20if%20the%20request%20is%20%22HEAD%20bucket/folder%22%20or%20%22HEAD%20bucket/folder/%22%2C%20the%20response%20will%20be%20200%20OK.%20It%20confuses%20our%20program%20because%20we%20think%20the%20%22folder%22%20and%20%22folder/%22%20both%20exist.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/filer_server_handlers_read.go
+++ b/weed/server/filer_server_handlers_read.go
@@ -125,6 +125,12 @@ func (fs *FilerServer) GetOrHeadHandler(w http.ResponseWriter, r *http.Request) 
 			writeJsonQuiet(w, r, http.StatusOK, entry)
 			return
 		}
+		// The input `r.URL.Path` is like `abc`, while filer.FindEntry() returns an entry `abc/`,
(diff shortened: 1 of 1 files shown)


YichuanSun:
Later I will add the test results manually.

kmlebedev:
@YichuanSun Would it be difficult for you to add the corresponding s3test to the pipeline, which solves the compatibility issue?
https://github.com/seaweedfs/seaweedfs/blob/master/.github/workflows/s3tests.yml#L55 
Не уверен что такой, что он есть для этого кейса

YichuanSun:
The behavior is a little bit confusing, it doesn' work.

PR #6195, opened 2024-11-04
Refactor `ec.balance` logic into a `weeed/shell/command_ec_common.go`…

proton-lisandro-pin:

https://github.com/seaweedfs/seaweedfs/discussions/6179

This PR lays the foundation to unify the (re)balancing logic for `ec.balance` and `ec.encode`. Following changes will:

1. Modify `ec.encode` to use the same balancing logic as `ec.balance`.
2. _(maybe?)_ Move `weed/shell/command_ec_common.go` into `weed/storage/erasure_coding`.
3. Rework balancing logic to be topology-aware.

# How is the PR tested?

No functional changes, so no new tests are introduced. Existing tests should remain unaffected.

# Checks


weed/shell/command_ec_common.go
```diff
@@ -377,3 +378,424 @@ func groupBy(data []*EcNode, identifierFn func(*EcNode) (id string)) map[string]
...
+			// ignore if not in current collection
+			if shardInfo.Collection == collection {
+				vidLocations[needle.VolumeId(shardInfo.Id)] = append(vidLocations[needle.VolumeId(shardInfo.Id)], ecNode)
+			}
+		}
+	}
+	return vidLocations
+}
+
+func Balance(commandEnv *CommandEnv, collections []string, dc string, applyBalancing bool) (err error) {
```
chrislusf: use EcBalance or BalanceEc, since this is in the same package as the normal volume balance.

chrislusf (approved):

weed/shell/command_ec_common.go
ghost: `EcBalance` it is. Thanks!

Issue #6204, opened 2024-11-05
[Helm Chart] filer-statefulset template flag s3.allowEmptyFolder and s3-deployment template flag allowEmptyFolder bug


FQHSLycopene:
<img width="651" alt="image" src="https://github.com/user-attachments/assets/fc6edb60-0ac8-4883-a23c-5da550f40b19">
<img width="606" alt="image" src="https://github.com/user-attachments/assets/9a2d7b72-f3b4-469f-927f-0097b5196aeb">

here two flag default value is true, and if helm value is false ,those flags were not render。
Can I submit a PR?

chrislusf:
> Can I submit a PR?

Of course.

PR #6206, opened 2024-11-05
fix helm: s3 allowEmptyFolder flag if (#6204)

FQHSLycopene:
the issue of conditional rendering based on a boolean value.
helm chart s3 allowEmptyFolder value not Effective
[#6204](https://github.com/seaweedfs/seaweedfs/issues/6204)

Added type check for boolean values: The if condition has been updated to check whether the value of .Values.someBool is of type bool using the typeOf function. This ensures that the flag were rendered when the value is explicitly set as true or false, and prevents rendering when the value is not a boolean or is missing.

# How is the PR tested?
helm template test

# Checks

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
+++ b/k8s/charts/seaweedfs/templates/filer-statefulset.yaml
@@ -209,7 +209,7 @@ spec:
               -s3.cert.file=/usr/local/share/ca-certificates/client/tls.crt \
               -s3.key.file=/usr/local/share/ca-certificates/client/tls.key \
               {{- end }}
-              {{- if .Values.filer.s3.allowEmptyFolder }}
+              {{- if eq (typeOf .Values.filer.s3.allowEmptyFolder) "bool" }}
               -s3.allowEmptyFolder={{ .Values.filer.s3.allowEmptyFolder }} \
               {{- end }}
               {{- if .Values.filer.s3.enableAuth }}
--- a/k8s/charts/seaweedfs/templates/s3-deployment.yaml
+++ b/k8s/charts/seaweedfs/templates/s3-deployment.yaml
@@ -134,7 +134,7 @@ spec:
               {{- if .Values.s3.domainName }}
               -domainName={{ .Values.s3.domainName }} \
               {{- end }}
-              {{- if .Values.s3.allowEmptyFolder }}
+              {{- if eq (typeOf .Values.s3.allowEmptyFolder) "bool" }}
               -allowEmptyFolder={{ .Values.s3.allowEmptyFolder }} \
               {{- end }}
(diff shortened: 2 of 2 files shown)


k8s/charts/seaweedfs/templates/filer-statefulset.yaml
```diff
@@ -209,7 +209,7 @@ spec:
               -s3.cert.file=/usr/local/share/ca-certificates/client/tls.crt \
               -s3.key.file=/usr/local/share/ca-certificates/client/tls.key \
               {{- end }}
-              {{- if .Values.filer.s3.allowEmptyFolder }}
+              {{- if eq (typeOf .Values.filer.s3.allowEmptyFolder) "bool" }}
```
chrislusf, line 212: this does not seem a good fix. Better just define `.Values.filer.s3.allowEmptyFolder`.

FQHSLycopene: allowEmptyFolder is a boolean type, and it serves not only as a condition for evaluation but also as the value of a flag. However, when its value is false, Helm treats it as an empty value, and using an if condition will prevent it from being rendered.

The logic needed here should be that if allowEmptyFolder is either true or false, it should be rendered, while other empty values should not be.

the flag delault value is true
it can not be like this
```
if .Values.filer.s3.allowEmptyFolder
s3.allowEmptyFolder=true
else
s3.allowEmptyFolder=false
```
this is conflicted with default
maybe change the default value

chrislusf: `allowEmptyFolder` default is false. The logic is correct?

FQHSLycopene: the default i mean is the flag's default in here,not chart default value
image
the flag default value is true,but chart default is false ,and now helm false no render,mean use flag default true,that's incorrect
if flag default is false ,chart value  false no render ,it's still false,the logic is correct

mx-sylveon: I can confirm this is an issue as shown above - we are completely unable to set the filer's S3 implementation to adhere to AWS S3 specifications because of this bug.

The helm template excluding the CLI flag entirely if the helm value `filer.s3.allowEmptyFolder` is `false` only works as expected if the default in `weed/command/filer.go` is also `false`. Since it's actually defaulting to `true` that is what we're stuck with when using the filer via the helm chart.

I've worked around this by just pulling the chart locally to make the needed adjustment, but I recommend you take another look at this @chrislusf - it's a trivial fix, whether you choose to do what @FQHSLycopene has provided or not, that's needed for the filer's S3 API to work correctly when deployed this way.

PR #6207, opened 2024-11-05
nested worm config

27149chen:
worm is enabled under /a, we need to
1. disable worm under /a/b, by adding a new config with prefix /a/b and set worm to disable
2. remove a worm file /a/test.txt, by adding a new config with prefix /a/test.txt and set worm to disable first, so we are able to delete it

set worm with three values: true, false, not set

# How is the PR tested?

# Checks


chrislusf:
I do not understand what the problem is.

27149chen:
> I do not understand what the problem is.

two cases as I mentioned above:
1. worm is enabled in parent path, but I want to disable worm in one of its child path
2. worm file is readonly and can not being deleted, if I upload a large file into a worm path but failed at middle, I can not retry it since the file already exists. A workaround is adding a fs configure for this file with worm disabled, so I can remove it and upload again

chrislusf:
`fs.configure` works. What problem this PR is solving?

27149chen:
> `fs.configure` works. What problem this PR is solving?

it does not work without this pr

27149chen:
problem is in this line
```
a.Worm = b.Worm || a.Worm
```

a is /a， worm is true
b is /a/b, worm is false
result is true, which is not as expected

27149chen:
@chrislusf PTAL

chrislusf:
Allowing this would mean breaking the WORM promise?

27149chen:
break which promise? worm config is per path, parent path and child path can have different config

chrislusf:
> break which promise? worm config is per path, parent path and child path can have different config

the child path is part of the parent directory.

if any directory is marked WORM, it should be reliable so that legal department can rely on the promise.

27149chen:
take another field in filer conf for example
```
a.Collection = util.Nvl(b.Collection, a.Collection)
```
a is /a， collection is a
b is /a/b, collection is b
result is b, which means collection is changed in a child path, if we support nested config, it should works in this way

chrislusf:
what is your use case for WORM folders?

27149chen:
mainly case2

> 
> two cases as I mentioned above:
> 
> 1. worm is enabled in parent path, but I want to disable worm in one of its child path
> 2. worm file is readonly and can not being deleted, if I upload a large file into a worm path but failed at middle, I can not retry it since the file already exists. A workaround is adding a fs configure for this file with worm disabled, so I can remove it and upload again

chrislusf:
what is the business use case?

27149chen:
we have a file sync agent to sync files from file server/local filesystem to sw fuse/s3, worm is enabled in sw. we need to delete partial failure files manually

chrislusf:
The partial files are in the ".upload" folder?

27149chen:
we now work on fuse mount，not the  partial files generated by sw.
For example, I have a video 1.mp4, I start to copy it from local path in host 1.2.3.4 to remote fuse path in host 5.6.7.8, and then stop the connection suddenly. there will be a 1.mp4 in fuse path in worm mode, we need to disable it so that we can delete it and copy again

chrislusf:
why not just leave it there and skip copying if it already exists?

WORM mode is to protect against possible data tampering. This PR would break this promise.

27149chen:
> why not just leave it there and skip copying if it already exists?
> 
> WORM mode is to protect against possible data tampering. This PR would break this promise.

For example, I have a video 1.mp4 (10GiB), I start to copy it from local path in host 1.2.3.4 to remote fuse path in host 5.6.7.8, and then stop the connection suddenly. there will be a 1.mp4((3GiB)) in fuse path in worm mode which is different from the source file, we need to disable it so that we can delete it and copy again

chrislusf:
This is similar to writing to CD-ROM. If write fails, write to a different file name.

27149chen:
> This is similar to writing to CD-ROM. If write fails, write to a different file name.

This is one approach; however, from a commercial product perspective, this usage may sometimes be unacceptable.

Also, worm is a field in filer conf, it It should exhibit consistent behavior with other fields like collection, replication

chrislusf:
> This is one approach; however, from a commercial product perspective, this usage may sometimes be unacceptable.
> 
> Also, worm is a field in filer conf, it It should exhibit consistent behavior with other fields like collection, replication

The listed reasons are not relevant to a dependable implementation of WORM.
This PR would easily fail any security test.

What you want may be a time duration, before which a new file can be read and write, and after which a file would be readonly.

chrislusf:
Or it could be a "commit" operation to mark the file is complete.

27149chen:
A time duration is difficult to define, we might take a long time to realize the break. A commit sounds good, but how to implement it and how the interaction will be

27149chen:
what is versioned files in seaweedfs? and how to solve this problem?

27149chen:
@chrislusf Kindly proceed with your review.

27149chen:
@chrislusf so what should we do now? I still think worm is a field in filer conf, it It should exhibit consistent behavior with other fields like collection, replication

chrislusf:
WORM mode can add an option: If updated, the older version will be renamed instead of deleted.

This is sort of an easier implementation of versioned file.

27149chen:
> WORM mode can add an option: If updated, the older version will be renamed instead of deleted.
> 
> This is sort of an easier implementation of versioned file.

@chrislusf normally, worm files should not be updated/deleted, it is meanless to keep various versions of a worm file. My approach is a fallback solution, allowing administrators to bypass certain restrictions (by temporarily adding a new sub conf) to delete the worm file in case of issues.

PR #6214, opened 2024-11-07
[master] avoid crowded more writable for auto grow

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/6121
image

unexpected behavior crowded more writable

removed auto grow in this case added logging of deletion from crowded

# How is the PR tested?

# Checks


chrislusf:
"unexpected behavior crowded more writable" I do not understand this sentence without going through the issue. Could you please explain this better?

kmlebedev:
> "unexpected behavior crowded more writable" I do not understand this sentence without going through the issue. Could you please explain this better?

the crowded map in some cases did not allocate volume id, which led to the fact that the number of volumes in the map grew infinitely until the restart of the master and led to an incorrect calculation of the difference between vratibles and led to an writable growth of volumes.

I'm still not sure that the problem has been solved, since it was not reproduced on the dev cluster. Today I'll do a PR and turn the growth logic back on and roll it out to production.

kmlebedev:
> "unexpected behavior crowded more writable" I do not understand this sentence without going through the issue. Could you please explain this better?

https://github.com/seaweedfs/seaweedfs/pull/6267

PR #6222, opened 2024-11-09
[shell] refactor and add parallelBalancing

kmlebedev:

For large installations where each node has 60 slow HDD disks, rebalancing in one thread takes forever

Perform rebalancing in parallel on each HDD disk

# How is the PR tested?

# Checks


kmlebedev:
@chrislusf You can add this refactoring for now, the parallelism there is quite naive, it requires improvement, but it will be more difficult to review

chrislusf:
* could you make it into 2 PRs, one for refactoring and one for the new feature?
* parallel balancing can be an integer, to adjust the level of parallelism.

PR #6224, opened 2024-11-11
fix rename event

### LINKED ISSUE #6187: File rename event has no old_entry
**Describe the bug**
File rename event returned by SubscribeMetadata has no old_entry

As previously discussed in this context: https://github.com/seaweedfs/seaweedfs/discussions/6178

**System Setup**
server
```
weed server -master.port=9333 -volume.port=8080 -dir="/Users/lou/tmp/mnt/weed1" -master.volumeSizeLimitMB=1024 -filer -filer.defaultReplicaPlacement=000  -filer.maxMB=4
```

mount
```
weed -v=4 mount -filer=localhost:8888 -dir=/Users/lou/tmp/loutest -filer.path=/Users/lou/tmp/loutest -collection=user1 -chunkSizeLimitMB=8
```

version: master

**Expected behavior**
File rename event should have old_entry

**Screenshots**
here is an example event:
```
{
  "directory": "/Users/lou/tmp/loutest",
  "event_notification": {
    "new_entry": {
      "name": "4",
      "chunks": [
        {
          "size": 2,
          "modified_ts_ns": 1726116263107347000,
          "e_tag": "sCYyTGkEsqnLS4jW1hyB0Q==",
          "fid": {
            "volume_id": 8,
            "file_key": 94,
            "cookie": 1156223458
          },
          "is_compressed": true
        }
      ],
      "attributes": {
        "file_size": 2,
        "mtime": 1726116263,
        "file_mode": 420,
        "uid": 501,
        "gid": 20,
        "crtime": 1726116263,
        "mime": "text/plain; charset=utf-8",
        "inode": 14219030884291337544
      }
    },
    "delete_chunks": true,
    "new_parent_path": "/Users/lou/tmp/loutest",
    "signatures": [
      -1470161406,
      -979691841
    ]
  },
  "ts_ns": 1730348208701552000
}
```

**Additional context**

27149chen:
fix https://github.com/seaweedfs/seaweedfs/issues/6187

fuse calls StreamRenameEntry to rename a file, StreamRenameEntry calls filer.CreateEntry to create a new entry, so only a create evnent will be generated which has not old path info and old entry(with old name) info. I add a new method RenameEntry and fix the logic

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/filer.go
+++ b/weed/filer/filer.go
@@ -240,6 +240,43 @@ func (f *Filer) CreateEntry(ctx context.Context, entry *Entry, o_excl bool, isFr
 	return nil
 }
 
+func (f *Filer) RenameEntry(ctx context.Context, oldEntry, newEntry *Entry, o_excl bool, isFromOtherCluster bool, signatures []int32, skipCreateParentDir bool, maxFilenameLength uint32) error {
+	if string(newEntry.FullPath) == "/" {
+		return nil
+	}
+
+	if newEntry.FullPath.IsLongerFileName(maxFilenameLength) {
+		return fmt.Errorf("entry name too long")
+	}
+
+	currentEntry, _ := f.FindEntry(ctx, newEntry.FullPath)
+	if currentEntry != nil && o_excl {
+		glog.V(3).Infof("EEXIST: entry %s already exists", newEntry.FullPath)
+		return fmt.Errorf("EEXIST: entry %s already exists", newEntry.FullPath)
+	}
+
+	if currentEntry == nil && !skipCreateParentDir {
+		dirParts := strings.Split(string(newEntry.FullPath), "/")
+		if err := f.ensureParentDirectoryEntry(ctx, newEntry, dirParts, len(dirParts)-1, isFromOtherCluster); err != nil {
+			return err
+		}
+	}
+
+	glog.V(4).Infof("renameEntry %s: old entry: %v", newEntry.FullPath, oldEntry.FullPath)
+	if err := f.renameEntry(ctx, oldEntry, newEntry, currentEntry != nil); err != nil {
+		glog.Errorf("rename entry %s: %v", newEntry.FullPath, err)
+		return fmt.Errorf("rename entry %s: %v", newEntry.FullPath, err)
+	}
+
+	f.NotifyUpdateEvent(ctx, oldEntry, newEntry, true, isFromOtherCluster, signatures)
+
+	f.deleteChunksIfNotNew(currentEntry, newEntry)
+
+	glog.V(4).Infof("RenameEntry %s: renamed", newEntry.FullPath)
+
+	return nil
+}
+
 func (f *Filer) ensureParentDirectoryEntry(ctx context.Context, entry *Entry, dirParts []string, level int, isFromOtherCluster bool) (err error) {
 
 	if level == 0 {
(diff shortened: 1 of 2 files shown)


chrislusf:
What problem are we solving?

27149chen:
> What problem are we solving?

plese see details in https://github.com/seaweedfs/seaweedfs/issues/6187

chrislusf:
Why need to change current way if it is already working?

27149chen:
> Why need to change current way if it is already working?

It does not work as the doc says: https://github.com/seaweedfs/seaweedfs/wiki/Filer-Metadata-Events#what-kind-of-events-it-contains

in this doc, it says:
```
* Rename Event: similar to Update Event, but also need `SubscribeMetadataResponse.directory` and `EventNotification.new_parent_path`.
```
but in fact, there is no rename event, just a create event, which has no old_entry and SubscribeMetadataResponse.directory always equals to EventNotification.new_parent_path

we need the rename event to do something different

chrislusf:
there are 3 entries here in the rename.
1. old entry
2. new entry
3. possible previous entry in the new entry position

This PR skips the 3rd entry and will cause problems downstream.

27149chen:
I didn't skip the 3rd entry, Which lines will cause what kind of problems? please add your comments in code

weed/filer/filer.go
```diff
@@ -313,6 +350,23 @@ func (f *Filer) UpdateEntry(ctx context.Context, oldEntry, entry *Entry) (err er
...
+		glog.Errorf("existing %s is a directory", oldEntry.FullPath)
+		return fmt.Errorf("existing %s is a directory", oldEntry.FullPath)
+	}
+	if !oldEntry.IsDirectory() && newEntry.IsDirectory() {
+		glog.Errorf("existing %s is a file", oldEntry.FullPath)
+		return fmt.Errorf("existing %s is a file", oldEntry.FullPath)
+	}
+
+	newEntry.Attr.Crtime = oldEntry.Attr.Crtime
+	if newEntryExists {
```
chrislusf, line 364: oldEntry is not deleted

27149chen: oldEntry is deleted in this function: moveSelfEntry
line 204-227

chrislusf: I do not see it.

27149chen: https://github.com/seaweedfs/seaweedfs/blob/3003c9e17e06db92f5a426c191ebaab15a1764a1/weed/server/filer_grpc_server_rename.go#L204

weed/filer/filer.go
```diff
@@ -240,6 +240,43 @@ func (f *Filer) CreateEntry(ctx context.Context, entry *Entry, o_excl bool, isFr
...
+		}
+	}
+
+	glog.V(4).Infof("renameEntry %s: old entry: %v", newEntry.FullPath, oldEntry.FullPath)
+	if err := f.renameEntry(ctx, oldEntry, newEntry, currentEntry != nil); err != nil {
+		glog.Errorf("rename entry %s: %v", newEntry.FullPath, err)
+		return fmt.Errorf("rename entry %s: %v", newEntry.FullPath, err)
+	}
+
+	f.NotifyUpdateEvent(ctx, oldEntry, newEntry, true, isFromOtherCluster, signatures)
```
chrislusf, line 271: the `currentEntry` removal would be missing from the metadata events.

27149chen: if currentEntry exists, it is updated, not deleted.
for example: rename from a(inode=1234) to b(inode=1234), and b(inode=5678) exists, after renaming, a rename event of b(inode=1234) will be generated, b is still there (but inode is changed), so the rename event is also a update event of old b

chrislusf: no. There are a, b, b_old 3 entries in this example. One event can only have info for 2 entries.

27149chen: yes, there are two events, rename: oldEntry: a(1234), newEntry b(1234), update: oldEntry: b(5678), newEntry: b(1234). only the 1st one is generated, do I need to generate the 2nd one? if yes, what is the order? I think they are exactly the same thing, so one event is enough.

27149chen:
@chrislusf  Kindly proceed with your review.

chrislusf:
> @chrislusf Kindly proceed with your review.

this PR missed events as comment above. To fix it, it would be the same as current implementation.

27149chen:
> > @chrislusf Kindly proceed with your review.
> 
> this PR missed events as comment above. To fix it, it would be the same as current implementation.

please take a look about this comment: https://github.com/seaweedfs/seaweedfs/pull/6224#discussion_r1839977289

weed/filer/filer.go
27149chen: @chrislusf there are two events as I said above, a rename and an update, new entry are all b, and these two events happen exactly the same time. If we generate two events, what is the order?  a=>b and then b_old=>b or b_old=>b and then a=>b, both are meaningless, I think an event a=>b is enough.

In other words, what is a rename event? inode is the same between old entry and new entry, but name is different, it is a rename event.
If inode is different between old entry and new entry, but name is the same, what kind of event is it? it is not a update event since inode is changed. or an override event? or just a side effect of a rename event?

PR #6241, opened 2024-11-15
Fix/parse upload filename

KevinBLT:
Refactoring `needle_parse_upload` and better handle filenames (`content-disposition` header).
Also a small fix to admin UI when `publicUrl` already contains `https://`.

Use `mime` lib to parse the header and handle variants of the given header value.
Also combine the different upload methods (`PUT`/ `POST`) and
content type `form-data` or `{mime-type}` e.g `image/png`.

# How is the PR tested?
I tested that the variants worked and viewed in browser and used these commands:

```cli
curl 'http://localhost:9333/dir/assign?count=3'
{"fid":"6,62109ceee3","url":"192.168.2.23:8080","publicUrl":"192.168.2.23:8080","count":3}

curl -X PUT -H 'Content-Disposition: filename="1.png"' -H 'Content-Type: image/png' --data-binary @1.png 'http://192.168.2.23:8080/6,62109ceee3_0'
{"name":"1.png","size":306542,"eTag":"5e530dd3","mime":"image/png"}                                                                                                                                                   

curl -X POST -H 'Content-Disposition: filename="2.png"' -H 'Content-Type: image/png' --data-binary @2.png 'http://192.168.2.23:8080/6,62109ceee3_1'
{"name":"2.png","size":293856,"eTag":"b28728ee","mime":"image/png"}

curl -X POST -F file=@3.png 'http://192.168.2.23:8080/6,62109ceee3_2' 
{"name":"3.png","size":525586,"eTag":"bb9b7005","mime":"image/png"}

```

# Checks

### DIFF
--- a/weed/operation/needle_parse_test.go
+++ b/weed/operation/needle_parse_test.go
@@ -44,7 +44,7 @@ func TestCreateNeedleFromRequest(t *testing.T) {
 	{
 		mockClient.needleHandling = func(n *needle.Needle, originalSize int, err error) {
 			assert.Equal(t, nil, err, "upload: %v", err)
-			assert.Equal(t, "", string(n.Mime), "mime detection failed: %v", string(n.Mime))
+			assert.Equal(t, "text/plain; charset=utf-8", string(n.Mime), "mime detection failed: %v", string(n.Mime))
 			assert.Equal(t, true, n.IsCompressed(), "this should be compressed")
 			assert.Equal(t, true, util.IsGzippedContent(n.Data), "this should be gzip")
 			fmt.Printf("needle: %v, originalSize: %d\n", n, originalSize)
--- a/weed/storage/needle/needle_parse_upload.go
+++ b/weed/storage/needle/needle_parse_upload.go
@@ -43,19 +43,8 @@ func ParseUpload(r *http.Request, sizeLimit int64, bytesBuffer *bytes.Buffer) (p
 		}
 	}
 
-	if r.Method == http.MethodPost {
-		contentType := r.Header.Get("Content-Type")
+	e = parseUpload(r, sizeLimit, pu)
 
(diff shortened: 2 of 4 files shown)


weed/storage/needle/needle_parse_upload.go
```diff
@@ -108,170 +97,157 @@ func ParseUpload(r *http.Request, sizeLimit int64, bytesBuffer *bytes.Buffer) (p
...
 		if dotIndex > 0 {
 			ext = strings.ToLower(pu.FileName[dotIndex:])
 			mtype = mime.TypeByExtension(ext)
 		}
-		contentType := r.Header.Get("Content-Type")
 
 		if contentType != "" && contentType != "application/octet-stream" && mtype != contentType {
 			pu.MimeType = contentType // only return mime type if not deducible
-			mtype = contentType
+		} else if mtype != "" && pu.MimeType == "" && mtype != "application/octet-stream" {
```
KevinBLT, line 246: Without this, there would never be a content type set if the one actually set in the form part is the same as the one detected.

weed/operation/needle_parse_test.go
```diff
@@ -44,7 +44,7 @@ func TestCreateNeedleFromRequest(t *testing.T) {
 	{
 		mockClient.needleHandling = func(n *needle.Needle, originalSize int, err error) {
 			assert.Equal(t, nil, err, "upload: %v", err)
-			assert.Equal(t, "", string(n.Mime), "mime detection failed: %v", string(n.Mime))
+			assert.Equal(t, "text/plain; charset=utf-8", string(n.Mime), "mime detection failed: %v", string(n.Mime))
```
KevinBLT, line 47: I see no reason, why it was not supposed to actually detect the type from `.txt` file previously?

PR #6254, opened 2024-11-18
Introduce logic to resolve volume replica placement within EC rebalancing.

proton-lisandro-pin:

Make EC balancing logic topology-aware. See https://github.com/seaweedfs/seaweedfs/discussions/6179 .

This PR introduces logic to resolve the replication settings for volumes whose EC shards need re-balancing; it will be used
to ensure those shards are moved respecting topology requirements.

# How is the PR tested?

Added unit tests for new function `volumeIdToReplicaPlacement()`

# Checks


weed/shell/command_ec_common.go
```diff
@@ -12,11 +11,31 @@ import (
...
 	"golang.org/x/exp/slices"
 	"google.golang.org/grpc"
 )
 
+type RackId string
+type EcNodeId string
+
+type EcNode struct {
+	info       *master_pb.DataNodeInfo
+	dc         string
```
chrislusf: add typing for `dc` also?

chrislusf (approved):

weed/shell/command_ec_common.go
ghost: Done.

PR #6267, opened 2024-11-21
[master] Revert back automatic volume grow in case of compensate for crowded volumes

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/6121

Returned the advanced growth of volumes to compensate for the volumes leaving due to crowded
Also added a limit on the growth of volumes available for writable, if suddenly something goes wrong with the calculation of crowded volumes

# How is the PR tested?

On dev cluster

# Checks


kmlebedev:
> I think here we need a struct to track how many volumes are to be created, based on the number of file ids assigned and maybe some other info, to pace the volume creation. Need a design for this.

I thought about it, but now it is more important to receive directly from the filer the volumeGrow configuration without waiting for gRPC requests for volumeGrow from the filer in the master

PR #6270, opened 2024-11-21
Improve EC shards rebalancing logic across racks

proton-lisandro-pin:
- Favor target shards with less preexisting shards, to ensure a fair distribution.
  - Randomize selection when multiple possible target shards are available.
  - Add logic to account for replication settings when selecting target shards (currently disabled).

Improve EC shards re-balancing logic across racks.
  
- Favor target shards with less preexisting shards, to ensure a fair distribution.
- Randomize selection when multiple possible target shards are available.
- Add logic to account for volume replication settings when selecting target shards (currently disabled).

# How is the PR tested?

New test for `pickRackToBalanceShardsInto()`.

# Checks

### DIFF
--- a/weed/shell/command_ec_common.go
+++ b/weed/shell/command_ec_common.go
@@ -3,6 +3,7 @@ package shell
 import (
 	"context"
 	"fmt"
+	"math/rand"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/operation"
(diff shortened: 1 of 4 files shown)


weed/shell/command_ec_common.go
```diff
@@ -521,23 +527,44 @@ func doBalanceEcShardsAcrossRacks(commandEnv *CommandEnv, collection string, vid
...
+		if shards == targetShards {
+			targets = append(targets, rackId)
+		}
 	}
 
-	return ""
+	if len(targets) == 0 {
+		return ""
+	}
+	return targets[rand.Intn(len(targets))]
```
chrislusf, line 567: Nit: use math rand v2

chrislusf (approved):

weed/shell/command_ec_common.go
chrislusf: updated

Issue #6278, opened 2024-11-25
[master] failed to find writable volumes in specific DC and Rack


kmlebedev:
**Describe the bug**

6 RPS errors:
```

I1124 22:23:24.564072 master_grpc_server_assign.go:133 assign count:1 replication:"100" collection:"js" data_center:"dc7" rack:"F14" {"collection":"js","replication":{"dc":1},"ttl":{"Count":0,"Unit":0},"dataCenter":"dc7","rack":"F14"}: failed to find writable volumes for collection:js replication:100 ttl: error: No writable volumes in DataCenter:dc7 Rack:F14 DataNode:

```

**System Setup**
```
3.77
```

**Expected behavior**

We check that there is writable volume in each DC + Rack  layout

PR #6280, opened 2024-11-25
use ShouldGrowVolumesByDcAndRack

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/6278

We check the DC and the Rack

# How is the PR tested?

# Checks

### DIFF
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -369,6 +369,19 @@ func (t *Topology) ListDataCenters() (dcs []string) {
 	return dcs
 }
 
+func (t *Topology) ListDCAndRacks() (dcs map[string][]string) {
+	t.RLock()
+	defer t.RUnlock()
+	dcs = make(map[string][]string)
+	for _, dcNode := range t.children {
+		dcNodeId := string(dcNode.(*DataCenter).Id())
+		for _, rackNode := range dcNode.Children() {
+			dcs[dcNodeId] = append(dcs[dcNodeId], string(rackNode.(*Rack).Id()))
+		}
+	}
+	return dcs
+}
+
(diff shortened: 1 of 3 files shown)


weed/topology/topology.go
```diff
@@ -369,6 +369,19 @@ func (t *Topology) ListDataCenters() (dcs []string) {
 	return dcs
 }
 
+func (t *Topology) ListDCAndRacks() (dcs map[string][]string) {
```
chrislusf: Use types instead of strings

weed/topology/volume_layout.go
```diff
@@ -399,6 +386,14 @@ func (vl *VolumeLayout) GetWritableVolumeCount() (active, crowded int) {
 	return len(vl.writables), len(vl.crowded)
 }
 
+func (vl *VolumeLayout) GetWritableVolumes() (writables []needle.VolumeId) {
```
chrislusf:
```suggestion
func (vl *VolumeLayout) CloneWritableVolumes() (writables []needle.VolumeId) {
```

kmlebedev: ok

weed/topology/topology.go
kmlebedev: > Use types instead of strings

done

chrislusf:
How about simplify this by adding a structure to each volume layout, tracking sent volume grow requests that are not reported back by the volume servers?

kmlebedev:
> How about simplify this by adding a structure to each volume layout, tracking sent volume grow requests that are not reported back by the volume servers?

@chrislusf 
I didn't quite understand how this would help.
But now, there is a problem in the synchronous creation of volumes where there are none, precisely because of the lack of a structure for each layer with a list of volumes available for writing. Because it's difficult to tie it together.

Maybe later I'll make another approach to this

PR #6289, opened 2024-11-26
change comment for expect header

CodeRusher:
golang server now do not strips off 'Expect' header, so we can remove this piece of code now.

# How is the PR tested?

```
// curl -v -X POST "http://localhost:8080" -H "Expect: 100-continue" --data "Hello, World"
func Test100ContinueServer(t *testing.T) {
	http.HandleFunc("/", handler)
	http.ListenAndServe(":8080", nil)
}

func handler(w http.ResponseWriter, r *http.Request) {
	expectValue := r.Header.Get("Expect")
        // output: Expect Header: 100-continue
	fmt.Printf("Expect Header: %s\n", expectValue)

	body, err := io.ReadAll(r.Body)
	if err != nil {
		http.Error(w, "Failed to read request body", http.StatusInternalServerError)
		return
	}
	defer r.Body.Close()

	fmt.Printf("Request Body: %s\n", string(body))

	fmt.Fprintf(w, "Request received and processed")
}

```

# Checks

### DIFF
--- a/weed/s3api/auth_signature_v4.go
+++ b/weed/s3api/auth_signature_v4.go
@@ -665,24 +665,6 @@ func extractSignedHeaders(signedHeaders []string, r *http.Request) (http.Header,
 			continue
 		}
 		switch header {
-		case "expect":
-			// Golang http server strips off 'Expect' header, if the
-			// client sent this as part of signed headers we need to
-			// handle otherwise we would see a signature mismatch.
-			// `aws-cli` sets this as part of signed headers.
-			//
-			// According to
-			// http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.20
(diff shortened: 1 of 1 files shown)


chrislusf:
Which go version has the fix? Any doc?

CodeRusher:
> Which go version has the fix? Any doc?

test on go 1.16.15 go 1.22.9, 
Check the Git commit history, the previous code corresponds to Go version 1.12.

CodeRusher:
debug in weed，can find the expect header,too
image

CodeRusher:
The default setting for retaining the Expect header needs to be preserved. In NGINX v1.1, the Expect header is removed when handling 100-continue requests. The comment needs to be updated, as previous comment was confusing.

Issue #6290, opened 2024-11-26
[master] avoid timeout when Assigning for main request with filter by DC or rack


kmlebedev:
**Describe the bug**

in case of [errors in finding the volume](https://github.com/seaweedfs/seaweedfs/issues/6278), we wait too long for a timeout instead of immediately sending an alternative request

<img width="545" alt="image" src="https://github.com/user-attachments/assets/ea94a786-654e-468e-8494-cecfc258d03f">

**System Setup**
```
3.77
```

**Expected behavior**

We don't wait for the timeout on the assign fro main request, but immediately send an alternative request

PR #6291, opened 2024-11-26
[master] avoid timeout when assigning for main request with filter by DC or rack

kmlebedev:

https://github.com/seaweedfs/seaweedfs/issues/6290

avoid timeout when assigning for main request with filter by DC or rack

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/master_grpc_server_assign.go
+++ b/weed/server/master_grpc_server_assign.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/stats"
+	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


weed/server/master_grpc_server_assign.go
```diff
@@ -100,6 +101,9 @@ func (ms *MasterServer) Assign(ctx context.Context, req *master_pb.AssignRequest
 			glog.V(1).Infof("assign %v %v: %v", req, option.String(), err)
 			stats.MasterPickForWriteErrorCounter.Inc()
 			lastErr = err
+			if (req.DataCenter != "" || req.Rack != "") && strings.Contains(err.Error(), "No writable volumes") {
```
chrislusf: Use a constant string.

kmlebedev: done

PR #6297, opened 2024-11-27
Improve EC shards rebalancing logic across nodes

proton-lisandro-pin:

- Favor target nodes with less preexisting shards, to ensure a fair distribution.
- Randomize selection when multiple possible target nodes are available.
- Add logic to account for replication settings when selecting target nodes (currently disabled).

# How is the PR tested?

Added new unit tests for `pickEcNodeToBalanceShardsInto()`.

# Checks

### DIFF
--- a/weed/shell/command_ec_common_test.go
+++ b/weed/shell/command_ec_common_test.go
@@ -146,8 +146,8 @@ func TestPickEcNodeToBalanceShardsInto(t *testing.T) {
 		wantOneOf []string
 		wantErr   string
 	}{
-		{topologyEc, "", "", nil, "invalid source/destination nodes"},
-		{topologyEc, "idontexist", "12737", nil, "invalid source/destination nodes"},
+		{topologyEc, "", "", nil, "INTERNAL: missing source nodes"},
+		{topologyEc, "idontexist", "12737", nil, "INTERNAL: missing source nodes"},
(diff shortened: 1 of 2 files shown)


weed/shell/command_ec_common_test.go
```diff
@@ -131,3 +138,84 @@ func TestPickRackToBalanceShardsInto(t *testing.T) {
...
+func TestPickEcNodeToBalanceShardsInto(t *testing.T) {
+	testCases := []struct {
+		topology  *master_pb.TopologyInfo
+		nodeId    string
+		vid       string
+		wantOneOf []string
+		wantErr   string
+	}{
+		{topologyEc, "", "", nil, "invalid source/destination nodes"},
+		{topologyEc, "idontexist", "12737", nil, "invalid source/destination nodes"},
```
chrislusf: Error message is not clear.

ghost: Done. Note these are internal errors, so they should really hit an end user under normal circumstances.

proton-lisandro-pin:
> Do you want to try some generic constraint satisfaction problem solver?
> 
> The only one I found so far is https://github.com/gnboorse/centipede

Interesting. Probably once i finish cleaning this logic up a bit?

chrislusf:
* May need to consider spreading to different disks later.
* Data shards and parity shards can have different constraints.

PR #6303, opened 2024-11-29
Resolve replica placement for EC volumes from master server defaults.

proton-lisandro-pin:

Implements a stopgap solution for EC re-balancing based on system topology, as Seaweed currently has no support for shard placement strategy options.

The plan is to rely on volume replication settings to filter destination racks/nodes for EC shards which need to move. Since EC shards have no replication information, we read EC replica settings from command line arguments, or master defaults if left unspecified.

Moving forward, we'll need a better solution in place - maybe as simple as a new `shard_placement` field.

# How is the PR tested?

New unit test for `parseReplicaPlacementArg()`, updated unit tests for `getDefaultReplicaPlacement()`.

# Checks

### DIFF
--- a/weed/shell/command_ec_balance.go
+++ b/weed/shell/command_ec_balance.go
@@ -17,6 +17,7 @@ func (c *commandEcBalance) Name() string {
 	return "ec.balance"
 }
 
+// TODO: Update help string and move to command_ec_common.go once shard replica placement logic is enabled.
 func (c *commandEcBalance) Help() string {
 	return `balance all ec shards among all racks and volume servers
 
--- a/weed/shell/command_ec_balance.go
+++ b/weed/shell/command_ec_balance.go
@@ -20,7 +20,7 @@ func (c *commandEcBalance) Name() string {
 func (c *commandEcBalance) Help() string {
 	return `balance all ec shards among all racks and volume servers
 
-	ec.balance [-c EACH_COLLECTION|<collection_name>] [-force] [-dataCenter <data_center>]
+	ec.balance [-c EACH_COLLECTION|<collection_name>] [-force] [-dataCenter <data_center>] [-shardReplicaPlacement <replica_placement>]
 
 	Algorithm:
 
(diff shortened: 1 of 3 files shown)


chrislusf:
Or use a command line parameter?

proton-lisandro-pin:
> Or use a command line parameter?

As in, for the EC default replication placement? Or to control use/don't use replication data overall?

_Some_ sort of solution without requiring additional config knobs would be nice. As noted, we currently don't have a better solution than relying on master defaults to do so.

chrislusf:
for the EC default replication placement

proton-lisandro-pin:
> for the EC default replication placement

Doable. I'd prefer something relying on _existing_ topology configuration, even if imperfect, instead of a new config switch. It would be easy to miss for end users.

Let me know which one you prefer though, and i'll update the PR.

chrislusf:
Default value for the command can derive from the master if necessary.

proton-lisandro-pin:
> Default value for the command can derive from the master if necessary.

That works :smile: Fixed, PTAL.

weed/shell/command_ec_balance.go
```diff
@@ -100,6 +101,7 @@ func (c *commandEcBalance) Do(args []string, commandEnv *CommandEnv, writer io.W
 	balanceCommand := flag.NewFlagSet(c.Name(), flag.ContinueOnError)
 	collection := balanceCommand.String("collection", "EACH_COLLECTION", "collection name, or \"EACH_COLLECTION\" for each collection")
 	dc := balanceCommand.String("dataCenter", "", "only apply the balancing for this dataCenter")
+	shardReplicaPlacement := balanceCommand.String("shardReplicaPlacement", "", "replica placement for EC shards, or master default if empty (currently unused)")
```
chrislusf, line 104: The notation for shard placement may need a better design.

chrislusf: `ec.encode` would also need this parameter?

ghost: It will - haven't gotten to unifying the rebalancing logic yet, but that's coming soon.

PR #6338, opened 2024-12-10
[shell] volume.list show only writable volumes 

kmlebedev:

It is necessary to track the position of the writable volumes on the disks

Add writable. filer

# How is the PR tested?

# Checks

### DIFF
--- a/weed/shell/command_volume_list.go
+++ b/weed/shell/command_volume_list.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/storage/erasure_coding"
+	"github.com/seaweedfs/seaweedfs/weed/storage/types"
 	"golang.org/x/exp/slices"
 	"path/filepath"
 	"strings"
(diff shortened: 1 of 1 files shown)


weed/shell/command_volume_list.go
```diff
@@ -172,6 +176,11 @@ func (c *commandVolumeList) writeDiskInfo(writer io.Writer, t *master_pb.DiskInf
 		if c.isNotMatchDiskInfo(vi.ReadOnly, vi.Collection, vi.Id) {
 			continue
 		}
+		if *c.writable {
+			if c.volumeSizeLimitMb >= vi.Size {
```
chrislusf: there could be other conditions to determine whether a volume is writable, e.g., being readonly, or a remote volume, ec volumes, etc.

dsd2077:
hi, @kmlebedev 
I found the logic of judging if a volume is writable has a little problem.
https://github.com/seaweedfs/seaweedfs/blob/da2a234b0040bcc59e31c3a851006e45efc9ff32/weed/shell/command_volume_list.go#L172

the condition should change to 
```shell
c.volumeSizeLimitMb <= uint64(volumeSize)
```

PR #6342, opened 2024-12-11
Begin implementing EC balancing parallelization support.

proton-lisandro-pin:

Begin the implementation of parallel EC re-balancing operations within `shell.EcBalance()`.

This impacts both `ec.encode` and `ec.balance`.

# How is the PR tested?

No algorithmic changes, other than running EC sub-tasks in parallel. Following changes will likely require at least unit testing.

# Checks


proton-lisandro-pin:
> how to prevent one volume server running multiple EC encoding processes?

Right now there's no parallelization in place for the EC encoding phase - just the re-balancing happening after.

Was thinking about running encodings in parallel though :cry: Is this a problem? I assumed volume servers could handle multiple EC encoding requests without issues.

chrislusf:
Later need to add a parallel limit on each volume server for the ec encoding/decoding.

PR #6346, opened 2024-12-12
[shell] only apply the balancing for writable volumes

kmlebedev:

Often new volumes are allocated on some one node, but there is no possibility to do a full rebalancing
https://github.com/seaweedfs/seaweedfs/issues/5735
https://github.com/seaweedfs/seaweedfs/pull/6338

Perform partial balancing of volumes of unfilled and available for writing new files using a filter `writable`

# How is the PR tested?

Not yet

# Checks


chrislusf:
test failed:

```
FAIL	github.com/seaweedfs/seaweedfs/weed/shell [build failed]
Error: shell/command_volume_balance_test.go:259:12: undefined: balanceVolumeServers
```

PR #6349, opened 2024-12-13
fix: record and delete bucket metrics after inactive

zouyixiong:

Metrics related to collections and buckets should be deleted after they become inactive. If not, we will end up with more and more huge metrics as the system runs, especially in environments where buckets are created, uploaded to, and deleted over time.

1 Delete collection metrics after the collection is deleted.
2 Record and delete bucket metrics after they become inactive for 10 minutes.

# How is the PR tested?

Tested in my development environment on the master branch.

# Checks

### DIFF
--- a/weed/stats/metrics.go
+++ b/weed/stats/metrics.go
@@ -23,11 +23,11 @@ const (
 	NoWriteOrDelete  = "noWriteOrDelete"
 	NoWriteCanDelete = "noWriteCanDelete"
 	IsDiskSpaceLow   = "isDiskSpaceLow"
-	bucketAtiveTTL   = 10 * 60 * 1000 // 10 min
+	bucketAtiveTTL   = 10 * time.Minute
 )
 
 var readOnlyVolumeTypes = [4]string{IsReadOnly, NoWriteOrDelete, NoWriteCanDelete, IsDiskSpaceLow}
-var bucketLastActiveTime map[string]int64 = map[string]int64{}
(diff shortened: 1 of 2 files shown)


weed/stats/metrics.go
```diff
@@ -23,9 +23,11 @@ const (
 	NoWriteOrDelete  = "noWriteOrDelete"
 	NoWriteCanDelete = "noWriteCanDelete"
 	IsDiskSpaceLow   = "isDiskSpaceLow"
+	bucketAtiveTTL   = 10 * 60 * 1000 // 10 min
```
chrislusf: use time.Duration here

weed/stats/metrics.go
```diff
@@ -23,9 +23,11 @@ const (
...
 )
 
 var readOnlyVolumeTypes = [4]string{IsReadOnly, NoWriteOrDelete, NoWriteCanDelete, IsDiskSpaceLow}
+var bucketLastActiveTime map[string]int64 = map[string]int64{}
```
chrislusf: use nano seconds and change name to bucketLastActiveTsNs, to be clear about the unit.

weed/stats/metrics.go
zouyixiong: updated

weed/stats/metrics.go
zouyixiong: updated

PR #6357, opened 2024-12-16
Update store.go

coffeecloudgit:
Add lock for location.ecVolumes when for each location.ecVolumes

fatal error: concurrent map iteration and map write

Add lock

# How is the PR tested?
no test

# Checks


coffeecloudgit:
No fix

PR #6374, opened 2024-12-17
Parallelize volume replica operations within `ec.encode`.

proton-lisandro-pin:

Make `ec.encode` to parallelize flagging volume replicas as writeable.

# How is the PR tested?

No algorithmic changes; this PR parallelizes existing EC encoding operations, which are already unit-tested.

# Checks

### DIFF
--- a/weed/shell/command_ec_encode.go
+++ b/weed/shell/command_ec_encode.go
@@ -119,7 +119,7 @@ func (c *commandEcEncode) Do(args []string, commandEnv *CommandEnv, writer io.Wr
 
 	// encode all requested volumes...
 	for _, vid := range volumeIds {
-		if err = doEcEncode(commandEnv, *collection, vid); err != nil {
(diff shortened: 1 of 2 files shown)


weed/shell/command_ec_encode.go
```diff
@@ -119,7 +119,7 @@ func (c *commandEcEncode) Do(args []string, commandEnv *CommandEnv, writer io.Wr
 
 	// encode all requested volumes...
 	for _, vid := range volumeIds {
-		if err = doEcEncode(commandEnv, *collection, vid); err != nil {
+		if err = doEcEncode(commandEnv, *collection, vid, *parallelize); err != nil {
```
chrislusf, line 122: how about using an integer of the number of paralleling goroutines?

ghost: Sure, see https://github.com/seaweedfs/seaweedfs/pull/6376 . I'll rebase this PR once the change is in.

PR #6377, opened 2024-12-18
Fix volume replica parallelization within `ec.encode`.

proton-lisandro-pin:

Fix volume replica parallelization for `ec.encode`. In particular, this PR makes the number of parallel tasks configurable.

See 826edd5d.

# How is the PR tested?

No functional / test changes.

# Checks


proton-lisandro-pin:
@chrislusf , the weed shell is currently broken at head :disappointed:  This PR fixes it.

PR #6387, opened 2024-12-21
Replace bolt dependency with bbolt

hack3ric:

There are some dependencies that still depends on the unmaintained bolt after https://github.com/seaweedfs/seaweedfs/pull/3554. Replace it so that we can have one less dependency, and build on newer architecture like RISC-V.

Two lines of `replace` are required because Go refuses to replace `github.com/boltdb/bolt` with `go.etcd.io/bbolt` when the latter is also present in the dependency tree. Replace it again with `github.com/etcd-io/bbolt` solves the issue.

# How is the PR tested?

We are able to build SeaweedFS and pass tests on Arch Linux RISC-V: https://github.com/felixonmars/archriscv-packages/pull/4376

# Checks

The PR does not introduce visible changes to usage.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -369,3 +369,8 @@ require (
 )
 
 // replace github.com/seaweedfs/raft => /Users/chrislu/go/src/github.com/seaweedfs/raft
+
+replace (
+	github.com/boltdb/bolt => go.etcd.io/bbolt v1.3.10
+	go.etcd.io/bbolt => github.com/etcd-io/bbolt v1.3.10
+)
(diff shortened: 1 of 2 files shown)


chrislusf:
@hack3ric This breaks "go install" in the common path.

Is it possible to conditionally add the replacement? For example, add this replacement in some risk-v branch?

hack3ric:
Seems Go 1.16 changed this behaviour and prevents any module with `replace`: https://github.com/golang/go/issues/44840

I personally think it is unnecessary to add such restriction in the first place, but if we do need `go install` to work properly, we may have to revert this PR until Go addresses the problem, and include this `replace`s downstream in our distros (which we do now). Not the most ideal solution but it works.

Another way would be guiding users to clone first and run `go install` against local repo like https://github.com/golang/go/issues/44840#issuecomment-1828537390.

chrislusf:
@hack3ric thanks for the details! Supporting `go install` will help to reduce the friction.

Please send a PR to revert this when the downstream repo is ready.

chrislusf:
@hack3ric I am going to revert this change and hopefully the downstream can already handle this.

hack3ric:
#Sorry for the late reply! We can already handle this and you can revert it anytime.

PR #6422, opened 2025-01-09
[weed] add parallel for volume.fix.replication

dsd2077:
Assume a scenario: If one of my volume server machines fails and cannot be restored for a while, and if there is a lot of data on this server, executing volume.fix.replication in this situation will be a very lengthy process. Therefore, it is necessary to add concurrent operations to volume.fix.replication.

# How is the PR tested?
shut down a volume server，executing `volume.fix.replication`，it will  replicating volumes concurrently.

# Checks

### DIFF
--- a/weed/shell/command_volume_fix_replication.go
+++ b/weed/shell/command_volume_fix_replication.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"path/filepath"
 	"strconv"
+	"sync"
 	"time"
 
 	"slices"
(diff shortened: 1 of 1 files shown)


chrislusf (changes requested):
each fix will change the topology, and the following fix would need to use the new topology to plan the move.

this PR would cause volumes to be placed on the most empty volume server for the concurrent fixes.

chrislusf (changes requested):
still the same problem. this is not about concurrent access.

dsd2077:
> still the same problem. this is not about concurrent access.

 @chrislusf Can you give me some ideas?
 In `fixOneUnderReplicatedVolume` I add lock to find a target node. Once found, `addVolumeCount` will change the topology, so the next concurrent thread will use the new topology to plan the move.
 
What's wrong with the logic?

```go
func findTargetNode(allLocations []location, replica *VolumeReplica, replicaPlacement *super_block.ReplicaPlacement, replicas []*VolumeReplica) *location {
	keepDataNodesSorted(allLocations, types.ToDiskType(replica.info.DiskType))
	fn := capacityByFreeVolumeCount(types.ToDiskType(replica.info.DiskType))
	for _, dst := range allLocations {
		if fn(dst.dataNode) > 0 && satisfyReplicaPlacement(replicaPlacement, replicas, dst) {
			addVolumeCount(dst.dataNode.DiskInfos[replica.info.DiskType], 1)
			return &dst
		}
	}
	return nil
}
```

PR #6444, opened 2025-01-15
Add bucket's traffic metrics

hadizamani021:
We added metrics to calculate the sent and received traffic of a bucket, meaning the number of bytes sent to and received from a bucket are counted. There are somethings like that in MinIo and other storages.
https://min.io/docs/minio/container/operations/monitoring/metrics-and-alerts.html#bucket-api

By adding metrics and increasing them only if the bytes are transferred to or from the filer successfully

# How is the PR tested?
It is tested locally.
image

# Checks

### DIFF
--- a/weed/stats/metrics.go
+++ b/weed/stats/metrics.go
@@ -319,6 +319,22 @@ var (
 			Name:      "in_flight_requests",
 			Help:      "Current number of in-flight requests being handled by s3.",
 		}, []string{"type"})
+
+	S3BucketTrafficReceivedBytes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: "s3",
+			Name:      "bucket_traffic_received_bytes",
+			Help:      "Total number of bytes sent from an S3 bucket to clients.",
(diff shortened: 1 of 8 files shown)


weed/stats/metrics.go
```diff
@@ -319,6 +319,22 @@ var (
 			Name:      "in_flight_requests",
 			Help:      "Current number of in-flight requests being handled by s3.",
 		}, []string{"type"})
+
+	S3BucketTrafficReceivedBytes = prometheus.NewGaugeVec(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: "s3",
+			Name:      "bucket_traffic_received_bytes",
+			Help:      "Total number of bytes sent from an S3 bucket to clients.",
```
chrislusf: the help messages seem wrong

chrislusf:
Please also try to adjust the grafana pages.

hadizamani021:
I added dashboards.
dash2

weed/stats/metrics.go
hadizamani021: I changed them.

PR #6443, opened 2025-01-15
fix S3 per-user-directory Policy

tcrasset:
This fixes two bugs:
- We were not allowing a non-bucket path to be provided as a `Resource` to `PolicyDocument`
- We were not picking up the correct Key from the bucket when checking for authentication.

`test.py` (as an example) shows that it works now, and when running in on `master`, did not work. Output of that script below.

I will remove `test.py` it once you approve, so that I can keep using that script if you decide that some changes are necessary.

```
❯ uv run test.py
Reading inline script metadata from `test.py`
SENTINEL file created at /tmp/SENTINEL
Creating Power User...
I0116 15:03:02.740972 masterclient.go:228 master localhost:9333 redirected to leader master:9333
.Power User created with key: power_user_key and secret: power_user_secret
Creating Bucket test-bucket-30e7c2ca...
Bucket test-bucket-30e7c2ca created.
Uploading /tmp/SENTINEL for Alice... on user-id-Alice
File /tmp/SENTINEL uploaded for Alice.
Uploading /tmp/SENTINEL for Bob... on user-id-Bob
File /tmp/SENTINEL uploaded for Bob.
Listing files of s3://test-bucket-30e7c2ca/...
	- user-id-Alice/SENTINEL
	- user-id-Bob/SENTINEL
Creating user Alice...
User Alice created with access key: BXTWKZIDSDS0IS769YQN3
Creating user Bob...
User Bob created with access key: CYQVHCK09JIKZK4ICQ9GG
Listing files of s3://test-bucket-30e7c2ca/...
	- user-id-Alice/SENTINEL
	- user-id-Bob/SENTINEL
Listing files of s3://test-bucket-30e7c2ca/user-id-Alice...
	- user-id-Alice/SENTINEL
Listing files of s3://test-bucket-30e7c2ca/user-id-Bob...
	- user-id-Bob/SENTINEL
Creating policy for Alice on test-bucket-30e7c2ca...
Policy for Alice on test-bucket-30e7c2ca created.
Creating policy for Bob on test-bucket-30e7c2ca...
Policy for Bob on test-bucket-30e7c2ca created.
Listing files of s3://test-bucket-30e7c2ca/user-id-Alice/...
	- user-id-Alice/SENTINEL
Listing files of s3://test-bucket-30e7c2ca/user-id-Bob/...
	- user-id-Bob/SENTINEL
Creating policy for Alice on test-bucket-30e7c2ca...
Policy for Alice on test-bucket-30e7c2ca created.
[...]

PR #6446, opened 2025-01-16
improve iam error handling

tcrasset:
While working on https://github.com/seaweedfs/seaweedfs/pull/6443, I noticed lots of errors are silently ignored by the `IamApiServer`, most notably in `GetActions`.

I had a lot of trouble, as a user of the Iam API, to understand why my `PolicyDocument` was not being parsed correctly. 

Instead of silently ignoring errors when parsing the PolicyDocument, I fail and return an error to the user.

Along the way, I improved the error management to better return the error to the user.


weed/iamapi/iamapi_handlers.go
```diff
@@ -1,31 +1,45 @@
...
+	"net/http"
+
 	"github.com/aws/aws-sdk-go/service/iam"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
-	"net/http"
 )
 
-func writeIamErrorResponse(w http.ResponseWriter, r *http.Request, err error, object string, value string, msg error) {
-	errCode := err.Error()
```
tcrasset, line 12: `writeIamErrorResponse` should not be responsible for creating error messages. It should only determine the HTTP Status code depending on the error code, and forward the message (or not) to the user.

The logic for determining the error code and message has been moved down to the respective IamApiServer functions.

weed/iamapi/iamapi_management_handlers.go
```diff
@@ -196,12 +200,12 @@ func GetPolicyDocument(policy *string) (policyDocument PolicyDocument, err error
...
 }
 
-func (iama *IamApiServer) CreatePolicy(s3cfg *iam_pb.S3ApiConfiguration, values url.Values) (resp CreatePolicyResponse, err error) {
+func (iama *IamApiServer) CreatePolicy(s3cfg *iam_pb.S3ApiConfiguration, values url.Values) (resp CreatePolicyResponse, iamError *IamError) {
 	policyName := values.Get("PolicyName")
 	policyDocumentString := values.Get("PolicyDocument")
 	policyDocument, err := GetPolicyDocument(&policyDocumentString)
 	if err != nil {
-		return CreatePolicyResponse{}, err
+		return CreatePolicyResponse{}, &IamError{Code: iam.ErrCodeMalformedPolicyDocumentException, Error: err}
```
tcrasset, line 208: I decided that not being able to parse a PolicyDocument warrants a `iam.ErrCodeMalformedPolicyDocumentException`

weed/iamapi/iamapi_management_handlers.go
```diff
@@ -212,37 +216,47 @@ func (iama *IamApiServer) CreatePolicy(s3cfg *iam_pb.S3ApiConfiguration, values
 	policyLock.Lock()
 	defer policyLock.Unlock()
 	if err = iama.s3ApiConfig.GetPolicies(&policies); err != nil {
-		return resp, err
+		return resp, &IamError{Code: iam.ErrCodeServiceFailureException, Error: err}
 	}
 	policies.Policies[policyName] = policyDocument
 	if err = iama.s3ApiConfig.PutPolicies(&policies); err != nil {
-		return resp, err
+		return resp, &IamError{Code: iam.ErrCodeServiceFailureException, Error: err}
```
tcrasset, line 223: But in this case, we can't load the config from the s3 container, so I decided it's better to respons as `iam.ErrCodeServiceFailureException` --> `HTTP 500`

weed/iamapi/iamapi_test.go
```diff
@@ -152,6 +153,53 @@ func TestPutUserPolicy(t *testing.T) {
...
+}
+
+func extractErrorCodeAndMessage(response *httptest.ResponseRecorder) (string, string) {
+	pattern := `<Error><Code>(.*)</Code><Message>(.*)</Message><Type>(.*)</Type></Error>`
+	re := regexp.MustCompile(pattern)
+
+	code := re.FindStringSubmatch(response.Body.String())[1]
+	message := re.FindStringSubmatch(response.Body.String())[2]
+	return code, message
+}
```
tcrasset, line 201: I added this because I noticed that `xml.Unmarshal`  of `executeRequest` does not work with the response, whether it is `ErrorResponse`, or any other `Response` e.g. `PutUserPolicyResponse`

weed/iamapi/iamapi_test.go
```diff
@@ -152,6 +153,53 @@ func TestPutUserPolicy(t *testing.T) {
 	assert.Equal(t, http.StatusOK, response.Code)
 }
 
+func TestPutUserPolicyError(t *testing.T) {
```
tcrasset, line 156: I added this test to verify that we can actually return a valid error code and message.

weed/iamapi/iamapi_management_handlers.go
```diff
@@ -293,38 +307,38 @@ func (iama *IamApiServer) GetUserPolicy(s3cfg *iam_pb.S3ApiConfiguration, values
...
-			continue
+			return nil, fmt.Errorf("not a valid effect: '%s'. Only 'Allow' is possible", statement.Effect)
 		}
 		for _, resource := range statement.Resource {
 			// Parse "arn:aws:s3:::my-bucket/shared/*"
 			res := strings.Split(resource, ":")
 			if len(res) != 6 || res[0] != "arn" || res[1] != "aws" || res[2] != "s3" {
-				glog.Infof("not match resource: %s", res)
-				continue
+				return nil, fmt.Errorf("not a valid resource: '%s'. Expected prefix 'arn:aws:s3'", res)
```
tcrasset, line 335: Instead of continuing, we return an error

PR #6404, opened 2024-12-31
worm grace period and retention time support

27149chen:
# What problem are we solving?

grace period and retention time are important features for WORM
grace period: The time from the creation of the file to its transformation into WORM. files will become WORM if they do not change for the grace_period duration
retention_time: A period of worm protection,  if the retention time is exceeded, the file becomes a normal file

Implementation references for some products:
[ONTAP SnapLock](https://docs.netapp.com/us-en/ontap/snaplock/commit-files-worm-state-manual-task.html): support grace period, retention_time, commit files to WORM manually/automatically
[OceanStor HyperLock](https://support.huawei.com/enterprise/zh/doc/EDOC1100264905/8776298f): support grace period, retention_time, commit files to WORM manually/automatically
[CTERA Vault](https://www.ctera.com/blog/immutable-file-systems-ctera-worm-storage/): support grace period, retention_time, commit files to WORM automatically

# How are we solving the problem?
this pr:
1. add two new fields in filer conf: worm_grace_period_seconds and worm_retention_time_seconds
2. add a new field in Entry: worm_enforced_at_ts_ns
3. support grace_period, in the grace_period duration, a file can be modified/deleted even worm is enabled
4. support to commit files to WORM manually by running: chmod -w
5. support to commit files to WORM automatically if the file do not change for the grace_period duration (TODO)
6. support retention_time, if the retention time is exceeded, the file becomes a normal file, if retention_time is zero, the file will never expire

# How is the PR tested?

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #6464, opened 2025-01-21
Commit files to WORM automatically

27149chen:

This pr implements the WORM auto-commit feature mentioned in this pr: https://github.com/seaweedfs/seaweedfs/pull/6404

1. create a queue and several workers to handle the auto-commit logic
2. when a file is created or updated, send it to the queue if it is worm enabled
3. one of the worker will pick the file from the queue, check if it should be committed right now. If yes, commit it by setting WORMEnforcedAtTsNs field, if no, re-add it to the queue later
4. when filer starts, loop all worm paths to find the files which need to commit and send them to the queue.

about the queue:
1. it is a priority queue, file with old mtime will be handled first
2. it ensures that one file can only be added once, if the file is processing, new add reqest will be dropped, if the file is in the queue, it will be replaced by new add request

# How is the PR tested?

# Checks

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -155,6 +155,7 @@ require (
 	go.uber.org/atomic v1.11.0
 	golang.org/x/sync v0.10.0
 	google.golang.org/grpc/security/advancedtls v1.0.0
+	k8s.io/utils v0.0.0-20241210054802-24370beab758
 )
 
 require (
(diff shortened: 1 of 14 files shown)


27149chen:
1. add a startup params: enableWORMAutoCommit, defuault to false, so user can enable it in one filer

27149chen:
2. a worker process includes the following things:
a. get the entry from store
b. update entry and save it to store if it is necessary

there are only one or two store query/update, I think it is fast enough. also we can increase the number of workers.
And the queue also can help to dedup the items

27149chen:
3. a full loop is necessary because during the filer down time , some files may have exceeded the grace period, if no new changes, they will never been committed as worm

the loop happens in backupgroud, and only choose the files which ate not worm enforced yet, so I think the impact is relatively small

27149chen:
@chrislusf please take a look

chrislusf:
S3 now has full support for this.

Issue #6517, opened 2025-02-06
implement aws s3 signature of type `STREAMING-UNSIGNED-PAYLOAD-TRAILER`


tcrasset:
We currently only support 
```
// Streaming AWS Signature Version '4' constants.
const (
	emptySHA256            = "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
	streamingContentSHA256 = "STREAMING-AWS4-HMAC-SHA256-PAYLOAD"
	signV4ChunkedAlgorithm = "AWS4-HMAC-SHA256-PAYLOAD"

	// http Header "x-amz-content-sha256" == "UNSIGNED-PAYLOAD" indicates that the
	// client did not calculate sha256 of the payload.
	unsignedPayload = "UNSIGNED-PAYLOAD"
)
```

but it seems like there are some calls that have `STREAMING-UNSIGNED-PAYLOAD-TRAILER`, which makes our signing process fail.


See some logs below.


```
│ I0206 17:22:49.230051 auth_credentials.go:362 v4 auth type                                                                                                                                                                                                                                                                                  │
│ I0206 17:22:49.230081 auth_signature_v4.go:45 reqSignatureV4Verify &{PUT /6b488237-3a75-4f77-9481-9f2f7df51e2a/69d2d994-af21-471e-95d1-e9ba2f3ecde5/6b488237-3a75-4f77-9481-9f2f7df51e2a/input/iris.csv HTTP/1.1 1 1 map[Accept:[*/*] Accept-Encoding:[identity] Amz-Sdk-Invocation-Id:[bcd0b94c-b480-470c-ba70-3b84fabcc478] Amz-Sdk-Reque │
│ st:[attempt=1] Authorization:[AWS4-HMAC-SHA256 Credential=5JP0QRRKADPXG08Q0FJU2/20250206/us-west-1/s3/aws4_request, SignedHeaders=content-encoding;host;transfer-encoding;x-amz-content-sha256;x-amz-date;x-amz-decoded-content-length;x-amz-sdk-checksum-algorithm;x-amz-trailer, Signature=135b10430d6b23ebf05407e4a370a582b10b6ec895d1cb │
[...]

PR #6525, opened 2025-02-07
add s3 signature tests and prepare implementation of STREAMING-UNSIGNED-PAYLOAD-TRAILER

tcrasset:
- **add tests for s3 signature**
- **add test for newSignV4ChunkedReader.Read()**

We add tests so that we can tackle the following issues, without breaking existing stuff:
- https://github.com/seaweedfs/seaweedfs/issues/6526
- https://github.com/seaweedfs/seaweedfs/issues/6517

# How is the PR tested?

# Checks
- [c] I have added unit tests if possible.

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -364,6 +364,9 @@ func (iam *IdentityAccessManagement) authRequest(r *http.Request, action Action)
 		glog.V(3).Infof("post policy auth type")
 		r.Header.Set(s3_constants.AmzAuthType, "PostPolicy")
 		return identity, s3err.ErrNone
+	case authTypeStreamingUnsigned:
+		glog.V(3).Infof("unsigned streaming upload")
+		return identity, s3err.ErrNone
 	case authTypeJWT:
 		glog.V(3).Infof("jwt auth type")
 		r.Header.Set(s3_constants.AmzAuthType, "Jwt")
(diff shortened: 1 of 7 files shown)


weed/s3api/auth_credentials.go
```diff
@@ -412,6 +415,10 @@ func (iam *IdentityAccessManagement) authUser(r *http.Request) (*Identity, s3err
 	var authType string
 	switch getRequestAuthType(r) {
 	case authTypeStreamingSigned:
+		glog.V(3).Infof("signed streaming upload")
+		return identity, s3err.ErrNone
+	case authTypeStreamingUnsigned:
+		glog.V(3).Infof("unsigned streaming upload")
```
tcrasset, line 421: as far as I understand, we do verify the user during the signature calculation of the streaming upload, we do not need to do it here.

weed/s3api/auth_credentials.go
```diff
@@ -364,6 +364,9 @@ func (iam *IdentityAccessManagement) authRequest(r *http.Request, action Action)
 		glog.V(3).Infof("post policy auth type")
 		r.Header.Set(s3_constants.AmzAuthType, "PostPolicy")
 		return identity, s3err.ErrNone
+	case authTypeStreamingUnsigned:
+		glog.V(3).Infof("unsigned streaming upload")
+		return identity, s3err.ErrNone
```
tcrasset, line 369: but I am less certain about this part

Issue #6526, opened 2025-02-07
allow verifying trailing checksums


tcrasset:
There are multiple values allowed for the `x-amz-content-sha256` header as defined below:

 https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html#sigv4-auth-header-overview
![Image](https://github.com/user-attachments/assets/22cf1a62-940a-412d-b529-1e92a48f6e4a)

It seems like payload trailer types are only [mainly over HTTPS](https://github.com/aws/aws-sdk-go-v2/issues/1667), which is why they don't appear locally on my end, but in my prod setup they do.

The payload trailer types that add a checksum are
- STREAMING-AWS4-HMAC-SHA256-PAYLOAD-TRAILER
- STREAMING-UNSIGNED-PAYLOAD-TRAILER
- STREAMING-AWS4-ECDSA-P256-SHA256-PAYLOAD-TRAILER

Currently, we support None of them.
The first two would be a good first step, the later would need the new signature SigV4A to be implemented.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/checking-object-integrity.html

Example HTTP Request
```
PUT /Key+ HTTP/1.1
Host: amzn-s3-demo-bucket
Content-Encoding: aws-chunked
x-amz-decoded-content-length: 17408
x-amz-content-sha256: STREAMING-UNSIGNED-PAYLOAD-TRAILER
x-amz-trailer: x-amz-checksum-crc32

2000\r\n                                   // Object body chunk 1 (8192 bytes)
object-bytes\r\n
2000\r\n                                   // Object body chunk 2 (8192 bytes)
object-bytes\r\n
400\r\n                                    // Object body chunk 3 (1024 bytes)
object-bytes\r\n
0\r\n                                      // Completion chunk
x-amz-checksum-crc32:YABb/g==\n\r\n\r\n    // Trailer chunk (note optional \n character)
\r\n                                         // CRLF
```

The current code is able to read streaming HTTP requests like the following, but it it is currently hardcoded that they need a chunk-signature, which `STREAMING-UNSIGNED-PAYLOAD-TRAILER` does not.

https://github.com/seaweedfs/seaweedfs/blob/9ebc132ffc1a4c6a07864a8cec05cc380a8f00e0/weed/s3api/chunked_reader_v4.go#L172

of

[...]

PR #6539, opened 2025-02-10
implement s3 streaming-unsigned-payload-trailer

tcrasset:
Closes https://github.com/seaweedfs/seaweedfs/issues/6517, closes https://github.com/seaweedfs/seaweedfs/issues/6526

We are updating the state machine of `s3ChunkedReader`.

# How is the PR tested?

Using unit tests, and manually using `boto3`

# Checks


weed/s3api/chunked_reader_v4.go
```diff
@@ -139,26 +144,85 @@ var errLineTooLong = errors.New("header line too long")
...
+	case streamingUnsignedPayload:
+		glog.V(3).Infof("streaming unsigned payload")
+		if authorizationHeader != "" {
+			// We do not need to pass the seed signature to the Reader as each chunk is not signed,
+			// but we do compute it to verify the caller has the correct permissions.
+			_, _, _, _, errCode = iam.calculateSeedSignature(req)
+			if errCode != s3err.ErrNone {
+				return nil, errCode
+			}
+		}
```
tcrasset, line 178: `calculateSeedSignature` could be improved by splitting it into:
1. validating the signature
2. computing and returning the seed signature

weed/s3api/chunked_reader_v4.go
```diff
@@ -139,26 +144,85 @@ var errLineTooLong = errors.New("header line too long")
...
 
-// newSignV4ChunkedReader returns a new s3ChunkedReader that translates the data read from r
+// newChunkedReader returns a new s3ChunkedReader that translates the data read from r
 // out of HTTP "chunked" format before returning it.
 // The s3ChunkedReader returns io.EOF when the final 0-length chunk is read.
-func (iam *IdentityAccessManagement) newSignV4ChunkedReader(req *http.Request) (io.ReadCloser, s3err.ErrorCode) {
-	ident, seedSignature, region, seedDate, errCode := iam.calculateSeedSignature(req)
-	if errCode != s3err.ErrNone {
-		return nil, errCode
+func (iam *IdentityAccessManagement) newChunkedReader(req *http.Request) (io.ReadCloser, s3err.ErrorCode) {
```
tcrasset, line 150: Renamed because it does not necessarily needs the chunks to be signed. The reader is responsible for reading the content in chunks, whether they are signed or not.

weed/s3api/chunked_reader_v4.go
```diff
@@ -139,26 +144,85 @@ var errLineTooLong = errors.New("header line too long")
...
+
+	contentSha256Header := req.Header.Get("X-Amz-Content-Sha256")
+	authorizationHeader := req.Header.Get("Authorization")
+
+	var ident *Credential
+	var seedSignature, region string
+	var seedDate time.Time
+	var errCode s3err.ErrorCode
+
+	switch contentSha256Header {
```
tcrasset, line 161: This `contentSha256Header` determines the type of streaming, while the `authorizationHeader` determines whether or not we should check the auth.

weed/s3api/chunked_reader_v4.go
```diff
@@ -169,7 +233,9 @@ type s3ChunkedReader struct {
 	region            string
 	state             chunkState
 	lastChunk         bool
-	chunkSignature    string
+	chunkSignature    string // Empty string if unsigned streaming upload.
```
tcrasset, line 236: `chunkSignature` is how we determine if we are `unsigned-streaming-upload` or the signed version. Not sure if we should use an enumeration or switch on the type. It will probably be needed in the future if we implement `SigV4A` types of signature.

weed/s3api/chunked_reader_v4.go
```diff
@@ -179,8 +245,11 @@ type s3ChunkedReader struct {
 // Read chunk reads the chunk token signature portion.
 func (cr *s3ChunkedReader) readS3ChunkHeader() {
 	// Read the first chunk line until CRLF.
-	var hexChunkSize, hexChunkSignature []byte
-	hexChunkSize, hexChunkSignature, cr.err = readChunkLine(cr.reader)
+	var bytesRead, hexChunkSize, hexChunkSignature []byte
+	bytesRead, cr.err = readChunkLine(cr.reader)
+	// Parse s3 specific chunk extension and fetch the values.
+	hexChunkSize, hexChunkSignature = parseS3ChunkExtension(bytesRead)
```
tcrasset, line 251: `readChunkLine` was split into two:
1. reading the line
2. parsing the chunk extension.

This way, it can be reused for reading the trailing chunk line

weed/s3api/chunked_reader_v4.go
```diff
@@ -202,7 +277,9 @@ const (
 	readChunkHeader chunkState = iota
 	readChunkTrailer
 	readChunk
+	readTrailerChunk
 	verifyChunk
+	verifyChecksum
```
tcrasset, line 282: We are adding two new states.

weed/s3api/chunked_reader_v4.go
```diff
@@ -246,11 +327,81 @@ func (cr *s3ChunkedReader) Read(buf []byte) (n int, err error) {
...
+			}
+
+			// If we're using unsigned streaming upload, there is no signature to verify at each chunk.
+			if cr.chunkSignature != "" {
+				cr.state = verifyChunk
+			} else if cr.lastChunk {
+				cr.state = readTrailerChunk
+			} else {
+				cr.state = readChunkHeader
+			}
```
tcrasset, line 360: This is tricky, as explained by the comment, we should probably rewrite this in the future to track down the exact part that fails, but I didn't want to do a complete rewrite of the state machine, so I decided to do this implementation of conditionally consuming (or not) the CRLF depending on whether the fact we are in the trailing chunk or not.

weed/s3api/chunked_reader_v4.go
```diff
@@ -246,11 +327,81 @@ func (cr *s3ChunkedReader) Read(buf []byte) (n int, err error) {
...
+
+			computedChecksum := cr.checkSumWriter.Sum(nil)
+			base64Checksum := base64.StdEncoding.EncodeToString(computedChecksum)
+			if string(extractedChecksum) != base64Checksum {
+				// TODO: Return BadDigest
+				glog.V(3).Infof("payload checksum '%s' does not match provided checksum '%s'", base64Checksum, string(extractedChecksum))
+				cr.err = errors.New("payload checksum does not match")
+				return 0, cr.err
 			}
```
tcrasset, line 393: The current implementation of the state machine does not allow return `s3Error` codes, but we might need to do that in the future so that we can return `ErrBadDigest`

weed/s3api/chunked_reader_v4.go
```diff
@@ -246,11 +327,81 @@ func (cr *s3ChunkedReader) Read(buf []byte) (n int, err error) {
...
-			cr.state = verifyChunk
+
+			// TODO: Extract signature from trailer chunk and verify it.
+			// For now, we just read the trailer chunk and discard it.
+
+			// Reading remaining CRLF.
+			for i := 0; i < 2; i++ {
+				cr.err = readCRLF(cr.reader)
+			}
```
tcrasset, line 401: Not really necessary, as this is the final chunk and we switch to `cr.state = eofChunk` at the end, and don't read the error message, but I wanted to highlight that we are discarding the two final CRLF. In the "signed upload with additional checksum computation for each chunk", this will error with `errMalformedEncoding`.

weed/s3api/chunked_reader_v4.go
```diff
@@ -333,12 +489,29 @@ func (cr *s3ChunkedReader) getChunkSignature(hashedChunk string) string {
 
 // readCRLF - check if reader only has '\r\n' CRLF character.
 // returns malformed encoding if it doesn't.
-func readCRLF(reader io.Reader) error {
+func readCRLF(reader *bufio.Reader) error {
```
tcrasset, line 492: The correct type being passed in `*bufio.Reader` so I updated the signature, which allowed me to use `reader.Read` and `reader.Peek`. See the next comment to show where

weed/s3api/chunked_reader_v4.go
```diff
@@ -139,26 +144,85 @@ var errLineTooLong = errors.New("header line too long")
...
+	if err != nil {
+		glog.V(3).Infof("error extracting checksum algorithm: %v", err)
+		return nil, s3err.ErrInvalidRequest
 	}
+
+	checkSumWriter := getCheckSumWriter(checksumAlgorithm)
+
 	return &s3ChunkedReader{
 		cred:              ident,
 		reader:            bufio.NewReader(req.Body),
```
tcrasset, line 194: Here we are passing the reader instance of type `bufio.Reader`

weed/s3api/chunked_reader_v4_test.go
```diff
@@ -53,13 +61,110 @@ func NewRequest() (*http.Request, error) {
...
+	checksum := writer.Sum(nil)
+	base64EncodedChecksum := base64.StdEncoding.EncodeToString(checksum)
+	trailer := "x-amz-checksum-crc32:" + base64EncodedChecksum + "\n\r\n\r\n\r\n"
+	fmt.Println("Trailer:", trailer)
 
+	payload := chunk1 + chunk2 + chunk3 + chunk4 + trailer
+	return payload
+}
+
+func NewRequestStreamingUnsignedPayloadTrailer(includeFinalCRLF bool) (*http.Request, error) {
```
tcrasset, line 109: FYI, this test does not include an `Authorization` header, so we do not check the signature, but clients do send it, and I've tested manually that it is indeed working.

tcrasset:
If all is good, and merged, I would appreciate it if you were able to create a release so that I can use this in production. Have a good day.

chrislusf:
please adjust s3 tests to reproduce the issue and prove the fix https://github.com/seaweedfs/seaweedfs/blob/master/.github/workflows/s3tests.yml

tcrasset:
> please adjust s3 tests to reproduce the issue and prove the fix [`master`/.github/workflows/s3tests.yml](https://github.com/seaweedfs/seaweedfs/blob/master/.github/workflows/s3tests.yml)

Actually, it's a simple `client.put_object`. It is due to the [new integrity changes that landed in botocore 1.36](https://github.com/boto/boto3/issues/4392).

The `ceph/s3-tests` repo [pin their version of botocore to `botocore<1.28`](https://github.com/ceph/s3-tests/blob/7a1aa1b42251c94f120393062deb9531ff7c45ef/requirements.txt#L5), so I would not be able to reproduce that particular implementation using the workflow you provided. I'll try and see if I can force it using a previous version of botocore.

tcrasset:
It seems like adding trailers in the checksum appeared somewhere between January 2022 and June 2022, as shown by the archives of the AWS doc:
- [January 2022](https://web.archive.org/web/20220126125132/https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html)
- [June 2022](https://web.archive.org/web/20220610085617/https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html#expand)

So it is definitely not available in the pinned version of our `ceph/s3-tests`.
But I can write a manual test using `botocore1.36`

tcrasset:
>  But I can write a manual test using botocore1.36.

I can't seem to reproduce it without a HTTPS proxy, and that would need a lot of setup in this repo for this test. Isn't my integration test in `chunked_reader_v4_test.go` enough?

chrislusf:
Need to adjust the integration tests, to use recent tests, in separate PRs.

PR #6547, opened 2025-02-13
fix: s3 command ignore -tlsVerifyClientCert and -cacert.file arguments

KarolBedkowski:

s3 command ignore tlsVerifyClientCert and cacert.file arguments from command line, so there is now way use mTLS for s3 https endpoint.

Fix what values are checked in startS3Server - check valued in s3opt structure (set from command line) instead of defaults in s3Options. 

# How is the PR tested?

With local ca and certificates:
`weed s3  -cacert.file ./localca.crt -cert.file ./localhost.crt -key.file ./localhost.pem  -port.https=8334 -filer xxx -tlsVerifyClientCert`
Check with openssl s_client  for "Acceptable client certificate CA names"; without patch - list is empty; after - contain certificate from command line.
Check with curl is connection required client (right) certificate; without patch client certificate is ignored; after - with wrong / missing client certificate - ssl error; with right certificate - correct response.

# Checks

### DIFF
--- a/weed/command/s3.go
+++ b/weed/command/s3.go
@@ -312,7 +312,7 @@ func (s3opt *S3Options) startS3Server() bool {
 		}
 
 		caCertPool := x509.NewCertPool()
-		if *s3Options.tlsCACertificate != "" {
+		if *s3opt.tlsCACertificate != "" {
 			// load CA certificate file and add it to list of client CAs
 			caCertFile, err := ioutil.ReadFile(*s3opt.tlsCACertificate)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


Zen3515:
I'm not sure exactly how, I suspect that the argument is added only for s3 command, disregarding filer command, as a result, this PR seems to prevent s3 to start from filer.

```bash
weed filer \
  -s3 \
  -s3.port="443" \
 ...
```

will not work for version 3.85, but works fine in 3.84 due to nil pointer dereference

KarolBedkowski:
Ffixed in https://github.com/seaweedfs/seaweedfs/pull/6659
Sorry about it.

PR #6550, opened 2025-02-17
fix: The free disk size and percentage are quite different from the output of df -h.

zouyixiong:

The free disk size and percentage are quite different from the output of df -h.

image
image 
image

According to the man page: https://man7.org/linux/man-pages/man3/statvfs.3.html
fs.Bfree: Number of free blocks
fs.Bavail: Number of free blocks for unprivileged users

fs.Bavail is the better value to use here.
As the image3 shows you, the `avail` percentage is more match the output of `df -h`.

// disk.Free = fs.Bfree * uint64(fs.Bsize)
disk.Free = fs.Bavail * uint64(fs.Bsize)

# How is the PR tested?
see test code: test/disk/disk_status.go
Test env:
ubuntu: 22.04
Architecture: x86_64
go: go version go1.22.7 linux/amd64

By the way, other system implementations haven't been changed yet because I don't have a test environment.
such as: disk_solaris.go, disk_openbsd.go disk_windows.go

# Checks


test/disk/disk_status.go
chrislusf, line 1: move this file to the `unmaintained` folder since this is not for tests.

zouyixiong: updated

proton-lisandro-pin:
Note this is very likely the root cause for https://github.com/seaweedfs/seaweedfs/discussions/6372 - i was looking into the issue and narrowed it down to `fillInDiskStatus()` as well.

PR #6582, opened 2025-02-27
add cassandra2

chrislusf:

When listing directory children, Cassandra LSM storage would still need to traverse all sub folders.

Partition key adds dishash, so when listing a directory's direct children, no grand children are traversed by Cassandra LSM.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -139,12 +139,13 @@ connection_max_lifetime_seconds = 0
 enableUpsert = true
 upsertQuery = """UPSERT INTO "%[1]s" (dirhash,name,directory,meta) VALUES($1,$2,$3,$4)"""
 
-[cassandra]
+[cassandra2]
 # CREATE TABLE filemeta (
+#    dirhash bigint,
 #    directory varchar,
 #    name varchar,
 #    meta blob,
-#    PRIMARY KEY (directory, name)
+#    PRIMARY KEY ((dirhash, directory), name)
 # ) WITH CLUSTERING ORDER BY (name ASC);
 enabled = false
 keyspace = "seaweedfs"
--- a/weed/filer/cassandra2/README.txt
+++ b/weed/filer/cassandra2/README.txt
@@ -1,0 +1,15 @@
+1. create a keyspace
+
+CREATE KEYSPACE seaweedfs WITH replication = {'class':'SimpleStrategy', 'replication_factor' : 1};
+
+2. create filemeta table
+
+ USE seaweedfs;
+
+ CREATE TABLE filemeta (
+    dirhash bigint,
+    directory varchar,
+    name varchar,
+    meta blob,
+    PRIMARY KEY ((dirhash, directory), name)
+ ) WITH CLUSTERING ORDER BY (name ASC);
--- a/weed/filer/cassandra2/cassandra_store.go
+++ b/weed/filer/cassandra2/cassandra_store.go
@@ -1,0 +1,221 @@
+package cassandra2
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/gocql/gocql"
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/filer"
+	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+	"github.com/seaweedfs/seaweedfs/weed/util"
+)
+
+func init() {
+	filer.Stores = append(filer.Stores, &Cassandra2Store{})
+}
+
+type Cassandra2Store struct {
+	cluster                 *gocql.ClusterConfig
+	session                 *gocql.Session
+	superLargeDirectoryHash map[string]string
+}
+
+func (store *Cassandra2Store) GetName() string {
+	return "cassandra2"
+}
+
+func (store *Cassandra2Store) Initialize(configuration util.Configuration, prefix string) (err error) {
+	return store.initialize(
+		configuration.GetString(prefix+"keyspace"),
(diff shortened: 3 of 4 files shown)


faurik:
Hello @chrislusf !

Could you please explain a little bit on how this change will increase listing performance on Scylla/Cassandra?

Feels like partitioning key became thicker and that's it.

chrislusf:
- Previously partitioning by directory only would put data related to one directory to one Cassandra node. But on the Cassandra node, the data are still ordered by directory. Scanning one prefix would match all descendents. E.g., If you're listing `/a`, it will scan `/a/b/c`.
- The partition key (dirhash, directory) means Cassandra groups data only by exact (dirhash, directory) pairs. This prevents prefix-based queries from scanning hierarchical directories.

vndzzz:
Hi, @chrislusf! 
Sorry, but I don't understand in details how it impacts reading from cassandra. As far as I understand cassandra works like this:
1. Node is selected using consistent hashing and partitioner. By default cassandra uses Murmur3Partitioner, so `/a` and `/a/b/c` will go to different token ranges and different nodes.
2. If in some circumstances two directories go to the same sstable, partitions will be sorted (memtable sorting is based on `SkipListMemtable`, `ShardedSkipListMemtable`, `TrieMemtable` https://github.com/apache/cassandra/blob/4318e74180d710844b075fedc33e8ca008f87f63/src/java/org/apache/cassandra/db/memtable/Memtable_API.md) alphanumerically. 
3. When cassandra reads data from SSTable it uses bloom filter, then uses index to find which blocks of data file contains necessary partition. And then it read from data file only those blocks, that are necessary for retrieving data for request. More details of sstable are provided here - https://cassandra.apache.org/doc/latest/cassandra/architecture/storage-engine.html
4. In seaweedfs source code when directory is listed, directory parameter is set as one partition value. 

In summary, I understand that salting directory randomises partitions, containing same prefix. But I don't understand how it prevents prefix-based queries, as I don't see where it exists without salting. 

Could you, please, provide more details?

chrislusf:
1. with limited number of Cassandra nodes N, some folders, about 1/N, under /a will go to the same Cassandra node.
2. Yes. The records would be sorted as:
  1. /a
  2. /a/b/c/c.txt
  3. /a/b/d/d.txt
  4. /a/f.txt
  So to listing operation would not need to scan c.txt and d.txt after salting.
3. Reading one file would not be affected. Listing operation can not use bloom filter.

vndzzz:
Sorry, I need to go deeper 🙂
I still don't see where scanning happen. 

CQL request is like this 
`SELECT name, meta FROM filemeta WHERE directory=? AND name>=? ORDER BY NAME ASC LIMIT ?`. To execute it only one value for directory is provided. Scanning occur only for `name` column. But this column is not part of partition key, it is part of clustering key. So scan is running only through one partition, not others. 

For example, let's we have following sstable:
```
dir, name

/a, b
..., f.txt
..., g.txt
..., m.txt
/a/b, c
...,  d
/a/b/c, c.txt
........., other_file_in_c.txt
/a/b/d, d.txt
........, other_file_in_d.txt
```

Here, we have 4 partitions with partition key value (dir) equal to `/a`, `/a/b`, `/a/b/c`, `/a/b/d`. These partitions are in sorted order as they are in same sstable. Rows inside each partition are sorted according to clustering keys. So all rows of same partition will be near other rows of the exact same parition (this ordering happens when memtable is populating in memory of node, or when compaction is merging two sstables).

Now, let's see what happens, when we execute `SELECT name, meta FROM filemeta WHERE directory='/a' AND name>='g.txt' ORDER BY NAME ASC LIMIT`. In this example cassandra only fetches bytes, where partition with key `/a` exists, and then makes scan starting from `g.txt` row and finishing after `m.txt`. It finishes because after `m.txt` new partition starts. 

So, as far as I understand no scanning should be done without salting. Please correct me, if I'm wrong in my circumstances.

chrislusf:
there are no separate storage for each shard. All data are in the same LSM files.

vndzzz:
>there are no separate storage for each shard. All data are in the same LSM files.

Yes, sure. I didn't wrote anything about different files. In prev comment I tried to explain what happens with one file.

lapo-luchini:
> So, as far as I understand no scanning should be done without salting. Please correct me, if I'm wrong in my circumstances.

@vndzzz I think the same, i.e. that the old primary index
`PRIMARY KEY ((directory), name)`
and the new primary index
`PRIMARY KEY ((dirhash, directory), name)`
are functionally identical, only difference is the second is a bit bigger and involves some extra hashing.

Queries are done on (hash of) the full partition key and thus always an exact match, never on "prefix".

In the first case the primary node is something like `CASSANDRA_HASH(directory) % NUM_NODES`, in the second case is `CASSANDRA_HASH(directory || MD5(directory)) % NUM_NODES` which is a different number, but has exactly the same "specificity" and will only always match the exact directory just as well.

Maybe I'm missing something about Cassandra internals? 🤔

lapo-luchini:
[Here in the docs](https://cassandra.apache.org/doc/stable/cassandra/architecture/storage-engine.html#sstables):

> Within the `Data.db` file, rows are organized by partition. These partitions are sorted in token order (i.e. by a hash of the partition key when the default partitioner, `Murmur3Partition`, is used). Within a partition, rows are stored in the order of their clustering keys.

So the files from [comment](https://github.com/seaweedfs/seaweedfs/pull/6582#issuecomment-2758444226) would actually be sorted as this (they are not a single string, but two parts) and c and d needs not to be skipped:
- (/a/b/d, d.txt)
- (/a, a.txt)
- (/a, f.txt)
- (/a/b/c, c.txt)

(here I'm assuming that Murmur3 hash of `/a/b/d` is less than the one of `/a`, which is completely arbitrary, but just to represent the fact they are not alphabetically ordered)

PR #6715, opened 2025-04-16
Fix implementation of `master_pb.CollectionList` RPC call

proton-lisandro-pin:

Currently, `CollectionListRequest.include_normal_volumes` is ignored; the call always returns regular volumes regardless of value.

Modified `weed.topology.ListCollections()` so `includeNormalVolumes` is properly evaluated.

# How is the PR tested?

Added unit tests for `ListCollections()` verifying parameter behavior.

# Checks

### DIFF
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"math/rand/v2"
+	"slices"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


proton-lisandro-pin:
@chrislusf - should i change the default value for `CollectionListRequest.include_normal_volumes` as well here? I'm not sure if clients are relying on this field being ignored.

PR #6738, opened 2025-04-27
feat(redis): add mTLS support for Redis connection initialization

jadedevin13:
- Enhanced the Redis2Store initialization to support mutual TLS (mTLS) by adding configuration options for CA certificate, client certificate, and client key paths.
- Updated the Redis client setup to use TLS configuration when mTLS is enabled, ensuring secure connections to the Redis server.

# How is the PR tested?

# Checks


chrislusf:
Please add the examples in the scaffold generated filer.toml.

Issue #6752, opened 2025-05-02
[Feature Request] Add SFTP Server Support Backed by SeaweedFS Filer


mohsek:
I'd like to propose adding support for an SFTP (SSH File Transfer Protocol) server as a new access layer for SeaweedFS, similar to the existing S3 and WebDAV APIs. This feature would allow users to interact with the SeaweedFS filer using standard SFTP clients (e.g., sftp, FileZilla, WinSCP) while leveraging the full capabilities of the SeaweedFS backend.

🔍 Use Case
Provide secure file transfer over SSH directly into SeaweedFS.

Enable users to use familiar SFTP tools for uploading, downloading, and managing files.

Ideal for environments where SFTP is the standard protocol for file exchange and automation.

🛠️ Proposed Implementation
Implement a standalone or integrated SFTP server service.

Reuse the existing filer gRPC interface for backend interactions.

Handle basic SFTP operations such as authentication, file read/write, directory listing, etc.

Authentication could initially be handled with SSH keys or a simple user/password mapping.

Userstore (users with their homedir, permissions ..)  can be implemented statically though config file 

📦 Benefits
Broader protocol support and easier integration into existing infrastructure.

Brings SeaweedFS closer to feature parity with traditional file servers.

Enhances the flexibility of deployment in secure or regulated environments.

I'd be happy to contribute to this effort and submit a PR.

Thanks!

PR #6753, opened 2025-05-02
Add SFTP Server Support

mohsek:
# What problem are we solving? (issue https://github.com/seaweedfs/seaweedfs/issues/6752) 

This PR introduces native SFTP (SSH File Transfer Protocol) server support to SeaweedFS. Many users and enterprise environments rely on SFTP for secure file transfer and automation, and this feature aims to address that gap.

Implemented an SFTP server as a new service (weed sftp) using Go's golang.org/x/crypto and github.com/pkg/sftp packages.

Integrated the SFTP layer with the existing SeaweedFS filer via the gRPC client (filer_pb).

Supported core SFTP operations such as authentication, file read/write, list directories, rename, delete, and mkdir.

Authentication is handled via SSH keys and optional password login (configurable).

Introduced a static userstore defined via a JSON configuration file:

* Each user entry includes username, password (or public key), home directory, and optional permissions (read/write).

* Each user is isolated in their respective homeDir, mapped directly to a filer path

# How is the PR tested?

# Checks


chrislusf:
This is great! Thanks!

Please check why it is breaking the s3 tests https://github.com/seaweedfs/seaweedfs/actions/runs/14798435239/job/41563733489?pr=6753

PR #6771, opened 2025-05-09
Use the correct constant when computing the offset in SearchNeedleFromSortedIndex

Okhoshi:

`NeedleHeaderSize` happens to have the same size as `NeedleMapEntrySize`, except when running the 5 bytes offset variant of Seaweedfs, because it does not contain OffsetSize. This causes ECX corruption on deletes, due to the drifting offset computation (offset is always computed on a basis of 16 bytes per record instead of 17 bytes).

Using the correct constant.

# How is the PR tested?

None for now

# Checks

### DIFF
--- a/weed/storage/erasure_coding/ec_volume.go
+++ b/weed/storage/erasure_coding/ec_volume.go
@@ -3,14 +3,13 @@ package erasure_coding
 import (
 	"errors"
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"math"
 	"os"
+	"slices"
 	"sync"
 	"time"
 
-	"slices"
-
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks! This is an important fix!

kmlebedev:
@Okhoshi @chrislusf What did this bug affect? ​​How did you notice it? What problem does it solve?

chrislusf:
Affects EC volumes with deletes on the large_disk version.

Okhoshi:
We did notice the hard way, after ECX files were corrupted and reads and deletions were failing in weird ways. It indeed only affects EC volumes, only with the large_disk version, and when they receives deletes.

PR #6778, opened 2025-05-11
fix: ydb filer bugs

SmoothDenis:

1. Bucket deletion: deleting a single S3 bucket removed all buckets due to incorrect path extraction, problem with empty ls s3:// response ater restart filer
2. Cleanup: empty directories remained in per-bucket YDB schema after bucket deletion
3. Pagination hang: listing directories with more than 1000 entries hung because YDB limited pages to 1000 rows (seaweed filer operating with 1024 entries requests)
4. Session recovery: PUT failed on YDB BAD_SESSION in cases with transport errors because of losted sessions
5. useBucketPrefix=false: unwanted per-bucket tables and directories were created even when bucket prefixing was disabled
6. Incorrect creation path to filemeta tables due to MakeRecursive, so I implemented ensureTables

1. Path fixes in DeleteFolderChildren and OnBucketDeletion, so only the target bucket is removed from main table with buckets list
2. Directory removal in OnBucketDeletion call RemoveDirectory when SupportBucketTable=true to clean up empty schema folders related to that bucket
3. Chunked pagination fix in ListDirectoryPrefixedEntries, added chunkLimit (max 1000) and pass that to YDB, avoiding 1000+ requests in one call
4. Idempotent retries in doTxOrDB to wrap non-transactional calls in Table().Do with table.WithIdempotent(). It should be safe, but I might missing some details
5. Added early return in OnBucketCreation and OnBucketDeletion when SupportBucketTable is false, so no per-bucket artifacts are created in that mode
6. Look at ensureTables

# How is the PR tested?
I tested all the scenarios described above manually, as there are currently no tests that would check:
- listing over 1000 files
- deleting a bucket and checking ls s3://
- checking YDB structures in different modes
- simulation of network problems, restarts, and other problems

I also don't quite understand the purpose of the "useBucketPrefix=false" mode. However, I tried to preserve its functionality, restore it if possible and not disrupt it with new changes.

# Checks

### DIFF
--- a/weed/filer/ydb/ydb_store.go
+++ b/weed/filer/ydb/ydb_store.go
@@ -6,6 +6,12 @@ package ydb
 import (
 	"context"
 	"fmt"
+	"os"
+	"path"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/seaweedfs/seaweedfs/weed/filer"
 	"github.com/seaweedfs/seaweedfs/weed/filer/abstract_sql"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks for the fixes! There is a test build error to address.

SmoothDenis:
I will need to check a couple more things, then I will return the request from the draft

SmoothDenis:
@chrislusf @kmlebedev look at the changes, please

PR #6790, opened 2025-05-14
correctly report volume with input/output error to master

dongxufeng:

input/output error is wrapped and could not be checked, and volume has io error cannot be correctly reported as deleted to master in heartbeat

# How is the PR tested?

mock disk failure by command echo 1 | sudo tee /sys/block/sd*/device/delete to produce input/output error

# Checks


chrislusf:
how about using "%w" wrap errors, and use "errors.Is()" function instead of relying on error text?

see https://pkg.go.dev/errors

PR #6791, opened 2025-05-14
[volume] refactor and add metrics for flight upload and download data limit condition

kmlebedev:

It seems that the concurrent Downloads Limit parameter excessively blocks read requests.

Since the limit is set at the maximum disk speed, blocking occurs, and the disk is not loaded, this is possible when reading is actually done from the page cache and perhaps the actual disk utilization should be taken into account

In any case, I would like to monitor the inFlightDownloadSize and inFlightUploadSize size metrics
in order to select the optimal size for a specific disk

# How is the PR tested?

# Checks


chrislusf:
Also need to add panels in the dashboard

PR #6811, opened 2025-05-23
Make fuse command linux/MacOS only 

Victrid:

Make fuse only able on linux and macos to prevent futural platform-specific compile errors.

https://github.com/seaweedfs/seaweedfs/pull/6809#issuecomment-2903260954

This also change the signal to comply 

> it seems that windows does not have SIGUSR1 (macOS have them as POSIX requires) causing building errors, and commit [f3d44b1](https://github.com/seaweedfs/seaweedfs/commit/f3d44b1eb65328ac6aa66bd4b5687127175cbecc) changes the listening signal to SIGTERM. In this case, this code should also be changed to SIGTERM to send correct signal:
> 
> https://github.com/seaweedfs/seaweedfs/blob/f3d44b1eb65328ac6aa66bd4b5687127175cbecc/weed/command/mount_std.go#L274

This pull request refactors the `runFuse` implementation to improve platform-specific handling and updates the signal used for parent process notification in the `RunMount` function. The primary changes involve splitting the `runFuse` logic into platform-specific files and modifying the signal sent to the parent process.

### Platform-specific handling of `runFuse`:

* **Removed old implementation**: The `runFuse` function and its dependencies were removed from `weed/command/fuse.go` to allow for platform-specific implementations.
* **Added non-supported platform handling**: A new `weed/command/fuse_notsupported.go` file was added to handle cases where FUSE is not supported (e.g., non-Linux and non-Darwin platforms). It outputs an appropriate message and returns true.
* **Added Linux/Darwin-specific implementation**: A new `weed/command/fuse_std.go` file was added, containing the `runFuse` implementation for Linux and Darwin platforms. This file retains the logic from the original `runFuse` function.

### Signal modification in `RunMount`:

* **Changed signal type**: The signal sent to the parent process in `RunMount` was changed from `SIGUSR1` to `SIGTERM`, ensuring compatibility with the new `runFuse` implementation.

# How is the PR tested?

Running the following command:
```bash
for os in linux windows darwin; do 
  GOOS=$os go build -o weed-$os
done
```

No error reported.

# Checks


chrislusf:
What about BSDs?

Victrid:
I think the `weed mount` option is not available in freebsd and openbsd and so `weed fuse`. 
The following command compiles without error:

```bash
for os in freebsd openbsd; do         
  GOOS=$os go build -o weed-$os
done
```

What is wrong with failing Ceph S3 tests? I think the code I've changed would not touch these things 🤔

Issue #6804, opened 2025-05-21
Volume server OOMs on large number of concurrent writes


proton-lisandro-pin:
We've been recently experiencing issues with v3.87 volume servers OOMing when loaded with concurrent write requests. Per pprof, the culprit is `needle_map.CompactMap` (i.e. the in-memory needle ID index) which eat up ~99% of the process' allocated memory.

![Image](https://github.com/user-attachments/assets/bc8aa320-56da-473f-9550-4bc0678798fa)

Interestingly, re-starting the server and loading back the .IDX indeces works fine, without memory issues.

proton-lisandro-pin:
A couple IDX files from crashed servers are available at https://drive.proton.me/urls/BXN5CBHHAG#NDCGs4xeVTPE

proton-lisandro-pin:
A quick overview of these indeces show needle IDs being very out of order. F.ex. from `69.idx`:

```
...
#1583282: needle 1740359781002000178, offset {{} {222 208 115 102}}, size 570949632
#1583283: needle 6798280394539063, offset {{} {50 222 202 115}}, size 1713705216
#1583284: needle 3466038607872, offset {{} {55 50 222 213}}, size 1936074106
#1583285: needle 52193918976, offset {{} {0 55 50 222}}, size -546085337
#1583285: needle 216172782518992896, offset {{} {0 0 55 50}}, size -555977882
#1583285: needle 3029796649315084032, offset {{} {0 0 0 55}}, size 853474163
#1583286: needle 7362564055564815399, offset {{} {0 0 0 0}}, size 926080764
#1583286: needle 8315385018125910040, offset {{} {39 0 0 0}}, size 3617503
#1583287: needle 3347131311507636224, offset {{} {12 39 0 0}}, size 14130
#1583288: needle 16084170001982947328, offset {{} {0 24 39 0}}, size 55
#1583289: needle 3665638122222663936, offset {{} {0 0 24 39}}, size 0
#1583289: needle 3977486863058745666, offset {{} {0 0 0 12}}, size 654311424
#1583290: needle 15537058142709306, offset {{} {203 0 0 0}}, size 807862272
#1583291: needle 60691633828710, offset {{} {64 212 0 0}}, size 1582848
#1583292: needle 237076697459, offset {{} {102 67 221 0}}, size 6183
#1583293: needle 926080851, offset {{} {115 102 70 230}}, size 6
[...]

PR #6842, opened 2025-06-05
New `needle_map.CompactMap()` implementation for reduced memory usage

proton-lisandro-pin:

https://github.com/seaweedfs/seaweedfs/issues/6804

This PR introduces a new `CompactMap()` implementation for in-memory map of needle indices, optimized for memory usage.

The idea behind it is replacing the bucket+overflow approach with a map of sorted indices segments, which are in turn accessed through binary search. This guarantees a best-case scenario (ordered inserts/updates) of _O(n)_ and a worst case scenario of _O(log n)_ runtime; notably, memory usage is unaffected by index ordering.

Note that even at _O(log n)_, the clock time for both reads and writes is very low, with `weed benchmark` actually running slightly faster than the existing implementation.

The end result is 20-30% improvement in memory usage, compared with v3.89. See https://github.com/seaweedfs/seaweedfs/issues/6804 for more details and benchmark results.

# How is the PR tested?

Every single existing unit and integration test should pass without issues; the existing API and behavior for `CompactMap()` is unchanged.

# Checks

### DIFF
--- a/weed/storage/needle_map/compact_map.go
+++ b/weed/storage/needle_map/compact_map.go
@@ -2,6 +2,7 @@ package needle_map
 
 import (
 	"fmt"
+	"math"
 	"sort"
 	"sync"
 
@@ -9,14 +10,23 @@ import (
 )
 
 const (
-	SegmentChunkSize = 25000
+	MaxCompactKey    = math.MaxUint16
+	SegmentChunkSize = 50000 // should be <= MaxCompactKey
 )
 
+type CompactKey uint16
+type CompactOffset [types.OffsetSize]byte
+type CompactNeedleValue struct {
+	key    CompactKey
+	offset CompactOffset
+	size   types.Size
+}
+
 type CompactMapSegment struct {
-	// TODO: maybe a compact-er structure for needle values?
-	list     []NeedleValue
-	firstKey types.NeedleId
-	lastKey  types.NeedleId
+	list     []CompactNeedleValue
+	chunk    int
+	firstKey CompactKey
+	lastKey  CompactKey
 }
 
 type CompactMap struct {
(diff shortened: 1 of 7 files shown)


chrislusf:
Could you keep the old compact map implementation and tests in separate files, just for reference and easier comparison.

ghost:
> Could you keep the old compact map implementation and tests in separate files, just for reference and easier comparison.

As in, `compact_map_old.go` and `compact_map_old_test.go`?

chrislusf:
> As in, `compact_map_old.go` and `compact_map_old_test.go`?

That would work. Or name it as CompactMapWithOverflow, since `old` will change over time.

Also, the test can have a benchmark to compare the two implementations.

ghost:
> > As in, `compact_map_old.go` and `compact_map_old_test.go`?
> 
> That would work. Or name it as CompactMapWithOverflow, since `old` will change over time.

Done!

> Also, the test can have a benchmark to compare the two implementations.

I've uploaded benchmark results on https://github.com/seaweedfs/seaweedfs/issues/6804#issuecomment-2945164212.

chrislusf:
> > Also, the test can have a benchmark to compare the two implementations.
> 
> I've uploaded benchmark results on [#6804 (comment)](https://github.com/seaweedfs/seaweedfs/issues/6804#issuecomment-2945164212).

I was thinking about https://pkg.go.dev/testing#hdr-Benchmarks

chrislusf:
> > > As in, `compact_map_old.go` and `compact_map_old_test.go`?
> > 
> > 
> > That would work. Or name it as CompactMapWithOverflow, since `old` will change over time.
> 
> Done!
> 
> >

I prefer no go build tags for this. You can put it in a sub directory if you don't want to deal with naming conflicts.

ghost:
> I prefer no go build tags for this. You can put it in a sub directory if you don't want to deal with naming conflicts.

Done. Note that i still had to tweak a couple lines, as `compact_map.go` shares a *lot* of code with the rest of `weed/storage/needle_map` :disappointed:

Issue #6837, opened 2025-06-03
Add TTL Strategy Support


FQHSLycopene:
## Current Behavior
Currently, SeaweedFS calculates file TTL based on the file's creation time (Crtime). This can be seen in the following code:

```go
if entry.TtlSec > 0 {
    if entry.Crtime.Add(time.Duration(entry.TtlSec) * time.Second).Before(time.Now()) {
        // handle expiration
    }
}
```

## Problem
1. Inconsistency with Major Cloud Storage Services:
   - AWS S3 uses Last-Modified time
   - Alibaba Cloud OSS uses Last-Modified time
   - Tencent Cloud COS uses Last-Modified time
   - And other major cloud storage services

2. Limited Use Cases:
   - When a file is updated, users typically want the TTL countdown to restart
   - Creation time may not accurately reflect the actual lifecycle of a file
   - For frequently updated files, using modification time is more reasonable

## Proposed Solution
1. Add TTL Strategy Support:
   - Add a new field to specify TTL calculation strategy
   - Support both creation time and modification time
   - Allow configuration at both file level and path rule level

2. Implementation Approach:
   - Add TtlStrategy field to Entry structure
   - Update TTL check logic in filer and mount
   - Add TTL strategy to path rules
   - Ensure backward compatibility by treating zero value as creation time

## Impact Areas
1. Core Components:
   - Filer
   - Mount
   - Storage backends
   - Path rules

2. User Interface:
   - API changes
   - Configuration changes
   - Documentation updates

## Considerations
1. Backward Compatibility:
   - Zero value of TtlStrategy will be treated as creation time
   - Existing data will continue to use creation time
   - New data can use modification time by default
   - No migration needed

2. Performance:
   - Minimize impact on existing operations
   - Consider storage overhead
   - Evaluate query performance

3. User Experience:
   - Clear documentation
   - Intuitive configuration
   - No migration required

## Next Steps
1. Discuss and finalize the proposal
2. Create detailed implementation plan
3. Implement and test changes
4. Update documentation

chrislusf:
One concern is that if one file is partially updated, for example, in a file that is split into 2 chunks, and only one chunk is updated, than the older chunk would be auto TTLed because of the chunk expires.

FQHSLycopene:
> One concern is that if one file is partially updated, for example, in a file that is split into 2 chunks, and only one chunk is updated, than the older chunk would be auto TTLed because of the chunk expires.

Thank you for raising this concern.
[...]

PR #6845, opened 2025-06-06
feat: add UpdateVolumeLastModified RPC and implement timestamp logic

FQHSLycopene:

[see issues 6837](https://github.com/seaweedfs/seaweedfs/issues/6837)

Currently, SeaweedFS calculates file TTL based on the file's creation time (Crtime), which is inconsistent with major cloud storage services (AWS S3, Alibaba Cloud OSS, Tencent Cloud COS) that use Last-Modified time. This creates several issues:

1. When a file is updated, users typically want the TTL countdown to restart
2. Creation time may not accurately reflect the actual lifecycle of a file
3. When a file is partially updated (e.g., split into chunks), some chunks might be prematurely expired due to TTL

1. Added a new `UpdateVolumeLastModified` RPC interface to update volume's last modified time and make sure volume will not be expried. 
2. Modified the TTL calculation logic to use Mtime instead of Crtime
3. Implemented volume-level timestamp updates to handle partial file updates:
   - When a file is updated, we collect all volume IDs associated with the file's chunks
   - Call `UpdateVolumeLastModified` for each related volume
   - This ensures all chunks of a file have consistent TTL behavior
4. Updated protobuf definitions and storage backend interfaces to support the new timestamp logic
5. Added proper error handling and validation for timestamp updates
6.  Modified volume get interface to remove needle-level TTL checks:
   - Commented out needle TTL validation in volume get handler
   - Chunk deletion is now handled entirely by volume expiration
   - This ensures chunks are always returned when requested, regardless of their individual TTL status

# How is the PR tested?

1. Unit Tests:
2. Integration Tests:
   - Tested file updates with TTL
   - Tested UpdateVolumeLastModified` RPC interface works
   - Tested partial file updates
   - Tested TTL expiration
   - Tested backward compatibility

3. Manual Tests:
   - Verified TTL behavior with AWS S3 compatibility
   - Tested with different file sizes and chunk configurations
   - Verified timestamp updates in volume server logs

# Checks

The implementation ensures that:
1. TTL behavior is consistent with major cloud storage services
2. Partial file updates don't cause premature chunk expiration


chrislusf:
Sorry to reply late. 

This filer-layer requirement caused changes on the block storage layer about TTL and garbage collection.
There are other possible implications on data racing, replication, etc.

Besides, this PR has too many big changes to merge safely. Many of the changes are too impactful.

FQHSLycopene:
> Sorry to reply late.
> 
> This filer-layer requirement caused changes on the block storage layer about TTL and garbage collection. There are other possible implications on data racing, replication, etc.
> 
> Besides, this PR has too many big changes to merge safely. Many of the changes are too impactful.

Thank you for your response and detailed feedback. I completely understand your concerns - this PR indeed contains too many significant changes. I will reorganize the code and split it into two smaller, safer PRs:

**PR 1: Add UpdateVolumeLastModified RPC interface and storage backend support**
- Add new RPC interface definitions and implementations
- Update storage backend interfaces to support timestamp operations
- This PR focuses on the infrastructure layer, providing the foundation for subsequent features

**PR 2: Implement Mtime-based TTL logic**
- Implement core TTL timestamp logic
- Fix volume last modified time update logic
- This PR focuses on business logic implementation

This split offers the following advantages:
1. **Clear dependencies**: PR1 provides the infrastructure, PR2 implements features on top of it
2. **Easier rollback**: If there are issues with any part, they can be handled separately

I will ensure each PR has comprehensive test coverage and provide detailed explanations of the scope of changes. What do you think about this split approach?

chrislusf:
* Even when the volume last modified time is updated, the TTL volume may still be compacted and the entry can still be dropped.
* It's simpler to just create new copies of the chunks.

FQHSLycopene:
> * Even when the volume last modified time is updated, the TTL volume may still be compacted and the entry can still be dropped.
> * It's simpler to just create new copies of the chunks.

what's you mean "create new copies of the chunks",just update needle's updatetime， or re-upload same chunk and delete old chunk (read old chunk → create new needle → write new needle → delete old needle)？

chrislusf:
> > * Even when the volume last modified time is updated, the TTL volume may still be compacted and the entry can still be dropped.
> > * It's simpler to just create new copies of the chunks.
> 
> what's you mean "create new copies of the chunks",just update needle's updatetime， or re-upload same chunk and delete old chunk (read old chunk → create new needle → write new needle → delete old needle)？

re-upload same chunk and delete old chunk

FQHSLycopene:
> > > * Even when the volume last modified time is updated, the TTL volume may still be compacted and the entry can still be dropped.
> > > * It's simpler to just create new copies of the chunks.
> > 
> > 
> > what's you mean "create new copies of the chunks",just update needle's updatetime， or re-upload same chunk and delete old chunk (read old chunk → create new needle → write new needle → delete old needle)？
> 
> re-upload same chunk and delete old chunk

let me clarify the process of re-upload:
First network transfer: Read the old chunk data from the volume server
Second network transfer: Upload the same data as a new chunk to the volume server
This means for each chunk that needs to be updated, we'll have two network round trips with the full chunk data being transmitted twice.
I would like to confirm: Do you consider this network overhead acceptable for the update operation?

chrislusf:
the re-uploading would cost a lot of resources. that's the reason supporting modified time for TTL does not have an efficient solution. I would prefer not supporting than having poor performance.

SmoothDenis:
@FQHSLycopene I think you don't understand exactly how ttl implementation works here
ttl at the volume level is used to delete the entire volume if there was no new records

At the chunk level, volume checks at the time of file read transfer and expired files are not returned, but are stored until near vacuum
A volume file with a 1d lifetime may never be deleted if it is written more often, only expired records (needle) will be deleted when the compact level is reached after run vacuum
have expired ttl = be deleted after vacuum

https://github.com/seaweedfs/seaweedfs/blob/dd464cd339de354937bc6af4fec6e12211cece69/weed/storage/volume_vacuum.go#L467-L468
https://github.com/seaweedfs/seaweedfs/blob/dd464cd339de354937bc6af4fec6e12211cece69/weed/storage/volume_vacuum.go#L482-L483

PR #6890, opened 2025-06-18
ydb filer improvements

SmoothDenis:
- partitioning of tables is not used
- an outdated driver is being used
- ydb has a new query api with no restrictions on the length of the list operation, we use the old table
- reading a non-existent bucket would create a table for it
- primary key values were declared optional
- retries were not used
- some operations, in particular deletions, check only on the hash of the directory

- added partitioning settings with some default values
- migrated to the new YDB query client and updated the driver
- prevented automatic table creation when reading non-existent buckets
- most of the fields are required (NOT NULL)
- WithIdempotent() for retry in case of problems
- query statements include directory in every case

# How is the PR tested?
load testing on self-hosted ydb cluster
existing tests

# Checks

### DIFF
--- a/weed/filer/ydb/ydb_queries.go
+++ b/weed/filer/ydb/ydb_queries.go
@@ -22,19 +22,21 @@ const (
 	deleteQuery = `
 		PRAGMA TablePathPrefix("%v");
 		DECLARE $dir_hash AS int64;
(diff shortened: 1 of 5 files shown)


SmoothDenis:
if I understand correctly, manual transaction management are needed for fs.mv. I left them on Table API for now

PR #6920, opened 2025-06-24
[volume] refactor and add metrics for flight upload and download data limit condition

kmlebedev:

It seems that the concurrent Downloads Limit parameter excessively blocks read requests.

Since the limit is set at the maximum disk speed, blocking occurs, and the disk is not loaded, this is possible when reading is actually done from the page cache and perhaps the actual disk utilization should be taken into account

In any case, I would like to monitor the inFlightDownloadSize and inFlightUploadSize size metrics
in order to select the optimal size for a specific disk

# How is the PR tested?

# Checks

### DIFF
--- a/weed/server/volume_server_handlers.go
+++ b/weed/server/volume_server_handlers.go
@@ -2,14 +2,15 @@ package weed_server
 
 import (
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
-	"github.com/seaweedfs/seaweedfs/weed/util"
-	"github.com/seaweedfs/seaweedfs/weed/util/version"
 	"net/http"
 	"strconv"
 	"strings"
 	"sync/atomic"
 	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
+	"github.com/seaweedfs/seaweedfs/weed/util"
+	"github.com/seaweedfs/seaweedfs/weed/util/version"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/security"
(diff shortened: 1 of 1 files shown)


weed/stats/metrics_names.go
```diff
@@ -4,8 +4,15 @@ package stats
...
-	WriteToLocalDisk            = "writeToLocalDisk"
-	WriteToReplicas             = "writeToReplicas"
+	WriteToLocalDisk   = "writeToLocalDisk"
+	WriteToReplicas    = "writeToReplicas"
+	DownloadLimitCond  = "downloadLimitCond"
+	UploadLimitCond    = "uploadLimitCond"
+	ReadProxyReq       = "readProxyReq"
+	ReadRedirectReq    = "readRedirectReq"
+	EmptyReadProxyLoc  = "emptyReadProxyLoc"
+	FailedReadProxyReq = "failedReadProxyReq"
```
chrislusf:
```suggestion
	DownloadLimitCond  = "downloadLimitCondition"
	UploadLimitCond    = "uploadLimitCondition"
	ReadProxyReq       = "readProxyRequest"
	ReadRedirectReq    = "readRedirectRequest"
	EmptyReadProxyLoc  = "emptyReadProxyLocaction"
	FailedReadProxyReq = "failedReadProxyRequest"
```

weed/stats/metrics.go
```diff
@@ -278,6 +278,38 @@ var (
...
+			Subsystem: "volumeServer",
+			Name:      "concurrent_download_limit",
+			Help:      "Limit total concurrent download size.",
+		})
+
+	VolumeServerConcurrentUploadLimit = prometheus.NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: Namespace,
+			Subsystem: "volumeServer",
+			Name:      "concurrent_upload_limit_size",
```
chrislusf:
```suggestion
			Name:      "concurrent_upload_limit",
```

weed/stats/metrics.go
```diff
@@ -278,6 +278,38 @@ var (
...
+			Help:      "Limit total concurrent upload size.",
+		})
```
chrislusf: these two metrics are not used

weed/command/server.go
```diff
@@ -141,6 +141,8 @@ func init() {
 	serverOptions.v.pprof = cmdServer.Flag.Bool("volume.pprof", false, "enable pprof http handlers. precludes --memprofile and --cpuprofile")
 	serverOptions.v.idxFolder = cmdServer.Flag.String("volume.dir.idx", "", "directory to store .idx files")
 	serverOptions.v.inflightUploadDataTimeout = cmdServer.Flag.Duration("volume.inflightUploadDataTimeout", 60*time.Second, "inflight upload data wait timeout of volume servers")
+	serverOptions.v.inflightDownloadDataTimeout = cmdServer.Flag.Duration("inflightDownloadDataTimeout", 60*time.Second, "inflight download data wait timeout of volume servers")
```
chrislusf:
```suggestion
	serverOptions.v.inflightDownloadDataTimeout = cmdServer.Flag.Duration("volume.inflightDownloadDataTimeout", 60*time.Second, "inflight download data wait timeout of volume servers")
```

chrislusf:
the logic to handle limits is pretty complicated and worth some more refactoring.

kmlebedev:
> the logic to handle limits is pretty complicated and worth some more refactoring.

refactoring affected avoiding long queries getting stuck and marking queries on the client side

chrislusf:
attempted a refactoring. please help to review and test.

Issue #6924, opened 2025-06-27
Request S3 with https failed


zuzuviewer:
Version： 3.92
deploy seaweedfs with docker, s3 service with https, upload with rclone failed with error message: "Head "https://xxx.xxx.xx.xxx:8433/test/update.txt": unexpected EOF", and seaweedfs      
 don't have any error message with "-v=10"

zuzuviewer:
It seems to be related to the HTTP2 protocol.

chrislusf:
Need reproducing steps to debug and confirm the fix.

PR #6943, opened 2025-07-05
* Fix undefined http serve behaiver

zuzuviewer:
Undefined http serve behaiver.

Use new object http server in every goroutine

# How is the PR tested?

# Checks

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -5,7 +5,6 @@ import (
 	"crypto/tls"
 	"crypto/x509"
 	"fmt"
-	"github.com/seaweedfs/seaweedfs/weed/util/version"
 	"net"
 	"net/http"
 	"os"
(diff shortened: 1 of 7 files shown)


chrislusf:
Not sure what problem this is trying to fix.

zuzuviewer:
> Not sure what problem this is trying to fix.

#6924  , http server object is used by several goroutines with function Serve and ServeTLS, but these function is not concurrent safety for http server object, therefore, HTTP/2 access failures may occur with a certain probability.

chrislusf:
is there any doc on the concurrency safety problem?

zuzuviewer:
> is there any doc on the concurrency safety problem?

No, but check the code, Server.Serve function will set up http2 only once with change Server.TLSConfig.NextProtos and Server.TLSNextProto, but our code(i.e weed/command/s3.go) will change Server.TLSConfig in main  goroutine, and use one Server object with unix socket, http socket, https socket in several goroutines, it's inappropriate.

chrislusf:
sounds we should configure everything first before starting the server

zuzuviewer:
I prefer every listener have their own http server

PR #6960, opened 2025-07-10
refactor: Performance and readability improvement on isDefaultPort

baikjy0215:
Refactor the ‘isDefaultPort’ function in the S3 Signature V4 verification logic to improve performance and code clarity.

It is optimized to avoid memory allocations by replacing ‘strings.ToLower’ with ‘strings.EqualFord’. Also, it now uses a switch statement for better readability and performance improvement.

# How is the PR tested?
Performance was validated by using benchmark tests comparing the current implementation with the new one. The benchmarks were run using 

```
go test -bench=. -benchmem
```

```go
package main

import (
	"strings"
	"testing"
)

func isDefaultPortV1(scheme, port string) bool {
	if port == "" {
		return true
	}

	lowerCaseScheme := strings.ToLower(scheme)
	return (lowerCaseScheme == "http" && port == "80") ||
		(lowerCaseScheme == "https" && port == "443")
}

func isDefaultPortV2(scheme, port string) bool {
	if port == "" {
		return true
	}

	switch port {
	case "80":
		return strings.EqualFold(scheme, "http")
	case "443":
		return strings.EqualFold(scheme, "https")
	default:
		return false
	}
}

// --- Benchmarks ---

func BenchmarkV1_HTTP_Match(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV1("HTTP", "80")
	}
}

func BenchmarkV2_HTTP_Match(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV2("HTTP", "80")
	}
}

func BenchmarkV1_HTTPS_Match(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV1("Https", "443")
	}
}

func BenchmarkV2_HTTPS_Match(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV2("Https", "443")
	}
}

func BenchmarkV1_NoMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV1("http", "8080")
	}
}

func BenchmarkV2_NoMatch(b *testing.B) {
	for i := 0; i < b.N; i++ {
		isDefaultPortV2("http", "8080")
	}
}
```

The results below show that the new implementation is **~6.7x faster** for matching cases and, most importantly, **completely eliminates memory allocations**.

### Benchmark Results

| Benchmark Scenario | Old (V1) ns/op | New (V2) ns/op | Speedup | Old (V1) Allocs/op | New (V2) Allocs/op |
| :----------------- | :------------- | :------------- | :------ | :----------------- | :----------------- |
| HTTP Match         | 41.66          | 6.256          | ~6.7x   | 1                  | **0**              |
| HTTPS Match        | 41.61          | 7.071          | ~5.9x   | 1                  | **0**              |
| No Match           | 6.977          | 2.091          | ~3.3x   | 0                  | **0**              |

### DIFF
--- a/weed/s3api/auth_signature_v4.go
+++ b/weed/s3api/auth_signature_v4.go
@@ -763,9 +763,14 @@ func isDefaultPort(scheme, port string) bool {
 		return true
 	}
 
-	lowerCaseScheme := strings.ToLower(scheme)
-	return (lowerCaseScheme == "http" && port == "80") ||
-		(lowerCaseScheme == "https" && port == "443")
+	switch port {
+	case "80":
+		return strings.EqualFold(scheme, "http")
+	case "443":
+		return strings.EqualFold(scheme, "https")
+	default:
+		return false
+	}
 }
 
 // getSignedHeaders generate a string i.e alphabetically sorted, semicolon-separated list of lowercase request header names


chrislusf:
Nice to learn this! Thanks!

PR #6962, opened 2025-07-10
[Notifications] Support webhook notifications

imkonsowa:
- Implement webhook publisher with MessageQueue interface
- Support bearer token authentication
- Send notifications as JSON with key and protobuf message

Real-time file event notifications by Enabling external systems to react immediately to file operations

Introducing a new configurable webhook notifications queue

# How is the PR tested?

Manuall and unit testing

# Checks
- [yes] I have added unit tests if possible.
- [yes] I will add related wiki document changes and link to this PR after merging.

### DIFF
--- a/weed/notification/webhook/http.go
+++ b/weed/notification/webhook/http.go
@@ -1,0 +1,57 @@
+package webhook
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net/http"
+
+	util_http "github.com/seaweedfs/seaweedfs/weed/util/http"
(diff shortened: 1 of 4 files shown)


chrislusf:
This is nice for the initial version. It still needs more work.

The http call may need some retrying mechanism, and should be async since it will block normal filer operations.
On production, there could be billions of files. Users may be only interested in a few sets of files. Need to support filter configurations.

imkonsowa:
Thanks for approving @chrislusf 
I agree with you, I'm planning to work on extending that feature ASAP.

Issue #7064, opened 2025-08-01
Support Reverse-Proxies with prefixes


gtudan:
PR #6092 introduced support for the X-Forwarded-Prefix header. That would allow for scenarios where proxies have to introduce a URL prefix that is stripped during the forwarding. Sadly, it introduced a regression and was later reverted. Could we try to add this again, as the feature would still be useful

chrislusf:
added a fix and updated the wiki. Let me know whether it works for you.

gtudan:
thanks a lot for the quick response! I'll give it and try and report back

gtudan:
@chrislusf Sorry, didn't manage to test the fix this week and I'll be off for a couple of days. I haven't forgotten and really appreciate the fast response!

PR #7068, opened 2025-08-01
S3: support for the X-Forwarded-Prefix header

chrislusf:
- Re-implemented X-Forwarded-Prefix support that was originally introduced in PR #6092 but later reverted due to regression
- Added X-Forwarded-Host header support for proper host header extraction in reverse proxy scenarios
- Enhanced both regular and presigned signature verification to handle reverse proxy path prefixes
- Created comprehensive tests to verify the functionality works correctly for both signature types


chrislusf:
fix https://github.com/seaweedfs/seaweedfs/issues/7064

Issue #7074, opened 2025-08-02
3.88+ ish: filer regression: context canceled when uploading via POST


cottand:
With filer v3.87 I can upload files fine, but on 3.88 onwards it fails (I tried 3.89, 3.96).
This is uploading to filer directly (not S3, webdav...).

When trying, I see variations of the following in the filer logs:
```
E0802 16:35:20.212825 filer.go:225 request_id:f6180fb5-5baf-480d-b24f-369fb97a1c52 insert entry /buckets/documents/file: update existing entry /buckets/documents/file: find /buckets/documents/file: context canceled

I0802 16:34:51.916961 common.go:81 response method:POST URL:/buckets/documents/ with httpStatus:500 and JSON:{"error":"rpc error: code = Canceled desc = context canceled"}
```

I didn't see anything that looked relevant in the master logs.

I am confident the master-volumes setup itself is fine because
- uploading files without filer works
- older filer versions work

This essentially breaks filer for me for anything 3.88 or above!

My topology: 1 datacenter, with 5 racks with 1 single datanode each
Replication settings: 020
Running on: Docker (Nomad), running on NixOS 25.05 on various Linux x86 hosts

chrislusf:
What is the filer store backend?

cottand:
ah good question, should have mentioned sorry: it's CockroachDB, here is my store config:

```toml
[postgres2]
  enabled = true
  createTable = """
    CREATE TABLE IF NOT EXISTS "%s" (
      dirhash   BIGINT,
      name      VARCHAR(65535),
      directory VARCHAR(65535),
      meta      bytea,
      PRIMARY KEY (dirhash, name)
    );
  """
  hostname = "localhost"
  port = 5432
  username = "seaweed_filer"
  password = "{{with secret ... {{end}}"
  database = "seaweed_filer" 
  schema = ""
  sslmode = "require"
  connection_max_idle = 100
  connection_max_open = 100
  connection_max_lifetime_seconds = 0
```

I have not changed these settings between upgrades.
Full config can be found [here](https://github.com/cottand/selfhosted/blob/master/jobs/seaweed/filer.nix)

cottand:
Can confirm [#7075](https://github.com/seaweedfs/seaweedfs/pull/7075) fixed it

Issue #7060, opened 2025-07-31
Context canceled followed by error 500


adionditsak:
### Describe the bug

I have had a running cluster for about a week with about 7T stored and no errors until now. I experience a sudden burst of error 500s with `context canceled`. It is fragmented and have not figured how to reproduce. Usually glog is giving me descriptive errors of where to look, but not in this case. I have searched through wiki, discussions, issues and source code.

The structure in the errors are similar to the `"no free volumes left" ` from the [S3 API FAQ](https://github.com/seaweedfs/seaweedfs/wiki/S3-API-FAQ), but with no clear message. Just `context canceled`. What can `context canceled` be an indication of? Any not obvious timeouts or other default limits in SeaweedFS? 

Excuse me for not being more clear, but for now this leads me nowhere.

This is the output from a timestamp with a mix of both weed.INFO and weed.ERROR logs in sequence:
```
filer3 6 weed[961] I0731 06:12:16.474822 masterclient.go:122 Connection wait stopped: context canceled
filer3 6 weed[961] E0731 06:12:16.474880 filer_server_handlers_write.go:49 failing to assign a file id: rpc error: code = Canceled desc = context canceled
filer3 6 weed[961] E0731 06:12:16.474923 filer_server_handlers_write_upload.go:230 upload error: rpc error: code = Canceled desc = context canceled
filer3 6 weed[961] I0731 06:12:16.474960 filer_server_handlers_write_upload.go:152 upload file 1985e65d4f0 error: rpc error: code = Canceled desc = context canceled
filer3 6 weed[961] I0731 06:12:16.475036 common.go:81 response method:PUT URL:/buckets/syslogs/1985e65d4f0 with httpStatus:500 and JSON:{"error":"rpc error: code = Canceled desc = context canceled"}
filer3 6 weed[961] E0731 06:12:16.475244 s3api_object_handlers_put.go:147 upload to filer error: rpc error: code = Canceled desc = context canceled
```

### System setup

3.91 large version

3 masters, 3 filers, 9 volumes running in LXC containers.

Example of master weed process:
`/opt/seaweedfs/bin/weed -v 0 -logdir=/var/log/seaweedfs master -mdir=/data/seaweedfs/master -peers=master1:9333,master2:9333,master3:9333 -ip=master3 -defaultReplication=010 -volumeSizeLimitMB=300000 -volumePreallocate -metricsPort=19500`

Example of master weed process:
[...]

Issue #6893, opened 2025-06-19
Writing several small files in parallel lead to congestion of the s3 service


d-consoli:
Hi, our set up is the one mentioned in [this issue](https://github.com/seaweedfs/seaweedfs/issues/6121), after solving the issues related to usage single data center per server.

The problem with are observing occurs when coping several small local files in parallel with MinIO interface (mc cp) to the cloud storage. After some minutes in which no issues are observed, the process starts to become slower, then to hang and then to return errors. At that point also reading other files from http does not work anymore.

The log do not show any specific error, and we can only observe some vacuum. Did you observe something similar and do you know what can cause it? Can it be HW related, or relative to the DB bottleneck?

Thank you for your help.

chrislusf:
> The problem with are observing occurs when coping several small local files in parallel with MinIO interface (mc cp) to the cloud storage.

Please show the command line. I am confused about "local files", "cloud storage".

leandroleal:
@chrislusf We are using mc ([MinIO Client](https://min.io/docs/minio/linux/reference/minio-mc.html)) to copy "local files" to our SeaweedFS service (17 nodes) via S3 protocol. 

We are using 11 servers to copy "local files", using  96 process in parallel per server. 
This resulted in 1056 concurrent connections to our SeaweedFS service.

chrislusf:
are the requests spread out to all S3 servers?

leandroleal:
@chrislusf Yes, we are doing it.

chrislusf:
* what are the errors?
* adding `weed -v=4 s3 ...` can enable more detailed logs. Same for other components.
* are there any grafana monitoring setup for this?

leandroleal:
@chrislusf Sorry for my long delay. It took some time to setup Grafana and update to v3.92.

The error only appears when we send hundreds of S3 concurrent requests to SeaweedFS (spread out to all S3 servers). Our current setup is creating the S3 gateway together with `filer`:

```
weed -v=4 filer -master="192.168.49.30:9333" -ip 192.168.49.30 -ip.bind=0.0.0.0 -s3.config=/s3.json -s3.ip.bind=0.0.0.0 -s3.idleTimeout 60 -s3
```

> what are the errors?

[...]

PR #7075, opened 2025-08-03
separate context for filer store

chrislusf:

https://github.com/seaweedfs/seaweedfs/issues/7074
https://github.com/seaweedfs/seaweedfs/issues/7060
https://github.com/seaweedfs/seaweedfs/issues/6893

Recent change due to 
https://github.com/seaweedfs/seaweedfs/pull/6824

caused the http request context being propagated to filer store layer.

The cancelled context errors shows up.

create a new background context to replace the http context

# How is the PR tested?

# Checks

### DIFF
--- a/weed/filer/filerstore_wrapper.go
+++ b/weed/filer/filerstore_wrapper.go
@@ -112,6 +112,7 @@ func (fsw *FilerStoreWrapper) Initialize(configuration util.Configuration, prefi
 }
 
 func (fsw *FilerStoreWrapper) InsertEntry(ctx context.Context, entry *Entry) error {
+	ctx = context.Background()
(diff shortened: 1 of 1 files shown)


cottand:
hi @chrislusf  - thanks for working on this! Agreed with the AIs, ideally use `withoutCancel(ctx)` when you don't want the upstream context to cancel the new context.

`context.Background()` should only be used as the top-level context (like in the main function) and you should not need it if you are using an HTTP/gRPC server, since the server gives you the context in the handler (like [here](https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/filer_server_handlers_read.go#L88)).

PR #7157, opened 2025-08-23
weed/mount: refactor to use atomic type

cuiweixie:

refactor to use atomic type

refactor

# How is the PR tested?

unit test

# Checks


chrislusf:
@cuiweixie thanks for the fix! Just curious, you are testing the RDMA part?

PR #7160, opened 2025-08-24
S3 API: Advanced IAM System

chrislusf:

Advanced IAM System

add Unified IAM Manager for all identity providers, STS, and policy components

# How is the PR tested?

unit and integration tests

# Checks


chrislusf:
78op]\o-i9u]\oiuy
n

Issue #7187, opened 2025-09-02
tikv store range delete may delete other k/v pairs put by rawkv client


giftz:
# Describe the bug
We are deploying a 3 vol server seaweed and 3 instance tikv on a k8s cluster, and we have some other services running on same k8s cluster.

- seaweed filer is using tikv (txn client)
- another service is using rawkv client to put some kv paris into same tikv cluster

then we performed some batch write and delete tests via s3 API, then we hit this problem:
- some how we found that the kv pairs put by rawkv client was missing partially

For example, if rawkv client put 1000 kv pairs, then after some tests, nearly half of the kv pairs will end up missing from tikv.

# System Setup
- OS version: rocky Linux 9.5: 5.14.0-503.14.1.el9_5.x86_64
- output of weed version: version 30GB 3.77 linux amd64
- filer store is tikv 8.5.1 cluster

# Expected behavior
kv pairs put by another rawkv client should not be missing

giftz:
After  some debugging, recreate the issue with following scenario:

1. Write a batch of files (like 5000) via fuse mount or s3 to one directory, e.g., /buckets/fs1/test01
2. Write 10000 kv pairs using rawkv client to same tikv cluster used by filer
3. At middle of 2, delete /buckets/fs1/test01
4. Found rawkv pairs missing at last


And the root cause is: 

When a full directory is deleted, tikv range delete will end with "filer.store.id", and this will cause the inclusion of other kv pairs put by rawkv client, then range delete will delete those kv pairs together.

PR #7188, opened 2025-09-02
replace range delete with normal txn delete

giftz:
https://github.com/seaweedfs/seaweedfs/issues/7187

Replace range delete with entry delete and commit

# How is the PR tested?
Recreate the issue with following scenario:
1. Write a batch of files (like 5000) via fuse mount or s3 to one directory, e.g., /buckets/fs1/test01
2. Write 10000 kv pairs using rawkv client to same tikv cluster used by filer
3. At middle of 2, delete /buckets/fs1/test01
4. Found rawkv pairs missing at last

Above can recreate the issue at 100% if using old version of tikv store. After the fix, the problem is gone.

# Checks

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -383,6 +383,9 @@ enabled = false
 pdaddrs = "localhost:2379"
 # Enable 1PC
 enable_1pc = false
+# batch delete count, default 10000 in code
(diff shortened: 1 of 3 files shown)


chrislusf:
This PR is nice! But there is conflicts and I do not have permission to update this PR. So I duplicated this to #7557

Issue #7202, opened 2025-09-05
`WEED_CLUSTER_SW_*` Environment Variables should not be passed to `allInOne` config


dragonfleas:
# Overview

There's an issue currently that the default `values.yaml` presents, if the `allInOne` configuration is used, the global variables for `extraEnvironmentVars` are passed into it. This should _not_ be the case as the service name for both `filer` and `master` for the allInOne config will be the allInOne service name.

## Affected Code

https://github.com/seaweedfs/seaweedfs/blob/63f4bc64a3f032a436ff54c0f688ce99bb147c63/k8s/charts/seaweedfs/values.yaml#L40-L41

## Remediation

- **(Recommended Solution)** 
  Refactor the deployments/statefulsets to dynamically and imperatively derive the service name using templates inside of a `_helpers.tpl` file and propagate them across the helm chart.

- **(Hacky Solution)**
A potential solution is to _not_ pass in `.Values.global.extraEnvironmentVars` to the allInOne configuration at all. This is likely not a good idea as it's creating additional behaviors to the `allInOne` config and additional drift in consistency with the other template formats in the chart.

chrislusf:
added https://github.com/seaweedfs/seaweedfs/pull/7217
welcome to review it.

PR #7217, opened 2025-09-09
WEED_CLUSTER_SW_* Environment Variables should not be passed to allIn…

chrislusf:
…One config

https://github.com/seaweedfs/seaweedfs/issues/7202

* Added helpers in templates/shared/_helpers.tpl
* Compute and filter keys WEED_CLUSTER_<DEFAULT>_MASTER and WEED_CLUSTER_<DEFAULT>_FILER out of both .Values.allInOne.extraEnvironmentVars and .Values.global.extraEnvironmentVars.
* Inject computed endpoints for the default cluster

# How is the PR tested?

# Checks


dragonfleas:
It's a bit explicit but it LGTM! Thanks for this!

Issue #7204, opened 2025-09-06
`DeleteEntry` retries incorrectly on transient error


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that the retry mechanism of `DeleteEntry` is incorrect, when deleting bucket. After this RPC encounters transient error during the process of returning response, it will be retried without checking whether the target (bucket here) is already deleted. Then confusing `botocore.errorfactory.NoSuchBucket` will be thrown out, which is not expected and can mislead users.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22185235/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run this script.

```python
import boto3


client_config = Config(signature_version='s3v4')
client = boto3.client(service_name='s3',
                    aws_access_key_id=config.main_access_key,
                    aws_secret_access_key=config.main_secret_key,
                    endpoint_url=config.default_endpoint,
                    use_ssl=config.default_is_secure,
                    verify=config.default_ssl_verify,
                    config=client_config)
bucket_name = get_new_bucket_name()
client.create_bucket(Bucket=bucket_name)
client.delete_bucket(Bucket=bucket_name)
```

Transient error can be reproduced by gRPC interceptor.

chrislusf:
added a PR. Please help to test it.

Issue #7203, opened 2025-09-06
`CreateEntry` retries incorrectly on transient error when creating bucket


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that the retry mechanism of `CreateEntry` when creating bucket is incorrect. After this RPC encounters transient error during the process of returning response, it will be retried without checking whether the bucket is already created. Then confusing `botocore.errorfactory.BucketAlreadyExists` is thrown, which is not expected and can mislead users.

Here is the concrete call site of this RPC, i.e., `CreateEntry`. [call_site.log](https://github.com/user-attachments/files/22185102/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run this script.

```python
import boto3


s3 = boto3.resource('s3',
                    aws_access_key_id=config.main_access_key,
                    aws_secret_access_key=config.main_secret_key,
                    endpoint_url=config.default_endpoint,
                    use_ssl=config.default_is_secure,
                    verify=config.default_ssl_verify)
name = get_new_bucket_name()
bucket = s3.Bucket(name)
bucket_location = bucket.create()
```

Transient error can be reproduced by gRPC interceptor.

chrislusf:
added a PR. Please help to test it.

PR #7218, opened 2025-09-09
Idempotent bucket create and delete

chrislusf:

- https://github.com/seaweedfs/seaweedfs/issues/7203
- https://github.com/seaweedfs/seaweedfs/issues/7204

* If the bucket directory already exists and belongs to the same owner, immediately return success instead of 409.
* Treat NotFound as success to make delete idempotent for object deletes and bucket directory cleanup during retries.

# How is the PR tested?

# Checks

### DIFF
--- a/weed/s3api/filer_util.go
+++ b/weed/s3api/filer_util.go
@@ -65,18 +65,10 @@ func doDeleteEntry(client filer_pb.SeaweedFilerClient, parentDirectoryPath strin
 
 	glog.V(1).Infof("delete entry %v/%v: %v", parentDirectoryPath, entryName, request)
 	if resp, err := client.DeleteEntry(context.Background(), request); err != nil {
-		// Treat NotFound as success for idempotency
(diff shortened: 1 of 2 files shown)


qts0312:
We have tested it. This PR can achieve idempotent bucket create and delete operation.

qts0312:
I have tested it and find the problem has not been solved. Besides, the error of `CreateEntry` changes to `BucketAlreadyOwnedByYou`.

chrislusf:
@qts0312 how to reproduce it? Preferably not a manual process.

qts0312:
We are developing a tool for reproducing bugs like this and planning to open source it soon.

chrislusf:
Update? Is this still reproducible in latest code?

qts0312:
Since our tool relies on the stack trace of RPC invocation, it may not work on the latest code (but work on my SeaweedFS fork). We are trying to update it, thanks for your patience.

qts0312:
We have updated our tool and reproduced the bugs with my fork, which is synchronized to the latest code.

PR #7236, opened 2025-09-15
[stats]: added s3 bucket stats

MilanFun:
There is a need to monitor S3 buckets in SeaweedFS, specifically to collect metrics on total bucket size (in bytes) and the number of objects. While this information can be obtained using weed shell commands, it requires custom admin scripts and does not provide out-of-the-box metrics suitable for Prometheus/Grafana.
An earlier approach involved traversing all directories and fetching entry lists, but that solution had logarithmic complexity and generated a large number of database queries (equal to the number of files in the bucket).

We leveraged SeaweedFS functionality to implement a direct way of exposing bucket metrics without the need for administrative scripts or a full traversal of all entries. This avoids excessive queries to the database while still providing accurate metrics for monitoring.

# How is the PR tested?
The functionality was tested locally to confirm correctness of metrics collection and ensure that the exposed values for bucket size and object count match expected results.

# Checks

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -49,21 +49,22 @@ type MasterOptions struct {
 	volumePreallocate          *bool
 	maxParallelVacuumPerServer *int
 	// pulseSeconds       *int
-	defaultReplication *string
-	garbageThreshold   *float64
-	whiteList          *string
-	disableHttp        *bool
-	metricsAddress     *string
-	metricsIntervalSec *int
-	raftResumeState    *bool
-	metricsHttpPort    *int
-	metricsHttpIp      *string
-	heartbeatInterval  *time.Duration
-	electionTimeout    *time.Duration
-	raftHashicorp      *bool
-	raftBootstrap      *bool
-	telemetryUrl       *string
-	telemetryEnabled   *bool
+	defaultReplication     *string
+	garbageThreshold       *float64
+	whiteList              *string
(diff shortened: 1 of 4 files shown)


chrislusf:
In general, I prefer not to merge this with the following reasons:

* There are already metrics collected at volume level. The bucket stats can be aggregated from the volume metrics with grouping by collection.
* This extra goroutine would be duplicated for each master.
* Extra CLI option can be avoided

MilanFun:
@chrislusf Thank you for your attention and review, appreciate it!

I see your point, but this change has the advantage that it allows us to use the bucket statistics directly, without having to rely on grouping and aggregation from volume-level metrics.

Another important aspect is that the current volume-level metrics don’t easily reflect the user-visible occupied size. To derive that, we would need to divide the raw totals by the replication factor (which can change dynamically) and also account for server failures or data relocation. As a result, the existing metrics can fluctuate in ways that don’t match what a user expects to see. The bucket-level stats in this change always show the actual logical data size from the user’s perspective, regardless of system changes.

Yes, I agree that the goroutine will be launched on each master, but the benefit is simplified access to bucket-level stats. Do you see any alternative place where this logic could potentially be hosted, so we avoid duplication while still keeping the direct bucket metrics available?

MilanFun:
Implemented in https://github.com/seaweedfs/seaweedfs/pull/7776

PR #7254, opened 2025-09-18
CI: replace sed patch with heredoc Python to fix indentation in s3tests

MohamedYassin-J:
- The original workflow used sed to inject multi-line Python into `s3tests_boto3/functional/test_s3.py`, which produced malformed indentation and caused `IndentationError` during runs.
- See original snippet in upstream workflow for context: `https://github.com/seaweedfs/seaweedfs/blob/master/.github/workflows/s3tests.yml`.

- Replace the sed-based replacements with a heredoc Python script that:
  - Reads `s3tests_boto3/functional/test_s3.py`.
  - Uses regex with captured leading whitespace to preserve correct Python indentation.
  - Writes the modified file back safely.
- This avoids YAML/shell escaping pitfalls and ensures valid Python indentation.

# How is the PR tested?
- Executed the updated step in the GitHub Actions workflow; tests proceed without `IndentationError`.

# Checks


chrislusf:
Does IndentationError still happen?

MohamedYassin-J:
That issue has been resolved! I'll go ahead and close this PR now. Thank you!

PR #7279, opened 2025-09-25
s3: fix SigV4 permission check

ehooi:

s3 proxy returns AccessDenied, even though the caller only performs reads via read-only access-key.

add an action parameter to `reqSignatureV4Verify`

# How is the PR tested?

added some tests

# Checks


chrislusf:
the solution in this PR has already been implemented.

PR #7331, opened 2025-10-14
fix(helm): set securitycontext for idx move initcontainer if enabled

wittdennis:
When deploying the volume component with idx into a namespace that enforces pod security standards ([docs](https://kubernetes.io/docs/tasks/configure-pod-container/enforce-standards-namespace-labels/)) we can't set the securityContext for the idx initContainer and so the resource is blocked.

This pr aims to also apply the containerSecurityContext set for the volume service to the initContainer.

# How is the PR tested?

# Checks


chrislusf:
Thanks for fixing!

PR #7340, opened 2025-10-19
[weed] update volume.fix.replication description

nightcoffee:
According https://github.com/seaweedfs/seaweedfs/pull/6421
only add -force option will apply the fix change

# How is the PR tested?

# Checks


chrislusf:
this change would affect existing scheduled jobs.

nightcoffee:
> this change would affect existing scheduled jobs.

This commit only modifies the relevant descriptions and templates; what essentially affects the execution of the scheduled task is this commit https://github.com/seaweedfs/seaweedfs/pull/6421, which changes the semantics of the command.

PR #7341, opened 2025-10-20
Refactor connection locking to use read lock

marios1861:
Changed locking mechanism to use read lock for connection check.

# What problem are we solving?
Fixes #7192 

# How are we solving the problem?
The mutex lock was held for the entire duration of the Connect() function [defer unlock](https://github.com/seaweedfs/seaweedfs/blob/97f30287821e1c49b816f2e4c05be46728b06a0b/weed/worker/client.go#L77). This caused a deadlock when [attemptConnection tried to obtain the read lock](https://github.com/seaweedfs/seaweedfs/blob/97f30287821e1c49b816f2e4c05be46728b06a0b/weed/worker/client.go#L119). 

# How is the PR tested?
Fixes #7192 with admin server on v3.97

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

PR #7350, opened 2025-10-21
Fix deadlock in worker client Connect() method

### LINKED ISSUE #7192: worker discrepancies
This is all with version 8000GB 3.97

```
$ weed help worker
...
  -capabilities string
    	comma-separated list of task types this worker can handle (default "vacuum,ec,remote,replication,balance")
...
```

logs:
```
I0902 14:30:19.454258 worker.go:64 Capabilities: vacuum,remote,replication,balance
W0902 14:30:19.454291 worker.go:215 Unknown capability: remote
W0902 14:30:19.454839 worker.go:215 Unknown capability: replication
...
I0902 14:30:19.457124 worker.go:213 WORKER STARTING: Worker w-lgtm-s-6634 starting with capabilities [vacuum balance], max concurrent: 2
I0902 14:30:19.458208 client.go:161 Connected to admin server at localhost:33646
```

admin ui:

Image

chrislusf:
## What problem are we solving?

Fixes #7192

The  method in  has a deadlock that prevents workers from connecting to the admin server. This is caused by improper lock management.

## Root Cause

The original code held a write lock for the entire duration of  using . When  calls  at line 87, and  tries to acquire a read lock at line 119 to access `c.lastWorkerInfo`, a deadlock occurs because:

1. The same goroutine holds a write lock (from `Connect()`)
2. Tries to acquire a read lock (in `attemptConnection()`)
3. Read locks cannot be acquired while a write lock is held by the same goroutine

## How are we solving the problem?

The fix removes the `defer c.mutex.Unlock()` pattern and manually releases the lock after checking the connected state but before calling `attemptConnection()`. This allows `attemptConnection()` to safely acquire its own locks.

**Key changes:**
- Remove `defer c.mutex.Unlock()` 
- Add explicit `c.mutex.Unlock()` after the connected check
- Add comment explaining the lock management

This is a cleaner solution than PR #7341 which incorrectly used `RLock()` and would have created race conditions around state modifications.

## How is the PR tested?

This fixes the deadlock reported in #7192. The worker can now successfully connect to the admin server without hanging.

## Checks

### DIFF
--- a/weed/worker/client.go
+++ b/weed/worker/client.go
@@ -74,11 +74,12 @@ func NewGrpcAdminClient(adminAddress string, workerID string, dialOption grpc.Di
 // Connect establishes gRPC connection to admin server with TLS detection
 func (c *GrpcAdminClient) Connect() error {
 	c.mutex.Lock()
-	defer c.mutex.Unlock()
-
 	if c.connected {
+		c.mutex.Unlock()
 		return fmt.Errorf("already connected")
 	}
+	// Release lock before calling attemptConnection which needs to acquire locks internally
+	c.mutex.Unlock()
 
 	// Always start the reconnection loop, even if initial connection fails
 	go c.reconnectionLoop()


chrislusf:
Thanks @gemini-code-assist! Just to clarify for anyone looking at this review comment - the suggested code in your comment is exactly the same as what's already implemented in this PR. Your review is confirming that the fix is correct, which I appreciate!

The key changes in this PR are:
1. **Removed** `defer c.mutex.Unlock()` (which was holding the lock for the entire function)
2. **Added** explicit unlock after the connected check
3. **Added** explicit unlock before calling attemptConnection()

This prevents the deadlock where Connect() held a write lock while attemptConnection() tried to acquire a read lock.

weed/worker/client.go
```diff
@@ -74,11 +74,12 @@ func NewGrpcAdminClient(adminAddress string, workerID string, dialOption grpc.Di
...
 func (c *GrpcAdminClient) Connect() error {
 	c.mutex.Lock()
-	defer c.mutex.Unlock()
-
 	if c.connected {
+		c.mutex.Unlock()
 		return fmt.Errorf("already connected")
 	}
+	// Release lock before calling attemptConnection which needs to acquire locks internally
+	c.mutex.Unlock()
```
marios1861, line 82: The lock is released but `c.connected` hasn't yet been set. Race condition remains.

PR #7439, opened 2025-11-04
Add defaults to help text for S3 tlsVerifyClientCert

yahooguntu:
The S3 server doesn't automatically turn on tlsVerifyClientCert if a CA is provided. This was unexpected for me, since the behavior of other services configured via security.toml is to turn on TLS client verification if a CA is provided.

Add to the help text for the tlsVerifyClientCert options to indicate the default value is false.

# How is the PR tested?
Not tested

# Checks

## Summary by CodeRabbit

* **Documentation**
  * Updated S3 TLS client certificate verification flag help text to clarify the default behavior.


santurak1:
What do want

chrislusf:
Each CLI option has a default value. Explicitly adding default value to the help message is not necessary and does not align with other options.

PR #7426, opened 2025-11-03
do delete expired entries on s3 list request

kmlebedev:
# What problem are we solving?

https://github.com/seaweedfs/seaweedfs/issues/6837
In S3, objects must be deleted based on their modification time.

# How are we solving the problem?

Disable deletion by creation time on the filler side and enable deletion in s3

# How is the PR tested?

I'll finish writing

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.

## Summary by CodeRabbit

* **New Features**
  * S3 TTL header/flag support, bucket-wide TTL update operation, and new APIs for checking/removing empty directories.

* **Bug Fixes**
  * Preserve/propagate S3 TTL across uploads and multipart completes, disable native TTL deletion for S3-marked objects, batched expired-entry cleanup to avoid DB deadlocks, and safer bounded empty-directory removal during deletes.

* **Tests**
  * Added lifecycle expiration tests and CI updates to run them.

* **Chores**
  * Lifecycle interval variants and dependency module updates.

PR #7444, opened 2025-11-06
S3: Also delete parent directory if empty

chrislusf:

Client side code to clean up empty parent folders can more easily get into data inconsistency.

Follow up for https://github.com/seaweedfs/seaweedfs/pull/7426

Use filer grpc service to clean up empty folders, instead of using client side logic.

# How is the PR tested?

integration tests

# Checks

## Summary by CodeRabbit

* **New Features**
  * Optional automatic cleanup of empty parent directories after deletions, with a configurable stop path available on delete requests.
  * Batch deletions now trigger deferred, scoped cleanup after processing, and cleanup runs even if the original request is cancelled.

* **Bug Fixes**
  * Expired S3 objects: deletion failures now surface errors instead of returning not-found.
  * Non‑S3 TTL expirations: deletion failures are now logged and returned rather than silently ignored.

* **Refactor**
  * Consolidated deletion flow to ensure cleanup completes using a non‑cancellable operation context.

### DIFF
--- a/weed/s3api/s3api_object_handlers_delete.go
+++ b/weed/s3api/s3api_object_handlers_delete.go
@@ -215,6 +215,7 @@ func (s3a *S3ApiServer) DeleteMultipleObjectsHandler(w http.ResponseWriter, r *h
 	var deleteErrors []DeleteError
 	var auditLog *s3err.AccessLog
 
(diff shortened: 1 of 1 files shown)


chrislusf:
This is deprecated after PR #7614

PR #7500, opened 2025-11-18
de/compress the fs meta file if filename ends with gz/gzip

shaofeng66:
Add the ability to fs.meta.save/load directly from gzipped file, this reduce the requirement of disk space if the deflated file would be large.

Pass data thru gzip pipe if the filename provided ends with .gz or .gzip, also default to .gz  file when saving meta.

# How is the PR tested?
Tested simply run fs.meta.save/load in weed shell.

# Checks

## Summary by CodeRabbit

* **New Features**
  * Added gzip compression support for metadata load and save.
  * Metadata files can now be read from and written to .gz or .gzip formats.
  * Default metadata output filename now uses .meta.gz for automatic compression.

* **Improvements**
  * Improved stream handling and error propagation for compressed and uncompressed I/O.

### DIFF
--- a/weed/shell/command_fs_meta_save.go
+++ b/weed/shell/command_fs_meta_save.go
@@ -40,7 +40,7 @@ func (c *commandFsMetaSave) Help() string {
 	fs.meta.save .               # save from current directory
 	fs.meta.save                 # save from current directory
 
-	The meta data will be saved into a local <filer_host>-<port>-<time>.meta file.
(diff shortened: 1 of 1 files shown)


shaofeng66:
Code change in this pr could increases exposure of an issue addressed by https://github.com/seaweedfs/seaweedfs/pull/8240. Pls also include pr 8240 in release.

PR #7625, opened 2025-12-04
fix: Admin UI user creation fails before filer discovery (#7624)

### LINKED ISSUE #7624: Unable to create users with weed admin 4.01, same config works with 4.00 binary
When using v4.01 weed binary, I cannot create users with the Admin UI. This error appears all the time:
Image

A downgrade to 4.00 fixes the problem and I can create users.

chrislusf:
## Description

Fixes #7624 - Unable to create users with weed admin 4.01

### Problem

When using v4.01 weed binary, users cannot create users with the Admin UI immediately after startup. The error appears:
```
failed to create user: failed to load configuration: filer_etc: filer address function not configured
```

This is a **race condition** in the admin server startup:
1. Admin server starts and spawns a background goroutine to configure the filer address for the credential manager
2. The goroutine was polling every **5 seconds** waiting for filer discovery
3. If a user tried to create a user before the first successful poll, they would get the error

### Solution

1. **Faster filer discovery with exponential backoff** - Changed the retry logic from a fixed 5-second interval to start with 200ms retries and exponentially back off up to 5 seconds. This means the filer address will be configured much faster if a filer is available.

2. **Better error messages** - Updated the error messages to be more user-friendly:
   - `filer not yet available - please wait for filer discovery to complete and try again`
   - `no filer discovered yet - please ensure a filer is running and accessible`

### Testing

The fix has been verified to compile. Users should now see the credential manager configured within ~1 second of startup instead of waiting up to 5 seconds.

### Changes
- `weed/admin/dash/admin_server.go`: Exponential backoff for filer discovery
- `weed/credential/filer_etc/filer_etc_store.go`: Improved error messages

## Summary by CodeRabbit

* **Improvements**
  * Replaced polling wait with immediate configuration to reduce startup delay.
  * Filer address resolution is now HA-aware and returns the current active filer.
  * Clearer, more actionable error messages when filer services aren’t yet available.
  * Added informational logs confirming credential store configuration and support for filer address resolution.

### DIFF
--- a/weed/admin/dash/admin_server.go
+++ b/weed/admin/dash/admin_server.go
@@ -101,36 +101,18 @@ func NewAdminServer(masters string, templateFS http.FileSystem, dataDir string) 
 		server.credentialManager = credentialManager
 		glog.V(0).Infof("Credential manager initialized with store type: %s", credentialManager.GetStore().GetName())
 
-		// For stores that need filer address function, set them
+		// For stores that need filer address function, configure them
 		if store := credentialManager.GetStore(); store != nil {
 			if filerFuncSetter, ok := store.(interface {
 				SetFilerAddressFunc(func() pb.ServerAddress, grpc.DialOption)
 			}); ok {
-				glog.V(0).Infof("Credential store supports filer address function, starting discovery goroutine")
-				// Set up a goroutine to configure filer address function once we discover filers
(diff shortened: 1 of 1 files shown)


aldy505:
Hi @chrislusf I'm wondering when will the next release be? I really dependent on this fix.

PR #7632, opened 2025-12-05
Separate vacuum speed from replication speed

m-sementsov:

Current implementation of "compaction speed' variable is rather problematic on HDDs.
Setting it to a large value, like 128MB, results in 2 things happening:
1. Volume replication, in case of disk failure, and volume balancing are fast
2. Volume vacuuming is also fast

The second thing is the whole problem.
If you have a bucket of 1000-2000 volumes spanning across all of your drives - 128MB vacuum speed results in VERY intensive IO load on each drive during parallelized vacuum.

Separating these values into 2 different variables solves this issue:
1. We can set a high value on maintenance speed, like 128MB, so that volume replication (and so their data recovery) is fast.
2. We can set a lower value on vacuum speed that won't disrupt normal cluster operation.

# How is the PR tested?

I've been running a custom build with these changes for the last 2-3 months, it worked flawlessly, so far :)

# Checks

## Summary by CodeRabbit

* **New Features**
  * Added a new command-line configuration flag to control the maintenance I/O rate, allowing administrators to set throughput limits for maintenance operations.

* **Bug Fixes**
  * Updated default throttling behavior for file copy operations to use the maintenance I/O rate instead of compaction rate when no explicit limit is specified.


chrislusf:
@m-sementsov thanks! This is great! It was something I had in mind but did not get to it.

PR #7669, opened 2025-12-09
fix: restore volume mount when VolumeConfigure fails

### LINKED ISSUE #7666: randomly failled on ` volume.configure.replication` with `unexpected token  oldFormat: failed to unmarshal old version volume info: proto: syntax error (line 1:1): unexpected token`
**this may cause data loss in some case**

I was running `volume.configure.replication -collectionPattern ... -replication 014` and I got this error:

`nmarshal error: proto: syntax error (line 1:1): unexpected token  oldFormat: failed to unmarshal old version volume info: proto: syntax error (line 1:1): unexpected token`, then the volume disappear.

This message appear randomly on different volumes, and I can't find a reason.

the volumes are all in version 3

the log of volume server:

```
volume-hdd-4-8t-1  | I1208 16:29:04.202162 volume_grpc_client_to_master.go:256 volume server 192.168.2.18:8074 deletes volume 10
volume-hdd-4-8t-1  | I1208 16:29:36.083750 store.go:607 UnmountVolume 9
volume-hdd-4-8t-1  | W1208 16:29:36.084431 volume_info.go:48 unmarshal error: proto: syntax error (line 1:1): unexpected token  oldFormat: failed to unmarshal old version volume info: proto: syntax error (line 1:1): unexpected token
volume-hdd-4-8t-1  | E1208 16:29:36.084456 volume_grpc_admin.go:135 volume configure volume_id:9 replication:"014": volume 9 failed to load vif: unmarshal error: proto: syntax error (line 1:1): unexpected token  oldFormat: failed to unmarshal old version volume info: proto: syntax error (line 1:1): unexpected token
volume-hdd-4-8t-1  | I1208 16:29:36.084566 volume_grpc_client_to_master.go:256 volume server 192.168.2.18:8074 deletes volume 9
volume-hdd-4-8t-1  | W1208 16:30:23.244256 volume_info.go:48 unmarshal error: proto: syntax error (line 1:1): unexpected token  oldFormat: failed to unmarshal old version volume info: proto: syntax error (line 1:1): unexpected token
```

chrislusf:
## Problem

When running `volume.configure.replication` command, if the `.vif` file fails to parse (e.g., corrupted or unexpected format), the volume disappears from the cluster, causing potential data loss.

### Root Cause

The `VolumeConfigure` function:
1. Calls `UnmountVolume()` which immediately notifies the master that the volume is "deleted"
2. Calls `ConfigureVolume()` to modify the `.vif` file
3. If step 2 fails, the function returns **without re-mounting the volume**
4. The volume remains unmounted and the master thinks it's deleted

## Solution

When `ConfigureVolume` fails, attempt to re-mount the volume to restore its state before returning the error. This ensures the volume remains available even if the configuration change fails.

## Changes

- Added recovery logic in `VolumeConfigure` to re-mount the volume when `ConfigureVolume` fails

Fixes #7666

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced volume configuration reliability by automatically attempting to remount volumes when initial configuration fails, with improved error logging for better troubleshooting.

### DIFF
--- a/weed/server/volume_grpc_admin.go
+++ b/weed/server/volume_grpc_admin.go
@@ -137,6 +137,7 @@ func (vs *VolumeServer) VolumeConfigure(ctx context.Context, req *volume_server_
 		// Try to re-mount to restore the volume state
 		if mountErr := vs.store.MountVolume(needle.VolumeId(req.VolumeId)); mountErr != nil {
 			glog.Errorf("volume configure failed to restore mount %v: %v", req, mountErr)
+			resp.Error += fmt.Sprintf(". Also failed to restore mount: %v", mountErr)
 		}
 		return resp, nil
 	}


chrislusf:
Regarding the concern about `.vif` parsing errors in `MountVolume`:

Looking at the code flow, the mount should **succeed even with a corrupted `.vif` file**:

1. `MountVolume` → `LoadVolume` → `loadExistingVolume` → `NewVolume` → `load()` → `maybeLoadVolumeInfo()`

2. In `volume_tier.go`, `maybeLoadVolumeInfo()` only **logs** the error but doesn't fail the volume load:
```go
if err != nil {
    glog.Warningf("load volume %d.vif file: %v", v.Id, err)
    return
}
```

3. The `.vif` file is optional - it's used for tiered storage metadata. The volume's core data (`.dat` and `.idx` files) can be loaded independently.

However, I've already addressed the second concern in the latest commit - if remount fails for any other reason, the error is now included in the response:
```go
resp.Error += fmt.Sprintf(". Also failed to restore mount: %v", mountErr)
```

PR #7703, opened 2025-12-10
fix: add pagination to list-object-versions for buckets with >1000 ob…

jfburdet:
## What problem are we solving?

Using rclone to send kopia repository to seaweedfs leed to inconsistent kopia repository when using an object lock seaweedfs bucket.

After investigation it turned out that when a bucket contains more than 1000 objects, list-object-versions only
  returns versions for objects in the first 1000 entries (alphabetically sorted).
  This causes S3 clients like rclone or kopia to incorrectly detect missing
  versions and re-upload files that already exist, or fail to list/restore
  versioned objects.

## How are we solving the problem?

The findVersionsRecursively() function was using a fixed limit of 1000
entries without pagination. Objects beyond the first 1000 entries
(alphabetically) were never returned by list-object-versions.

This fix adds pagination following the same pattern used in
filer_delete_entry.go and filer_util.go (processDirectoryTTL).

## How is the PR tested?

No additionnal test is provided. The project should be populated with integration test covering hedge cases : bucket with a lot of files, lot of versions, etc.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 object versioning to handle large directories via paginated listings and recursive traversal
  * Better deduplication across versions and null-versions, preventing duplicate entries
  * Correct handling of delete markers and explicit directory entries in version listings
  * Enhanced support for pre-versioning objects with synthetic null-version creation and consistent metadata (ETag/size)
  * Improved logging and end-of-list detection for more reliable processing

### DIFF
--- a/.github/workflows/metadata-subscribe-tests.yml
+++ b/.github/workflows/metadata-subscribe-tests.yml
@@ -1,0 +1,92 @@
+name: "Metadata Subscribe Integration Tests"
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - 'weed/filer/**'
+      - 'weed/pb/filer_pb/**'
+      - 'weed/util/log_buffer/**'
+      - 'weed/server/filer_grpc_server_sub_meta.go'
+      - 'weed/command/filer_backup.go'
+      - 'test/metadata_subscribe/**'
+      - '.github/workflows/metadata-subscribe-tests.yml'
+  pull_request:
+    branches: [ master ]
+    paths:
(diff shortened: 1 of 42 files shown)


jfburdet:
LLM disclaimer : Tracking the bug and change proposal made with huge help from claude code.

jfburdet:
Hold on. It seems that this fix does not fix my initial problem of rcloning my kopia backup. 

I'm doing more tests.

jfburdet:
Made another better and extended PR. Closing this one.

Issue #7748, opened 2025-12-14
ec.decode fails on volumes when all files in collection are deleted


jrcichra:
I am testing erasure coding and believe I found a decoding bug that results in this error message on `ec.decode`:

```
error: delete ec shards for volume 1: mountVolumeAndDeleteEcShards mount volume 1 on seaweedfs-server:8080: rpc error: code = Unknown desc = volume 1 not found on disk
```

Using this `docker-compose.yml`:

```yml
x-seaweedfs-version: &seaweedfs_version chrislusf/seaweedfs:4.02
services:
  seaweedfs:
    image: *seaweedfs_version
    container_name: seaweedfs-server
    command: "server -master.volumeSizeLimitMB=100 -dir=/data -s3 -ip=seaweedfs-server -ip.bind=0.0.0.0"
    ports:
      - "9333:9333"  # Master port
      - "8080:8080"  # Volume port
      - "8888:8888"  # Filer port
      - "8333:8333"  # Filer port
    networks:
      - seaweedfs-network

  seaweedfs-admin:
    image: *seaweedfs_version
    container_name: seaweedfs-admin
    command: "admin -masters=seaweedfs-server:9333"
    ports:
      - "23646:23646"  # Admin port
    networks:
      - seaweedfs-network
    depends_on:
      - seaweedfs

networks:
  seaweedfs-network:
    driver: bridge

```

To recreate the bug - create an S3 bucket, fill it with some data (in my case the same test file with a few copies with different names), encode it, delete all the files, then try to decode it.

The goal for this test is to have enough data to fill up a few volumes. I've set ` -master.volumeSizeLimitMB=100` to hit this quicker so encode will run.

```
$ rc copyto test.mkv test:foo/test3.mkv --verbose --progress
2025/12/14 16:34:50 INFO  : S3 bucket foo: Bucket "foo" created with ACL ""
2025/12/14 16:34:53 INFO  : test.mkv: Multi-thread Copied (new) to: test3.mkv
Transferred:      383.956 MiB / 383.956 MiB, 100%, 129.389 MiB/s, ETA 0s
Transferred:            1 / 1, 100%
Elapsed time:         2.6s
2025/12/14 16:34:53 INFO  :
Transferred:      383.956 MiB / 383.956 MiB, 100%, 129.389 MiB/s, ETA 0s
Transferred:            1 / 1, 100%
Elapsed time:         2.6s
```

[...]

PR #7749, opened 2025-12-14
fix(ec.decode): purge EC shards when volume is empty

chrislusf:
Fixes https://github.com/seaweedfs/seaweedfs/issues/7748

When all entries in an EC volume are deleted, decoding should be a no-op: do not generate/mount an empty normal volume; instead, purge EC shards and return success.

Changes:
- Volume server detects empty EC index (no live entries) and signals the client
- ec.decode handles this by unmounting+deleting shards without mounting a reconstructed volume
- Adds a unit test for empty detection

## Summary by CodeRabbit

* **Bug Fixes**
  * EC volume decode now treats volumes with no live index entries as a successful no-op and automatically unmounts/removes shards instead of producing an empty volume.

* **New Features**
  * Added detection to check for live entries in EC indexes and a coordination marker for the empty-entry case.
  * Added centralized cross-location shard cleanup helpers.

* **Refactor**
  * Consolidated mount-and-cleanup flow to use the new cleanup helpers.

* **Tests**
  * Added tests validating live-entry detection and empty-volume behavior.

### DIFF
--- a/weed/shell/command_ec_decode.go
+++ b/weed/shell/command_ec_decode.go
@@ -159,6 +159,7 @@ func unmountAndDeleteEcShardsWithPrefix(prefix string, grpcDialOption grpc.DialO
 
 	// unmount and delete ec shards in parallel (one goroutine per location)
 	for location, ecIndexBits := range nodeToEcIndexBits {
+		location, ecIndexBits := location, ecIndexBits // capture loop variables for goroutine
 		ewg.Add(func() error {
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed remaining comments:\n- Replaced the 'has no live entries' magic substring with a local constant in ec.decode\n- Added TestHasLiveNeedles_WithLiveEntryIsTrue to cover the positive path\n

PR #7790, opened 2025-12-16
s3: fix remote object not caching

G-OD:
# Problem
Currently when calling the S3 API when a remote is mounted it will not cache the remote object at all. Instead it will just return an error 500 on the S3 API.

# Solution
By checking if the object is remote only and then caching the remote object.

## Summary by CodeRabbit

* **New Features**
  * Remote objects are cached to the local cluster on first access with deduplicated concurrent caching to improve read performance.

* **Bug Fixes**
  * Better caching error handling with graceful fallback to streaming from remote when caching fails; negative bucket cache entries are cleared after bucket updates to avoid false "not found".

* **Tests**
  * Added end-to-end integration tests covering basic, concurrent, large-object, range, and not-found remote-cache scenarios.

* **CI / Docs**
  * New CI job, test Makefile, README, and .gitignore entries to run and collect remote-cache test logs.

### DIFF
--- a/weed/s3api/s3api_object_versioning.go
+++ b/weed/s3api/s3api_object_versioning.go
@@ -1064,26 +1064,13 @@ func (s3a *S3ApiServer) getObjectOwnerFromVersion(version *ObjectVersion, bucket
 		return CanonicalUser{ID: version.OwnerID, DisplayName: ownerDisplayName}
 	}
 
-	// Fallback: try to get owner from the current version of the object
-	// This handles cases where older versions might not have owner metadata
-	if version.VersionId == "null" {
-		// For null version, check the regular object file
-		bucketDir := s3a.option.BucketsPath + "/" + bucket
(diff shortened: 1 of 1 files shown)


G-OD:
@chrislusf you absolute legend. You got right onto it. I got no idea how most of this stuff works so apologies for my subpar PR :stuck_out_tongue:

chrislusf:
> @chrislusf you absolute legend. You got right onto it. I got no idea how most of this stuff works so apologies for my subpar PR 😛

Thanks for noticing the bug! After 4.01 when s3 component skips filer in the data path, there are several regressions like this. Issue reports and PRs are invaluable.

G-OD:
> > @chrislusf you absolute legend. You got right onto it. I got no idea how most of this stuff works so apologies for my subpar PR 😛
> 
> Thanks for noticing the bug! After 4.01 when s3 component skips filer in the data path, there are several regressions like this. Issue reports and PRs are invaluable.

Awesome. Nice work. I'll be on the lookout for anything else

PR #7822, opened 2025-12-19
master: enable dual-write to both seaweedfs/raft and hashicorp/raft for seamless migration

chrislusf:
## Summary

Enable dual-write to both raft implementations when using seaweedfs/raft, allowing seamless future migration to hashicorp/raft without manual intervention.

## Problem

Existing SeaweedFS deployments using seaweedfs/raft need a way to migrate to hashicorp/raft without service interruption or manual state migration.

## Solution

When using seaweedfs/raft (`-raftHashicorp=false`, the default), the master now:

1. Initializes hashicorp raft in parallel (in a `hashicorp/` subdirectory)
2. Writes all state changes to both raft implementations
3. Keeps hashicorp raft in sync automatically

## Migration Path

1. **Current version**: Upgrade to this version, masters continue running with seaweedfs/raft
2. **Automatic sync**: hashicorp raft receives all state changes in the background
3. **When ready**: Switch to `-raftHashicorp=true` - hashicorp raft already has all the state
4. **Future version**: Remove seaweedfs/raft entirely

## Changes

- `raft_server.go`: Add `initHashicorpRaftForDualWrite()` to initialize hashicorp raft alongside old raft
- `topology.go`: `NextVolumeId()` now writes to both raft implementations when both exist
- `master_server.go`: `SetRaftServer()` sets both `RaftServer` and `HashicorpRaft` in topology
- `master.go`: Register hashicorp transport manager for gRPC when dual-writing

## Testing

The dual-write is designed to be transparent. If hashicorp raft initialization fails, a warning is logged and the master continues with seaweedfs/raft only (no breaking change).

### DIFF
--- a/weed/server/master_server.go
+++ b/weed/server/master_server.go
@@ -209,6 +209,8 @@ func (ms *MasterServer) SetRaftServer(raftServer *RaftServer) {
 			if ms.Topo.RaftServer.Leader() != "" {
 				glog.V(0).Infof("[%s] %s becomes leader.", ms.Topo.RaftServer.Name(), ms.Topo.RaftServer.Leader())
 				ms.Topo.LastLeaderChangeTime = time.Now()
+				// Sync hashicorp raft leadership to match seaweedfs/raft leader for dual-write
+				raftServer.SyncHashicorpRaftLeadership(ms.Topo.RaftServer.Leader())
 			}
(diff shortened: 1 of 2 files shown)


chrislusf:
auto migration is complicated.

PR #7833, opened 2025-12-21
feat: auto-configure optimal volume size limit based on available disk space

chrislusf:
## Summary
Auto-configure the optimal volume size limit for 'weed mini' command based on available disk space.

## Changes
- Add calculateOptimalVolumeSizeMB() function with OS-independent disk detection
- Reuses existing stats.NewDiskStatus() which works across Linux, macOS, Windows, BSD, Solaris
- Algorithm: available disk / 100, rounded up to nearest power of 2 (64MB, 128MB, 256MB, 512MB, etc.)
- Volume size capped to maximum of 1GB (1024MB) for better stability
- Minimum volume size is 64MB
- Master now starts with dynamically configured volume size limit
- Welcome message displays the calculated optimal volume size

## Example Behavior
- 10GB disk → 64MB (minimum)
- 100GB disk → 128MB
- 1TB disk → 256MB
- 10TB disk → 512MB
- 100TB+ disk → 1024MB (capped at 1GB)

## Benefits
- Improves usability for mini mode by adapting to different disk sizes
- No hardcoded values - automatically optimizes based on available resources
- Cross-platform compatible using existing OS-independent utilities
- Predictable behavior with min/max constraints

## Summary by CodeRabbit

* **New Features**
  * Automatic volume size configuration at startup based on available disk space (64MB–1GB) and shown in the welcome/“Optimized Settings” message
  * Configuration file load/save and CLI flag detection to persist and recognize user options

* **Improvements**
  * Startup messaging and logs indicate whether the volume size was auto-calculated or explicitly provided
  * Preserves and logs any user-specified volume size instead of overriding it
  * Welcome output updated to display the computed volume size value

### DIFF
--- a/weed/command/mini.go
+++ b/weed/command/mini.go
@@ -263,7 +263,7 @@ func init() {
 }
 
 // calculateOptimalVolumeSizeMB calculates optimal volume size based on total disk capacity.
-// 
+//
 // Algorithm:
 // 1. Read total disk capacity using the OS-independent stats.NewDiskStatus()
 // 2. Divide total disk capacity by 100 to estimate optimal volume size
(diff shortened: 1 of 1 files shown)


andrewazores:
@chrislusf is there any way to use this new auto sizing function with the older `weed server` subcommand? Or alternatively, can `weed mini` be used with admin UI and webdav disabled somehow?

chrislusf:
> @chrislusf is there any way to use this new auto sizing function with the older `weed server` subcommand? Or alternatively, can `weed mini` be used with admin UI and webdav disabled somehow?

Explain a bit more about your use case?

andrewazores:
https://github.com/cryostatio/cryostat-storage/blob/c03b731228d7a80648cb865514b4e6bb4daead08/cryostat-entrypoint.bash#L62

In the "batteries included" case where Cryostat is deployed with a single-container SeaweedFS container, I have some crappy bash script logic that has a similar goal of figuring out a default volume size and volume count that fills the available disk. This is expected to be a k8s PersistentVolume of some kind which is claimed solely for the storage container, so it makes sense to try to automatically divide up all available space for seaweed volumes. Obviously the master, filer, and volume servers are required, and we're also using S3, but have no need for webdav and do not want an admin UI to be enabled/exposed (even if we just don't expose the port) since all configuration should be static and handled separately.

chrislusf:
@andrewazores I think you want to disable UI for admin server and web dav? There would be some maintenance operations done by admin + workers in the future.

andrewazores:
Yes, exactly. Something like `weed mini -webdav=false -admin.ui=false` would, I think, be a useful configuration for `mini` mode. Just a single-node S3-compatible storage server ready to drop in to a Docker or k8s environment, which will try to use all available disk space.

I guess other users might also want the option to disable S3 and use only Webdav instead, too.

PR #7931, opened 2026-01-01
helm: Support multiple hosts for S3 ingress

sheyabernstein:

Users need to expose S3 ingress through multiple hostnames but the current template only supports a single host value.

Updated the S3 ingress template to accept both string and array values for the `host` field. The template uses `kindIs "slice"` to detect the type and generates appropriate ingress rules for each host. Maintains full backwards compatibility with existing single-host configurations.

# How is the PR tested?

Tested with helm template using:
- Single host string (backwards compatibility): `host: "seaweedfs.cluster.local"`
- Multiple hosts array: `host: ["seaweedfs.cluster.local", "seaweedfs.domain.com"]`
- Wildcard/no host: `host: false`

# Checks

## Summary by CodeRabbit

* **New Features**
  * S3 ingress configuration now supports multiple hosts while maintaining backwards compatibility with single-host setups.

* **Documentation**
  * Clarified ingress host parameter to indicate support for both single hostname and hostname array formats.

### DIFF
--- a/k8s/charts/seaweedfs/templates/s3/s3-ingress.yaml
+++ b/k8s/charts/seaweedfs/templates/s3/s3-ingress.yaml
@@ -11,7 +11,13 @@
 {{- else if .Values.s3.ingress.host }}
   {{- $hosts = list .Values.s3.ingress.host }}
 {{- end }}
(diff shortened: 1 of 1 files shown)


sheyabernstein:
Hey @chrislusf, for your review 🙏 

Do you have a timeline for the next release?

PR #7971, opened 2026-01-05
feat: add flags to disable WebDAV and Admin UI in weed mini

chrislusf:
Addresses https://github.com/seaweedfs/seaweedfs/pull/7833#issuecomment-3711924150

## Summary

Adds flags to allow disabling WebDAV and Admin UI in `weed mini` mode. The Admin server still runs for gRPC access and API operations when the UI is disabled, allowing minimal overhead deployments.

## Changes

- Add `-webdav` flag (default: true) to optionally disable the WebDAV server
- Add `-admin.ui` flag (default: true) to optionally disable the Admin UI only (server still runs for gRPC/API)
- Conditionally skip WebDAV service startup based on flag
- Pass disableUI flag to SetupRoutes to skip UI route registration while keeping admin server running

## Use Cases

This enables deployments like:
- Single-node S3-compatible storage in containers/Kubernetes without WebDAV or admin UI overhead
- API-only admin access (gRPC endpoints still available when UI is disabled)
- Lightweight deployments with minimal resource footprint

## Example Usage

```bash
# S3-only mode (no WebDAV, no Admin UI, but API still works)
weed mini -dir=/data -webdav=false -admin.ui=false

# WebDAV-only mode (no Admin UI, but API still works)
weed mini -dir=/data -admin.ui=false

# Default (all services enabled)
weed mini -dir=/data
```

### DIFF
--- a/weed/command/mini.go
+++ b/weed/command/mini.go
@@ -1165,7 +1165,11 @@ func printWelcomeMessage() {
 	} else if *miniEnableAdminUI {
 		fmt.Fprintf(&sb, credentialsInstructionTemplate, *miniIp, *miniAdminOptions.port)
 	} else {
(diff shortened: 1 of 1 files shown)


andrewazores:
Awesome, thanks @chrislusf !

PR #7992, opened 2026-01-09
fix: Invalid volume mount conditional in filer template

nichobi:
There is a mistmatch in the conditionals for the definition and mounting of the `config-users` volume in the filer's template. 

Volume definition:
```
        {{- if and .Values.filer.s3.enabled .Values.filer.s3.enableAuth }}
```
Mount:
```
            {{- if .Values.filer.s3.enableAuth }}
```

This leads to an invalid specification in the case where s3 is disabled but the enableAuth value is set to true, as it tries to mount in an undefined volume.

I've switched updated the volume mount to use the same conditional as the volume definition, so that both s3.enabled and s3.enableAuth.

# How is the PR tested?

# Checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Kubernetes deployment configuration to correctly handle S3 authentication setup. The authentication configuration is now properly mounted only when both S3 and S3 authentication are enabled, preventing potential misconfiguration scenarios.

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer/filer-statefulset.yaml
+++ b/k8s/charts/seaweedfs/templates/filer/filer-statefulset.yaml
@@ -229,7 +229,7 @@ spec:
             - name: seaweedfs-filer-log-volume
               mountPath: "/logs/"
             {{- end }}
-            {{- if .Values.filer.s3.enableAuth }}
+            {{- if and .Values.filer.s3.enabled .Values.filer.s3.enableAuth }}
             - name: config-users
               mountPath: /etc/sw
               readOnly: true


nichobi:
> While this change is correct, a related issue remains in `k8s/charts/seaweedfs/templates/filer/filer-statefulset.yaml`. The command-line argument `-s3.config` at line 217 is still conditional only on `.Values.filer.s3.enableAuth`. This should also be updated to `{{- if and .Values.filer.s3.enabled .Values.filer.s3.enableAuth }}` to prevent the filer from trying to access a configuration file that doesn't exist when the corresponding volume is not mounted. I recommend applying this additional change to fully resolve the inconsistency.

That if statement is nested within another conditional (starting at line 203), which already checks `s3.enabled`. There's no need to check it again.

Issue #8004, opened 2026-01-12
volume.fsck -forcePurging may mis-identify in-use chunks as orphan due to partial failures in Filer FID collection; TraverseBfs error propagation race


kisow:
### volume.fsck -forcePurging -reallyDeleteFromVolume
`volume.fsck -forcePurging -reallyDeleteFromVolume` removes volume data that has no corresponding metadata reference in the filer, i.e., orphan data.

The concern is that if the orphan-detection logic has failed, it can lead to two types of errors, one of which is critical:

- **Type I Error (False Positive)**: data is still referenced (in-use) but is incorrectly classified as orphan → **data loss** → application **I/O errors**
- **Type II Error (False Negative)**: data is truly orphan but is incorrectly treated as in-use → not cleaned up → wasted disk space

From reading the code, orphan detection appears to be based on a set subtraction:

- Let **|A|** be the **set of file IDs derived from filer metadata**
- Let **|B|** be the **set of chunk/file ID entries present on the volume servers**
- Orphans are identified as **|B| − |A|** (for -reallyDeleteFromVolume)

Because this command is destructive with `-forcePurging -reallyDeleteFromVolume`, the correctness of |A| is especially important:

- If building **|A|** (filer-derived set) is **incomplete due to partial failure**, then valid in-use entries can appear in **|B| − |A|** → **False Positive** → deletion of good data.
- If building **|B|** (volume-derived set) is **incomplete due to partial failure**, then some orphan entries may be missing from **|B|** → **False Negative** → orphan data not cleaned.

In my understanding, when running `volume.fsck -forcePurging`, if we encounter a partial failure while building **|A|** (the filer-derived FID set), it would be safer to abort immediately to avoid accidentally purging valid in-use entries.

However, from reading the current code, it looks like some partial errors may not be propagated, and the `-forcePurging` workflow could continue using a partially-built |A|.

Could you please confirm whether my reading is incorrect? If not, is this behavior intentional by design?

We are seeing orphans continuously in production, so we need to run `volume.fsck -forcePurging` periodically. This makes us concerned about the risk of incorrect orphan detection and unintended data loss.

chrislusf:
* the `cutoffTimeAgo` is default to 5 hours now. previously it was default to 5 minutes, which may cause problem if there is a weed mount process writing a large file longer than 5 minutes.
* the `volume.fsck` does not need to be run frequently.

jdblack:
[...]

PR #8015, opened 2026-01-13
Fix volume.fsck -forcePurging -reallyDeleteFromVolume to fail fast on filer traversal errors

kisow:

* #8004 

Orphan set is determined by the difference between the **full filer FID set** `|A|` and **the full volume FID set** `|B|`, 
i.e., `|B| - |A|`. If omissions occur while collecting `|A|`, **the omitted FIDs may be incorrectly classified as orphans, which can lead to severe data loss.**

**Therefore, if an error occurs while collecting `|A|`, we should not ignore it; instead, we should propagate the error and stop the operation.**

- Add TraverseBfsWithContext function to support context cancellation
- **Fixed a race where `TraverseBfs` could overwrite errors by synchronizing error handling.**
- Improve error handling with fail-fast behavior and proper error propagation (genFn and saveFn)
- Enhance error messages in readFilerFileIdFile with detailed context

# How is the PR tested?

# Checks

## Summary by CodeRabbit

* **Bug Fixes**
  * Fail-fast traversal and I/O: operations stop promptly on first error and return clearer failures.
  * Improved read/write error reporting and propagation for more reliable recovery.
  * System/internal directories are skipped during traversal to avoid spurious processing.

* **Chores**
  * Added context-aware cancellation across traversal and saving flows to improve responsiveness.

* **Refactor**
  * Reworked traversal/saver concurrency for robust worker coordination, deterministic shutdown, and consistent error handling.

### DIFF
--- a/weed/shell/command_fs_meta_save.go
+++ b/weed/shell/command_fs_meta_save.go
@@ -171,10 +171,7 @@ func doTraverseBfsAndSaving(filerClient filer_pb.FilerClient, writer io.Writer, 
 			Dir:   parentDir,
 			Entry: e,
 		}
-		if hasErr.Load() {
-			// fail-fast: do not continue emitting partial results after first error
-			return firstErr
-		}
+
 		if genErr := genFn(protoMessage, outputChan); genErr != nil {
 			once.Do(func() {
 				firstErr = genErr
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks for the PR! (But later do not generate PR by AI.)

Issue #7977, opened 2026-01-06
Maintenance mode for volume servers


ghost:
At times, it would be great to have a way to turn an entire volume server in read only mode - meaning, reads are allowed, but no new volumes / EC shards should be created on it.

The idea is to limit the blast radius of bad disks, and allow evacuation without having to deal with stuff being written at the time.

ghost:
Ideally, maintenance mode should be a sticky property of volume servers to be controlled via `weed shell` or similar.

ghost:
Maintenance mode is now fully implemented ❤️ 

Added a small documentation snippet for this new feature at https://github.com/seaweedfs/seaweedfs/wiki/Volume-Management as well.

ghost:
All done.

PR #8016, opened 2026-01-13
Return volume server state flags via `VolumeServerStatus()` RPCs.

ghost:

Add support for maintenance (read-only) mode on volume servers.

See https://github.com/seaweedfs/seaweedfs/issues/7977 for details.

Expose volume server state flags on `VolumeServerStatus()` RPCs responses.

# How is the PR tested?

Covered by other volume server state tests - this PR just exposes the state proto message on `VolumeServerStatusResponse`.

# Checks

## Summary by CodeRabbit

* **New Features**
  * Volume server status responses now include server state information, improving visibility into current conditions.
  * This additional state is surfaced in status and admin views, enhancing monitoring, diagnostics, and observability for operators.


chrislusf:
https://github.com/seaweedfs/seaweedfs/pull/8016#pullrequestreview-3655150591 this should be addressed

ghost:
> The `State` field is assigned directly from `vs.store.State.Pb`, a pointer initialized to nil in `NewState()`. The public `LoadState()` method exists to populate this field, but it is never invoked during volume server initialization. Only `SaveState()` guarantees the field is non-nil by creating an empty instance if needed. Either call `LoadState()` during initialization or confirm that nil state is acceptable behavior. Note that while protobuf handles nil message fields as empty messages, the codebase treats nil state as a warning condition in `SaveState()`, suggesting it is unintended.

This is intentional. I will upload a separate PR so  `LoadState()` is called on volume initialization, which will effectively enable this new feature in a single change.

PR #8026, opened 2026-01-14
[shell]: volume balance capacity by min volume density

kmlebedev:

When balancing, sorting by capacity by maxVolumes is used, which does not always correspond to disk occupancy

Added the option to sort by volume density on the volume server.

# How is the PR tested?

localy

# Checks

## Summary by CodeRabbit

* **New Features**
  * Added -capacityBy and -volumeBy options and a -v (verbose) flag.
  * Support for density-aware capacity strategies and per-volume size limits.

* **Bug Fixes / Improvements**
  * More accurate capacity and free-volume checks across balancing, replication, evacuation, and tier moves.
  * Selection respects ACTIVE/FULL/ALL modes and size thresholds for safer relocations.

* **Tests**
  * Updated tests to exercise multiple capacity strategies.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -293,7 +293,7 @@ require (
 	cloud.google.com/go/compute/metadata v0.9.0 // indirect
 	cloud.google.com/go/iam v1.5.3 // indirect
 	cloud.google.com/go/monitoring v1.24.2 // indirect
-	filippo.io/edwards25519 v1.1.0 // indirect
+	filippo.io/edwards25519 v1.1.1 // indirect
 	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.21.0
 	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.13.1
 	github.com/Azure/azure-sdk-for-go/sdk/internal v1.11.2 // indirect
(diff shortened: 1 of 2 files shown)


chrislusf:
Here are my thoughts. Welcome different ideas!

This may have too many options for users. There could be many more options, such as different CPU types, disk types, network throughput, read/write traffic. Keep supporting more options would not make it easy to manage.

I think ideally the balancing should factor in as few as possible. For example, these 2 factors:
* the ratio of the active writing volumes over total max volumes.
* the ratio of the storage free capacity over total max volumes.

These 2 factors correspond to computing and storage, and the priority is that computing is over storage. 

What not to cover:
* Network usage is hard to quantify and shared among all servers. We may tackle it later.
* Read traffic may be mitigated by external caching layer, e.g., CDN.

kmlebedev:
@chrislusf Thank you for your detailed comment. If I understand correctly, your point is to reduce the number of parameters and, by default, select the most preferable option for most cases, without paying attention to backward compatibility and not giving users time to adapt their installations to the new type of balancing. 

I apologize in advance for any toxicity, but I need to spell this out as I understand it so that we can think clearly and be on the same page.

chrislusf:
We can try to add some new CLI options, or even new commands, and keep existing options as is for backward compatibility.

kmlebedev:
@chrislusf I switched the new sorting function to default.
I left the volumeby option because there are fewer active volumes and they can be quickly sorted and a full balancing is not necessary.

chrislusf:
Thanks a lot!

Issue #7545, opened 2025-11-25
[volume] fatal error: concurrent map writes


kmlebedev:
fatal error: concurrent map writes

```
version 8000GB 3.93 fd6c4951 linux amd64
```

chrislusf:
any thread dump?

kmlebedev:
@chrislusf 

> any thread dump?

`kill -3` it doesn't work

kmlebedev:
<img width="1393" height="440" alt="Image" src="https://github.com/user-attachments/assets/2da11b42-494c-405f-8bff-7aae3c434499" />

It plays regularly at 5am, possibly due to admin scripts. I'll turn on the debug log and look at the shell logs.

kmlebedev:
@chrislusf trace log from another server
```
 internal/runtime/maps.fatal({0x3361f45?, 0x28?})
         /usr/local/go/src/runtime/panic.go
 github.com/seaweedfs/seaweedfs/weed/storage/needle_map.(*CompactMap).segmentForKey(...)
         /builds/infrastructure/utils/storage/seaweedfs/weed/storage/needle_map/compact_map.go
 github.com/seaweedfs/seaweedfs/weed/storage/needle_map.(*CompactMap).Get(0xc00174a020, 0x67e7213)
         /builds/infrastructure/utils/storage/seaweedfs/weed/storage/needle_map/compact_map.go
 github.com/seaweedfs/seaweedfs/weed/storage.(*NeedleMap).Get(0x419854?, 0xc0027a9470?)
         /builds/infrastructure/utils/storage/seaweedfs/weed/storage/needle_map_memory.go
 github.com/seaweedfs/seaweedfs/weed/storage.(*Volume).readNeedle(0xc000a626e0, 0xc0027a95a0, 0x0, 0x0)
         /builds/infrastructure/utils/storage/seaweedfs/weed/storage/volume_read.go
 github.com/seaweedfs/seaweedfs/weed/storage.(*Store).ReadVolumeNeedle(0xc0027a9658?, 0x24, 0xc0027a95a0, 0x0, 0x0)
         /builds/infrastructure/utils/storage/seaweedfs/weed/storage/store.go
 github.com/seaweedfs/seaweedfs/weed/server.(*VolumeServer).VolumeNeedleStatus(0xc00031e000, {0x2b63ac0?, 0xc001e7a9b8?}, 0xc001e4b680)
[...]

PR #8029, opened 2026-01-14
fix: CompactMap race condition causing runtime panic

chrislusf:
## Problem

This PR fixes a critical race condition in `CompactMap` that causes `runtime/maps.fatal` panics under concurrent load.

Fixes https://github.com/seaweedfs/seaweedfs/issues/7545

### Root Cause

The `Set()` and `Delete()` methods in `compact_map.go` incorrectly used **read locks (`RLock()`)** instead of **write locks (`Lock()`)**. Since `segmentForKey()` can create new map segments (modifying `cm.segments`), performing this write operation under a read lock caused concurrent map write panics when multiple goroutines accessed the map simultaneously.

### Stack Trace Example
```
internal/runtime/maps.fatal({0x3361f45?, 0x28?})
    /usr/local/go/src/runtime/panic.go
github.com/seaweedfs/seaweedfs/weed/storage/needle_map.(*CompactMap).segmentForKey(...)
    /builds/.../compact_map.go
github.com/seaweedfs/seaweedfs/weed/storage/needle_map.(*CompactMap).Get(0xc00174a020, 0x67e7213)
    /builds/.../compact_map.go
```

## Changes

- **`Set()` method**: Changed `RLock()/RUnlock()` to `Lock()/Unlock()`
- **`Delete()` method**: Changed `RLock()/RUnlock()` to `Lock()/Unlock()`
- **`Get()` method**: Correctly continues using `RLock()/RUnlock()` (read-only operation)

## Testing

✅ All tests passed with race detector enabled:
```bash
go test -v -race ./weed/storage/needle_map/...
# PASS (82.977s) - no race conditions detected
```

## Impact

This fix resolves production panics that occur when concurrent gRPC requests (e.g., `VolumeNeedleStatus`) trigger map modifications under read locks.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved thread safety in data storage operations by ensuring exclusive access during updates, preventing potential race conditions.

### DIFF
--- a/weed/storage/needle_map/compact_map.go
+++ b/weed/storage/needle_map/compact_map.go
@@ -236,8 +236,8 @@ func (cm *CompactMap) segmentForKey(key types.NeedleId) *CompactMapSegment {
 // Set inserts/updates a NeedleValue.
 // If the operation is an update, returns the overwritten value's previous offset and size.
 func (cm *CompactMap) Set(key types.NeedleId, offset types.Offset, size types.Size) (oldOffset types.Offset, oldSize types.Size) {
-	cm.RLock()
-	defer cm.RUnlock()
+	cm.Lock()
+	defer cm.Unlock()
 
 	cs := cm.segmentForKey(key)
 	return cs.set(key, offset, size)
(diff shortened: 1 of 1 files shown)


chrislusf:
## Updated Fix

I've updated the PR with an important additional fix and optimization:

### Additional Race Condition in Get()
The initial fix missed that `Get()` also called `segmentForKey()` which could create new segments under a read lock, causing the same race condition.

### Optimizations Applied
Both `Get()` and `Delete()` now check segment existence directly instead of calling `segmentForKey()`:

**Get() - Before:**
```go
cs := cm.segmentForKey(key)  // Could create segment under RLock!
```

**Get() - After:**
```go
chunk := Chunk(key / SegmentChunkSize)
cs, ok := cm.segments[chunk]
if !ok {
    return nil, false  // Early return, no segment creation
}
```

**Benefits:**
1. ✅ **Eliminates race condition** - No method calls `segmentForKey()` under read lock
2. ✅ **Memory optimization** - Avoids creating empty segments for non-existent keys
3. ✅ **Performance improvement** - Early return when segment doesn't exist

All tests still pass with race detector enabled.

PR #8052, opened 2026-01-18
fix(s3api): correct wildcard matching

sosweetham:

Currently only suffix `*` wildcard is matched

S3 Api however suggests
https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_policies_elements_resource.html#reference_policies_elements_resource_wildcards

```
The asterisk (*) character can expand to replace everything within a segment, including characters like a forward slash (/) that may otherwise appear to be a delimiter within a given service namespace. For example, consider the following Amazon S3 ARN as the same wildcard expansion logic applies to all services.


"Resource": "arn:aws:s3:::amzn-s3-demo-bucket/*/test/*"
The wildcards in the ARN apply to all of the following objects in the bucket, not only the first object listed.


amzn-s3-demo-bucket/1/test/object.jpg
amzn-s3-demo-bucket/1/2/test/object.jpg
amzn-s3-demo-bucket/1/2/test/3/object.jpg 
amzn-s3-demo-bucket/1/2/3/test/4/object.jpg
amzn-s3-demo-bucket/1///test///object.jpg
amzn-s3-demo-bucket/1/test/.jpg
amzn-s3-demo-bucket//test/object.jpg
amzn-s3-demo-bucket/1/test/
```

We are solving it by replacing the current "*" suffix only matcher by replacing it with a (in my view) compliant wildcard pattern matcher that works in O(n) time.

Alternative considered was regex but, I learnt that it might be slower, so I settled for this.

# How is the PR tested?

added analogous tests and tested against a compiled copy with the following s3 config 

```
{
  "identities": [
    {
      "name": "anonymous",
      "actions": [
        "Read:syr/uploads/user:*/public/*"
      ]
    }
  ]
}
```
(it works)

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 API permission wildcard matching to handle glob-style patterns (* and ?) reliably across checks.

* **Enhancements**
  * Replaced regex-based wildcard handling with a unified, more robust string-based matcher for consistent behavior.

* **Tests**
  * Added comprehensive unit tests covering wildcard matching edge cases and patterns.

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -18,6 +18,7 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/iam_pb"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/policy_engine"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3_constants"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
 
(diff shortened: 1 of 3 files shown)


chrislusf:
@sosweetham Thanks!

PR #8072, opened 2026-01-21
fix(s3api): missing `Vary: Origin` header on non-CORS and `OPTIONS` requests

### LINKED ISSUE #8071: [S3] Missing `Vary: Origin` header on non-CORS requests causes browser caching issues
**Related**
#7544 

**Describe the bug**
The S3 gateway currently behaves inconsistently regarding the `Vary: Origin` response header. It correctly returns `Vary: Origin` when a request contains an `Origin` header (CORS request), but fails to include it when the request is a standard, non-CORS request (e.g., a direct browser navigation or a same-origin fetch).

**The Problem**
This inconsistency leads to browser caching collisions. If a user (or a CDN) first requests a resource without an `Origin` header, the browser caches this response (which lacks CORS headers). If a subsequent request for the same resource is made from a different origin (requiring CORS), the browser may serve the cached response. Since the cached response lacks the necessary `Access-Control-Allow-Origin` headers (and the cache wasn't told to `Vary` by `Origin`), the CORS check fails in the browser, breaking the application.

**Steps to Reproduce**

1. Upload a file to SeaweedFS S3.
2. Request the file **without** an Origin header (simulating a direct access):
```bash
curl -I http://localhost:8333/bucket/test.jpg

```

**Result:** The response does **not** contain `Vary: Origin`.
3. Request the file **with** an Origin header:
```bash
curl -I -H "Origin: http://example.com" http://localhost:8333/bucket/test.jpg

```

**Result:** The response contains `Vary: Origin`.

**Expected behavior**
The S3 gateway should return `Vary: Origin` on **all** responses for resources where CORS is enabled or possible, regardless of whether the incoming request initiated a CORS handshake. This ensures that downstream caches (browsers, CDNs, proxies) understand that the response content (specifically the headers) depends on the request's Origin.

**Actual behavior**

* **Request with Origin:** Returns `Vary: Origin`.
* **Request without Origin:** Does not return `Vary: Origin`.

**Additional Context**
According to HTTP specifications and best practices for CORS, if a response differs based on the value of the `Origin` header (i.e., whether CORS headers are added or not), the server implies a variation. Omitting the `Vary` header puts clients at risk of receiving opaque responses when they expect CORS-enabled responses.

llc1123:
Fixes #8071.

The S3 gateway behaved inconsistently regarding the `Vary: Origin` response header, leading to potential caching issues:
- **Non-CORS Requests:** It failed to return `Vary: Origin` for requests without an `Origin` header, even when CORS was enabled. This could cause caches to serve non-CORS responses to subsequent CORS requests.
- **OPTIONS Requests:** The `HandleOptionsRequest` handler explicitly used for preflight requests also lacked the logic to add `Vary: Origin`, causing preflight responses to be cached incorrectly across different origins.

- **Consolidated Logic:** Modified both `Handler` (for standard requests) and `HandleOptionsRequest` (for preflight OPTIONS requests) in `weed/s3api/cors/middleware.go` to always add the `Vary: Origin` header if a valid CORS configuration exists for the bucket.
- **Refactoring:** Refactored both `Handler` and `HandleOptionsRequest` to share a consistent, linear control flow. This improves readability and ensures that configuration lookup and header application happen early in the process for all request types.
- **Cleanup:** Removed redundant logic in `weed/s3api/cors/cors.go` that was conditionally adding the header.

# How is the PR tested?
Added comprehensive unit tests in `weed/s3api/cors/middleware_test.go`:

- **`TestMiddlewareVaryHeader`**: Verifies that `Vary: Origin` is correctly added for standard requests with specific `AllowedOrigins`, wildcard origins, and correctly omitted when no config exists.
- **`TestHandleOptionsRequestVaryHeader`**: Verifies that `Vary: Origin` is correctly added for `OPTIONS` preflight requests.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * CORS behavior standardized: Vary: Origin is now applied when a bucket CORS configuration exists; requests without a config proceed without CORS headers and preflight requests are denied. ApplyHeaders no longer conditionally sets Vary based solely on Allow‑Origin.

* **Refactor**
  * Consolidated CORS request handling into a single, unified flow for consistent responses.

* **Tests**
  * Added/updated tests validating Vary: Origin across specific, wildcard, no‑CORS, and OPTIONS scenarios.

### DIFF
--- a/weed/s3api/cors/middleware.go
+++ b/weed/s3api/cors/middleware.go
@@ -66,70 +66,28 @@ func (m *Middleware) getCORSConfig(bucket string) (*CORSConfiguration, bool) {
 // Handler returns the CORS middleware handler
 func (m *Middleware) Handler(next http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		corsReq := ParseRequest(r)
-		bucket, _ := s3_constants.GetBucketAndObject(r)
-
-		// 1. Basic Validation
-		if bucket == "" {
+		m.processCORS(w, r, func() {
 			next.ServeHTTP(w, r)
-			return
-		}
-
-		// 2. Load Configuration
-		config, hasConfig := m.getCORSConfig(bucket)
-
-		// 3. Apply Vary Header (Always applied if config exists)
-		if hasConfig {
-			w.Header().Add("Vary", "Origin")
-		}
-
-		// 4. Handle Non-CORS Requests
-		if corsReq.Origin == "" {
-			next.ServeHTTP(w, r)
-			return
-		}
-
-		// 5. Handle Missing Configuration
-		if !hasConfig {
-			if corsReq.IsPreflightRequest {
-				s3err.WriteErrorResponse(w, r, s3err.ErrAccessDenied)
-			} else {
-				next.ServeHTTP(w, r)
-			}
-			return
-		}
-
-		// 6. Evaluate CORS Request
-		corsResp, err := EvaluateRequest(config, corsReq)
-		if err != nil {
-			glog.V(3).Infof("CORS evaluation failed for bucket %s: %v", bucket, err)
-			if corsReq.IsPreflightRequest {
-				s3err.WriteErrorResponse(w, r, s3err.ErrAccessDenied)
-			} else {
-				next.ServeHTTP(w, r)
-			}
-			return
-		}
-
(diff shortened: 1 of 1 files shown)


llc1123:
@coderabbitai review

llc1123:
@coderabbitai review

llc1123:
@coderabbitai review

PR #8130, opened 2026-01-26
Implement IAM propagation to S3 servers

chrislusf:
This PR implements the propagation of IAM changes (Users, AccessKeys, Policies, ServiceAccounts) from the Filer (Admin/Shell) to known S3 servers.

## Changes
- **CredentialStore Wrapper**: Introduced `PropagatingCredentialStore` in `weed/credential/propagating_store.go` which wraps an underlying `CredentialStore`. It intercepts write operations and propagates them to all S3 servers via gRPC in parallel.
- **CredentialManager**: Updated to use `PropagatingCredentialStore` when a `MasterClient` is provided (i.e., when running on Filer).
- **Protos**: Added `PutPolicy`, `GetPolicy`, `DeletePolicy`, `ListPolicies` RPCs to `s3.proto` to support policy propagation.
- **S3ApiServer**: Implemented the new Policy RPCs.
- **FilerServer**: Wired up to set the `MasterClient` on .

## Verification
- Verified build passes.
- Propagation logic uses parallelism with a timeout to avoid blocking writes.

## Summary by CodeRabbit

* **New Features**
  * Credential and policy changes now propagate automatically to all S3 servers via a dedicated propagation channel.
  * S3 gateway adds explicit bind IP and gRPC port options and a configurable IAM read-only mode (s3.iam.readOnly).

* **Refactor**
  * IAM now uses an in-memory identity/policy cache; static identities are preserved across dynamic reloads.

* **Tests**
  * Tests updated for propagation, read-only enforcement, and static identity protection.

### DIFF
--- a/weed/credential/propagating_store.go
+++ b/weed/credential/propagating_store.go
@@ -103,8 +103,15 @@ func (s *PropagatingCredentialStore) UpdateUser(ctx context.Context, username st
 		return err
 	}
 	s.propagateChange(ctx, func(tx context.Context, client s3_pb.SeaweedS3IamCacheClient) error {
-		_, err := client.PutIdentity(tx, &iam_pb.PutIdentityRequest{Identity: identity})
-		return err
+		if _, err := client.PutIdentity(tx, &iam_pb.PutIdentityRequest{Identity: identity}); err != nil {
+			return err
+		}
+		if username != identity.Name {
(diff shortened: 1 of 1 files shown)


wzrdtales:
Just noticed readOnly is enabled by default and a warning is shown that policies are not propogated to other s3 instances? Is this temporary?

PR #8165, opened 2026-01-29
shell: allow spaces in arguments via quoting (#8157)

### LINKED ISSUE #8157: Not possible to configure options with empty spaces in the  string in weed shell
When using weed shell, it is not possible to configure options that contain empty spaces. For example, this command will fail silently:

> s3.configure -user=Test_number_004 -account_display_name="Test number 004" -actions=write -apply

Similarly, this one will also fail:

> s3.configure -user=Test_number_004 -account_display_name='Test number 004' -actions=write -apply

To work around this, I have to use an underscore instead of a space:

> s3.configure -user=Test_number_004 -account_display_name=Test_number_004 -actions=write -apply

This results in:

json
```
{
  "name": "Test_number_004",
  "credentials": [
    {
      "accessKey": "",
      "secretKey": "",
      "status": ""
    }
  ],
  "actions": [
    "write"
  ],
  "account": {
    "id": "",
    "displayName": "Test_number_004",
    "emailAddress": ""
  },
  "disabled": false,
  "serviceAccountIds": [],
  "policyNames": []
}
```
However, when using the Admin UI, empty spaces are handled without any issues—even in the user name:

json
```
{
  "name": "Test number 005",
  "credentials": [],
  "actions": [],
  "account": null,
  "disabled": false,
  "serviceAccountIds": [],
  "policyNames": []
}
```
It would be helpful if weed shell could also accept empty spaces—perhaps within quotes—at least for the displayName field.

Thanks in advance!

chrislusf:
This PR fixes #8157 where weed shell was unable to parse arguments containing spaces even when quoted.

Summary of changes:
- Updated the argument splitting regex in weed/shell/shell_liner.go to correctly handle quoted segments within a flag.
- Added a  function to remove matching quotes from flags/values.
- Added a comprehensive unit test in weed/shell/shell_liner_test.go to verify the fix and prevent regressions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved shell command parsing: more reliable tokenization, better handling of quoted and escaped arguments, and correct processing of multiple semicolon-delimited commands so flags and values are interpreted as intended.

* **Tests**
  * Added unit tests validating tokenization, quote stripping, escaped-quote handling, and flag parsing to ensure parsing behavior.

### DIFF
--- a/weed/shell/shell_liner.go
+++ b/weed/shell/shell_liner.go
@@ -187,7 +187,7 @@ func parseShellInput(line string, split bool) (args []string, unbalanced bool) {
 		args = append(args, current.String())
 	}
 
-	return args, inDoubleQuotes || inSingleQuotes
+	return args, inDoubleQuotes || inSingleQuotes || escaped
 }
 
 func printGenericHelp() {
--- a/weed/shell/shell_liner_test.go
+++ b/weed/shell/shell_liner_test.go
@@ -59,6 +59,7 @@ func TestStripQuotes(t *testing.T) {
 		{input: `"unbalanced`, expected: `"unbalanced`},
 		{input: `'unbalanced`, expected: `'unbalanced`},
 		{input: `-name="a\"b`, expected: `-name="a\"b`},
+		{input: `trailing\`, expected: `trailing\`},
 	}
 
 	for _, tt := range tests {


chrislusf:
Updated the implementation with a more robust state machine parser to handle escaped quotes (e.g., `-name="a\"b"`) and single-quote literal behavior, as suggested. Comprehensive unit tests added.

PR #8434, opened 2026-02-24
s3api: preserve Host header port in signature verification

chrislusf:
Avoid stripping default ports (80/443) from the Host header in extractHostHeader. 

This fixes SignatureDoesNotMatch errors when SeaweedFS is accessed via a proxy (like Kong Ingress) that explicitly includes the port in the Host header or X-Forwarded-Host, which S3 clients sign.

The fix aligns SeaweedFS with MinIO behavior and AWS SDK expectations.

Key changes:
- Refactored `extractHostHeader` to keep explicit ports.
- Removed unused `isDefaultPort` and variables.
- Updated unit tests to cover new port preservation logic.

## Bug Fixes

* Improved S3 API authentication to correctly handle requests with explicit port numbers
* Enhanced IPv6 address support in S3 API authentication calculations


chrislusf:
FYI: going to revert this since this PR did not solve the problem and introduced breaking change.

PR #8444, opened 2026-02-25
Fix S3 signature verification behind reverse proxies

blitt001:
## Summary

When SeaweedFS is deployed behind a reverse proxy (nginx, Kong, Traefik, etc.), AWS S3 Signature V4 verification fails because the Host header the client signed with (e.g. `localhost:9000`) differs from the Host header SeaweedFS receives on the backend (e.g. `seaweedfs:8333`).

This PR adds a new **`-s3.externalUrl`** parameter (and **`S3_EXTERNAL_URL`** environment variable) that tells SeaweedFS what public-facing URL clients use to connect. When set, SeaweedFS uses this host value for signature verification instead of the Host header from the incoming request.

### New parameter

```
-s3.externalUrl=http://localhost:9000    # CLI flag
S3_EXTERNAL_URL=https://s3.example.com   # environment variable (useful in Docker/K8s)
```

The flag takes precedence over the environment variable. Default ports are stripped at startup to match AWS SDK behavior (`http://:80` → host only, `https://:443` → host only).

### Bugs fixed

- Default port stripping was removed by a prior PR, causing signature mismatches when clients connect on standard ports (80/443)
- `X-Forwarded-Port` was ignored when `X-Forwarded-Host` was not present
- Scheme detection now uses proper precedence: `X-Forwarded-Proto` > TLS connection > URL scheme > `"http"`
- `expectedHost` field in `TestSignatureV4WithForwardedPort` was declared but never actually checked (self-referential test)
- Several test expectations for standard port stripping were incorrect

### Architecture

```
AWS CLI (signs with Host: localhost:9000)
    │
    ▼
nginx / Kong / Traefik (:9000)
    │  proxy_pass → seaweedfs:8333
    │  Sets: X-Forwarded-Host, X-Forwarded-Port, X-Forwarded-Proto
    ▼
SeaweedFS S3 (:8333, -s3.externalUrl=http://localhost:9000)
    │  externalHost = "localhost:9000" (parsed at startup)
    │  extractHostHeader() returns "localhost:9000"
    │  Matches what AWS CLI signed with → verification succeeds
```

## Test plan

- [x] All existing unit tests pass (`go test ./weed/s3api/...`)
- [x] New `TestParseExternalUrlToHost` — 12 cases covering port stripping, IPv6, invalid URLs
- [x] New `TestExternalUrlSignatureVerification` — 6 cases using real AWS SDK v4 signer with externalUrl
- [x] New `TestRealSDKSignerWithForwardedHeaders` — 8 cases validating extracted host matches what AWS SDK signed with (non-self-referential)
- [x] New `auth_proxy_integration_test.go` — full HTTP stack test with httptest.Server + httputil.ReverseProxy + real AWS SDK signer
- [x] New Docker Compose integration test (`test/s3/proxy_signature/`) — nginx reverse proxy with real SeaweedFS, validated with AWS CLI
- [x] `go vet ./weed/s3api/...` clean (pre-existing warning in unmodified file only)
- [x] Full build passes

## Summary by CodeRabbit

* **New Features**
  * S3 gateway adds external URL support (s3.externalUrl) to enable correct signature verification behind reverse proxies.

* **Tests**
  * New unit and integration tests validating signature verification with proxies, forwarded headers, and various host/port cases.
  * End-to-end script-based tests exercising bucket/object flows via a proxy.

* **Documentation**
  * Added README with proxy-signature test setup and run instructions.

* **Chores**
  * CI workflow and containerized test environment added to run proxy signature tests.

### DIFF
--- a/.github/workflows/s3-proxy-signature-tests.yml
+++ b/.github/workflows/s3-proxy-signature-tests.yml
@@ -35,9 +35,6 @@ jobs:
         run: |
           set -x
           CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -buildvcs=false -v -o test/s3/proxy_signature/weed ./weed
-          ls -al test/s3/proxy_signature/weed
-          file test/s3/proxy_signature/weed
-          ldd test/s3/proxy_signature/weed || true
 
       - name: Run S3 Proxy Signature Tests
         timeout-minutes: 10
@@ -50,22 +47,6 @@ jobs:
           # Check if containers are running
           echo "Checking container status..."
           docker compose ps
-          
-          # Inspect binary in container image
-          echo "Inspecting binary in SeaweedFS container image:"
-          IMAGE_ID=$(docker compose images -q seaweedfs)
-          if [ -n "$IMAGE_ID" ]; then
-            docker run --rm "$IMAGE_ID" ls -al /usr/bin/weed || true
-            docker run --rm "$IMAGE_ID" ldd /usr/bin/weed || true
-          else
-            echo "Skipping image inspection: seaweedfs image not found"
-          fi
-          
-          # Capture initial logs to diagnose startup issues
-          echo "SeaweedFS logs:"
-          docker compose logs seaweedfs || true
-          echo "Nginx logs:"
-          docker compose logs nginx || true
           
           # Wait for services to be ready
           echo "Waiting for nginx proxy to be ready..."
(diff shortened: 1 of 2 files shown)


chrislusf:
FYI: going to revert https://github.com/seaweedfs/seaweedfs/pull/8434 since it did not solve the problem and introduced breaking change.

blitt001:
FYI, I did revert your changes already in my solution. Furthermore I added other exceptions that the code did not cover. For example http on port 443 or https on port 80 is also an exception where port number should be in the S3 signature. With Minio we had similar problems in the past and there they introduced the MINIO_SERVER_URL in order to bypass all the problem that do occur with reverse proxys, headers etc.

chrislusf:
@blitt001 thanks! Please help to resolve the build errors and merge conflicts.

blitt001:
Done

Issue #8442, opened 2026-02-25
Consistency fails when deleting files via mount


kisow:
**After file deletions complete, 
directory listings seen through weed mount remain inconsistent across nodes, and this inconsistency persists over time.** 

In other words, when accessing the same directory from different mount nodes, each node can continue to show a different set/count of remaining files even though filer-side deletion has already succeeded (as confirmed by a fresh mount showing 0).

### **Environment**
- SeaweedFS version : 4.13  

### Reproduction Steps 
- Mount target: `/mnt/dfs-dev/test.parallel_delete`
- Mount hosts: `host1`, `host2`

1. On `host1`, create 10,000 files in `/mnt/dfs-dev/test.parallel_delete`.
2. Run parallel deletion:
   - `host1`: delete 5,000 files
   - `host2`: delete 5,000 files
3. Check remaining file counts from both mounts.
4. (Optional cross-check) Mount the same path on a fresh node (e.g., `host3`) and count files.

**Expected**  
All mounts should report `0` remaining files.

**Actual**  
- `host1` and `host2` can report different non-zero counts.
- Fresh mount (`host3`) reports `0`.

**It appears that stale cache entries remain after deletion**, 
meaning some cache state is not synchronized even after the delete operation has completed.

chrislusf:
@kisow do you want to try out https://github.com/seaweedfs/seaweedfs/pull/8531 ?

kisow:
Looking at the changes in PR #8531, it seems there is a protocol change, 
so it might not be compatible with version 4.13. Is that correct?

Our DFS is currently running on version 4.13, so we will try it once we upgrade in the near future.

PR #8510, opened 2026-03-05
Fix metadata cache synchronization issue when deleting files in weed mount

kisow:

Fix a bug(#8442) where **stale entries in the metadata cache were never removed** due to a synchronization issue in the delete logic, causing files deleted on the Filer to remain visible on some weed mount instances.

* **Fix meta cache refresh consistency with atomic directory replace**
  * Fix a bug in `doEnsureVisited` where only inserts were performed without deleting existing entries, causing stale entries to remain once created.
    * Clear existing directory cache entries using `DeleteFolderChildren` before repopulating (clear-then-insert).
    * Add `ReplaceDirectoryEntries` for atomic directory replacement.
  * Synchronize insert operations in `doEnsureVisited` with a lock.
    * Entries deleted by subscriber events from a previous cache cycle could be reinserted due to a race condition with batch inserts.
    * Keep `ReadDirAllEntries` (which involves network I/O) outside the lock, and apply locking only to the local metadata cache update phase (clear-then-insert).
* **Allow local delete events to be processed by the mount subscriber**
* **Filter refresh reinsertions using a pending delete journal**
  * Entries deleted between `ReadDirAllEntries` and `ReplaceDirectoryEntries` could be reinserted during refresh.
  * Add logic to record deletions occurring in that window and filter them out during the insert phase.

# How is the PR tested?

* Mount the same directory on **node1** and **node2** using `weed mount`.
* Create 10,000 files in the directory from **node1**.
* Delete 5,000 files from each node concurrently.
* All 10,000 files should be deleted, but previously each node still showed a few hundred files remaining.
  * The files were already deleted on the Filer.

```
      CLUSTER=dev
      mount_path=/mnt/dfs-dev
      target_dir=/mnt/dfs-dev/test.parallel_delete
[1/5] validating nodes: node1, node2
[2/5] creating 10000 files on node1:/mnt/dfs-dev/test.parallel_delete
[3/5] collecting and splitting file list from node1
      total=10000 nodeA_delete=5000 nodeB_delete=5000
[4/5] deleting in parallel on node1 and node2
[5/5] tree summary after delete
      node1: 0 directories, 0 files
      node2: 0 directories, 0 files
```

* watch -n 1 'tree /mnt/dfs-dev/test.parallel_delete` for node[1-3] 
```
node1 : 0 directories, 0 files
node2 : 0 directories, 0 files
node3 : 0 directories, 52 files # 4.13
```

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Per-directory refresh lifecycle controls to better coordinate directory refreshes and avoid stale state during updates.

* **Bug Fixes**
  * Prevented “ghost” entries during directory refreshes by tracking pending deletions and filtering reinserts.
  * Improved atomic update handling to record overwritten/removed entries reliably.
  * Adjusted file removal behavior to ensure stale entries are cleared by subscription handlers.

### DIFF
--- a/weed/mount/meta_cache/meta_cache.go
+++ b/weed/mount/meta_cache/meta_cache.go
@@ -159,47 +159,35 @@ func (mc *MetaCache) DeleteFolderChildren(ctx context.Context, fp util.FullPath)
 	return mc.localStore.DeleteFolderChildren(ctx, fp)
 }
 
-// ReplaceDirectoryEntries atomically clears all children of dirPath and inserts
-// the provided entries under a single lock. This prevents race conditions where
-// concurrent DeleteEntry (from Unlink) or AtomicUpdateEntryFromFiler (from
-// subscription) could interleave with streaming batch inserts, causing ghost entries.
-func (mc *MetaCache) ReplaceDirectoryEntries(ctx context.Context, dirPath util.FullPath, entries []*filer.Entry) error {
+// insertBatchFiltered inserts a batch of entries under lock, filtering out any
+// entries that were deleted while the directory refresh was in progress.
+func (mc *MetaCache) insertBatchFiltered(ctx context.Context, dirPath util.FullPath, entries []*filer.Entry) error {
 	mc.Lock()
 	defer mc.Unlock()
+
 	pendingDeletes := mc.pendingDeletes[dirPath]
-	delete(mc.pendingDeletes, dirPath)
-
-	if err := mc.localStore.DeleteFolderChildren(ctx, dirPath); err != nil {
-		return err
-	}
-
-	n := len(entries)
-	for i := 0; i < n; i += batchInsertSize {
-		end := i + batchInsertSize
-		if end > n {
-			end = n
-		}
-		batch := entries[i:end]
-		if len(pendingDeletes) > 0 {
-			filtered := make([]*filer.Entry, 0, len(batch))
-			for _, entry := range batch {
-				if _, deleted := pendingDeletes[entry.Name()]; deleted {
-					glog.V(2).Infof("[ghost_trace] replace skip reinsert dir=%s name=%s", dirPath, entry.Name())
-					continue
-				}
-				filtered = append(filtered, entry)
+	if len(pendingDeletes) > 0 {
+		filtered := make([]*filer.Entry, 0, len(entries))
+		for _, entry := range entries {
+			if _, deleted := pendingDeletes[entry.Name()]; deleted {
+				glog.V(2).Infof("[ghost_trace] streaming skip reinsert dir=%s name=%s", dirPath, entry.Name())
+				continue
 			}
-			batch = filtered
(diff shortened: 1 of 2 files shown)


chrislusf:
thanks a lot for the fix! 

I was thinking there could be some metadata updates missed when fuse mount receives from the filer.

This particular fix seems specific for deletions, which may not reveal the actual problem.

Not sure. Need some effort to debug.

kisow:
Yes, this fix only addresses the bug where **deletions can be semi-permanently missed in the metadata cache**.

In our use case, **large batch deletions are frequent**, and we observed a persistent state where deletions appeared **inconsistently across nodes** (some nodes showing files as deleted while others still showed them). This caused service issues, so we reported and fixed it urgently.

In tests, the issue reproduced **very frequently**, leaving **dozens to hundreds of entries** each time.

After adding detailed logs and analyzing **10,000 create/delete operations**, I found that even when **all filer subscription events were delivered**, deleted entries could still reappear in the metadata cache for two reasons:

* **Directory cache refresh performs inserts without corresponding deletes**, so if a cache delete is missed once (e.g., due to a network error), stale entries may remain indefinitely.
* **There is no synchronization between cache inserts and deletes triggered by `Unlink`**, so a deleted entry can be reinserted into the metadata cache.
  * LevelDB writes are thread-safe, but the **operation order is non-deterministic**, so synchronization is needed.

If there is an **event loss issue**, it likely requires separate debugging. However, even without event loss, the bug where **deleted entries can reappear in the metadata cache** still needs to be addressed.


```mermaid
sequenceDiagram
    autonumber
    participant U as FUSE Unlink(Y)
    participant MC as MetaCache (LevelDB)
    participant F as Filer
    participant E as doEnsureVisited
    participant S as Subscription Handler

    Note over S,E: Deletion on Host2 → 64+ subscription events/2s<br/>→ RecordDirectoryUpdate → maybeRefreshDirectory

    Note over E: InvalidateChildrenCache(dir)



    E->>F: ReadDirAllEntries(dir) start
    Note over F: Y exists at the iterator snapshot point

    Note over U: Host1 requests deletion of file Y
    U->>F: filer_pb.Remove(Y, sig=[Host1])
    F-->>U: OK (Y deletion completed in filer)

    U->>MC: DeleteEntry(Y) [🔒 acquire]
    Note over MC: Y not inserted yet → no-op ⚠️
    U-->>MC: [🔓 release]

    F-->>E: [..., Y, ...] streaming response
    E->>MC: doBatchInsertEntries([...Y...])
    Note over MC: ⛔ lock not held!<br/>→ Y gets inserted (ghost)

    E->>MC: markCachedFn(dir)

    Note over S: Y deletion event received
    S->>S: sig == selfSignature → return nil
    Note over S: ❌ Entire event ignored

    Note over MC: ❌ Y = permanent ghost entry<br/>No reconciliation path

```

PR #8512, opened 2026-03-05
Remote storage: hot-reload mount mappings and propagate errors

blitt001:
## Summary

- **Hot-reload remote storage mount mappings**: After `remote.mount`, the filer picks up new mount mappings without a restart. Adds `Reset()` to `FilerRemoteStorage` to clear and rebuild the ptrie when `/etc/remote` changes.
- **Propagate remote storage errors instead of silent 404**: Previously all remote errors (network, auth, timeout) were silently swallowed in `maybeLazyFetchFromRemote`, causing `FindEntry` to return `ErrNotFound`. Now only `ErrRemoteObjectNotFound` returns nil (genuine 404); all other errors are propagated so callers can distinguish "file doesn't exist" from "remote is down."

## Test plan

- [ ] Verify `remote.mount` picks up new mappings without filer restart
- [ ] Verify remote timeout/auth errors return 5xx, not 404
- [ ] Verify genuinely missing files still return 404
- [ ] Unit tests included for both changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Better error handling for remote fetches: network errors are propagated, and remote "not found" is distinguished and translated appropriately.
  * Remote storage reloads now trigger correctly on relevant metadata events.

* **Refactor**
  * Improved concurrency safety for in-memory remote storage state and added safe reset behavior.

* **Tests**
  * Added tests covering error propagation, not-found handling, and remote storage reload/mapping behavior.

### DIFF
--- a/weed/filer/filer_on_meta_event.go
+++ b/weed/filer/filer_on_meta_event.go
@@ -130,11 +130,11 @@ func (f *Filer) LoadRemoteStorageConfAndMapping() {
 	}
 }
 func (f *Filer) maybeReloadRemoteStorageConfigurationAndMapping(event *filer_pb.SubscribeMetadataResponse) {
-	dir := event.Directory
-	if event.EventNotification.NewParentPath != "" {
-		dir = event.EventNotification.NewParentPath
+	msg := event.GetEventNotification()
+	if msg == nil {
+		return
 	}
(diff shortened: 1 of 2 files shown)


blitt001:
Closing to rework and resubmit after thorough local testing.

PR #8531, opened 2026-03-06
mount: make metadata cache rebuilds snapshot-consistent

### LINKED ISSUE #8442: Consistency fails when deleting files via mount

chrislusf:
## Summary
- expose filer metadata events and list snapshot timestamps so mounts can consume authoritative mutation events and stable directory listings
- route local mount mutations and subscriber events through a single sequenced metadata-cache applier instead of direct LevelDB writes
- keep hot directories in read-through mode, and rebuild cold directories from a listing snapshot while buffering and replaying post-snapshot mutations before publishing the cache
- dedupe repeated filer events in the applier so self events can flow through the same path without double-applying
- add unit tests for event ordering, local-vs-remote side effects, buffered build replay, and uncached rename behavior

Closes #8442

## Testing
- go test ./weed/pb/filer_pb ./weed/filer ./weed/server ./weed/mount/...

## Release Notes

* **New Features**
  * Added snapshot-based consistency for directory listings
  * Introduced metadata event propagation in file operations (create, update, delete)
  * Added direct directory read mode to improve performance under certain conditions

* **Bug Fixes**
  * Enhanced nil-safety checks across multiple file operations to prevent crashes
  * Improved error handling in directory listing and metadata operations
  * Fixed inconsistent timestamp handling in file metadata

* **Improvements**
  * Strengthened metadata consistency during rename operations
  * Better error messages for metadata application failures

### DIFF
--- a/weed/filer/filer_notify.go
+++ b/weed/filer/filer_notify.go
@@ -71,6 +71,9 @@ func (f *Filer) notifyUpdateEvent(ctx context.Context, oldEntry, newEntry *Entry
 }
 
 func (f *Filer) newMetadataEvent(oldEntry, newEntry *Entry, deleteChunks, isFromOtherCluster bool, signatures []int32) *filer_pb.SubscribeMetadataResponse {
+	if oldEntry == nil && newEntry == nil {
+		return nil
+	}
 	var fullpath util.FullPath
 	if oldEntry != nil {
 		fullpath = oldEntry.FullPath
--- a/weed/mount/meta_cache/meta_cache.go
+++ b/weed/mount/meta_cache/meta_cache.go
@@ -535,6 +535,17 @@ func (mc *MetaCache) applyMetadataResponseLocked(ctx context.Context, resp *file
 
 	mc.Lock()
 	err := mc.atomicUpdateEntryFromFilerLocked(ctx, oldPath, newEntry, allowUncachedInsert)
+	// When a directory is deleted or moved, remove its cached descendants
+	// so stale children cannot be served from the local cache.
+	if err == nil && oldPath != "" && message.OldEntry != nil && message.OldEntry.IsDirectory {
+		isDelete := message.NewEntry == nil
+		isMove := message.NewEntry != nil && (message.NewParentPath != resp.Directory || message.NewEntry.Name != message.OldEntry.Name)
+		if isDelete || isMove {
+			if deleteErr := mc.localStore.DeleteFolderChildren(ctx, oldPath); deleteErr != nil {
+				glog.V(2).Infof("delete descendants of %s: %v", oldPath, deleteErr)
+			}
+		}
+	}
 	mc.Unlock()
 	if err != nil {
 		return metadataResponseSideEffects{}, err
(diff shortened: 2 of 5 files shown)


chrislusf:
Follow-up on the concurrency model this series is trying to fix.

Before this change, mount metadata consistency had three unsynchronized paths:
1. local FUSE ops mutated the cache directly
2. subscriber events mutated the cache directly
3. directory refresh could clear and repopulate a cached directory with no filer snapshot boundary

The bug in #8442 is one manifestation of that model: a cached directory becomes hot, refresh starts, a delete/create/rename lands while the directory is being cleared or relisted, and the cache can end up serving a state that never existed atomically. In practice that showed up as ghost entries after delete, but the underlying issue was broader than delete handling.

What this PR changes:
- hot directories stop doing in-place refresh and instead fall back to read-through
- local RPC responses and subscriber events go through one sequenced cache applier
- cold directory rebuilds use a filer listing snapshot timestamp, buffer any metadata events that touch the directory while the build is in progress, and replay only events newer than the snapshot before the directory is marked cached
- self events are deduped inside the applier, so the mount no longer depends on filtering them out at subscribe time

That should close the main lost-update window: the cache is no longer cleared and repopulated concurrently with unrelated writers, and a rebuild publishes only after post-snapshot mutations have been replayed.

Tradeoffs and remaining limits:
- Hot directories are intentionally served read-through instead of staying aggressively cached. This is a correctness-first tradeoff and can be slower under heavy churn.
- The replay buffer is per-mount and in-memory. If the subscriber stream disconnects or there is any gap, the design still relies on invalidation + rebuild/read-through rather than trying to guarantee gap-free cache continuity.
- Correctness depends on the filer timestamps used for listing snapshots and emitted metadata events being monotonic enough for the `event.TsNs > snapshotTsNs` cutoff to be meaningful.
- The self-event dedupe is bounded and local to recent events. It is meant to suppress the mount's own response-event plus subscribe-event double-apply path, not to be a durable replay journal.

So the goal here is not "make the cache always perfectly current". The goal is narrower and safer: either serve from a cache generation that corresponds to a stable snapshot plus replayed post-snapshot events, or bypass cache and go back to the filer.

PR #8645, opened 2026-03-15
fix(remote): don't send empty StorageClass in S3 uploads

chrislusf:
## Summary

- When `S3StorageClass` is empty (the default), `aws.String("")` was passed as the `StorageClass` in S3 upload requests. While AWS S3 treats this as "use default," S3-compatible providers (e.g. SharkTech) reject it with `InvalidStorageClass: The storage class you specified is not valid`.
- Only set `StorageClass` on the upload input when a non-empty value is configured, letting the provider use its bucket default.

Fixes https://github.com/seaweedfs/seaweedfs/discussions/8644

## Test plan

- [ ] Verify `filer.remote.sync` works with S3-compatible providers that don't support storage classes (e.g. SharkTech)
- [ ] Verify `filer.remote.sync` still respects `S3StorageClass` when explicitly configured (e.g. `STANDARD_IA`)
- [ ] Verify uploads to AWS S3 with default (empty) config still use the bucket's default storage class

### DIFF
--- a/weed/remote_storage/s3/s3_storage_client.go
+++ b/weed/remote_storage/s3/s3_storage_client.go
@@ -275,13 +275,16 @@ func (s *s3RemoteStorageClient) WriteFile(loc *remote_pb.RemoteStorageLocation, 
 	}
 
 	// Upload the file to S3.
-	_, err = uploader.Upload(&s3manager.UploadInput{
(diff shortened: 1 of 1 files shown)


jfmatth:
Just cloned down the latest code, compiled and works great with SharkTech.

PR #8647, opened 2026-03-16
filer: add conditional update preconditions

chrislusf:

s3tables metadata commits only had a client-side compare-and-swap check. Two writers could read the same table xattr version and still race on UpdateEntry.

- add filer UpdateEntry preconditions
- validate them on the server and handle missing expected xattrs correctly
- use a dedicated small metadata-version xattr as the CAS token for iceberg table metadata updates
- make the stale-update test hook return errors instead of failing from a non-test goroutine

Store-level atomic CAS across all filer backends is still a follow-up because it requires FilerStore interface changes or transactional enforcement in every store implementation.

# How is the PR tested?

- go test ./weed/plugin/worker/iceberg -run 'TestMetadataVersionCAS|TestCommitWithRetry' -count=1
- go test ./weed/server -run TestDoesNotExist -count=1

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added server-side precondition validation for entry updates, enabling compare-and-swap semantics to prevent data conflicts.
  * Introduced metadata version tracking to detect and prevent concurrent update conflicts with proper error reporting.

* **Bug Fixes**
  * Enhanced concurrent update handling to return FailedPrecondition errors when metadata versions mismatch, guiding clients to retry with updated state.

### DIFF
--- a/weed/plugin/worker/iceberg/exec_test.go
+++ b/weed/plugin/worker/iceberg/exec_test.go
@@ -38,7 +38,7 @@ type fakeFilerServer struct {
 
 	mu           sync.Mutex
 	entries      map[string]map[string]*filer_pb.Entry // dir → name → entry
-	beforeUpdate func(*fakeFilerServer, *filer_pb.UpdateEntryRequest)
+	beforeUpdate func(*fakeFilerServer, *filer_pb.UpdateEntryRequest) error
 
 	// Counters for assertions
 	createCalls int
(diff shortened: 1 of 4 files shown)


chrislusf:
Addressed the actionable review items in the latest push:

- fixed the missing-xattr precondition case in `validateUpdateEntryPreconditions`
- changed the fake `beforeUpdate` hook to return an error instead of calling `t.Fatal*` from the gRPC goroutine
- switched the s3tables CAS path to a compact `s3tables.metadataVersion` token so the request no longer re-sends the previous full metadata blob as the precondition value

I did not try to close the storage-layer TOCTOU gap in this PR. That requires widening the filer/store write contract so the compare-and-swap check is enforced atomically inside each backend transaction, rather than in the gRPC layer. That is the right long-term fix, but it is a broader filer/storage change across all stores, so I left it as a follow-up instead of partially implementing it here.

PR #8648, opened 2026-03-16
iceberg: detect compaction for spec-evolved tables

chrislusf:

Compaction detection skipped Iceberg tables whose current manifest list contained data manifests from more than one partition spec, even though the compaction executor already bins files by spec.

- remove the planner-side mixed-spec rejection
- extend the manifest-writing helper so tests can build mixed-spec manifest lists
- add a regression test for a spec-evolved table with compactable small files in multiple specs

# How is the PR tested?

- go test ./weed/plugin/worker/iceberg -run 'TestDetectSchedulesCompactionWithoutSnapshotPressure|TestDetectSchedulesCompactionWithDeleteManifestPresent|TestDetectSchedulesCompactionForSpecEvolvedTable' -count=1

# Checks

# Checks for AI generated PRs

### DIFF
--- a/weed/plugin/worker/iceberg/detection.go
+++ b/weed/plugin/worker/iceberg/detection.go
@@ -333,18 +333,13 @@ func hasEligibleCompaction(
 	}
 
 	var dataManifests []iceberg.ManifestFile
-	specIDs := make(map[int32]struct{})
 	for _, mf := range manifests {
 		if mf.ManifestContent() != iceberg.ManifestContentData {
 			continue
 		}
 		dataManifests = append(dataManifests, mf)
-		specIDs[mf.PartitionSpecID()] = struct{}{}
 	}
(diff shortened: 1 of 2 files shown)


chrislusf:
Follow-up on the skipped/remaining review comments:

- CodeRabbit skipped this PR because it targets a non-default base branch in the stack. I did a manual pass instead; there were no additional actionable issues from that skip.
- The open Gemini threads on the helper/test are descriptive rather than new defects. The helper already fails fast on mismatched `manifests/specs` lengths and writes each manifest with its matching spec.
- I did not broaden this regression to cover schema/type/order/count evolution variants. This PR only removes the planner-side mixed-spec early exit; it does not change schema projection or compaction execution semantics. Broader schema-evolution coverage belongs in the compaction-path tests, not in this detection regression.

PR #8649, opened 2026-03-16
iceberg: add lifecycle maintenance parity options

chrislusf:

The maintenance worker was still missing a few practical Iceberg lifecycle features: snapshot expiration did not respect refs, orphan cleanup could not run in dry-run mode, and expired metadata JSON files were never cleaned up.

- make expiration planning ref-aware for branches and tags
- add bounds-checked retention planning and cleanup metrics
- add clean_expired_metadata and remove_orphans_dry_run worker options plus UI defaults
- add tests for tagged snapshots, expired metadata cleanup, dry-run orphan reporting, and config parsing

# How is the PR tested?

- go test ./weed/plugin/worker/iceberg -count=1

# Checks

# Checks for AI generated PRs

### DIFF
--- a/weed/plugin/worker/iceberg/config.go
+++ b/weed/plugin/worker/iceberg/config.go
@@ -23,15 +23,18 @@ const (
 	defaultOperations             = "all"
 
 	// Metric keys returned by maintenance operations.
-	MetricFilesMerged        = "files_merged"
-	MetricFilesWritten       = "files_written"
-	MetricBins               = "bins"
-	MetricSnapshotsExpired   = "snapshots_expired"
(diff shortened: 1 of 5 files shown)


chrislusf:
Addressed the remaining review items in the latest push:

- renamed `numSnapshots` to `ancestorCount`
- replaced the unchecked `int64 -> int` conversion with an explicit bounds-checked helper before using `MaxSnapshotsToKeep` in retention planning

GitHub Advanced Security comments do not expose a normal review-thread resolve path, but the code-scanning finding itself is handled by the follow-up commit on this branch.

PR #8662, opened 2026-03-16
docs: add s3tables planning index design

chrislusf:

The S3 tables maintenance roadmap calls out indexed planning as the highest-leverage scale improvement after parity work, but the current roadmap only outlines the idea. We need a concrete design for the index scope, storage model, read/write behavior, safe skip rules, and failure handling before implementation starts.

This PR adds a dedicated design doc for a per-table planning index. It defines the cached fields, storage shape, skip-decision rules, best-effort update path, schema versioning, metrics, and implementation sequence.

# How is the PR tested?

- documentation-only change

## Checklist

- [x] Updated documentation accordingly
- [x] No protobuf changes
- [x] No user-facing behavior changes in this PR

## Release Notes

* **Documentation**
  * Added a new design document outlining the S3 Tables Planning Index, which details the structure, read and write paths, failure handling, metrics, and testing strategy for an optimization feature.

### DIFF
--- a/docs/design/s3tables-planning-index.md
+++ b/docs/design/s3tables-planning-index.md
@@ -1,0 +1,162 @@
+# S3 Tables Planning Index Design
+
+## Problem
+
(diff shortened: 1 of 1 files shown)


chrislusf:
Replacing the docs-only design PR with an implementation PR for this feature.

PR #8679, opened 2026-03-18
feat: improve aio support for admin/volume ingress and fix UI links

jr200:

The allInOne deployment mode lacks feature parity with standalone component deployments for admin and volume ingress configuration:

1. The admin ingress template requires `admin.enabled`, which deploys a standalone admin pod. In allInOne mode, the admin server runs inside the allInOne pod but has no way to expose an ingress through the chart.

2. The volume ingress template requires `volume.enabled`, while the filer and s3 ingress templates already support allInOne mode via an `or .Values.{component}.enabled .Values.allInOne.enabled` pattern.

This inconsistency forces users to create manual ingress manifests for the volume UI when using allInOne.

3. The legacy master UI templates always link to the internal pod IP (`Url`) for volume server links, ignoring the `PublicUrl` field. 
This makes the "Open Volume Server UI" link unusable when the master UI is accessed through an ingress. The new admin UI already handles this correctly.

- Add `allInOne.admin` config section (enabled by default, ports inherit from `admin.*`) and expose admin ports in the allInOne deployment and service when enabled.
- Update the admin ingress template to use the same or-based enablement and ternary service name pattern used by filer and s3 ingress templates.
- Apply the same pattern to the volume ingress template so it renders when `allInOne.enabled` is true, targeting the all-in-one service.
- Fix `master.html` and `masterNewRaft.html` to prefer `PublicUrl` over `Url` when `PublicUrl` is set and differs from `Url`.

# How is the PR tested?

- Verified with `helm template` that admin and volume ingresses render correctly in allInOne mode, targeting the all-in-one service.
- Confirmed that standalone mode (volume.enabled / admin.enabled) continues to work unchanged.
- Tested in a live allInOne K3s deployment: volume server UI is accessible through the ingress, and the master UI volume link correctly uses the PublicUrl.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Dashboard enriched with public URL mappings so data nodes show the most appropriate reachable address.
  * Ingress now selects the correct service for volume or all-in-one deployments automatically.

* **Bug Fixes**
  * Topology view shows a single, correct link per data node (prefers PublicUrl when available).
  * Minor Kubernetes manifest formatting cleanup.

### DIFF
--- a/weed/admin/dash/cluster_topology.go
+++ b/weed/admin/dash/cluster_topology.go
@@ -65,6 +65,7 @@ func (s *AdminServer) fetchPublicUrlMap() map[string]string {
 
 	body, err := io.ReadAll(resp.Body)
 	if err != nil {
+		glog.V(1).Infof("Failed to read /dir/status response body: %v", err)
 		return nil
 	}
 


jr200:
thanks @chrislusf !

PR #8819, opened 2026-03-29
docker: default published images to seaweed user

### LINKED ISSUE #8816: AquaScanner/Tivy Vulnerability Scanner tipped off by user switch not being performed in Dockerfile using USER keyword
Image

Corporate security scanners such as AquaScanner/Tivy, are being kicked off by the user not being root. Reason being, these recognize the user switch via docker inspect --format='{{.Config.User}}' chrislusf/seaweedfs:4.17, which for this image, due to not using the 'USER' Dockerfile keyword, returns ''

chrislusf:
## Summary
- set `USER seaweed` in the published Docker images so `docker inspect` reports a non-root default user
- keep the existing runtime fallback path for operators who start the container as `root` so `/data` ownership can still be repaired before dropping privileges

## Why
Issue #8816 reports that scanners such as Trivy and Aqua flag the image because it drops privileges only at runtime and does not set a Dockerfile `USER`. That leaves the image metadata with an empty `Config.User`, which is treated as a root default.

Using `USER seaweed` fixes the scanner signal without removing the compatibility path for deployments that still need to run the entrypoint as `root` to `chown` mounted storage before switching users.

## Impact
- default container startup now runs as the non-root `seaweed` user
- deployments that rely on startup-time ownership repair can still opt in with `--user root`

## Validation
- `git diff --check`
- `/bin/sh -n docker/entrypoint.sh`

Closes #8816.

## Release Notes

* **Chores**
  * Container now defaults to running as a non-root user for improved security. Users can override this with `docker run --user root` if root access is needed for volume permission management.

### DIFF
--- a/.github/workflows/container_latest.yml
+++ b/.github/workflows/container_latest.yml
@@ -26,7 +26,6 @@ on:
 
 permissions:
   contents: read
-  security-events: write
 
 jobs:
   setup:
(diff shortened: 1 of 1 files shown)


docker/Dockerfile.go_build
```diff
@@ -79,5 +79,9 @@ RUN mkdir -p /data/filerldb2 && \
 VOLUME /data
 WORKDIR /data
 
-# Entrypoint will handle permission fixes and user switching
+# Run as non-root by default (satisfies security scanners).
+# Use `docker run --user root` if you need the entrypoint to fix
+# /data volume ownership before dropping privileges.
+USER seaweed
+
 ENTRYPOINT ["/entrypoint.sh"]
```
lfritz-seweurodrive, line 87: Positive news: The security scanner is now happy

Negative news:
I've tried patching my image the same way, by just adding the 'USER seaweed' entry.

I'm afraid, just changing the USER alone before calling /entrypoint.sh might not be enough, because now the chown -R calls which might have occured, do not occur anymore in K8S deployments, leading to e.g. seaweedfs-master being unable to write in /logs and /data. 

See also here:
image

For K8S deployments, there either needs to be an initContainer which performs the chmoding for the seaweed user.

      initContainers:
        - command:
            - /bin/sh
            - '-ec'
            - |
              chown -R 1000:1000 /data || true
              chown -R 1000:1000 /logs || true
          image: ${IMAGESEAWEEDFS}
          imagePullPolicy: Always
          name: givefolderperms
          resources: {}
          securityContext:
            runAsGroup: 0
            runAsNonRoot: false
            runAsUser: 0
          terminationMessagePath: /dev/termination-log
          terminationMessagePolicy: File
          volumeMounts:
            - mountPath: /data
              name: pvc-for-data
            - mountPath: /logs/
              name: seaweedfs-master-log-volume

e.g. adding this initContainer to all statefulsets of seaweedfs, in my case, enabled the seaweedfs to work with the seaweedfs user without the permission errors.

Issue #8908, opened 2026-04-03
S3 API: Docker Registry blob finalization fails silently, resulting in corrupted/missing blobs


os-pradipbabar:
## Summary

When Docker Registry (or similar S3-compatible clients) finalize blob uploads, SeaweedFS S3 API silently fails to move/copy the uploaded data from the temporary upload location to the final blob location. The S3 API returns HTTP 201 Created (success), but the blob data never reaches its destination. Later, when the blob is requested, it either doesn't exist or returns corrupted data.

## Environment

- **SeaweedFS Version**: 4.18 (latest)
- **Storage Backend**: S3 API (used by Docker Registry v2)
- **Client**: Docker Registry v2 with S3 storage driver
- **Deployment**: Kubernetes
- **Registry Config**: S3 storage driver pointing to SeaweedFS S3 endpoint

### Symptoms

1. **Upload succeeds** - Docker Registry uploads image layer (blob)
2. **Registry returns success** - HTTP 201 Created, manifest created successfully
3. **Download fails** - When pulling the image, SHA256 mismatch occurs

```bash
# Registry reports blob exists with correct size
$ curl -I https://registry/v2/.../blobs/sha256:719388d1...
HTTP/1.1 200 OK
Content-Length: 215324229  # ~205 MB

# But download returns wrong content
$ curl https://registry/v2/.../blobs/sha256:719388d1... | sha256sum
bd3e01d91af27bcc8a5cb822c1a7a66d582177ff0a6078fb89920ad311e78eff
# ❌ Expected: 719388d129af2e5e9d19ea2528319c8db508d35f8e8ddedcc35e7abc9f86ad9e

# Pod fails to pull image
Error: ImagePullBackOff
Reason: Digest did not match
  Expected: sha256:719388d1...
  Got:      sha256:bd3e01d9...
```

### Frequency

- Observed in **2-5%** of large image uploads (>200MB)
- Reproducible with Docker Registry syncing from ECR to SeaweedFS-backed registry

## Reproduction

Using Docker Registry with S3 storage driver backed by SeaweedFS:

```yaml
# registry/config.yml
storage:
  s3:
    regionendpoint: https://seaweedfs-s3:8666
    bucket: registry
    region: us-east-1
```

Then:
1. Push a large image (>200MB) to the registry
2. Wait for push to complete (HTTP 201)
3. Try to pull the image
4. Observe SHA256 mismatch ~2-5% of the time

## Impact

[...]

PR #8909, opened 2026-04-03
fix: Prevent orphaned metadata from cancelled S3 operations (Issue #8908)

os-pradipbabar:

Fixes silent data corruption in Docker Registry blob uploads and other S3-compatible clients by adding pre-cancellation checks to write operations in filer store wrapper. This prevents orphaned metadata entries when client connections are interrupted during S3 CopyObject, CompleteMultipartUpload, and similar operations.

A regression introduced in **v3.97** added `context.WithoutCancel(ctx)` to **all** filer store operations, including **write paths**. This causes write operations to ignore request cancellation signals.

- Docker Registry blob uploads return HTTP 201 Created (success) but blobs are missing or corrupted
- SHA256 mismatch errors during image pulls (ImagePullBackOff)
- Observed in 2-5% of large uploads (>200MB)
- `EmptyFolderCleaner` deletes upload directories that should have been cleaned by clients
- No error logs during the failed upload operations

added a **pre-cancellation safety check** to all **write operations** in `FilerStoreWrapper` **before** calling `context.WithoutCancel(ctx)`.

### Key change

For all write-related methods:

```go
if ctx.Err() != nil {
    return ctx.Err()
}
ctx = context.WithoutCancel(ctx)
```

# How is the PR tested?
Added comprehensive coverage to `weed/filer/filerstore_wrapper_test.go`.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * File store write operations now immediately return when the context is cancelled, eliminating unnecessary processing and improving responsiveness during cancellation events.

* **Tests**
  * Added comprehensive test coverage to verify context cancellation handling across file store operations, including writes, reads, and transaction control methods.


chrislusf (commented):
## Review: Refactoring suggestions & a bug fix

The overall approach is sound — pre-cancellation checks before `context.WithoutCancel` on write paths is the right fix. A few things to address:

### Bug: `RollbackTransaction` must NOT be guarded

`RollbackTransaction` is a cleanup operation, often called *because* something went wrong. If the context is cancelled during a write and rollback is also blocked, you get exactly the orphaned state this PR aims to prevent. See `filer_grpc_server_rename.go`:

```go
moveErr := fs.moveEntry(ctx, ...)
if moveErr != nil {
    fs.filer.RollbackTransaction(ctx)  // must succeed even if ctx is cancelled
    return nil, ...
}
```

**Remove the cancellation guard from `RollbackTransaction`.**

### Refactor: Extract a helper to eliminate repetition

The same 3-line check is repeated 11 times. Extract it:

```go
func checkContextCancelled(ctx context.Context) error {
    return ctx.Err()
}
```

Each call site becomes:
```go
if err := checkContextCancelled(ctx); err != nil {
    return err
}
```

### Test improvement: Add `context.DeadlineExceeded` coverage

`ctx.Err()` returns either `context.Canceled` or `context.DeadlineExceeded`. The tests only cover cancellation — add deadline-exceeded cases too.

### Minor: Inconsistent blank lines

Some guarded methods have a spurious blank line before the check while `InsertEntry` does not.

---

I've pushed a refactored version to branch `industrious-thrill` that addresses all of the above — feel free to cherry-pick or use as reference.

PR #9079, opened 2026-04-14
perf(filer.sync): don't serialize descendants behind dir attribute updates

chrislusf:
## Summary

`MetadataProcessor` in `filer.sync` treated every in-flight directory job as a subtree barrier: any active dir job at `/foo` forced every subsequent file event under `/foo` to wait, and because `AddSyncJob`'s admit loop runs on the single `stream.Recv()` goroutine, a stalled descendant also stalled the whole gRPC stream. For syncs over a large directory, every attribute-only dir event (mtime / xattr / chmod bumps) became a full-subtree pinch point — and this was invisible to `-concurrency`, since the stall happens upstream of the worker pool.

This PR loosens the rule: a dir event is classified as **barrier** (create / delete / rename) or **non-barrier** (`filer_pb.IsUpdate` on a directory — same parent, same name, i.e. an in-place attribute update). Only barrier dirs block descendants and get blocked by ancestor barrier dirs. Non-barrier dir updates still bump the ancestor `descendantCount`, so an incoming barrier dir on an ancestor still waits for them — preserving the "delete `/a` waits for in-flight `/a/b` update" safety.

Context: https://github.com/seaweedfs/seaweedfs/discussions/8771#discussioncomment-16559394

### What changed

- New `jobKind` enum (`kindFile`, `kindBarrierDir`, `kindNonBarrierDir`) on `syncJobPaths`, replacing `isDirectory bool`.
- `activeDirPaths` → `activeBarrierDirPaths`. Only barrier dirs land there, and that's the only index the ancestor check reads — so file events no longer wait behind attribute-only dir updates on ancestors.
- Non-barrier dir updates are not tracked at their own path, but still increment `descendantCount` for ancestors so barrier ancestors wait for them.
- `pathConflicts` dispatches on kind: files check `activeFilePaths`, barrier dirs check `descendantCount`, non-barrier dirs check nothing at their own path; all three still check `activeBarrierDirPaths` on ancestors.
- `extractPathsFromMetadata` → `extractJobInfo`, which also returns the kind via `classifyDirEvent`.

### What stays the same

- File-vs-same-file serialization.
- File-under-active-barrier-dir serialization (barrier dir create/delete/rename still blocks descendants).
- Barrier dirs still wait for in-flight descendants of any kind.
- Rename behavior: both source and destination paths are registered, and a dir rename is a barrier on both sides.

## Test plan

- [x] `go build ./weed/command/...`
- [x] `go test ./weed/command/ -count=1` — full package green.
- [x] New `TestNonBarrierDirUpdateDoesNotBlockDescendants` covers:
  - attribute-only dir update at `/dir1` does **not** conflict with a file under `/dir1` (or deeply nested);
  - a barrier dir create at the same path **still** blocks a file descendant;
  - a barrier dir delete **still** waits for an in-flight file descendant;
  - a barrier ancestor dir delete **still** waits for an in-flight non-barrier descendant update.
- [x] Existing `TestFileVsFileConflict`, `TestFileUnderActiveDirConflict`, `TestDirWithActiveFileUnder`, `TestDirVsDirConflict`, `TestRenameConflict`, `TestActiveRenameConflict`, `TestRootDirConflict`, `TestIndexCleanup`, `TestWatermarkWithHeap`, `TestPathAncestors` all still pass.
- [ ] Reviewer: worth running the existing `filer_sync_subscription_test` stream benchmarks to sanity-check throughput on a dir-attribute-heavy workload.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync conflict detection: directory updates that only change attributes no longer block descendant file operations; barrier directory creates/deletes/renames still block descendants and now enforce stricter same-path serialization with files and other barrier ops.

* **Tests**
  * Expanded tests covering non-blocking directory attribute updates, same-path barrier serialization, and updated benchmarks.

### DIFF
--- a/weed/command/filer_sync_jobs.go
+++ b/weed/command/filer_sync_jobs.go
@@ -71,6 +71,13 @@ type MetadataProcessor struct {
 	// deliberately invisible to the ancestor check so that they don't
 	// serialize every file descendant.
 	activeBarrierDirPaths map[util.FullPath]int
+	// activeNonBarrierDirPaths counts active non-barrier dir jobs at each
+	// exact path. This is read *only* by incoming barrier dirs, so a
+	// delete/rename/create at p correctly waits for an in-flight chmod/
+	// xattr/mtime update at the same p. It is deliberately invisible to the
+	// ancestor check, so non-barrier updates still don't serialize file
+	// descendants.
+	activeNonBarrierDirPaths map[util.FullPath]int
 	// descendantCount counts active jobs (of any kind) strictly under each
 	// directory. Read by incoming barrier dirs so they wait for their whole
 	// subtree to drain before running, regardless of descendant kind.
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the high-priority review comment in 23a477435. `pathConflicts` now serializes:

- any incoming job (file, barrier dir, or non-barrier update) against an in-flight barrier dir at the same path;
- incoming file or barrier-dir jobs against an in-flight file at the same path.

Non-barrier dir-vs-non-barrier-dir at the same path still overlaps (attribute bumps are last-writer-wins — intentional).

`TestDirVsDirConflict` and `TestFileUnderActiveDirConflict` flipped their "same path does not conflict" assertions; new `TestSamePathBarrierSerialization` covers all five same-path cases explicitly. `./weed/command/` suite green.

On CodeRabbit's nitpick about rename classification — the rename branch uses `oldEntry.IsDirectory` as CodeRabbit noted, and the filer never reshapes an entry's directory status through a rename, so I'm leaving that as-is.

The `FUSE Integration Testing` check failure on the prior commit was `TestPosixFileLocking/ConcurrentLockContention` — unrelated to this PR (no FUSE code is touched). Looks like a flake; CI re-run on the new commit should confirm.

PR #9100, opened 2026-04-16
feat(mount): singleflight dedup for concurrent chunk reads

chrislusf:
## Summary

- Add a lightweight `SingleFlightGroup` in `weed/util/singleflight.go` that deduplicates concurrent function calls keyed by a string. When multiple goroutines call `Do` with the same key, only one executes the function; the rest wait and share the result.
- Integrate the singleflight group into `ReaderCache` so that the network fetch (lookup + HTTP download) inside `SingleChunkCacher.startCaching` is deduplicated per fileId. If a chunk gets evicted from the `downloaders` map and re-requested concurrently, only one volume server round-trip occurs.
- Add unit tests for the singleflight (dedup, independent keys, error propagation, fresh call after completion) and an integration test confirming concurrent `ReadChunkAt` calls for the same fileId produce only one lookup.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./weed/util/... -race` passes (singleflight unit tests)
- [x] `go test ./weed/filer/... -race` passes (reader cache tests including new dedup test)
- [x] Singleflight dedup test is stable over 100 runs with `-count=100 -race`
- [ ] Manual smoke test: mount a SeaweedFS volume, read a large file with multiple concurrent readers (e.g. `fio --numjobs=8`), confirm volume server access logs show no duplicate chunk fetches

## Summary by CodeRabbit

* **Performance**
  * Concurrent requests for the same chunk now share a single in-flight fetch, reducing duplicate downloads and contention.

* **Tests**
  * Added concurrency tests verifying deduplication of concurrent chunk fetches, correct error propagation to waiters, and fresh execution for subsequent requests after completion.

* **Documentation**
  * Clarified caching behavior to explain deduplication guarantees.

### DIFF
--- a/weed/filer/reader_cache.go
+++ b/weed/filer/reader_cache.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/util"
 	"github.com/seaweedfs/seaweedfs/weed/util/chunk_cache"
 	util_http "github.com/seaweedfs/seaweedfs/weed/util/http"
 	"github.com/seaweedfs/seaweedfs/weed/util/mem"
@@ -18,8 +19,9 @@ type ReaderCache struct {
 	chunkCache     chunk_cache.ChunkCache
 	lookupFileIdFn wdclient.LookupFileIdFunctionType
 	sync.Mutex
-	downloaders map[string]*SingleChunkCacher
-	limit       int
+	downloaders  map[string]*SingleChunkCacher
+	limit        int
+	fetchGroup   util.SingleFlightGroup
 }
 
 type SingleChunkCacher struct {
@@ -178,6 +180,10 @@ func newSingleChunkCacher(parent *ReaderCache, fileId string, cipherKey []byte, 
 // startCaching downloads the chunk data in the background.
 // It does NOT hold the lock during the HTTP download to allow concurrent readers
 // to wait efficiently using the done channel.
+//
+// The network fetch is wrapped in a per-ReaderCache singleflight group keyed
+// by fileId, so concurrent downloads of the same chunk share a single HTTP
+// round-trip.
 func (s *SingleChunkCacher) startCaching() {
 	s.wg.Add(1)
 	defer s.wg.Done()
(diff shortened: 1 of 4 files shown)


utafrali (commented):
The dedup approach is architecturally sound and the test coverage is thorough, but the PR introduces a hand-rolled singleflight implementation when `golang.org/x/sync/singleflight` is already in the module's dependency graph. The more pressing concern is that the returned `[]byte` is shared across all concurrent callers with no documented ownership contract, which is a latent correctness hazard if any caller mutates the buffer.

weed/util/singleflight.go
```diff
@@ -0,0 +1,59 @@
+package util
```
utafrali, line 1: This custom `SingleFlightGroup` reimplements functionality that `golang.org/x/sync/singleflight` already provides, and that package is already a direct dependency used in `weed/filer/filer.go`. A thin typed wrapper avoids maintaining a parallel implementation:

```go
import "golang.org/x/sync/singleflight"

type SingleFlightGroup struct{ g singleflight.Group }

func (sg *SingleFlightGroup) Do(key string, fn func() ([]byte, error)) ([]byte, error) {
    v, err, _ := sg.g.Do(key, func() (interface{}, error) { return fn() })
    if err != nil {
        return nil, err
    }
    return v.([]byte), nil
}
```

This also grants `Forget()` and `DoChan()` for free. The standard library version is race-tested and covers edge cases (e.g. panicking functions) that the custom version does not handle.

weed/util/singleflight.go
```diff
@@ -0,0 +1,59 @@
...
+// call represents an in-flight or completed function call.
+type call struct {
+	wg  sync.WaitGroup
+	val []byte
+	err error
+}
+
+// SingleFlightGroup provides deduplication of concurrent function calls
+// keyed by a string. If multiple goroutines call Do with the same key
+// concurrently, only one executes the function; the others wait and
```
utafrali, line 16: Missing a doc comment on `SingleFlightGroup` (and the `Do` method) explaining the ownership contract: all concurrent callers for the same key receive the **same** `[]byte` slice, not independent copies. Any caller that mutates the returned buffer (in-place decryption, decompression, etc.) will silently corrupt the data seen by other concurrent callers. This constraint is non-obvious and must be documented.

weed/filer/reader_cache.go
```diff
@@ -191,25 +197,28 @@ func (s *SingleChunkCacher) startCaching() {
...
-	if err != nil {
-		s.Lock()
-		s.err = fmt.Errorf("operation LookupFileId %s failed, err: %v", s.chunkFileId, err)
-		s.Unlock()
-		return
-	}
+	// Use singleflight to deduplicate concurrent network fetches for the
+	// same chunk. If another goroutine is already fetching this fileId,
+	// we wait for that result instead of issuing a redundant request.
+	data, err := s.parent.fetchGroup.Do(s.chunkFileId, func() ([]byte, error) {
```
utafrali: When two `SingleChunkCacher` instances race for the same `chunkFileId`, the one that loses the `fetchGroup.Do` race receives the same underlying `[]byte` as the winner and stores it in `s.data`. If downstream code ever operates on `s.data` in-place (e.g. cipher XOR, decompression into the same buffer), both cacers' data will be corrupted. If in-place mutation is possible anywhere on the read path, assign a defensive copy here:

```go
s.data = append([]byte(nil), data...)
```

If the buffer is guaranteed read-only, add a comment stating that so future maintainers don't introduce mutations.

weed/util/singleflight_test.go
```diff
@@ -0,0 +1,159 @@
+package util
```
utafrali, line 1: The dedup test checks `callCount == 1` but does not assert that all goroutines received the **same** result value (identical slice content). Adding a content check alongside the call-count check would catch a regression where dedup fires but the shared slice is incorrectly copied or zeroed before some callers read it.

PR #9107, opened 2026-04-16
fix(filer): drop stale master gRPC cache on stream death (#9102)

### LINKED ISSUE #9102: Filer gRPC stream to master becomes stuck with 'context canceled' after master pod restart; pod recreate required
## Version

SeaweedFS 3.95 (image `chrislusf/seaweedfs:3.95`, commit `394e42cd5`).

Deployed on Kubernetes (k3s v1.x, 2 nodes, Linux kernel 6.12). Master, volume servers, and filers are separate StatefulSets. Master is single-replica; filer is replicated with bidirectional `weed filer.sync`. S3 API enabled on the filer (`-s3`).

## Summary

After a master pod restart, existing filer pods keep their established gRPC connection to `seaweedfs-master:9333` (a Kubernetes ClusterIP Service resolving to the new master pod), but request-level streams (`AssignVolume`, `SubscribeToTopic`, etc.) immediately return `rpc error: code = Canceled desc = context canceled` for **every** new request.

Symptoms observed on the filer:

```
masterclient.go:122 Connection wait stopped: context canceled
filer_server_handlers_write.go:49 failing to assign a file id: rpc error: code = Canceled desc = context canceled
s3api_object_handlers_put.go:244 upload to filer error: rpc error: code = Canceled desc = context canceled
```

Volume servers recover transparently on master restart (they re-register). Filers do not. Recreating the filer pod (e.g. `kubectl delete pod seaweedfs-filer-X-0`) clears the state and operation resumes normally.

## Reproduction (approximate)

1. Deploy master + volume(s) + filer(s) + filer-sync as StatefulSets/Deployments (k8s).
2. Induce master restart: `kubectl delete pod seaweedfs-master-0`.
3. New master starts, volume servers re-register, `weed shell` → `volume.list` shows full topology.
4. Attempt any write via filer S3 (`mc cp`) or REST (`PUT /buckets/.../key`).
5. Filer logs show the `context canceled` pattern above; S3 handler returns 500 `InternalError`.
6. Delete the filer pod; new pod replaces it; subsequent uploads succeed.

## Hypothesis

The filer's `masterclient.KeepConnectedToMaster` loop appears to hold the gRPC channel open across master TCP resets because k8s ClusterIP re-routes to the new pod with the same IP:port target. The underlying `ClientConn` may remain `READY`, but the in-flight request context (or the child stream the filer opened at startup) was tied to the previous master process and is permanently canceled. New requests reuse the (stale) channel and inherit the canceled context.

## Impact

- CNPG `barman-cloud-backup` multipart uploads fail 100% until filer is recreated.
- Any high-frequency client (bihaw photo upload at >50 RPS) returns 500 for the same reason.
- Masquerades as an auth/multipart/size issue during triage because the error message doesn't point at master reconnection.

## Workaround

Recreate filer pods after any master restart. In practice, adding a liveness probe tied to a synthetic `AssignVolume` round-trip would self-heal, but none of the stock probe endpoints catch this condition.

## Suggested fix

Either (a) rebuild the gRPC channel to the master on `context canceled` errors from request calls, or (b) expose a filer `/masterconn/healthz` endpoint that does a synthetic `AssignVolume` and returns non-200 when the upstream stream is dead, so k8s liveness probes can trigger pod recreate automatically.

Happy to test patches against this environment if helpful.

chrislusf:
## Summary

Fixes #9102 — after a master pod restart behind a k8s Service VIP (or any other stable L4 endpoint), existing filers keep returning `rpc error: code = Canceled desc = context canceled` on every `AssignVolume` / `LookupVolume` until the filer pod is recreated.

Root cause: `MasterClient.tryConnectToMaster` uses a dedicated streaming channel for `KeepConnected`, but request-path RPCs (`operation.Assign`, etc.) share a cached `*grpc.ClientConn` in `pb.grpcClients`. When the TCP reset is masked by a stable VIP, that cached channel can stay in `READY` while actually being dead, and `Assign` doesn't retry on `Canceled` so the error reaches the client.

Changes:
- `pb.InvalidateGrpcConnection(address)` — new exported helper that closes and removes a cached `ClientConn` by address.
- `wdclient.MasterClient.tryConnectToMaster` — after `KeepConnected` returns, drop the cached request-path channel for that master so unrelated callers dial fresh.
- `operation.Assign` — extend the retry predicate to include `codes.Canceled` / `codes.DeadlineExceeded` while the caller context is still live. The first failure invalidates the stale cached conn via the existing `shouldInvalidateConnection` path; the retry hits a freshly dialed channel.

## Test plan

- [x] `go build ./...`
- [x] `go test ./weed/wdclient ./weed/operation ./weed/pb`
- [x] Docker-compose repro with nginx TCP proxy simulating a stable VIP:
  - Before: `docker pause sw-master` → next filer PUT returns `context canceled` immediately (matches issue).
  - After: same pause → PUT retries for the 30s budget and returns `DeadlineExceeded` with a clearer message; after `docker unpause` (or full container replacement), subsequent PUTs succeed in ~20 ms with no residual stuck state.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved retry behavior for gRPC operations to better handle network interruptions, timeout scenarios, and deadline-related failures.
  * Enhanced automatic recovery mechanism for failed gRPC connections to prevent cascading errors and improve overall operation reliability.

### DIFF
--- a/weed/filer/meta_aggregator.go
+++ b/weed/filer/meta_aggregator.go
@@ -107,6 +107,13 @@ func (ma *MetaAggregator) loopSubscribeToOneFiler(f *Filer, self pb.ServerAddres
 				errLvl = glog.Level(4)
 			}
 			glog.V(errLvl).Infof("subscribing remote %s meta change: %v", peer, err)
+			// The streaming subscribe uses a dedicated connection, but
+			// doSubscribeToOneFiler also issues non-streaming calls
+			// (readFilerStoreSignature, updateOffset, ...) over the shared
+			// cached ClientConn to this peer. When the peer restarts behind
+			// a stable VIP, that cached channel can look healthy yet return
+			// cancelled RPCs; drop it so the next iteration dials fresh.
+			pb.InvalidateGrpcConnection(peer.ToGrpcAddress())
 		}
 		if lastTsNs < nextLastTsNs {
 			lastTsNs = nextLastTsNs
--- a/weed/filer/meta_aggregator.go
+++ b/weed/filer/meta_aggregator.go
@@ -107,13 +107,6 @@ func (ma *MetaAggregator) loopSubscribeToOneFiler(f *Filer, self pb.ServerAddres
 				errLvl = glog.Level(4)
 			}
 			glog.V(errLvl).Infof("subscribing remote %s meta change: %v", peer, err)
-			// The streaming subscribe uses a dedicated connection, but
-			// doSubscribeToOneFiler also issues non-streaming calls
-			// (readFilerStoreSignature, updateOffset, ...) over the shared
-			// cached ClientConn to this peer. When the peer restarts behind
-			// a stable VIP, that cached channel can look healthy yet return
-			// cancelled RPCs; drop it so the next iteration dials fresh.
-			pb.InvalidateGrpcConnection(peer.ToGrpcAddress())
 		}
 		if lastTsNs < nextLastTsNs {
 			lastTsNs = nextLastTsNs
--- a/weed/mq/broker/broker_connect.go
+++ b/weed/mq/broker/broker_connect.go
@@ -23,7 +23,7 @@ func (b *MessageQueueBroker) BrokerConnectToBalancer(brokerBalancer string, stop
 	}
 
 	// connect to the lock owner
-	return pb.WithBrokerGrpcClient(true, brokerBalancer, b.grpcDialOption, func(client mq_pb.SeaweedMessagingClient) error {
+	err := pb.WithBrokerGrpcClient(true, brokerBalancer, b.grpcDialOption, func(client mq_pb.SeaweedMessagingClient) error {
 		stream, err := client.PublisherToPubBalancer(context.Background())
 		if err != nil {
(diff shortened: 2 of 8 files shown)


chrislusf:
Refactored in 3ac66848e: moved the invalidate-cached-non-streaming-peer-conn wrapper into \`pb.WithGrpcClient\`'s streaming branch. Every streaming caller is now covered by default, so the six per-site wrappers (masterclient, meta_aggregator, FollowMetadata, on_each_partition, connect_to_sub_coordinator, broker_connect) are deleted.

- Net: -78 / +37 lines across 7 files, most of it removed duplicated boilerplate + comments.
- Behavior preserved: healthy leader redirect in masterclient returns \`nil\` from \`fn\`, so no invalidation fires (same as the earlier \`gprcErr != nil\` guard).
- Verified against the docker-compose repro: paused master → 30s \`DeadlineExceeded\` (expected retry budget), unpaused → subsequent PUTs succeed in ~20ms.

Unit tests under \`weed/pb\`, \`weed/wdclient\`, \`weed/operation\`, \`weed/filer\`, \`weed/mq/...\` all pass.

PR #9117, opened 2026-04-17
feat(shell): add fs.distributeChunks command for even chunk distribution

kisow:
Add a new weed shell command that redistributes a file's chunks evenly across volume server nodes.

Supports three distribution modes via -mode flag:
- primary:     balance chunk ownership across nodes (default)
- replica:     balance both ownership and replica copies
- round-robin: assign chunks by offset order for sequential read
               optimization (chunk[0]->A, chunk[1]->B, chunk[2]->C, ...)

Additional options:
- -nodes=N to target specific number of nodes
- -apply to execute (dry-run by default)

Usage:
  fs.distributeChunks -path=/buckets/file.dat fs.distributeChunks -path=/buckets/file.dat -mode=round-robin -apply fs.distributeChunks -path=/buckets/file.dat -mode=replica -apply fs.distributeChunks -path=/buckets/file.dat -nodes=5 -apply

스크린샷 2026-04-17 오전 10 21 48

We deploy large-scale LLM models across hundreds of GPU servers using `weed mount`. A key architectural advantage of SeaweedFS in this scenario is its **full-mesh connectivity**: each `weed mount` client can directly connect to and receive chunks from all volume servers simultaneously, without routing through an intermediate proxy or gateway node. This is a significant advantage over S3-based distributed file systems, where reads are funneled through a relay layer.

To fully leverage this architecture, when multiple `weed mount` clients concurrently read the same large model file (e.g., during model loading across hundreds of GPU servers), the reads must be able to **fan out across all volume server nodes**. This requires the file's chunks to be distributed evenly across all nodes.

However, when writing large files via `weed mount`, the master server selects volumes using a random pick (`rand.IntN(len(writables))`) among writable volumes. In practice, this can result in chunks for a single file being concentrated on only a few nodes — leaving the full-mesh read throughput largely unrealized.

#### `primary` (default)
Balances chunk count per node using topology-based ownership mapping. For each chunk's volume, the owning node is determined as `volume_nodes[vid % len(volume_nodes)]` where `volume_nodes` is the sorted list of nodes holding that volume (= replication copy count). Chunks are moved from nodes exceeding the ideal count (`totalChunks / totalNodes`) to nodes below it.

**Limitation**: With replication, the computed owner may not match the actual `Assign` target node, since new volumes receive IDs that may hash to a different node. Results are approximate in replicated environments.

#### `replica`
Performs primary-mode balancing (Step 1), then additionally moves chunks to balance the total copies-per-node count including replicas (Step 2). The copies target is `totalCopies / totalNodes`.

**Limitation**: The replica placement of each new volume is determined by the master at volume grow time based on replication policy — it cannot be specified externally. Therefore perfect convergence is not guaranteed in a single pass; multiple runs converge progressively.

#### `round-robin`
Sorts chunks by file offset and assigns them to nodes in round-robin order: `chunk[i] → nodeList[i % N]`. This ensures consecutive chunks are placed on different nodes, enabling I/O pipelining during sequential reads.

This mode does not depend on ownership calculation. The `Assign(DataNode=X)` call places the primary on the intended node regardless of how the ownership mapping resolves afterward.

**Note**: As with all modes, replica placement is controlled by the master at volume grow time and cannot be specified externally. The read-path may use any replica, so the exact A→B→C pipeline pattern depends on client behavior.

### Implementation

Each redistribution operation per chunk follows this sequence:

1. **Download**: `HTTP GET http://{source_volume_server}/{old_fid}`
2. **Assign**: `gRPC Assign(DataNode={target_node})` to get a new FID on the target node
3. **Upload**: `HTTP PUT http://{target_volume_server}/{new_fid}`
4. **Update metadata**: `gRPC filer_pb.UpdateEntry` with the updated chunk FID list
5. **Delete original**: `HTTP DELETE http://{source_volume_server}/{old_fid}`
   - Only sent to one volume server; `ReplicatedDelete` propagates to all replicas automatically

The delete is sent to a single volume server because `ReplicatedDelete` in `topology/store_replicate.go` calls `LookupVolumeId` to find all replica locations and propagates the deletion to each one. Sending DELETE to all replicas individually would cause 404 errors on the replicas that have already been deleted by propagation.

Filer metadata is updated via gRPC `UpdateEntry` after all chunk copies succeed. If the metadata update fails, the original chunks remain intact and the newly uploaded copies are orphaned (no data loss). Old chunks are deleted only after metadata update succeeds.

### Patterns Reused from Existing Commands

| Pattern | Source command |
|---------|---------------|
| `moveChunk`: download → upload → update FID | `fs.mergeVolumes` |
| `filer_pb.UpdateEntry` for metadata update | `fs.meta.changeVolumeId` |
| `filer_pb.LookupEntry` for file entry lookup | `fs.cat` |
| `collectTopologyInfo` for cluster topology | `ec_common.go` |
| Single-server DELETE with `ReplicatedDelete` propagation | `volume_server_handlers_write.go` |

# How is the PR tested?

Tested on a 3-node dev cluster (replication `001`, 2 copies per volume):

| File | Size | Mode | Chunks moved | md5 before | md5 after | Result |
|------|------|------|-------------|------------|-----------|--------|
| `cs8_1G` | 1 GB | `primary` | 14 | `379efd6f...` | `379efd6f...` | ✓ |
| `cs8_1G` | 1 GB | `replica` | 8 | `379efd6f...` | `379efd6f...` | ✓ |
| `cs8_1G` | 1 GB | `round-robin` | 86 | `379efd6f...` | `379efd6f...` | ✓ |
| `test_10g.dat` | 10 GB | `round-robin` | 1703 | `3875bfd3...` | `3875bfd3...` | ✓ |

## Limitations

- **Chunk manifests not supported**: Files stored as chunk manifests (very large files using manifest indirection) are not yet handled and will return an error.
- **Replica placement is not controllable**: All three modes can only specify the primary node via `Assign(DataNode=X)`. The replica node is determined by the master at volume grow time based on replication policy. Redistribution is best-effort for replica distribution.
- **`primary` and `replica` modes use approximate ownership**: The `vid % len(volume_nodes)` ownership mapping is a heuristic and may not reflect the actual `Assign` target, causing the post-redistribution view to appear imbalanced. 

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added a CLI command to analyze and redistribute file chunks across volume servers. Validates inputs, rejects directory targets, flattens manifests, handles legacy IDs, enforces cluster/node constraints, and supports dry-run vs apply. Three strategies: primary, replica, round-robin. Shows assignment preview, reports when no moves are needed, and performs concurrent moves when applied (with metadata update and warning on failure).

* **Tests**
  * Added tests for node selection, distribution counting, planning for all three strategies, and output/summary verification.


chrislusf:
Nice idea! Thanks! How large are the LLM model files usually? In your case, how many volume servers? what is the chunk size used? all nvme disks? planning to use RDMA?

kisow:
Model sizes and individual weight file sizes vary significantly, making it difficult to generalize. For example:

GLM-5.1: total size ~1.5 TB (each shard ~5.xx GB)
Qwen3.5-397B-A17B: total size ~807 GB (each shard ~8.xx GB)
Qwen3.5-0.8B: total size ~1.77 GB

kisow:
Initially, we operated with 50 volume servers, but we occasionally experienced chunk loss and found live updates of SeaweedFS to be operationally burdensome. As a result, we split the setup into **two clusters of 25 volume servers each** and configured them in an active-passive mode using `filer.sync`.

From our experiments, smaller chunk sizes perform better for cold reads. However, if the chunk size is too small, the number of chunks increases significantly, putting pressure on the metadata database. We are currently using a **16 MB chunk size**. (We plan to raise a separate issue later—there is a problem when creating a 100 GB file with `chunksize=8MB`.)

Due to infrastructure constraints, all volume servers are **HDD-based**. While it would be ideal to use technologies like InfiniBand, RDMA, or high-performance setups, infrastructure is outside our control as we primarily handle software engineering. In practice, **10GbE** is the best we can use.
For this reason, we chose SeaweedFS, as its architecture allows a full-mesh–like data access pattern.

chrislusf:
For chunk manifests — could we resolve the manifest first and redistribute the underlying data chunks? Large files are exactly the ones that would benefit most from this command, so it feels odd to bail out on them. Happy to help in a follow-up if it's out of scope here.

chrislusf:
@kisow merged! Thanks!

Also, added a P2P chunk loading feature among "weed mount" instances. See https://github.com/seaweedfs/seaweedfs/wiki/P2P-reading-in-weed-mount

Let me know whether it can help for this use case.

kisow:
@chrislusf Thanks!

I just got back from my summer vacation today and am checking this now. The weed mount P2P feature sounds like a great capability.

However, in our case, when reading files, we’re currently hitting the 10GbE NIC bandwidth limit on the weed mount side before the volume server becomes a bottleneck, so the P2P approach may not provide immediate benefits for us.

At the moment, our priority—on top of model serving and deployment—is to ensure stable storage and sharing for other internal platforms as well, including data like .git repositories, logs, and monitoring data. So for now, we have higher expectations for full POSIX compliance.

Issue #8928, opened 2026-04-05
volume server: Vacuum fails with "cannot hydrate needle from file: EOF"


madalee-com:
Not sure how long this has been going on.  I've seen it before, thought it was better, but it's back.
Currently on 4.18:innovative-pitcher

It seems like most or all volumes are failing to vacuum with error: "cannot hydrate needle from file: EOF"

I'm thinking it might be related to the fact that these volumes were originally 1GB volumes, and I changed the configs to 10GB.

Log except at v2
```
Apr 04 20:36:16 us00 seaweedfs-volume-ext1401[1636934]: E0404 20:36:16.232811 needle_read.go:46 /mnt/backing/ext1401/9778.dat read 0 dataSize 71808 offset 639982616 fileSize 324180408: EOF
Apr 04 20:36:20 us00 seaweedfs-volume-ext1401[1636934]: I0404 20:36:20.276147 store_vacuum.go:76 volume 32473 compaction space check: volume=426007792, index=15856, space_needed=426023648, free_space=2239876333568
Apr 04 20:36:20 us00 seaweedfs-volume-ext1401[1636934]: E0404 20:36:20.992546 needle_read.go:46 /mnt/backing/ext1401/32473.dat read 0 dataSize 153512 offset 565338632 fileSize 426007792: EOF
Apr 04 20:36:23 us00 seaweedfs-volume-ext1401[1636934]: E0404 20:36:23.653030 volume_grpc_vacuum.go:69 failed compact volume 9778: cannot hydrate needle from file: EOF
Apr 04 20:36:25 us00 seaweedfs-volume-ext1401[1636934]: I0404 20:36:25.033967 store_vacuum.go:76 volume 9729 compaction space check: volume=746237800, index=14000, space_needed=746251800, free_space=2239876333568
Apr 04 20:36:25 us00 seaweedfs-volume-ext1401[1636934]: E0404 20:36:25.212294 needle_read.go:46 /mnt/backing/ext1401/9729.dat read 0 dataSize 2097248 offset 746237800 fileSize 746237800: EOF
Apr 04 20:36:32 us00 seaweedfs-volume-ext1401[1636934]: E0404 20:36:32.178352 volume_grpc_vacuum.go:69 failed compact volume 9729: cannot hydrate needle from file: EOF
[...]

PR #9118, opened 2026-04-17
fix(volume): tolerate v4 off-by-one in startup structural .idx check

chrislusf:
## Summary

Follow-up to #9115. The integrated `MaxNeedleEnd` check at volume load fires whenever the `.idx` claims `(offset + actual size)` past the `.dat` end. Enterprise has a v4→v3 compaction bug papered over by an off-by-one tolerance in `readNeedleTail` / `ReadBytes` that can inflate `MaxNeedleEnd` by up to one needle padding cycle (8 bytes) — small enough to be a known recoverable case, but enough to trip the structural check and mark affected volumes read-only.

Allow `NeedlePaddingSize` bytes of slack so the structural check matches the existing tolerance. Real corruption from #8928 overshoots by megabytes, not bytes, so the check still catches it.

## Test plan

- [x] `go test ./weed/storage/...` — green
- [x] Confirmed downstream enterprise tests (`TestOffByOneToleranceVolumeReload`) pass with the same change

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined volume loading structural validation with adjusted tolerance thresholds to prevent unnecessary read-only status assignment while maintaining data integrity during storage operations.

### DIFF
--- a/weed/storage/volume_loading.go
+++ b/weed/storage/volume_loading.go
@@ -248,9 +248,15 @@ func (v *Volume) load(alsoLoadIndex bool, createDatIfMissing bool, needleMapKind
 		// disk I/O. A violation marks the volume read-only so a corrupt
 		// .idx left over from a crashed batched write does not silently
 		// power vacuum to drop reachable data. See issue #8928.
+		//
+		// Allow a single needle-padding cycle (NeedlePaddingSize) of slack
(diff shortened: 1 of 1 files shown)


chrislusf:
Not needed for OSS — the off-by-one tolerance only exists for enterprise's v4→v3 compaction bug. OSS stays with the strict structural check. Closing.

PR #9197, opened 2026-04-23
fix(admin): use basePath for API fetches when urlPrefix is set

faspix:
Many client-side fetch('/api/...') calls still used root-relative URLs, so the browser requested /api/... on the site origin instead of {prefix}/api/.... Those requests missed the admin app (or returned HTML), which showed up as failures such as “Failed to load policies” / JSON parse errors on the Object Store Users page and broke create/update/delete flows on Buckets, IAM Policies, Groups, and Service Accounts when a URL prefix was configured.

Route all affected fetch() URLs through basePath(), the same helper defined in admin.js as (window.__BASE_PATH__ || '') + path.

# How is the PR tested?
1. Build and run seaweedfs without prefix: smoke-test Users, Buckets, Policies, Groups, Service accounts - modals and actions still work

2. Run with a prefix -urlPrefix=/seaweedfs (and proxy so the UI is served under that path). Open the same pages and repeat actions (load Users page, policies dropdown, create/edit bucket, policy view/edit, groups, service accounts)

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin pages now send API requests through the configured base path, fixing routing for groups, policies, S3 buckets, object-store users, topics, and service-account management.
* **Chores**
  * Removed an obsolete page-level script from the Groups page to streamline frontend loading.


faspix:
Removed the extra <script src="/static/js/iam-utils.js"> from the Groups page because layout.templ already loads iam-utils.js with the URL prefix, and the duplicate used a root-only path that 404s under -urlPrefix

PR #9223, opened 2026-04-25
fix(volume): cap leveldb OpenFilesCacheCapacity per index DB (#9139)

chrislusf:
## Summary

- Reported via [#9139 (comment)](https://github.com/seaweedfs/seaweedfs/issues/9139#issuecomment-4319740414): on a busy volume server (3 processes × `-max=3785`), the user's mount log floods with `failed to write leveldb: open .../000006.log: too many open files` — even with a generous ulimit.
- Each leveldb index opens with `opt.Options{...}` that never set `OpenFilesCacheCapacity`, so goleveldb defaults to **500 FDs per DB**. Across thousands of volumes, that ceiling stacks up and starves WAL rotation.
- Fix: cap per-DB `OpenFilesCacheCapacity` at **16 / 32 / 64** for `NeedleMapLevelDb` / `Medium` / `Large`. `CompactionTableSizeMultiplier=10` already keeps SST counts low, so the smaller cache is sufficient and per-DB FD usage is bounded.

## Test plan

- [x] `go build ./weed/storage/...`
- [x] `go test ./weed/storage/` passes

## Summary by CodeRabbit

* **Chores**
  * Improved database performance and resource efficiency through optimized caching behavior across different volume sizes.

### DIFF
--- a/weed/storage/volume_loading.go
+++ b/weed/storage/volume_loading.go
@@ -199,6 +199,12 @@ func (v *Volume) load(alsoLoadIndex bool, createDatIfMissing bool, needleMapKind
 					BlockCacheCapacity:            2 * 1024 * 1024, // default value is 8MiB
 					WriteBuffer:                   1 * 1024 * 1024, // default value is 4MiB
 					CompactionTableSizeMultiplier: 10,              // default value is 1
+					// goleveldb defaults this cache to 500 FDs per DB; with
+					// thousands of volumes per server that ceiling stacks up
+					// and starves WAL rotation (open .../00000N.log: too many
+					// open files). CompactionTableSizeMultiplier=10 already
+					// keeps the SST count low, so a small cache is sufficient.
+					OpenFilesCacheCapacity: 16,
 				}
 				if v.tmpNm != nil {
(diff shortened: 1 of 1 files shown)


chrislusf:
Thanks for the cross-cutting suggestion. Holding scope here intentionally — the reported symptom (#9139) is volume-server specific because that process hosts **one leveldb per volume**, so the per-DB default of 500 stacks across thousands of DBs and exhausts FDs. The other call sites you flagged don't stack:

- `weed/filer/leveldb{,2,3}_store.go` — one (or a small fixed number of) DB per filer process; capping would hurt filer metadata throughput without solving any FD problem.
- `weed/util/chunk_cache/chunk_cache_on_disk.go` — one DB per cache instance.
- `weed/storage/needle_map_leveldb.go` `generateLevelDbFile` / `DoOffsetLoading` — short-lived (open → write/walk → `Close()`); FDs don't accumulate.
- `weed/storage/needle_map/memdb.go` — in-memory backend.

`UpdateNeedleMap` (line 298) already inherits the capped opts from `volume_loading.go` via its `opts *opt.Options` parameter, so the volume path is fully covered.

PR #9256, opened 2026-04-28
fix(admin): use protocol-relative URLs for component links so HTTPS clusters don't break clicks

pmiriyev:

The admin UI hardcodes `http://` when constructing browser links to other cluster components (master, volume server, filer, EC shard server, Iceberg REST catalog). When the cluster runs HTTPS-only — i.e., `security.toml` has `[https.master]`, `[https.volume]`, `[https.filer]` configured — clicking any of these links sends a plain HTTP request to a TLS-only endpoint:

- Master receives `client sent an HTTP request to an HTTPS server` and returns HTTP 400 in the browser.
- Volume / Filer return the same.

So the clickable "Open Master UI" / "Open Volume Server UI" / etc. buttons silently fail in any HTTPS deployment.

The same pattern caused the issue fixed in #9227 (admin's own backend `/dir/status` fetch). This PR is the browser-facing equivalent.

## Reproduction

1. Deploy SeaweedFS with `security.toml`:
   ```toml
   [https.master]
   cert = "/etc/seaweedfs/ssl/master.crt"
   key  = "/etc/seaweedfs/ssl/master.key"
   ca   = "/etc/seaweedfs/ssl/ca.crt"
   [https.volume]
   ...
   [https.filer]
   ...
   ```
2. Open admin UI (`https://admin:23646/`) and navigate to **Cluster → Masters / Volume Servers / Filers**.
3. Click any component link.
4. Browser sends `http://master:9333/ui/index.html` → master returns 400.

Replace hardcoded `http://` with **protocol-relative URLs** (`//host:port/path`) in admin UI templates. The browser uses the **same scheme as the page hosting the link** — so when admin UI is loaded over HTTPS, browser-initiated clicks use HTTPS too. When admin is loaded over plain HTTP (existing default deployments), behavior is unchanged.

This is a minimal, fully backwards-compatible change:
- HTTPS-only deployments: links now work.
- HTTP-only deployments: identical behavior to before.
- Mixed deployments (e.g., admin HTTPS but volume HTTP): edge case — acknowledged limitation, addressed properly by future work on per-component public-URL configuration.

Affected templates (9 files), each accompanied by its generated `_templ.go` (kept in lockstep so `templ generate` is not required for reviewers):

- `weed/admin/view/app/admin.templ` (master, volume, filer links on dashboard)
- `weed/admin/view/app/cluster_filers.templ`
- `weed/admin/view/app/cluster_masters.templ` (Go templ + JS modal)
- `weed/admin/view/app/cluster_volume_servers.templ` (Go templ + JS modal)
- `weed/admin/view/app/cluster_volumes.templ`
- `weed/admin/view/app/ec_volume_details.templ`
- `weed/admin/view/app/volume_details.templ`
- `weed/admin/view/app/iceberg_catalog.templ`
- `weed/admin/view/app/s3tables_buckets.templ`

# How is the PR tested?

- `go build ./weed/admin/...` — clean.
- Verified against SeaweedFS 4.22 with HTTPS-only master/volume/filer:
  - **Before**: clicking "Open Master UI" in admin → browser shows `400 Bad Request: client sent an HTTP request to an HTTPS server`.
  - **After**: same click → browser uses HTTPS, hits master's HTTPS UI cleanly.
- Verified against HTTP-only deployment: behavior unchanged (browser fills in `http://` to match admin page scheme).

# Checks
- [x] The PR is kept as minimum as possible — single-character substitution (`http:` → empty) in 17 link constructions across 9 templates + their generated `_templ.go` siblings (32 lines total).
- [ ] I have added unit tests if possible. *(The change is a single-character template substitution; the affected code paths are pure URL string construction with no business logic. Verified end-to-end in the browser against both HTTPS and HTTP deployments. Happy to add a test if requested.)*

## Notes on follow-up

This PR fixes the **scheme** for the dominant deployment patterns (homogeneous all-HTTPS or all-HTTP). Two further deployment patterns remain unsupported and warrant separate work — happy to follow up if the maintainer agrees on direction:

- **Per-component public URL** for filer/master/broker (volume already has `-publicUrl`, but no equivalent on filer/master). Would let admin UI link operators to a publicly-reachable host:port instead of the internal one.
- **Operator-side URL templates** (`-volume.publicUrl.template=https://volumes-proxy.example.com/{volumeId}`) for path-routed reverse proxies and Kubernetes Ingress, where the `host:port` model itself doesn't fit.

Both can be additive: today's defaults (use `host:port` from heartbeat / `-publicUrl`) stay correct.

## Summary by CodeRabbit

* **Bug Fixes**
  * Admin dashboard external links now use protocol-relative URLs so Master, Volume Server, Filer, EC volume and API links follow the current page protocol (http/https).
* **UI**
  * Connection examples and REST endpoint snippets now reflect the current protocol dynamically, ensuring displayed endpoints match how the admin interface was loaded.


pmiriyev:
Thanks @gemini-code-assist — both flagged JS overrides fixed in [a6b6952](https://github.com/seaweedfs/seaweedfs/pull/9256/commits/a6b695274). The post-render `href = \`http://...\`` template literals in iceberg_catalog.templ and s3tables_buckets.templ now use the same protocol-relative form as the templ side, so the browser scheme is preserved end-to-end. Both .templ source and the generated _templ.go updated in lockstep.

pmiriyev:
Thanks @coderabbitai — addressed in [8388194](https://github.com/seaweedfs/seaweedfs/pull/9256/commits/83881942a).

The displayed REST endpoint text (`<code>http://host:port/v1</code>`) in both `iceberg_catalog.templ` and `s3tables_buckets.templ` now wraps the scheme in a `<span id="...-protocol">http</span>` and the existing inline script sets its `innerText` from `window.location.protocol`. Mirrors the dynamic-host pattern that was already there.

The SQL/JSON code-block examples (`CREATE EXTERNAL TABLE ... ENDPOINT 'http://...'`, `\"uri\": \"http://...\"`) are intentionally left as-is — they're starter snippets the user adapts, not copy-paste-into-runtime. Happy to follow up with server-side scheme threading if you want.

PR #9260, opened 2026-04-28
filer(mysql): TLS hostname/SNI knobs + MariaDB upsert documentation

chrislusf:
## Summary

Two filer-store fixes that come up when pointing the `[mysql]` filer at a managed MariaDB / MySQL cluster that requires SSL:

- **TLS verification flexibility.** Add `tls_insecure_skip_verify` and `tls_server_name` config knobs to the `[mysql]` block. The driver previously offered no way to override SNI / cert hostname, so any cluster whose server cert SAN didn't match the connection `hostname` (very common with internal LB hostnames or self-signed cluster certs) failed the handshake with no recourse short of reissuing the cert.
- **CA optional + fail-loud client cert load.** `ca_crt` is now only required when verification is on. Conversely, if `client_crt` / `client_key` are set but unreadable, the store now fails loudly instead of silently falling back to non-mTLS — which previously produced confusing handshake errors that looked like server-side problems.
- **MariaDB upsert documentation.** The default `upsertQuery` (introduced in #4096 to silence MySQL 8.0+ deprecation warnings) uses the `AS new` row alias, which MariaDB does not support and which causes `ERROR 1064 (42000)` on the first write — frequently misdiagnosed as a connection issue. Added a ready-to-paste MariaDB-compatible override in scaffold/filer.toml. Default unchanged so existing MySQL 8 deployments are not affected.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./weed/filer/mysql/...` clean
- [ ] Manual: connect filer to a MariaDB cluster with SSL using new `tls_server_name` override and confirm handshake succeeds
- [ ] Manual: confirm `client_crt` set to a bogus path now errors instead of silently proceeding

## Summary by CodeRabbit

* **New Features**
  * Added tls_insecure_skip_verify and tls_server_name options for MySQL TLS.
  * Added a commented alternate upsertQuery example for MariaDB and MySQL 5.7.

* **Improvements**
  * Enforced minimum TLS 1.2 for MySQL connections.
  * Stronger TLS validation: client cert/key must both load if provided; CA use is optional.

* **Documentation**
  * Clarified PEM path semantics and mTLS pairing requirements in MySQL TLS docs.

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -58,9 +58,11 @@ enabled = false
 # [username[:password]@][protocol[(address)]]/dbname[?param1=value1&...&paramN=valueN]
 dsn = "root@tcp(localhost:3306)/seaweedfs?collation=utf8mb4_bin"
 enable_tls = false
-ca_crt = ""                # ca.crt dir when enable_tls set true
-client_crt = ""            # mysql client.crt dir when enable_tls set true
-client_key = ""            # mysql client.key dir when enable_tls set true
+ca_crt = ""                       # path to CA cert (PEM) — optional; if empty, the system trust store is used
+client_crt = ""                   # path to client cert (PEM) — only when server requires mTLS; must be set together with client_key
+client_key = ""                   # path to client key (PEM)  — only when server requires mTLS; must be set together with client_crt
+tls_insecure_skip_verify = false  # skip server cert verification (use only for testing or with self-signed certs)
+tls_server_name = ""              # override SNI / cert hostname; leave empty to use `hostname` above
 hostname = "localhost"
 port = 3306
 username = "root"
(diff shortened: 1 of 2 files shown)


chrislusf:
## End-to-end verified against real MariaDB

Built the patched binary, started **MariaDB 11.8** in Docker with `require_secure_transport=ON` and a server cert whose only SAN is `mariadb-real-san` (deliberately *not* matching the connection address), then exercised `MysqlStore.Initialize` directly.

| # | Scenario | Expected | Result |
|---|----------|----------|--------|
| 1 | Connect to `127.0.0.1`, no override | TLS verify fails (SAN mismatch) | `tls: failed to verify certificate: x509: cannot validate certificate for 127.0.0.1 because it doesn't contain any IP SANs` |
| 2 | Same + `tls_server_name = "mariadb-real-san"` | Connects | `Initialize err=<nil>` |
| 3 | Same + `tls_insecure_skip_verify = true` | Connects | `Initialize err=<nil>` |
| 4 | `enable_tls=true`, empty `ca_crt`, no skip | Loud config error | `mysql tls enabled but ca_crt is empty; set ca_crt or tls_insecure_skip_verify=true` |
| 5 | Bogus `client_crt`/`client_key` paths | Loud error (not silent fallback) | `load mysql client keypair (...): no such file or directory` |
| 6 | MariaDB-friendly upsertQuery (`VALUES(meta)` form) | Two upserts collide and the row stores latest | `meta="v2"` |
| 7 | Default upsertQuery (`AS new` form) on MariaDB | Fails with syntax error (motivates the doc note) | `Error 1064 (42000): ... near 'AS \`new\` ON DUPLICATE KEY UPDATE ...'` |

Scenario 1 reproduces the original handshake failure; scenarios 2 and 3 confirm both new escape hatches work; 4 and 5 validate the new fail-loud paths; 6 and 7 confirm the documented MariaDB upsert override is necessary and correct.

Test scaffolding was outside the repo (no test artifacts added to the PR); test environment torn down.

PR #9262, opened 2026-04-28
fix(nfs): reject NFSv4 calls with PROG_MISMATCH so clients fall back to v3

chrislusf:
## Summary

`mount -t nfs <weed-nfs-host>:<export> /mnt` against a `weed nfs` deployment fails on Linux clients with:

> mount.nfs: requested NFS version or transport protocol is not supported

…because the upstream `willscott/go-nfs` library dispatches RPC calls by `(program, procedure)` only — it never validates the program version. An NFSv4 probe (`prog=100003 vers=4 proc=1 COMPOUND`) lands on the same handler map as v3 and gets routed to v3 `SETATTR`, which parses the COMPOUND args as `SETATTR3args` and writes a malformed reply. The kernel returns `EPROTONOSUPPORT` and `mount.nfs` never retries v3.

This PR adds a thin listener wrapper that peeks the first RPC frame on each new TCP connection and, if it's `prog=NFS|MOUNT` and `vers != 3`, writes a protocol-correct `PROG_MISMATCH` reply (supported range `3..3`, per RFC 5531) directly to the socket and closes the connection. v3 frames are replayed unchanged via a `bufio.Reader`, so go-nfs sees the original bytes.

This makes `weed nfs` behave the same way `rclone serve nfs` (which uses the same go-nfs library) effectively does for users who pin `nfsvers=3` in their mountOptions: the v4-first fallback path now produces a clean rejection instead of a corrupted reply, and Linux clients fall back to v3 on their own.

## Verified with `tcpdump` against a real Linux client

**Before:**
```
CALL  prog=100003 vers=4 proc=1
REPLY accept_stat=SUCCESS   # garbage body — v3 SETATTR encoding of a v4 COMPOUND request
```
mount.nfs reports `Input/output error` and gives up.

**After:**
```
CALL  prog=100003 vers=4 proc=0
REPLY PROG_MISMATCH (low=3 high=3)
```
mount.nfs reports `Protocol not supported` for v4 and falls back to v3, which mounts cleanly via portmap.

## Commits (split for review)

1. **`feat(nfs): add NFSv3-only RPC version filter`** — new `rpc_version_filter.go` with the listener/conn wrapper, plus 4 unit tests (NFSv4 reject, MOUNTv4 reject, NFSv3 pass-through, unknown-program pass-through). The filter is not yet wired in.
2. **`fix(nfs): wire NFSv3 version filter into the listener chain`** — one-line activation in `server.serve()`, placed after the existing client-allowlist wrapper so unauthorized peers are still rejected by IP/CIDR before we inspect RPC content.

## Test plan

- [x] `go test ./weed/server/nfs/ -count=1 -timeout=180s` (all existing + new tests pass)
- [x] `go vet ./weed/server/nfs/` clean
- [x] End-to-end: `weed nfs -portmap.bind=0.0.0.0` in a container; mount from another container with `nfs-utils`. Verified `mount -t nfs -o nfsvers=3,nolock <host>:<export> /mnt` succeeds, and explicit `vers=4` is now rejected with `Protocol not supported` instead of hanging on a malformed reply.
- [x] `tcpdump`-decoded RPC trace confirms PROG_MISMATCH reply with `low=3 high=3`.

## Summary by CodeRabbit

* **New Features**
  * Incoming NFS TCP connections are now peek-inspected and validated for RPC version; unsupported NFS or MOUNT version probes receive a protocol-shaped rejection and the connection is closed so clients can retry with a compatible version (NFSv3 supported).

* **Tests**
  * Added comprehensive tests for version rejection, pass-through for supported/unknown programs, handling of short/fragmented frames, transient accept errors, and avoidance of head-of-line blocking.

### DIFF
--- a/weed/server/nfs/rpc_version_filter.go
+++ b/weed/server/nfs/rpc_version_filter.go
@@ -255,6 +255,16 @@ func filterFirstRPCFrame(conn net.Conn) (net.Conn, bool) {
 		// Not a CALL — odd, but pass through.
 		return &peekedConn{Conn: conn, reader: r}, true
 	}
+	if rpcVers := binary.BigEndian.Uint32(hdr[12:16]); rpcVers != 2 {
+		// ONC RPC v2 is the only version we and go-nfs speak; if the
+		// rpcvers field is anything else the rest of the header is
+		// untrusted (could be a non-RPC protocol that happens to share
+		// the port, or simply garbled traffic). Don't synthesize a
+		// PROG_MISMATCH that lies about supporting NFS — pass it
+		// through and let go-nfs / RFC 5531 §9 RPC_MISMATCH handling
+		// in the upstream library do the right thing.
+		return &peekedConn{Conn: conn, reader: r}, true
+	}
 
 	prog := binary.BigEndian.Uint32(hdr[16:20])
 	vers := binary.BigEndian.Uint32(hdr[20:24])
--- a/weed/server/nfs/rpc_version_filter.go
+++ b/weed/server/nfs/rpc_version_filter.go
@@ -1,0 +1,181 @@
+package nfs
+
+import (
+	"bufio"
+	"encoding/binary"
+	"io"
+	"net"
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/glog"
+)
+
+// The upstream willscott/go-nfs library dispatches RPC calls by (program,
+// procedure) only — it does not validate the RPC program version. That means
+// a Linux client speaking NFSv4 (program 100003 vers 4) lands on the same
+// handler map as NFSv3: proc=1 routes to NFSv3 SETATTR, which parses the
+// NFSv4 COMPOUND args as if they were SETATTR3args and writes a malformed
+// reply. The client cannot decode that reply, the kernel returns
+// EPROTONOSUPPORT, and mount.nfs prints "requested NFS version or transport
+// protocol is not supported" without ever falling back to v3.
+//
+// The default Linux mount.nfs path is to try NFSv4 first, so this affects
+// every plain `mount -t nfs <host>:<export> /mnt` against a `weed nfs`
+// deployment. To make the v4→v3 fallback work, we wrap the listener so the
+// first RPC frame on each new TCP connection is inspected: if the program is
+// NFS or MOUNT and the version is not 3, we synthesize a PROG_MISMATCH reply
+// (with the supported version range 3..3) directly to the socket and close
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed all four review comments — pushed three commits, each scoped to one concern:

**1. `test(nfs): exercise MOUNT v4 in the v4-rejection test, not v1`** (ce9132a94) — coderabbit minor.
`TestVersionFilterRejectsMOUNTv4WithProgMismatch` was sending version `1`. Bumped to `4` so the case the name asserts is what's actually exercised.

**2. `refactor(nfs): reuse package RPC constants and io.ReadFull in version filter`** (60f3c7f2b) — gemini medium.
The RPC numeric constants (`rpcMsgCall`/`rpcMsgReply`/`rpcMsgAccepted`/`rpcAcceptProgMismatch`/`rpcAuthNone`/`nfsProgram`/`mountProgram`) are already named in `portmap.go` next to the portmap responder. Removed the parallel set in `rpc_version_filter.go` and reuse those instead — one source of truth per package. Filter-only constants (peek timeout, peek length, `supportedNFSVer`) stay local. Also dropped the bespoke `readFull` loop in the test in favour of `io.ReadFull`.

**3. `fix(nfs): move RPC peek off the Accept path`** (530308274) — coderabbit major (head-of-line blocking).
This was a real bug. Restructured `versionFilterListener` so a background goroutine drives the inner `Accept()` loop and hands each raw conn to its own peek goroutine. Validated conns flow through a channel that `Accept()` reads from; rejected conns finish their `PROG_MISMATCH` reply and disappear without ever reaching the channel. The wrapper now also implements `Close()`/`Addr()` directly with a `sync.WaitGroup` coordinating shutdown.

Added a regression test (`TestVersionFilterDoesNotHeadOfLineBlockOnSlowConn`) that opens a slow conn (TCP-only, never writes) and a fast conn (sends a v3 frame) and asserts the fast conn reaches the inner accept handler in under 2s — well below the 10s peek timeout that would expose a regression to the old serial behaviour.

Verified:
- `go test ./weed/server/nfs/ -count=1 -timeout=180s` (all tests including the new HOL regression)
- `go test ./weed/server/nfs/ -count=1 -race -run TestVersionFilter` (race-clean)
- `go vet ./weed/server/nfs/` clean
- End-to-end: rebuilt the binary, ran `weed nfs -portmap.bind=0.0.0.0` in a container, verified `mount -t nfs -o nfsvers=3,nolock` succeeds and `mount -t nfs -o vers=4` is rejected with `Protocol not supported` (PROG_MISMATCH visible in server logs).

PR #9269, opened 2026-04-28
fix(s3api): clear stale latest-version pointer when .versions dir cleanup is blocked

chrislusf:
## Summary

When a versioned object's `.versions` directory contains an orphan (a `v_<id>` file that lacks the `Seaweed-X-Amz-Version-Id` extended attribute and so isn't classified as a version), two cleanup paths leave the directory's `LatestVersionId` pointer referencing a now-missing file:

1. `updateLatestVersionAfterDeletion` — after deleting the only remaining tagged version, the scan finds zero version-id-tagged entries; the non-recursive `rm` of `.versions` then fails because the orphan blocks it.
2. `healStaleLatestVersionPointer` — the pointer already references a missing file, and the rescan finds nothing to promote.

Both branches previously left the pointer untouched. Every subsequent read of the object then walks `getLatestObjectVersion`'s 13× retry loop on the missing pointer file *plus* the self-heal rescan, only to return the same 404.

## Why it matters in practice

Veeam's checkpoint flow does a tight GET-PUT-GET-DELETE cycle against a small lock object (`…/Metadata/Lock/create.checkpoint/try`) on every checkpoint:

```
GET try                                        # check current lock state
PUT try                                        # claim the lock (creates v_<new>)
GET try                                        # confirm
DELETE try?versionId=<new>                     # release the lock
  -> updateLatestVersionAfterDeletion
     -> 1 entry remaining (orphan)
     -> "no versions left"
     -> rm .versions  → fails (non-empty)
     -> pointer left as v_<new> (just deleted)
```

Once an orphan exists, the cycle is self-perpetuating: each release leaves the pointer pointing at the version it just removed, and the *next* run's first GET 404s through the slow heal path. We saw this in production logs as `Failed to preprocess target / Storage not found` on every Veeam job — fixed only by manually deleting the `.versions` subtree.

## Fix

Make both paths fall through to a best-effort pointer clear when no recoverable version remains:

- delete `Seaweed-X-Amz-Latest-Version-Id` and `Seaweed-X-Amz-Latest-Version-File-Name` from the `.versions` directory entry's Extended map
- clear cached list metadata so it doesn't carry a stale snapshot
- persist the cleared entry via `mkFile`

The orphan files themselves are left in place (an operator can audit and remove them); from the S3 API perspective the object is now correctly absent and subsequent reads short-circuit to `ErrNotFound` on the fast path.

## Test plan

- [x] `go build ./weed/s3api/...`
- [x] `go test ./weed/s3api/` — existing tests pass
- [ ] Manual: reproduce orphan state on a versioned bucket, verify next GET returns 404 directly without the 13× retry loop in logs.
- [ ] Veeam: confirm `Failed to preprocess target` clears once orphan-blocked pointers are auto-cleared on the next run.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale version metadata in S3 object versioning when all object versions are removed.
  * Prevents orphaned version pointers from causing repeated self-healing attempts, improving stability and performance.
  * Better handling of incomplete or inconsistent version directory entries to maintain a consistent object state.

### DIFF
--- a/weed/s3api/s3api_object_versioning.go
+++ b/weed/s3api/s3api_object_versioning.go
@@ -1132,27 +1132,45 @@ func (s3a *S3ApiServer) updateLatestVersionAfterDeletion(bucket, object string) 
 			return nil
 		}
 		glog.Warningf("updateLatestVersionAfterDeletion: failed to delete .versions directory for %s/%s: %v", bucket, object, err)
-
-		_, hasId := versionsEntry.Extended[s3_constants.ExtLatestVersionIdKey]
-		_, hasFile := versionsEntry.Extended[s3_constants.ExtLatestVersionFileNameKey]
-		if !hasId && !hasFile {
-			return nil
-		}
-		delete(versionsEntry.Extended, s3_constants.ExtLatestVersionIdKey)
-		delete(versionsEntry.Extended, s3_constants.ExtLatestVersionFileNameKey)
-		clearCachedVersionMetadata(versionsEntry.Extended)
-		if mkErr := s3a.mkFile(bucketDir, versionsObjectPath, versionsEntry.Chunks, func(updatedEntry *filer_pb.Entry) {
-			updatedEntry.Extended = versionsEntry.Extended
-			updatedEntry.Attributes = versionsEntry.Attributes
-			updatedEntry.Chunks = versionsEntry.Chunks
-		}); mkErr != nil {
-			glog.Warningf("updateLatestVersionAfterDeletion: failed to clear stale pointer for %s/%s: %v", bucket, object, mkErr)
-		} else {
-			glog.V(1).Infof("updateLatestVersionAfterDeletion: cleared stale latest-version pointer for %s/%s (orphan entries remain in .versions directory)", bucket, object)
-		}
+		s3a.clearStaleLatestVersionPointer(bucket, object, bucketDir, versionsObjectPath, versionsEntry, "updateLatestVersionAfterDeletion")
 	}
 
 	return nil
+}
+
+// clearStaleLatestVersionPointer best-effort clears the latest-version
+// pointer on a .versions directory entry when no recoverable version
+// remains and the directory cannot be removed (e.g. because of orphan
+// entries that lack the version-id extended attribute).
+//
+// Persists with mkFile, snapshotting the caller's versionsEntry.Extended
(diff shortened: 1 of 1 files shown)


chrislusf:
Addressed reviewer feedback in f8dce98:

## Summary of the change

Both inline clear-pointer blocks now go through a new helper `clearStaleLatestVersionPointer` that does the persist CAS-style:

1. **Re-scan** `.versions` for any version-id-tagged entry. If one now exists, abort — a concurrent writer populated the directory and either already promoted, or the next read's self-heal will.
2. **Re-fetch** the live `.versions` directory entry and only proceed if its `Seaweed-X-Amz-Latest-Version-Id` still matches the stale id the caller observed. A concurrent pointer update is detected here.
3. **Persist** with `mkFile` using the *live* `Extended` map (minus the two pointer fields and cached metadata) so any other Extended fields written concurrently between (2) and the persist are preserved.

## Reviewer-by-reviewer

**@coderabbitai (TOCTOU, Critical)** — addressed via the re-scan + re-fetch + match-id pattern in steps 1–2 above. The persist in step 3 still has a millisecond-level window where the filer doesn't expose CAS, but it operates on the freshly-read live `Extended` rather than the caller's stale snapshot, so concurrent updates to *other* Extended fields are preserved. To get a true atomic CAS we'd need a filer-level conditional update, which is a larger change.

**@gemini-code-assist line 1148/1421 (race in mkFile callback, High)** — *not* fixed by mutating `updatedEntry.Extended` inside the callback as suggested, because `mkFile` constructs a fresh `*filer_pb.Entry` (see `weed/pb/filer_pb/filer_client.go:247`) rather than reading the live entry first. The callback's `updatedEntry.Extended` is `nil` at invocation, so a `delete` on it would be a no-op; we'd lose all Extended fields, not just the two we want to clear. The helper instead does an explicit re-fetch before persist, which achieves the same intent (don't overwrite with stale state) and works with how `mkFile` actually behaves.

**@gemini-code-assist line 1426 (refactor duplicate, Medium)** — addressed by extracting both clear blocks into `clearStaleLatestVersionPointer`. The `caller` parameter carries the source-function name so the log lines operators were already grepping (`updateLatestVersionAfterDeletion: cleared stale latest-version pointer …` / `healStaleLatestVersionPointer: cleared stale latest-version pointer …`) keep the same prefix.

`go test ./weed/s3api/` still green.

PR #9279, opened 2026-04-29
 fix(shell): use exact match for volume.balance -racks/-nodes filter

qzhello:
## Summary
Two small fixes in `weed/shell/command_volume_balance.go` surfaced during a review of the volume.* shell commands.

  ### 1. Typo in user-facing error message

  `maybeMoveOneVolume` returned `"does not move volume in remove storage"` when refusing to move a remote-tier volume. Should read `remote storage`.

  ### 2. `-racks` / `-nodes` filter used substring match

  `collectVolumeServersByDcRackNode` filtered with:

  ```go
  if selectedRacks != "" && !strings.Contains(selectedRacks, r.Id) { continue }
```

  strings.Contains(haystack, needle) treats the user's flag value as one big string and asks whether the rack id appears anywhere inside it. That quietly mis-matches whenever one rack id
  is a substring of another listed id.

  Concrete failure cases:
```
  ┌───────────────────┬───────────┬──────────────┬──────────┐
  │ -racks= / -nodes= │ actual id │ old behavior │ expected │
  ├───────────────────┼───────────┼──────────────┼──────────┤
  │ rack10            │ rack1     │ included ❌  │ excluded │
  ├───────────────────┼───────────┼──────────────┼──────────┤
  │ prod-rack-1       │ prod      │ included ❌  │ excluded │
  ├───────────────────┼───────────┼──────────────┼──────────┤
  │ node10            │ node1     │ included ❌  │ excluded │
  └───────────────────┴───────────┴──────────────┴──────────┘
```
  In practice, node ids that include a :port suffix happen to be protected by the colon boundary (e.g. host1:8080 is not a substring of host10:8080), so the bug surfaces more often on the
  rack filter. Bare-id deployments still trip the node filter.

## Fix

  Introduce splitCSVSet(csv) map[string]bool that parses the comma-separated flag into an exact-match set (trims whitespace, treats empty as "no filter"). Both -racks and -nodes now use
  set membership instead of substring search.

  Behavior is preserved for all inputs where no listed id was a prefix/substring of another (the common case).

## Tests

  Added to command_volume_balance_test.go:

  - TestSplitCSVSet — table-driven coverage of the helper: empty input, single value, multi-value, whitespace trimming, empty-item skipping.
  - TestCollectVolumeServersByDcRackNode_RackFilter — regression: -racks=rack10 against a topology containing {rack1, rack10, rack2} selects only rack10; -racks=rack1,rack2 selects exactly
   those two and excludes rack10.
  - TestCollectVolumeServersByDcRackNode_NodeFilter — same regression for -nodes, using bare ids so the substring relationship is real.

  Both filter tests fail on the old strings.Contains implementation and pass after this change.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed rack/node filter handling to require exact ID matches (prevents `rack1` matching `rack10`).
  * Improved CSV list parsing to trim whitespace and ignore empty tokens, making list-based filters more reliable.
  * Updated user-facing error text to say “remote storage” when skipping moves for volumes with remote storage.

* **Tests**
  * Added unit and regression tests validating CSV parsing and exact rack/node filtering behavior.

### CODE REVIEW

File: weed/shell/command_volume_balance.go [hunk 1]
```diff
@@ -215,17 +215,34 @@ func (c *commandVolumeBalance) balanceVolumeServersByDiskType(diskType types.Dis
 	return nil
 }
 
+// splitCSVSet parses a comma-separated list into a set for exact-match filtering.
+// Returns nil when the input is empty so callers can treat nil as "no filter".
+func splitCSVSet(csv string) map[string]bool {
+	if csv == "" {
+		return nil
```
chrislusf: just return empty map. keep the code cleaner.

weed/shell/command_volume_balance.go
```diff
@@ -239,11 +236,11 @@ func collectVolumeServersByDcRackNode(t *master_pb.TopologyInfo, selectedDataCen
 			continue
 		}
 		for _, r := range dc.RackInfos {
-			if rackSet != nil && !rackSet[r.Id] {
+			if len(rackSet) > 0 && !rackSet[r.Id] {
```
chrislusf, line 239: `len(rackSet) > 0` is not needed

qzhello: If we drop the guard, `volume.balance` with no `-racks` flag would `continue` on every rack and do nothing. `splitCSVSet("")` returns an  empty (non-nil) map, and `!rackSet[r.Id]` is always `true` on an empty  map, so every rack gets skipped.

PR #9346, opened 2026-05-07
docs(s3/lifecycle): event-driven redesign

chrislusf:
Design doc for the S3 lifecycle redesign that replaces the synchronous `PutBucketLifecycleConfigurationHandler` walk (currently times out on ~60k-object buckets) with an event-driven worker.

Highlights:
- Cluster-singleton meta-log reader subscribed to one filer; client-side heap merge with per-filer-shard `MessagePosition` cursors. `LogEntry.Offset` is per-buffer, so a single global cursor is incorrect — per-filer cursors are required.
- Bucket-level bootstrap walker with inline delete for currently-due age rules and `SCAN_AT_DATE` for date rules.
- Two-phase rule activation (`bootstrap_complete` + `mode == EVENT_DRIVEN`), atomic per-rule completion, target-mode-aware cursor seeding.
- Blocked-cursor model: deterministic per-event failures pause the failing stream and write a durable `BlockerRecord`; operator resolves via `s3.lifecycle.blockers retry|resume|quarantine`. No automatic dead-lettering — for lifecycle, silently skipping a delete decision is unsafe.
- Durable `retry_budget` file promotes sustained `RETRY_LATER` to `BLOCKED` after a configured threshold; `handleEvent` owns retry-target construction so the failing branch's context is correct.
- Retention mode gate: `metaLogRetention < eventLogHorizon(rule) + bootstrapLookbackMin` → `scan_only` with `degraded_reason = RETENTION_BELOW_HORIZON`. Applies to all reader-driven kinds, not just age-based.
- Lost-log handling: tail-drained shards GC'd safely (durable `tail_drained_streams` marker); shards whose logs vanished before catch-up downgrade affected rules to `scan_only` with `LOST_LOG`. Operator must explicitly `s3.lifecycle.reseed -ack-lost-log` to clear.
- Object lock check via `enforceObjectLockProtections` in the `LifecycleDelete` server. Compliance/legal-hold deferred — `SKIPPED_OBJECT_LOCK` logs and advances; safety-scan re-attempts.

Phased dev plan (Phase 0–8) and a four-layer test plan are at the bottom of the doc.

## Summary by CodeRabbit

* **New Features**
  * S3 lifecycle management: rules to expire objects (by days or fixed dates), handle incomplete multipart uploads, delete markers, and noncurrent versions.
  * Deterministic rule hashing, explicit lifecycle action kinds, and durable state + retry metadata for reliable scheduling and pause/resume.
  * Lifecycle task routing with resume hints, computed due-times, event-horizons and trigger ages for precise processing.

* **Tests**
  * Expanded unit tests covering evaluation, due-time logic, horizons, trigger ages, action kinds, and rule hashing.


chrislusf:
Addressed the four CodeRabbit findings in the latest three commits:

1. **`s3lifecycle/evaluate.go` + `due_at.go`** — non-current delete markers now correctly fall through to `NoncurrentVersionExpirationDays` (the `IsDeleteMarker` switch arm is gated on `IsLatest`). Added regression tests in `evaluate_test.go` and `due_at_test.go`. Real bug.
2. **`s3lifecycle/rule_hash.go`** — dropped `normalizePrefix`. `"logs"` and `"logs/"` match different objects under literal `strings.HasPrefix` semantics, so they must hash differently; collapsing them would let an XML edit silently bind a new rule to the previous rule's durable state. Test flipped to assert the new behavior.
3. **`s3_lifecycle.proto`** — added `_UNSPECIFIED = 0` sentinels to `RuleKind`, `RuleMode`, `DegradedReason` (replaces `NONE`), and `StreamKind`. Persisted state schemas need this so a partially-populated payload doesn't silently default to a semantically active value.
4. **`worker.proto`** — same fix for `S3LifecycleParams.Subtype`. Previously `READ = 0` made READ the implicit default; an unset subtype on a TaskParams payload would have routed to the cluster-singleton reader. Now `SUBTYPE_UNSPECIFIED = 0`, READ at 1.

No on-disk state has been written yet (worker is not wired up until Phase 2c-d), so the renumbering is safe. Local `go test ./weed/s3api/s3lifecycle/...` passes.

chrislusf:
Addressed the latest three findings:

1. **High — multi-action rules collapse to one compiled entry**: restructured to per-action throughout. One XML `<Rule>` now expands into N compiled actions (`RuleActionKinds`); each gets its own state directory `<bucket>/<rule_hash>/<action_kind>/`, its own delay group, mode, pending stream, and degraded_reason. Helpers `MinTriggerAge` and `EventLogHorizon` now take an `ActionKind` so a 90d `EXPIRATION_DAYS` sibling no longer shares its cursor advance with a 7d `ABORT_MPU` sibling. Design doc updated to match. Proto: top-level `ActionKind` enum, `LifecycleState` keyed by `(rule_hash, action_kind)`, `BootstrapKey`/`PendingKey`/`BlockerRecord`/`RetryTarget` carry `action_kind`. `NoncurrentVersionExpirationDays + NewerNoncurrentVersions` are AWS-paired conditions on a single non-current expiration action — `RuleActionKinds` subsumes `NEWER_NONCURRENT` into `NONCURRENT_DAYS` when both are set.
2. **Medium — `LifecycleState` proto vs. design contract**: aligned. Added `last_safety_scan_ts_ns`, `next_safety_scan_ts_ns`, and the four counters (`evaluated_total`, `expired_total`, `metadata_only_total`, `error_total`); dropped `last_evaluated_ns`, `deleted_total`, `skipped_object_lock_total`, `pending_size`.
3. **Low — `RuleHash` delimiter collisions**: switched to length-prefixed canonical encoding. Each scalar is `<field-tag-byte> <uvarint-length> <bytes>`. Tag pair `(a=b, c)` no longer collides with `(a, b=c)`, and a prefix containing `\nexp_days=99` no longer forges an action field. Three regression tests added.

No on-disk state has been written yet (worker not wired until Phase 2c-d), so the renumbering / rename / restructure is free at this point. `go test ./weed/s3api/s3lifecycle/...` passes; `go build ./...` clean.

chrislusf:
Two new findings addressed:

1. **High — engine still keyed by rule_hash alone**: finished the per-action restructure across the engine pseudocode, bootstrap completion, drain locks, detector paths, policy CAS, and the Phase 2 plan. Introduced \`ActionKey{rule_hash, action_kind}\` as the engine's primary identity. \`originalDelayGroups\`, \`predicateActions\`, \`dateActions\`, \`state[ActionKey]\`, \`engine.markActive\`, \`newly_completed\`, \`target_modes\`, drain locks, detector status walks, and policy CAS all now key off \`ActionKey\`. Sibling actions of one rule have separate cursors, separate completion bits, separate locks, separate metrics. An XML edit that adds an action sub-element produces a new \`ActionKey\` (same rule_hash, new kind) that goes through pending_bootstrap while the rule's other actions stay active.

2. **Medium — aggregate-rule evaluator API**: \`Evaluate(rule, info, now)\` and \`ComputeDueAt(rule, info)\` replaced with \`EvaluateAction(rule, kind, info, now)\` and \`ComputeDueAt(rule, kind, info)\`. Each helper now decides exactly one (rule, kind) compiled action against one entry. Asking for a kind the rule doesn't declare, or against the wrong object shape, returns \`ActionNone\` / zero — never silently routes to a sibling. Multi-action regression test pinned: a rule with \`ExpirationDays=90 + AbortMPU=7\` evaluates each kind independently and the 7d window has no influence on the 90d eligibility.

\`go test ./weed/s3api/s3lifecycle/...\` passes; \`go build ./...\` clean.

PR #9348, opened 2026-05-07
feat(s3/lifecycle): policy engine — XML conversion, Compile, decideMode, Match

chrislusf:
Phase 2b of the S3 lifecycle redesign (#9346). Builds the in-memory engine that the bootstrap walker (Phase 2c) and the cluster reader (Phase 3) drive against. No callers yet — engine only; safe to land standalone.

## Summary

- **`weed/s3api/s3api_lifecycle_canonical.go`** — converts the XML-deserialized `Lifecycle` into the flat `s3lifecycle.Rule` shape the engine compiles against. Handles `<Filter><And>` flattening, single `<Tag>` filter, top-level `<Prefix>` fallback, multi-action rules.
- **`weed/s3api/s3lifecycle/action_kind.go`** — adds `s3lifecycle.ActionKey{rule_hash, action_kind}`, the engine's primary identity. One XML rule with N populated action sub-elements expands into N `ActionKey`s sharing `RuleHash` but differing in `ActionKind` so sibling actions are scheduled, gated, and degraded independently.
- **`weed/s3api/s3lifecycle/engine/`** — new package:
  - `Engine`/`Snapshot`: atomic-swap immutable snapshot. `MarkActive` is per-action and visible to in-flight reader passes without recompile.
  - `Compile`: produces the snapshot from `[]CompileInput` + per-`ActionKey` `PriorState` + cluster `MetaLogRetention`.
  - `decideMode`: disabled rule → `DISABLED`; `EXPIRATION_DATE` → `SCAN_AT_DATE`; reader-driven kind whose `eventLogHorizon + bootstrapLookbackMin` exceeds `metaLogRetention` → `SCAN_ONLY`; otherwise `EVENT_DRIVEN`. `metaLogRetention=0` means unbounded (the SeaweedFS default), gate doesn't trip.
  - `MatchOriginalWrite` / `MatchPredicateChange` / `MatchPath`: routing for the reader's per-delay sweep, the predicate-change sweep, and the bootstrap walker. Filters on `IsActive()` so the MarkActive flip is observable without recompile.

## Test plan
- `go test ./weed/s3api/s3lifecycle/...` — passing locally; covers single-action / multi-action expansion, retention gate, sibling actions degrading independently, `ExpirationDate` SCAN_AT_DATE path, disabled rule, prefix/tag/size filters, MPU-init shape gating, predicate-change routing, bootstrap-walker `MatchPath`.
- `go build ./...` — clean.
- No runtime callers yet; behavior cannot regress prod paths until Phase 2c/2d wires the engine into the LifecycleDelete RPC server and the bootstrap walker.

Stacked on #9346 (already merged).

## Summary by CodeRabbit

* **New Features**
  * Compiled lifecycle engine with deterministic snapshots, routing and matching for scheduled and event-driven actions; supports prefix/tag/size filters and MPU-abort gating.

* **Bug Fixes**
  * Safer handling of noncurrent-version indexing and nil inputs; correct preservation of disabled/prior modes and expiration-date parsing.

* **Tests**
  * Expanded tests covering canonicalization, compilation, matching, snapshot behavior, and retention logic.

* **Documentation**
  * Concise updates to lifecycle and scheduling docs.

### DIFF
--- a/weed/s3api/s3api_lifecycle_canonical.go
+++ b/weed/s3api/s3api_lifecycle_canonical.go
@@ -1,0 +1,101 @@
+package s3api
+
+import (
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3lifecycle"
+)
+
+// LifecycleToCanonical converts the XML-deserialized Lifecycle into the flat
+// s3lifecycle.Rule shape the engine compiles against. One XML <Rule> maps to
+// exactly one s3lifecycle.Rule (with potentially multiple action sub-fields
+// populated); the engine then expands each into its compiled actions via
+// s3lifecycle.RuleActionKinds.
+//
+// Filter resolution mirrors AWS semantics: the optional <Filter> element may
+// contain a single <Prefix> | <Tag> | <And>, or be absent (in which case the
+// older top-level <Prefix> applies). When <And> is used, its sub-elements
+// (Prefix + multiple Tags + size filters) are flattened into the canonical
+// Rule's individual fields.
+func LifecycleToCanonical(lc *Lifecycle) []*s3lifecycle.Rule {
+	if lc == nil {
+		return nil
+	}
+	out := make([]*s3lifecycle.Rule, 0, len(lc.Rules))
+	for i := range lc.Rules {
+		out = append(out, ruleToCanonical(&lc.Rules[i]))
+	}
+	return out
+}
+
+func ruleToCanonical(r *Rule) *s3lifecycle.Rule {
+	out := &s3lifecycle.Rule{
+		ID:     r.ID,
+		Status: string(r.Status),
+	}
+
+	// Prefix: <Filter><Prefix> or <Filter><And><Prefix> or top-level <Prefix>.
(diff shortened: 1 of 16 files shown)


chrislusf:
Audit found one more missed review (`#pullrequestreview-4247145366`): added the requested nil-NoncurrentIndex regression tests in commit `843d75a6b`. Both `ActionKindNoncurrentDays` (with keep-N retention set) and `ActionKindNewerNoncurrent` now have explicit `NoncurrentIndex == nil` cases that pin the conservative ActionNone behavior — locks the *int pointer migration's safety contract.

chrislusf:
Addressed [review #4247947095](https://github.com/seaweedfs/seaweedfs/pull/9348#pullrequestreview-4247947095): the duplicate-bucket concern in compile.go is real but defensive — all callers (tests + the future bootstrap walker) pass one input per bucket. Documented the invariant on `CompileInput` rather than adding merge logic for a misuse case (commit 403ccff6c).

PR #9349, opened 2026-05-07
feat(s3/lifecycle): LifecycleDelete RPC server

chrislusf:
Phase 2c — adds the worker-to-S3 RPC the lifecycle worker calls to execute one (rule, action) verdict against one entry. Stacks on #9348.

## Summary

- **`weed/pb/s3_lifecycle.proto`** — adds \`SeaweedS3LifecycleInternal.LifecycleDelete\`, \`LifecycleDeleteRequest\` (bucket / object_path / version_id / rule_hash / action_kind / stream context / EntryIdentity CAS / engine_snapshot_id), \`LifecycleDeleteResponse\`, and \`LifecycleDeleteOutcome\` enum (DONE / NOOP_RESOLVED / SKIPPED_OBJECT_LOCK / RETRY_LATER / BLOCKED). Makefile emits gRPC stubs.
- **\`weed/s3api/s3api_internal_lifecycle.go\`** — handler implementing the five-step flow:
  1. Re-fetch live entry (NOT_FOUND → NOOP_RESOLVED).
  2. EntryIdentity CAS (mtime / size / head fid / length-prefixed sha256 over sorted Extended). Mismatch → NOOP_RESOLVED (STALE_IDENTITY).
  3. \`enforceObjectLockProtections\` with \`governanceBypassAllowed=false\` — lifecycle never bypasses legal-hold or compliance. Lock refusal → SKIPPED_OBJECT_LOCK (logged + counted, cursor advances).
  4. Dispatch by \`action_kind\`: createDeleteMarker / deleteUnversionedObjectWithClient / deleteSpecificObjectVersion. ABORT_MPU is stubbed pending Phase 5 wiring (returns RETRY_LATER).
  5. Helper-side failures the server can't classify as transient → BLOCKED (worker writes BlockerRecord); filer fetch transport errors → RETRY_LATER (worker's retry budget eventually promotes sustained transients).

## Why \`hashExtended\` is length-prefixed

A naively concatenated Extended hash lets a forged single-tag payload (\`{k1: "v1k2v2"}\`) collide with a real two-tag map (\`{k1: "v1", k2: "v2"}\`). Length prefixing each scalar prevents this. Regression test pins it.

## Test plan
- \`go test ./weed/s3api/ -run TestComputeEntryIdentity -run TestHashExtended -run TestIdentityMatches -run TestLifecycleDelete\` — passing locally; covers identity field comparison, hashExtended order/delimiter stability, empty-request rejection.
- \`go build ./...\` — clean.
- Full server integration tests (real filer, every action_kind end-to-end) come in Layer 2 once the worker side (Phase 2d) lands.

No callers yet — server handler exists but no client invokes it until the worker's bootstrap walker / reader land.

## Summary by CodeRabbit

* **New Features**
  * Implemented S3 lifecycle automated deletion capabilities with support for expiration rules, noncurrent version removal, and multipart upload cleanup operations.

* **Bug Fixes**
  * Fixed processing of noncurrent version expiration and multipart upload configuration fields in lifecycle rules.


chrislusf:
Addressed the suggestion: replaced the hand-rolled `bytesEqual` with `bytes.Equal` from the stdlib. Slightly faster (intrinsified on amd64/arm64) and one fewer surface to maintain. PR rebased onto the updated `lifecycle-engine` so the bucket-scoped `ActionKey` fix is included.

chrislusf:
Review findings:

- `LifecycleDeleteRequest` carries `rule_hash`, `action_kind`, and `engine_snapshot_id`, and the proto comment says the server verifies the current policy still contains that `(rule_hash, action_kind)`. The handler never does that check: after live fetch + identity CAS it goes straight to object-lock and `lifecycleDispatch`. A pending/drain item from a removed or changed lifecycle rule can still delete as long as the object identity matches. Please re-read the bucket lifecycle XML in the handler, canonicalize/expand it, and return `NOOP_RESOLVED` / `STALE_POLICY` when the current policy does not contain the requested action key.

- Step 1 currently cannot distinguish transient fetch failures from not-found. `LifecycleDelete` calls `getObjectEntry`, but that helper wraps every retrieval error as `ErrObjectNotFound` (`s3api_object_retention.go`), so a filer/RPC/read error will be returned to the worker as `NOOP_RESOLVED` and the cursor can advance past an unprocessed delete. Use a fetch path that preserves the underlying error, or change the helper to wrap the real error so only actual not-found maps to `NOOP_RESOLVED`; transport/IO errors should return `RETRY_LATER`.

- `identityMatches` treats a missing `expected_identity` as a match. That disables the CAS that makes this RPC safe; a caller that cannot provide identity should be rejected/blocked rather than allowed to delete. The bootstrap walker has the live entry in hand, so it should build the identity before dispatch.

- `ABORT_MPU` always returns `RETRY_LATER` with `"ABORT_MPU not yet wired"`. Since the engine already routes `ABORT_MPU` actions, the first due MPU action will stall that stream indefinitely. Either wire the abort path in this PR or keep `ABORT_MPU` out of dispatch/routing until the server can complete it.

Focused tests run locally: `go test ./weed/pb/... ./weed/s3api`.

chrislusf:
Audit found seven inline comments from gemini's second review (18:35 UTC) that I'd missed. Addressed in two commits:

1. `eb59c0b3a` — `retryLater` helper added alongside `done`/`noopResolved`/`blocked`; replaced the two inline RETRY_LATER literals with the helper for shape consistency.
2. `a30d1008c` — semantic fix: classify versioning lookup, createDeleteMarker, deleteUnversionedObjectWithClient, and deleteSpecificObjectVersion errors as RETRY_LATER (transient) by default. The previous BLOCKED was too aggressive — every transient filer hiccup would halt the stream until operator intervention. RETRY_LATER lets the worker's retry budget escalate sustained transients to BLOCKED at the right threshold. BLOCKED stays only for clearly deterministic errors: missing version_id and unknown action_kind.

Sorry I missed the whole batch on the first sweep.

chrislusf:
Audited the [linked review](https://github.com/seaweedfs/seaweedfs/pull/9349#pullrequestreview-4246764768) — 4 of 7 findings were already addressed by earlier commits (`retryLater` helper, `blocked` for empty request, `retryLater` for transport errors, MPU stub helper). Three were partial. Fixed in commit `d8a7467fe`:

1. Versioning lookup now treats `filer_pb.ErrNotFound` (bucket deleted between live-fetch and versioning-lookup) as `NOOP_RESOLVED` rather than `RETRY_LATER`.
2. `deleteUnversionedObjectWithClient` recognizes `ErrObjectNotFound` alongside `filer_pb.ErrNotFound` for the resolved-NOOP path.
3. `deleteSpecificObjectVersion` recognizes `ErrObjectNotFound` alongside `filer_pb.ErrNotFound` and `ErrVersionNotFound`.

Matches the live-fetch step's not-found classification (which already covered all three error variants).

PR #9350, opened 2026-05-07
feat(s3/lifecycle): bucket-level bootstrap walker

chrislusf:
Phase 2d — adds the bucket-level bootstrap walker as a callback-driven library. Stacks on #9349.

## Summary

- **\`weed/worker/types/task_types.go\`** — adds \`TaskTypeS3Lifecycle = \"s3_lifecycle\"\` constant. The string is already wired to \`LaneLifecycle\` in \`admin/plugin/scheduler_lane.go\`; this commit just gives the type a Go name so future task-handler wiring doesn't sprinkle string literals.
- **\`weed/s3api/s3lifecycle/bootstrap/walker.go\`** — \`Walk(ctx, snap, bucket, list, dispatch, opts)\`. Callback-driven so callers supply the listing source (real \`filer_pb.SeaweedList\` or test fake) and the dispatcher (real \`LifecycleDelete\` gRPC client or test fake). Keeps the walker free of filer_pb dependencies.

Walker semantics (matches design doc §\"Bootstrap — bucket-level, inline-delete\"):
- Per entry, calls \`Snapshot.MatchPath\` to get the active \`ActionKey\`s whose filter matches.
- Per matched action, runs \`s3lifecycle.EvaluateAction(rule, kind, info, now)\`. Currently-due → \`Dispatcher.Delete\`. Not-yet-due → skip (the meta-log reader picks them up).
- SCAN_AT_DATE actions are skipped (their date-triggered bootstrap is a separate task).
- Pending-bootstrap actions are skipped via the \`IsActive()\` filter inside \`MatchPath\`.
- Returns a \`Checkpoint{LastScannedPath, Completed}\` so the caller can persist resume state.
- \`opts.Resume\` skips entries with \`Path <= Resume\` for kill-resume.
- Dispatch error halts the walk; the checkpoint stays at the last successfully-processed entry so the next task retries the failing path.

## What's NOT in this PR
- Worker task framework integration (handler registration with \`base.RegisterTask\`, config schema for the admin UI, etc.).
- Real \`filer_pb.SeaweedList\` adapter — Phase 2e.
- Real \`LifecycleDelete\` gRPC client — the server handler is in #9349; the client wrapper lands in Phase 2e.
- Detector — Phase 2e.
- Shell commands — Phase 2e.

## Test plan
- \`go test ./weed/s3api/s3lifecycle/bootstrap/...\` — passing locally; covers prefix-mismatch / not-yet-due / date-kind / pending-bootstrap skip, multi-action rule (one rule, three actions, three dispatches — the regression per-action keying fixes), dispatch error halts at last-successful checkpoint, Resume skips entries up to the resume path.
- \`go build ./...\` — clean.

## Summary by CodeRabbit

* **New Features**
  * Added S3 lifecycle bootstrap processing to evaluate and execute lifecycle rules on bucket entries.
  * Added checkpoint and resume support for interrupted lifecycle scans.
  * Implemented proper handling of lifecycle rule status and action dispatching.

* **Tests**
  * Added comprehensive test coverage for lifecycle bootstrap walker functionality.

### DIFF
--- a/weed/s3api/s3lifecycle/bootstrap/walker.go
+++ b/weed/s3api/s3lifecycle/bootstrap/walker.go
@@ -17,14 +17,17 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3lifecycle/engine"
 )
 
-// Entry is the minimal slice of a filer entry the walker needs. SuccessorModTime
-// and NoncurrentIndex are populated only for versioned-bucket walks (Phase 5);
-// non-versioned walks leave them at zero / nil and the count-based retention
-// path bails out conservatively.
+// Entry is the minimal slice of a filer entry the walker needs.
+// IsDirectory marks SeaweedFS directory entries; the walker never
+// dispatches against directories. SuccessorModTime and NoncurrentIndex are
+// populated only for versioned-bucket walks (Phase 5); non-versioned walks
+// leave them at zero / nil and the count-based retention path bails out
+// conservatively.
 type Entry struct {
 	Path           string
 	ModTime        time.Time
 	Size           int64
+	IsDirectory    bool
 	IsLatest       bool
 	IsDeleteMarker bool
 	IsMPUInit      bool
@@ -72,6 +75,12 @@ func Walk(ctx context.Context, snap *engine.Snapshot, bucket string, list ListFu
 		if entry == nil || entry.Path == "" {
 			return nil
 		}
+		// Lifecycle rules only apply to objects; SeaweedFS directory
+		// entries can co-exist in the listing and must not be deleted.
+		if entry.IsDirectory {
+			cp.LastScannedPath = entry.Path
+			return nil
+		}
 		if err := walkEntry(ctx, snap, bucket, entry, dispatch, now); err != nil {
 			return err
 		}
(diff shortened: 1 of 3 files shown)


chrislusf:
Rebased onto the updated `lifecycle-rpc` (which carries the `ActionKey.Bucket` fix from `lifecycle-engine`). Walker test updated to construct ActionKeys with the new Bucket field. `go test ./weed/s3api/s3lifecycle/...` passes.

chrislusf:
Both quick wins applied: dropped the redundant `Resume` re-filter in Walk's callback (trusts ListFunc's contract instead) and hoisted the `ObjectInfo` allocation out of the per-action loop in walkEntry (one allocation per entry instead of one per (entry, kind) pair). `go test ./weed/s3api/s3lifecycle/bootstrap/...` passes.

chrislusf:
Review findings:

- The bootstrap walker only evaluates `ACTIVE` actions via `snap.MatchPath`, but new lifecycle actions are intentionally inactive / `bootstrap_complete=false` until bootstrap finishes. That means the initial bootstrap for a new action will scan the bucket and dispatch nothing, then the surrounding task can still mark the action complete and seed cursors. Currently `TestWalk_PendingBootstrapNotDispatched` pins that behavior, but it is the opposite of the bootstrap contract. The walker needs a way to evaluate the target action set for the bootstrap task, including pending-bootstrap and scan-only actions, without registering them in the event-reader indexes.

- `ModeScanAtDate` actions are always skipped in `walkEntry`. This also makes the date-triggered bootstrap a no-op: the detector can schedule the bucket walk at `rule.date`, but the walker will skip the only action it is supposed to enforce. Please pass the bootstrap purpose/action scope into the walker, or otherwise allow `EXPIRATION_DATE` actions during their scheduled `SCAN_AT_DATE` walk while still excluding them from regular age-rule bootstrap.

Focused tests run locally: `go test ./weed/s3api/s3lifecycle/... ./weed/worker/types`.

chrislusf:
Audit found three missed comments from gemini's second review. Addressed in two commits:

1. `8cdfd7e72` — two safety findings: (a) added `Entry.IsDirectory` and skip directory entries in Walk so the walker never tries to delete a SeaweedFS directory; (b) skip `engine.ModeDisabled` actions explicitly in `walkEntry` since operator pause is independent of XML rule status.
2. `432322ae9` — perf: hoisted ObjectInfo allocation out of the per-entry callback (one alloc per Walk, not per entry). Safe because EvaluateAction reads synchronously and doesn't retain.

Two regression tests pin the new safety behavior.

Issue #19, opened 2014-10-16
File Listing or Finding


mcqueenorama:
Is there a way to find or list the files?  Is there a simple plan so I can code it myself?  I do like the simplicity of this.

PR #9351, opened 2026-05-07
feat(pb): ReadLogFileRefsWithPosition for per-shard cursors

chrislusf:
Phase 0 / Phase 3 prerequisite (Task #19) — adds the per-shard variant of \`ReadLogFileRefs\` that the lifecycle reader (Phase 3) drives the meta log through. Lands on master directly (not stacked) since it's a generic pb-package addition with no dependencies on the lifecycle PRs.

## Why per-shard

\`LogEntry.Offset\` is per-buffer per-filer, not globally unique. A single uniform startTsNs (the existing \`ReadLogFileRefs\` API) is fine for a single subscriber, but the lifecycle reader heap-merges chunks from multiple filers and needs to skip per-filer past the last resolved cursor. A single global cursor would either miss events or replay them.

## API

- **\`MessagePosition{TsNs, Offset}\`** — local to pb to avoid pulling \`weed/util/log_buffer\`.
- **\`MaxMessagePosition\`** — sentinel meaning \"skip every event for this shard\". Used by the reader when a shard has an active \`BlockerRecord\`; the cursor stays at the position recorded in the blocker.
- **\`ReadLogFileRefsWithPosition(refs, newReader, startPositions, stopTsNs, filter, cb)\`**:
  - \`startPositions[filer_id]\` is the strict-\`<=\` skip predicate (events with \`(ts, offset) <= start\` are skipped).
  - \`cb(event, filer_id, position)\` lets the caller update its per-shard cursor map after each delivered event.
  - Returns \`lastByFiler\` so the worker's checkpoint write is atomic.
  - Cross-filer ordering: \`(event.ts, filer_id, event.offset)\`. The \`filer_id\` tiebreak makes retries from the same starting cursor see the same event sequence, which the per-event resolution-or-stop logic depends on.

## What's NOT in this PR

\`EarliestRetainedPositionPerShard\` and \`RetainedLogRangePerShard\` are the other two APIs Task #19 calls out; they walk \`SystemLogDir\` to build the lazy-seeding floor and tail-drain GC inputs. They're additions to the same package but a separate concern; they'll land in a follow-up so this PR stays focused.

## Test plan

- \`go test ./weed/pb/ -run TestReadLogFileRefsWithPosition -run TestMessagePosition\` — passing locally; covers order + cursor return, per-shard skip via \`(ts, offset)\` compound predicate, paused shard via \`MaxMessagePosition\`, \`stopTsNs\` cap, callback-error halt, and \`MessagePosition.LessOrEqual\` boundary cases.
- \`go build ./...\` — clean.

No callers yet — the existing \`ReadLogFileRefs\` is unchanged. The lifecycle reader (Phase 3) will be the first caller.

## Summary by CodeRabbit

* **New Features**
  * Resumable log reading with per-stream position cursors and a sentinel to fully pause a stream
  * Deterministic merge of events from multiple sources ordered by timestamp with a stable tie-break
  * Position-aware emission: per-event position info, stop-time cap, and callback-driven delivery; callback errors halt processing and are reported
  * Returns last-seen position per active stream

* **Tests**
  * Comprehensive tests for merge ordering, per-stream skipping/pausing, stop-time capping, ordering semantics, and callback error handling

### DIFF
--- a/weed/pb/filer_pb_direct_read_with_position.go
+++ b/weed/pb/filer_pb_direct_read_with_position.go
@@ -96,7 +96,7 @@ func ReadLogFileRefsWithPosition(
 	var wg sync.WaitGroup
 	for _, fid := range filerOrder {
 		startPos := startPositions[fid]
-		paused := startPos.TsNs == MaxMessagePosition.TsNs && startPos.Offset == MaxMessagePosition.Offset
+		paused := startPos == MaxMessagePosition
 		ch := make(chan *filer_pb.LogEntry, 256)
 		errCh := make(chan error, 1)
 		s := &filerStream{id: fid, ch: ch, errCh: errCh, startPos: startPos, paused: paused}
--- a/weed/pb/filer_pb_direct_read_with_position.go
+++ b/weed/pb/filer_pb_direct_read_with_position.go
@@ -12,25 +12,20 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 )
 
-// MessagePosition is the per-shard cursor (ts_ns, offset). LogEntry.Offset
-// is per-buffer per-filer, so a single global cursor can't safely resume from
-// a multi-filer merged stream. Local to pb to avoid pulling log_buffer.
+// MessagePosition is the per-shard cursor. LogEntry.Offset is per-filer, so
+// a single global cursor can't resume safely from a merged stream.
 type MessagePosition struct {
 	TsNs   int64
 	Offset int64
 }
 
-// MaxMessagePosition returns the sentinel that pauses a shard's emission
-// entirely; used when the shard has an active blocker. Returned as a value
-// rather than an exported var so an importer can't accidentally mutate it
-// (e.g. `pb.MaxMessagePosition.TsNs = 0` would silently break every
-// `startPos == MaxMessagePosition()` check).
+// MaxMessagePosition pauses a shard's emission. Function-not-var so an
+// importer can't mutate the sentinel.
 func MaxMessagePosition() MessagePosition {
 	return MessagePosition{TsNs: 1<<63 - 1, Offset: 1<<63 - 1}
 }
 
-// LessOrEqual is the strict-`<=` skip predicate so the last resolved event
-// isn't replayed.
+// LessOrEqual: strict `<=` so the last resolved event isn't replayed.
 func (p MessagePosition) LessOrEqual(other MessagePosition) bool {
 	if p.TsNs != other.TsNs {
 		return p.TsNs < other.TsNs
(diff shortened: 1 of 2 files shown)


chrislusf:
Dropped the dead empty-`if` block. The per-entry predicate already covers every event correctly; safely pruning whole files would need a `FileMaxTsNs` on `LogFileChunkRef` (currently we only have `FileTsNs`, the flush start time). Noted that requirement inline so a future commit can revisit if the optimization is wanted.

chrislusf:
Addressed all three nits from the missed review:

1. `paused := startPos == MaxMessagePosition` — `MessagePosition` is comparable, no need for the field-by-field check.
2. `TestReadLogFileRefsWithPosition_OrderAndCursor` now pins the documented `(ts, filerId)` tiebreak: tied `TsNs` must put alpha before beta. The previous assertion would have silently accepted swapped order, which would break retry determinism.
3. `TestReadLogFileRefsWithPosition_PerShardSkip` no longer sets alpha to `{0,1}` and immediately overwrites — merged into one map literal that uses `base` directly, with a clearer comment.

Sorry I missed this review on the first pass — the polling logic was only watching inline + top-level comments, not review *bodies*. That's the third bucket and where coderabbit's nit lists land. Polling now covers all three.

chrislusf:
Review findings:

- `dispatchOneEntryWithPosition` logs proto unmarshal failures and returns `nil`, so `ReadLogFileRefsWithPosition` records that position in `lastByFiler` and the lifecycle reader can advance past a corrupt metadata event. For lifecycle, a malformed persisted event is a deterministic unprocessable record, not something the reader may silently skip; it should surface an error with `(filerId, position)` so the worker can pause/block that stream and keep the cursor at the previous resolved position.

- `streamEntriesWithPosition` also skips `isChunkNotFound` errors. For this lifecycle replay path, a referenced log chunk disappearing is lost metadata history; treating it as an empty file advances the stream with no lost-log signal. Please return an error or explicit lost-log outcome here, and let the lifecycle reader degrade to `scan_only` / `LOST_LOG` instead of silently dropping the file.

Focused tests run locally: `go test ./weed/pb`.

chrislusf:
Audit found one more review I missed (`#pullrequestreview-4247181453`). Fixed in commit `9a29742e8`: `MaxMessagePosition` is now a function instead of an exported mutable var, so an importer can't accidentally do `pb.MaxMessagePosition.TsNs = 0` and silently break every paused-shard comparison. Each call returns a fresh value.

chrislusf:
Closing: per-shard cursors aren't load-bearing. Filer metadata is shared and not the bottleneck; Phase 3 reader will use the existing ReadLogFileRefs with a single global TsNs cursor. Re-open if/when ts-collision resume or per-shard pause becomes operationally needed.

PR #9352, opened 2026-05-07
feat(pb): listing helpers for lazy-seeding + tail-drain GC

chrislusf:
Second half of Task #19. Stacks on #9351.

## Summary

\`LogDirLister\` is the directory-listing interface. Tests inject in-memory fakes; production code wraps \`filer_pb.SeaweedFilerClient\`. Two consumers:

- **\`EarliestRetainedPositionPerShard(ctx, lister)\`** → \`map[filer_id]MessagePosition\` with each filer's earliest retained chunk. Used by lazy-seeding so a newly-discovered shard's cursor is floored at the earliest event the shard could possibly emit (otherwise a fresh shard seeded at \"now\" misses in-flight events).

- **\`RetainedLogRangePerShard(ctx, lister)\`** → \`map[filer_id]FilerShardRange\` with \`{Earliest, Latest}\`. Latest's \`Offset\` is \`int64-max\` so a \`cursor >= range.latest\` lex comparison (under the strict \`<=\` skip predicate) is satisfied only when the cursor is at or past the chunk's last entry. This matches the design's tail-drain criterion exactly.

The reader's \`gcDepartedShards\` (Phase 3) consumes \`RetainedLogRangePerShard\`:
- \`cursor[F] >= range.latest\` → safe to GC F's cursor (record in \`tail_drained_streams\`).
- \`F\` has no range AND \`tail_drained_streams[F]\` is unset → lost-log; downgrade reader-driven rules to \`scan_only\` with \`degraded_reason = LOST_LOG\`.

## Test plan
- \`go test ./weed/pb/ -run TestEarliestRetainedPositionPerShard -run TestRetainedLogRangePerShard -run TestWalkLogChunks -run TestGetFilerIdFromName -run TestParseChunkTime\` — passing locally; covers two-filer earliest, empty-dir, range earliest/latest for distinct chunks, single-chunk case, malformed-name skip, helper-function boundary cases.
- \`go build ./...\` — clean.

A \`filer_pb.SeaweedFilerClient\`-backed \`LogDirLister\` adapter will land alongside the Phase 3 reader (which is the first caller). This PR keeps the \`pb\` package free of gRPC streaming mechanics in the unit tests.

### DIFF
--- a/weed/pb/filer_pb_log_dir_listing.go
+++ b/weed/pb/filer_pb_log_dir_listing.go
@@ -44,7 +44,7 @@ type FilerShardRange struct {
 // entries at exactly the floor's TsNs with Offset>0 are not skipped.
 func EarliestRetainedPositionPerShard(ctx context.Context, lister LogDirLister) (map[string]MessagePosition, error) {
 	out := make(map[string]MessagePosition)
-	err := walkLogChunks(ctx, lister, false, func(filerId string, ts time.Time) bool {
+	err := walkLogChunks(ctx, lister, func(filerId string, ts time.Time) bool {
 		if _, seen := out[filerId]; !seen {
 			out[filerId] = MessagePosition{TsNs: ts.UnixNano(), Offset: 0}
 		}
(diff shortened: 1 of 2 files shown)


chrislusf:
Folded the two walks into one. The single-pass version tracks min and max per filer in the same callback, halves RPC cost, and removes the TOCTOU window (a filer that was present on the first walk and gone before the second would have left `latest` as zero `time.Time` and yielded a nonsense MessagePosition with TsNs = -6795364578871345152). `go test ./weed/pb/...` passes.

chrislusf:
Review findings:

- `EarliestRetainedPositionPerShard` seeds the cursor at `(earliest_chunk_ts, 0)`, but `ReadLogFileRefsWithPosition` skips events with `event_pos <= cursor`. That drops the first retained event when its offset is `0`. The design uses a before-first offset sentinel for exactly this reason. Please return `(earliest_chunk_ts, BEFORE_FIRST_OFFSET)` (for example offset `-1`) and add a test with an entry at the exact earliest chunk timestamp/offset 0.

- `RetainedLogRangePerShard` sets `Latest` to `(latest_chunk_filename_ts, MaxInt64)`. The filename timestamp is the minute bucket / flush-file start, while entries inside that chunk can have later `LogEntry.TsNs`. After processing just one event in that latest chunk, a cursor like `(filename_ts + 10s, 0)` already compares greater than `(filename_ts, MaxInt64)`, so tail-drain GC can conclude the shard is drained while later events in the same retained chunk remain unread. Use an actual last-entry position, or a conservative upper bound for the chunk interval such as end-of-minute plus max offset, and add a test where the latest chunk contains multiple event timestamps after the filename minute.

Focused tests run locally: `go test ./weed/pb`.

chrislusf:
Audit found one more inline comment I missed (`listAll` buffering). Fixed in commit `0a7da7957`: replaced the per-directory buffering with a `streamEntries` helper that paginates and invokes the per-entry callback inline. Per-day chunks (the hot path with hundreds of thousands of entries on busy clusters) are processed and discarded inline instead of accumulating in a slice. An `errStopWalk` sentinel propagates the "cb returned false" signal through the nested call stack without confusing it with real errors.

chrislusf:
Closing: per-shard cursors aren't load-bearing. Filer metadata is shared and not the bottleneck; Phase 3 reader will use the existing ReadLogFileRefs with a single global TsNs cursor. Re-open if/when ts-collision resume or per-shard pause becomes operationally needed.

PR #9357, opened 2026-05-07
fix(s3/lifecycle): address Phase 3 post-merge review (#9354 #9355 #9356)

chrislusf:
## Summary
Addresses gemini findings on the three Phase 3 PRs that landed before review (#9354, #9355, #9356):

### #9354 (reader)
- **HIGH** — `extractBucketKey` accepted `/buckets/` but rejected the bare `/buckets` (no trailing slash); some delete events emit the bare form, so bucket-root events were silently dropped. Now both forms resolve to a bucket-root event with the entry name as the bucket. (commit 6320ffd8f)
- **MED** — `BucketsPath` is pre-normalized once on `Run` instead of recomputing per event. (same commit)
- **MED** — `ShardID` reuses `sha256.Hash` via `sync.Pool`; the function runs on every meta-log event before the shard filter and was a per-event allocation. (commit 4c4dcb418)

### #9355 (router)
- **HIGH** — Skip hard deletes (`NewEntry == nil`); they carry no schedule-relevant state and would hit NOOP_RESOLVED at dispatch. (commit bf28174dd)
- **HIGH** — `EntryIdentity.MtimeNs` now holds true nanoseconds-since-epoch on both sides (router and `s3api_internal_lifecycle.go::computeEntryIdentity`). Previously both wrote `Mtime` (seconds) into `MtimeNs`; the CAS worked because both sides agreed, but the encoding contradicted the field name and would break if either side later started using true ns. (commits bf28174dd + b4511a84d)
- **MED** — Drop the unused `versioned` parameter from `buildObjectInfo`. (commit bf28174dd)
- **MED** — `buildObjectInfo` returns nil when `Attributes` are missing (otherwise `ModTime=0001-01-01` would make ExpirationDays fire immediately). (same commit)

### #9356 (dispatcher)
- **HIGH** — Distinguish `context.Canceled` / `context.DeadlineExceeded` from transport errors: re-queue the Match at its original DueTime with no retry-budget burn. Otherwise frequent restarts could escalate items to BLOCKED. (commit 20a16d6b5)

## Tests
- New: bare-prefix bucket-root event handled by reader
- New: hard-delete event skipped by router
- New: missing-Attributes event skipped by router
- New: identity captures Mtime in true nanoseconds
- New: ctx-cancel dispatch path doesn't burn budget or freeze cursor
- Updated: `TestComputeEntryIdentity_BasicFields` pinned to ns encoding
- All pre-existing tests still pass

## Test plan
- [x] go test ./weed/s3api/...
- [x] go build ./...

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of lifecycle operations during cancellations and timeouts so retries/budgets aren’t consumed
  * Fixed timestamp precision for lifecycle identity checks (uses full seconds+nanoseconds)
  * Corrected routing to skip delete-marker/hard-delete events and to ignore entries missing attributes
  * Broadened bucket path parsing to accept with/without trailing slash

* **Performance**
  * Reduced allocations in lifecycle sharding by reusing hashers

### DIFF
--- a/weed/s3api/s3api_internal_lifecycle.go
+++ b/weed/s3api/s3api_internal_lifecycle.go
@@ -135,7 +135,10 @@ func computeEntryIdentity(entry *filer_pb.Entry) *s3_lifecycle_pb.EntryIdentity 
 	}
 	id := &s3_lifecycle_pb.EntryIdentity{}
 	if entry.Attributes != nil {
-		id.MtimeNs = entry.Attributes.Mtime
+		// FuseAttributes splits the timestamp across Mtime (seconds) and
+		// MtimeNs (nanosecond component); EntryIdentity.MtimeNs is the
+		// combined nanoseconds-since-epoch value.
+		id.MtimeNs = entry.Attributes.Mtime*int64(1e9) + int64(entry.Attributes.MtimeNs)
 		id.Size = int64(entry.Attributes.FileSize)
 	}
 	if len(entry.GetChunks()) > 0 {
--- a/weed/s3api/s3api_internal_lifecycle_test.go
+++ b/weed/s3api/s3api_internal_lifecycle_test.go
@@ -10,15 +10,16 @@ import (
 
 func TestComputeEntryIdentity_BasicFields(t *testing.T) {
 	entry := &filer_pb.Entry{
-		Attributes: &filer_pb.FuseAttributes{Mtime: 1700000000, FileSize: 4096},
+		Attributes: &filer_pb.FuseAttributes{Mtime: 1700000000, MtimeNs: 123, FileSize: 4096},
 		Chunks: []*filer_pb.FileChunk{
 			{FileId: "1,abc"},
 			{FileId: "1,def"},
 		},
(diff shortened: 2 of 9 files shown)


chrislusf:
Addressed both gemini findings ([review #4248215368](https://github.com/seaweedfs/seaweedfs/pull/9357#pullrequestreview-4248215368)):
- Reader fallback prefix normalization now mirrors Run (commit 81873810c) — fixes the double-slash edge case when BucketsPath already ends with '/'.
- ObjectInfo.ModTime now carries the FuseAttributes.MtimeNs nanosecond component (commit e84c6db81) — keeps ModTime precision consistent with the EntryIdentity ns encoding.

chrislusf:
Addressed gemini [review #4248425391](https://github.com/seaweedfs/seaweedfs/pull/9357#pullrequestreview-4248425391): used `defer shardHashPool.Put(h)` so the hasher returns to the pool even if a Write/Sum panics (commit 7aab17a68).

PR #9358, opened 2026-05-08
feat(s3/lifecycle): filer-backed cursor Persister + drop BlockerStore

chrislusf:
## Summary
1. **`dispatcher.FilerPersister`** — persists per-shard cursor maps as JSON to `/etc/s3/lifecycle/cursors/shard-NN.json` via `filer.SaveInsideFiler`. One file per shard keeps Save atomic.
2. **Drop `BlockerStore`** entirely — the durable cursor is the durable block. A frozen cursor doesn't advance, so on worker restart the reader re-encounters the poison event at `MinTsNs`, the dispatcher walks the same retry budget to BLOCKED, and the cursor freezes at the same EventTs. Other in-flight events that already succeeded on the prior pass self-resolve via `NOOP_RESOLVED` (STALE_IDENTITY) at the LifecycleDelete server.

### Removed
- `BlockerStore` interface + `InMemoryBlockerStore` + `BlockerRecord`
- `Dispatcher.Blockers` field + `Dispatcher.ReplayBlockers`
- The `BlockerStore.Put` call inside `handleBlocked`
- `Pipeline.Blockers` field + the startup `ReplayBlockers` call

Operator visibility: a cursor whose `MinTsNs` hasn't advanced is the signal — surfaced via the durable cursor file, not a separate record store.

## Encoding
JSON, not proto. These files are internal worker state; no on-the-wire compat surface, and the existing `s3_lifecycle.proto` cursor messages are designed for the per-filer model that we replaced with per-(shard, ActionKey).

## Stack
This is Phase 3 follow-up wiring (a) of three:
- (a) **this PR** — filer-backed cursor durability
- (b) worker task registration: `RegisterS3LifecycleTask` with Detection/Scheduling/CreateTask handlers, dispatching `S3LifecycleParams.Subtype=READ` to `Pipeline.Run`
- (c) Phase 4: strip the existing TTL fast-path from `PutBucketLifecycleConfigurationHandler` once (b) is operational

## Test plan
- [x] `FilerPersister` round-trip: empty load, save/load, shard isolation, overwrite, corrupt-data error path
- [x] `TestDispatchRestartReFreezesNaturally` pins the self-recovery property: a fresh Dispatcher with a fresh Cursor, fed the same poison event, reaches the same frozen state at the same EventTs without any durable blocker store
- [x] All pre-existing dispatcher/router/reader tests pass
- [x] `go build ./...`

## Summary by CodeRabbit

* **Refactor**
  * Streamlined S3 lifecycle cursor persistence to durable filer-backed per-shard files with deterministic saves, and simplified stuck-operation handling by removing the separate blocker store so frozen cursors are re-applied on restart.
* **Tests**
  * Added comprehensive tests for cursor persistence, shard isolation, overwrite behavior, deterministic save output, and error handling for corrupted persisted state.


chrislusf:
Verified: `handleBlocked` did drop `reason` (the `_ = reason` comment was wrong — none of the three callers log it). Fixed in commit f1b2039d5: `glog.Warningf` with shard, key, eventTs, and the original reason before freezing the cursor. Same logging shape the rest of the package uses.

PR #9360, opened 2026-05-08
refactor(s3api): move Lifecycle XML structs to leaf package lifecycle_xml

chrislusf:
## Summary
The S3 BucketLifecycleConfiguration XML structs and the canonical-form converter lived in package \`weed/s3api\`. \`weed/s3api\` transitively imports \`weed/server\`, which imports \`weed/shell\` — so any caller outside \`s3api\` (shell command, future lifecycle worker) that wanted to parse a bucket's lifecycle XML hit an import cycle.

This is the unblocker for the Phase 3 follow-up shell command and any worker-side bucket-config discovery.

## Moves
- \`weed/s3api/s3api_policy.go\` → \`lifecycle_xml/types.go\` (s3api_policy.go was misnamed: its content was lifecycle XML, not S3 bucket policy)
- \`weed/s3api/s3api_lifecycle_canonical.go\` → \`lifecycle_xml/canonical.go\`
- \`s3api_lifecycle_canonical_test.go\` → \`lifecycle_xml/canonical_test.go\`
- \`s3api_policy_test.go\` → \`lifecycle_xml/round_trip_test.go\`

## Other changes
- Renames the public \`RuleStatus\` type (was unexported \`ruleStatus\`)
- Adds small accessor methods (\`Set\`, \`Val\`, \`AndSet\`, \`TagSet\`) so the s3api handler can read what it needs across the package boundary
- Adds \`NewPrefix\`, \`NewExpirationDays\` constructors for the GET handler
- Adds a \`Tag\` struct local to the package so it has zero internal seaweed deps
- Adds a one-shot \`ParseCanonical(xml []byte)\` helper for non-server callers (the shell command)

## Test plan
- [x] All pre-existing s3api lifecycle tests pass under their new home
- [x] s3api root tests pass (the handler change for accessor calls)
- [x] \`go build ./...\` clean

## Summary by CodeRabbit

* **New Features**
  * Improved S3 bucket lifecycle XML handling: more robust parsing, clearer rule status handling, and better support for filters, transitions, expirations, and edge cases—yielding more accurate lifecycle behavior.

* **Tests**
  * Expanded test coverage for canonicalization, XML round-trip, validation, and various filter/expiration scenarios to ensure reliable lifecycle behavior.

### DIFF
--- a/weed/s3api/lifecycle_xml/canonical.go
+++ b/weed/s3api/lifecycle_xml/canonical.go
@@ -1,8 +1,31 @@
-package s3api
+package lifecycle_xml
 
 import (
+	"bytes"
+	"encoding/xml"
+
 	"github.com/seaweedfs/seaweedfs/weed/s3api/s3lifecycle"
 )
+
+// Parse decodes a BucketLifecycleConfiguration XML body into the wire-form
+// Lifecycle struct.
+func Parse(xmlBytes []byte) (*Lifecycle, error) {
+	var lc Lifecycle
+	if err := xml.NewDecoder(bytes.NewReader(xmlBytes)).Decode(&lc); err != nil {
+		return nil, err
+	}
+	return &lc, nil
+}
+
+// ParseCanonical is the one-shot path most non-server callers want:
+// raw XML in, []*s3lifecycle.Rule out.
+func ParseCanonical(xmlBytes []byte) ([]*s3lifecycle.Rule, error) {
+	lc, err := Parse(xmlBytes)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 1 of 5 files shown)


chrislusf:
Both nitpicks valid; fixed:
- `canonical_test.go` parseLifecycle now calls the public `Parse` instead of `xml.Unmarshal` directly.
- `round_trip_test.go` asserts via `TagSet()`, `AndSet()`, `Set()` exporters instead of the internal `tagSet`, `andSet`, `set` fields. Same package so the private fields remain reachable; the change is about exercising the public contract that the s3api handler (and future external callers) rely on.

PR #9361, opened 2026-05-08
feat(shell): s3.lifecycle.run-shard + integration test

chrislusf:
## Summary
Operator-runnable harness for end-to-end validation of the Phase 3 event-driven lifecycle pipeline. Until the plugin-worker auto-scheduler lands, this is how an operator (or test) actually drives \`Pipeline.Run\` against a real cluster.

\`s3.lifecycle.run-shard -shard <0..15> -s3 <host:port> [-events 1000] [-dispatch 5s] [-checkpoint 30s]\`

- Subscribes to the filer meta-log filtered to one \`(bucket, key-prefix-hash)\` shard
- Routes events through \`engine.Snapshot\` + \`router.Route\`
- Dispatches due actions to the S3 server's \`LifecycleDelete\` RPC (registered by #9359)
- Cursor persists to \`/etc/s3/lifecycle/cursors/shard-NN.json\` (the \`FilerPersister\` from #9358)
- \`EventBudget\` bounds a single invocation so the command exits cleanly

Buckets are discovered by walking the configured \`DirBuckets\` path and reading each bucket entry's \`Extended[s3-bucket-lifecycle-configuration-xml]\` through \`lifecycle_xml.ParseCanonical\` (#9360 unblocked this import).

All compiled actions are seeded \`BootstrapComplete=true\` so the run dispatches immediately. Production bootstrap walks set this incrementally per bucket; this command is out-of-band of that flow.

## Stack
- #9358 (FilerPersister) — done
- #9359 (gRPC service registration) — done
- #9360 (lifecycle_xml leaf package) — done
- **this PR** — manual end-to-end harness

After this lands, Phase 4 (strip back-stamp from PutBucketLifecycleConfiguration) becomes safe to ship: lifecycle has a working dispatch path through the new code.

## Test plan
- [x] go build ./...
- [x] go test ./weed/shell/...

Manual validation requires a running cluster + S3 server, not exercised here.

## Summary by CodeRabbit

* **New Features**
  * Added a shell command to run S3 lifecycle dispatch across selected shard sets; dispatcher now supports multi-shard runs with per-shard checkpointing and configurable cadence and runtime limits.
* **Behavior**
  * Scheduling now respects object modification time (fixes backdated-object dispatch) and lifecycle identity includes extended metadata for accurate compare-and-swap.
* **CI**
  * New workflow job to run lifecycle integration tests and upload logs on failure.
* **Tests**
  * End-to-end lifecycle tests, shard-selection unit tests, and Makefile tooling added.
* **Documentation**
  * README for lifecycle test setup and shard behavior.

### DIFF
--- a/.github/workflows/s3-go-tests.yml
+++ b/.github/workflows/s3-go-tests.yml
@@ -245,6 +245,51 @@ jobs:
           path: test/s3/retention/weed-test*.log
           retention-days: 3
 
+  s3-lifecycle-tests:
+    name: S3 Lifecycle Tests
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v6
+
+      - name: Set up Go
+        uses: actions/setup-go@v6
+        with:
+          go-version-file: 'go.mod'
+        id: go
+
+      - name: Install SeaweedFS
+        run: |
+          go install -buildvcs=false
+
+      - name: Run S3 Lifecycle Tests
+        timeout-minutes: 12
+        working-directory: test/s3/lifecycle
+        run: |
+          set -x
+          make test-with-server
+
+      - name: Show server logs on failure
+        if: failure()
+        working-directory: test/s3/lifecycle
+        run: |
+          if [ -f weed-test.log ]; then
+            echo "=== Last 200 lines of server logs ==="
+            tail -200 weed-test.log
+          fi
+          ps aux | grep -E "(weed|test)" || true
+          netstat -tlnp 2>/dev/null | grep -E "(8333|9333|8080|8888)" || true
+
+      - name: Upload test logs on failure
+        if: failure()
+        uses: actions/upload-artifact@v7
+        with:
+          name: s3-lifecycle-test-logs
+          path: test/s3/lifecycle/weed-test*.log
+          retention-days: 3
+
   s3-checksum-tests:
     name: S3 Checksum Tests
     runs-on: ubuntu-22.04
--- a/test/s3/lifecycle/Makefile
+++ b/test/s3/lifecycle/Makefile
@@ -1,0 +1,80 @@
+# S3 Lifecycle Test Makefile
+# End-to-end test of the event-driven lifecycle worker, driven by
+# the s3.lifecycle.run-shard shell command.
+
+.PHONY: help build-weed start-server stop-server test test-with-server clean health-check
+
+WEED_BINARY := ../../../weed/weed_binary
+S3_PORT := 8333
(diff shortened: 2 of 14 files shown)


chrislusf:
Bundled the integration test in as you asked.

Three commits on this branch now:
- `b34a754bb` — `s3.lifecycle.run-shard` shell command
- (new) integration test under `test/s3/lifecycle/` (Makefile + Go test + README) and a `s3-lifecycle-tests` job in `s3-go-tests.yml`

The test creates a bucket with a 1-day expiration on a prefix, PUTs the target object, rewrites the entry's Mtime to 30 days ago via `filer_pb.UpdateEntry` (sidestepping the S3 API's `Expiration.Days >= 1` minimum), then runs the shell command for every shard `[0, 16)` and asserts:
- the backdated object is deleted within 30s
- the in-prefix-but-recent object remains

Running every shard keeps the test independent of the `ShardID(bucket, key)` hash function — only that *some* shard's pipeline does the work.

chrislusf:
All four findings addressed in commit f3fab4ebc:

- **MED (gemini)** `pb.GrpcDial` with `context.Background()` could hang the shell — bounded with a 30s timeout.
- **MED (gemini)** silently-skipped malformed XML configs — first three errors now print to the operator with bucket name + reason; rest summarized as a count.
- **MED (gemini)** `SeaweedList` page limit of 4096 silently dropped tail buckets on large clusters — now paginates with `startFrom`, loops until a page comes back short.
- **LOW (coderabbit)** negative `-events` accepted — rejected explicitly; help text already documented 0 as unbounded.

chrislusf:
Added -shards range/set support (commit 1401d1a4b).

```
s3.lifecycle.run-shard -shards 0-15 -s3 host:port    # contiguous range
s3.lifecycle.run-shard -shards 0,3,7 -s3 host:port   # explicit set
s3.lifecycle.run-shard -shard 0 -s3 host:port        # single, unchanged
```

Threads the same model through reader and pipeline:
- `reader.Reader` gains `ShardPredicate func(int) bool` and `StartTsNs`. `Event` carries `ShardID` so consumers route per-shard without rehashing.
- `dispatcher.Pipeline` gains `Shards []int`. When set, `Run` holds one Cursor + Schedule + Dispatcher per shard, opens **one** `SubscribeMetadata` stream with a predicate covering the whole set, and dispatches events to the matching shard's schedule from a **single** goroutine — no per-shard goroutine fan-out.
- `-shard` and `-shards` are mutually exclusive; flag-parsing tests in `command_s3_lifecycle_run_shard_test.go`.

Integration test now uses `-shards 0-15` in a single subprocess invocation instead of a 16-iteration loop.

PR #9362, opened 2026-05-08
feat(worker/s3_lifecycle): plugin handler with admin UI config

chrislusf:
## Summary
The lifecycle scheduler now lives behind the existing pluginworker descriptor pattern (same path Iceberg uses). Operators configure it via the admin UI; no CLI flags, no in-admin scheduler.

- `weed/worker/tasks/s3_lifecycle/`: new package
  - `handler.go`: registers a JobHandler in init(); declares Capability + Descriptor with AdminConfigForm and WorkerConfigForm
  - `config.go`: parses the descriptor values, applies defaults
  - `config_test.go`: parser unit tests
- `weed/plugin/worker/handlers/handlers.go`: blank-imports the new package so the worker binary loads it

### Admin form (cluster shape)
- `workers` 1..16 (default 1) — pipeline goroutines per executing worker
- `s3_grpc_endpoints` (comma list) — where to dispatch LifecycleDelete

### Worker form (operational tuning)
- `dispatch_tick_ms` (default 5000) — pipeline tick cadence
- `checkpoint_tick_ms` (default 30000) — cursor checkpoint cadence
- `refresh_interval_ms` (default 300000) — engine snapshot refresh
- `event_budget` (default 0 = unbounded; the daemon model)

### Detect / Execute
- Detect emits one proposal per cycle when both S3 endpoints + filer addresses are configured
- `MaxExecutionConcurrency=1` — only one daemon runs per worker; a fresh proposal next cycle restarts it after the prior Execute exits
- Execute dials the configured S3 endpoint + filer, builds a `scheduler.Scheduler` with the parsed config, runs it until ctx cancellation

### Reuses (no new core code)
- `weed/s3api/s3lifecycle/scheduler` (Scheduler + LoadCompileInputs)
- `weed/s3api/s3lifecycle/dispatcher` (Pipeline + FilerPersister)
- `weed/s3api/s3lifecycle/{reader,router,engine}`

### Bug fix bundled
- `cbaa3670b` already on PR #9361: Reader now accepts `StartTsNs=0 + Cursor=nil` (the legitimate "fresh subscription" state). CI logs showed the prior over-strict check made the scheduler retry-loop forever without ever subscribing.

### What's left
The integration test still drives the manual `s3.lifecycle.run-shard` shell command; auto-trigger via the admin plugin scheduler needs a config-pre-population path for `mini` so the test can configure `s3_grpc_endpoints` without an interactive UI step. Adding that as a follow-up rather than bundling here.

## Stack
On top of #9361 (shell command + multi-shard Pipeline + reader fix). Will rebase to master when #9361 lands.

## Release Notes

* **New Features**
  * Added S3 lifecycle policy processing and scheduling for automatic bucket lifecycle management.
  * New worker configuration options to tune processing performance, worker threads, and operation intervals.
  * Enhanced cluster context with S3 endpoint discovery for distributed lifecycle operations across nodes.

### DIFF
--- a/weed/admin/dash/plugin_api.go
+++ b/weed/admin/dash/plugin_api.go
@@ -14,7 +14,9 @@ import (
 
 	"github.com/gorilla/mux"
 	"github.com/seaweedfs/seaweedfs/weed/admin/plugin"
+	"github.com/seaweedfs/seaweedfs/weed/cluster"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/plugin_pb"
 	"google.golang.org/protobuf/encoding/protojson"
 	"google.golang.org/protobuf/proto"
@@ -738,6 +740,9 @@ func (s *AdminServer) parseOrBuildClusterContext(raw json.RawMessage) (*plugin_p
 	if len(contextMessage.VolumeGrpcAddresses) == 0 {
 		contextMessage.VolumeGrpcAddresses = append(contextMessage.VolumeGrpcAddresses, fallback.VolumeGrpcAddresses...)
 	}
+	if len(contextMessage.S3GrpcAddresses) == 0 {
+		contextMessage.S3GrpcAddresses = append(contextMessage.S3GrpcAddresses, fallback.S3GrpcAddresses...)
+	}
 	if contextMessage.Metadata == nil {
 		contextMessage.Metadata = map[string]string{}
 	}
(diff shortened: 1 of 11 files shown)


chrislusf:
Reworked per the directive: removed the CLI flags + in-admin scheduler, switched to the pluginworker handler pattern (same as Iceberg). Admin UI now renders the AdminConfigForm + WorkerConfigForm fields declared in the handler's Descriptor.

chrislusf:
Note: previous review comments referencing weed/command/admin.go and weed/command/mini.go are against the earlier admin-scheduler approach. That was replaced with the pluginworker handler pattern (matches Iceberg/Vacuum); the current diff is just three commits — scheduler runtime, run-shard refactor, and the worker handler with AdminConfigForm + WorkerConfigForm.

PR #9376, opened 2026-05-09
feat(s3/lifecycle): make volume-TTL routing operator-explicit (Phase 11)

chrislusf:
## Summary

Volume-TTL routing was a side effect of \`PutBucketLifecycleConfiguration\` (and stripped on \`DeleteBucketLifecycle\`). After this PR it's an explicit operator step:

- \`s3.lifecycle.route-ttl -bucket <name> [-apply]\`
- \`s3.lifecycle.unroute-ttl -bucket <name> [-apply]\`

\`PutBucketLifecycleConfiguration\` becomes O(1): validate XML, reject unimplemented Transition rules, store. No filer.conf read or write, no replication / volume-growth resolution. \`DeleteBucketLifecycle\` is symmetric — clears the stored XML and stops there. A transient DELETE no longer pulls TTL routing out from under in-flight TTL volumes.

## Why explicit

PUT auto-routing tied two unrelated concerns together:
- the lifecycle worker's expiration logic, which is what AWS PUT actually means, and
- volume-TTL routing as an unrelated SeaweedFS performance feature for Expiration.Days rules.

Operators couldn't see what routing was in place, couldn't change it without re-PUTting, and couldn't have lifecycle XML without the volume-TTL fast path. Now they can.

## Behavior

\`route-ttl\` reads the bucket's lifecycle XML, parses canonically, and proposes one filer.conf PathConf per simple Expiration.Days rule (prefix-only, no tag/size filter, status Enabled). Versioned buckets are skipped — TTL volumes destroy noncurrent versions as a unit. Idempotent: already-routed prefixes at the same TTL are a no-op; TTL drift overwrites.

\`unroute-ttl\` removes every PathConf under \`<bucketsPath>/<bucket>/\` with a day-suffixed TTL (the shape \`route-ttl\` writes). Manual operator config (fsync, readonly, etc.) on the same prefixes is left alone.

Both default to dry-run; \`-apply\` writes via \`SaveInsideFiler\`.

## Tests

\`planRouteTTL\`: simple add, tag/size filter skip, disabled / non-Expiration.Days skip, idempotent re-run no-op, TTL drift overwrite.
\`planUnrouteTTL\`: removes day-TTL entries, leaves foreign buckets alone, doesn't touch non-day-TTL operator config, no-op on empty FilerConf.

Dropped \`resolveLifecycleDefaultsFromFilerConf\` and its test — production no longer calls it. Operators get filer-level defaults via empty Replication / VolumeGrowthCount in the operator-written PathConf.

## Backward-compat note

Operators upgrading need to:
1. Run \`s3.lifecycle.route-ttl\` once on each bucket whose lifecycle XML they want backed by TTL volumes (one-shot migration).
2. Use \`s3.lifecycle.unroute-ttl\` (or \`fs.configure -delete\`) when retiring a route — \`DeleteBucketLifecycle\` no longer clears it.

## Summary by CodeRabbit

* **New Features**
  * Added CLI commands to route TTLs from S3 lifecycle rules and to unroute day-suffixed TTL routing, both supporting dry-run and an `-apply` option.
* **Behavior Change**
  * S3 lifecycle PUT/DELETE no longer auto-modifies routing configuration; lifecycle XML is stored/validated but routing is now an explicit operator action.
* **Tests**
  * Added unit tests covering routing/unrouting planning and idempotency.


chrislusf:
Closing in favor of a simpler approach. The S3 server already has the bucket's lifecycle XML at PUT time (it's cached via getBucketConfig), and AssignVolume already accepts an explicit TTL. Resolving the matching Expiration.Days at PUT time and passing the TTL through is cleaner than keeping a separate filer.conf projection in sync via operator commands. New PR forthcoming.

PR #9382, opened 2026-05-09
fix(volume): don't nuke local data on transient IO error (#9378)

### LINKED ISSUE #9378: [question] v.lastIoError would delete volume data?
Reading code of CollectHeartbeat(), if volume hit IO error, in next heartbeat this replica would be [deleted](https://github.com/seaweedfs/seaweedfs/pull/6790/changes#diff-e5309254490c2148e77b5dce11b70d51012e8b8abaf930aaddc120045592d4f3R273)? Is there possibility that multiple replicas of a volume would all be deleted due to some transient hw/network error? If the volume has no remote file, the data would be lost?

Is it more safe to mark replica readonly instead of nuke out, on successive IO error instead of single?

chrislusf:
Fixes #9378.

A single `syscall.EIO` from any volume read/write/delete set `v.lastIoError`, and the next `CollectHeartbeat` then both excluded the volume from the heartbeat *and* called `Volume.Destroy`, physically removing the `.dat`/`.idx`/`.vif`/`.sdx`/`.ldb`/`.rdb` files. A brief NFS / fabric / controller blip hitting several replicas in the same window could cascade into removal of the last healthy copy, with no recovery for non-tiered volumes — exactly the data-loss scenario the issue raises.

This change:

- Tracks consecutive EIOs (`lastIoErrorCount`) and resets it on the next successful op.
- Requires `IoErrorTolerance` (3) consecutive errors before acting, so a single transient blip is forgiven.
- On threshold: marks the replica read-only locally (so further writes fail fast instead of producing more EIOs) and stops announcing it to the master so re-replication kicks in from healthy peers.
- **Never** deletes the local data files. The on-disk copy stays put for operator inspection and recovery — even if the disk is genuinely bad, deletion was always premature.

The TTL-expiration deletion path is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improve handling of I/O failures: replicas are now quarantined only after sustained I/O errors instead of on any single error, and previously-quarantined replicas are re-enabled once I/O recovers.
* **New**
  * Introduced an I/O-error tolerance gate to avoid reacting to transient failures.
* **Tests**
  * Added unit and race tests validating consecutive I/O error tracking and tolerance behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9382)

### DIFF
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -415,34 +415,35 @@ func (s *Store) CollectHeartbeat() *master_pb.Heartbeat {
 			if maxFileKey < curMaxFileKey {
 				maxFileKey = curMaxFileKey
 			}
-			shouldDeleteVolume := false
 
-			if v.lastIoError != nil && v.lastIoErrorCount.Load() >= IoErrorTolerance {
+			if ioErr, ioCount := v.getIoErrorState(); ioErr != nil && ioCount >= IoErrorTolerance {
 				// Sustained EIO: stop announcing this replica so the master
 				// re-replicates from healthy peers, and mark it read-only so
 				// further writes fail fast instead of producing more EIOs.
 				// Never physically delete the data — the disk may be
 				// transiently bad and this could be the last good copy.
-				shouldDeleteVolume = true
 				glog.Warningf("volume %d has %d consecutive IO errors, marking read-only and unreporting from master: %v",
-					v.Id, v.lastIoErrorCount.Load(), v.lastIoError)
-				if !v.noWriteOrDelete {
-					v.noWriteLock.Lock()
-					v.noWriteOrDelete = true
-					v.noWriteLock.Unlock()
-				}
+					v.Id, ioCount, ioErr)
+				v.noWriteLock.Lock()
+				v.noWriteOrDelete = true
+				v.noWriteLock.Unlock()
+				// Skip per-volume size and read-only bookkeeping: a
+				// quarantined replica should not be summed into the
+				// collection's reported total nor counted in the
+				// read-only stats. Recovery via MarkVolumeWritable
+				// resets the error state so it can rejoin.
+				continue
+			}
+
+			shouldDeleteVolume := false
+			if !v.expired(volumeMessage.Size, s.GetVolumeSizeLimit()) {
+				volumeMessages = append(volumeMessages, volumeMessage)
 			} else {
-				if !v.expired(volumeMessage.Size, s.GetVolumeSizeLimit()) {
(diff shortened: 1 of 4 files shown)


chrislusf:
Addressed in c0e1cbe:

- **Data race on `lastIoError` (gemini, high)** — `lastIoError` and `lastIoErrorCount` are now guarded together by a `RWMutex` and accessed only through `noteIoError` / `clearIoError` / `getIoErrorState`. The heartbeat reader sees a consistent snapshot. `go test -race` clean.
- **Collection-size accounting (gemini, high)** — When a volume is quarantined for sustained EIO, the loop now `continue`s, skipping all per-volume size and read-only bookkeeping. The previous branch flipped `shouldDeleteVolume = true`, which made the bottom of the loop subtract a size that was never added — dragging the collection gauge to zero or negative.
- **Unrecoverable state via `MarkVolumeWritable` (gemini, high)** — `MarkVolumeWritable` now also calls `clearIoError`, so an operator can rejoin a quarantined replica. If the disk is still bad, the next failed op re-arms the streak.
- **Reset streak on non-EIO failure (coderabbitai, major)** — `checkReadWriteError` now clears the streak on any non-EIO error. A sequence `EIO, EIO, ENOSPC, EIO` ends with `count == 1`, not 3. New regression test in `TestCheckReadWriteErrorNonEIOResetsStreak`.

Not addressed: gemini's "extend tracking to read path" — `checkReadWriteError` is already called from `volume_read.go:65`, so successful reads do clear the streak today; the comment is stale relative to the current code.

chrislusf:
Both real, both fixed in af32985:

**1. Quarantine wasn't sticky.** Added an `ioErrorQuarantined` flag set the first time `CollectHeartbeat` observes the streak crossing `IoErrorTolerance` and cleared only by `MarkVolumeWritable` (now via a new `resetIoErrorState`). `clearIoError` (called from successful ops) continues to reset just the streak counter, so the heartbeat stays in the suppressed branch even after a stray good read. Recovery is operator-only, as the doc-comment promised. Covered by `TestQuarantineIsSticky`.

**2. Streamed reads bypassed the counter.** `readNeedleDataInto` now threads `n.ReadNeedleData` errors through `checkReadWriteError`, and `ReadNeedleBlob` does the same for `needle.ReadNeedleBlob`. A fully-clean `readNeedleDataInto` also clears the streak (the disk produced clean bytes; CRC failures still surface to the caller). Regression test `TestReadNeedleBlobTracksEIO` uses a fake EIO backend to verify the path bumps the streak by one per failed read.

PR #9409, opened 2026-05-10
test(s3/lifecycle): direct router + engine helper coverage

chrislusf:
Bundles direct tests for several pure helpers in \`router.go\` and accessors on \`engine.Snapshot\`. All previously exercised only indirectly through the larger \`Route\` / \`Compile\` tests, where a helper-level regression would surface as a downstream failure that's harder to diagnose.

24 tests across two files' worth of helpers.

**Router helpers (16 cases):**
- \`successorModTimeFromContainer\` (5): missing ext, empty value, non-numeric garbage, non-positive seconds, positive seconds round-trip
- \`logicalKeyFromVersionPath\` (4): extracts logical key, rejects non-\`.versions\` parent / root-level / no slash / bare-container
- \`isVersionsContainerKey\` (1 table): bucket-root \`.versions\` rejected, trailing \`.versions\` accepted, files inside container rejected
- \`isVersionFolderPath\` (1 table): child of \`.versions\` accepted, container itself rejected, trailing-slash edge documented
- \`isDeleteMarkerEntry\` (2): nil/empty extended false, only literal \`"true"\` matches (\`"True"\`, \`"TRUE"\`, \`"1"\` all rejected)
- \`extractTags\` (3): nil/empty input returns nil, only \`AmzObjectTagging\`-prefixed keys surface, no-tag input returns nil (not empty map)
- \`hasActiveEventDrivenAction\` (2): matches only active+event-driven kinds, scan-only kind rejected, unknown action skipped

**Engine \`Snapshot\` accessors (5):**
- \`BucketVersioned\` reflects the compiled flag, unknown bucket false
- \`BucketActionKeys\` returns all compiled keys, unknown bucket nil
- \`Action\` returns nil for unknown key
- \`AllActions\` enumerates every compiled kind regardless of active state
- \`SnapshotID\` is strictly monotonic across recompiles (the dispatcher's stale-snapshot check depends on this)

### DIFF
--- a/weed/s3api/s3lifecycle/router/helpers_test.go
+++ b/weed/s3api/s3lifecycle/router/helpers_test.go
@@ -1,0 +1,335 @@
+package router
+
+import (
+	"testing"
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3_constants"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3lifecycle"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3lifecycle/engine"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Direct coverage for the pure helpers in router.go that the larger
+// Route* tests exercise indirectly. Pinning each one separately catches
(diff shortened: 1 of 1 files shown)


chrislusf:
Superseded by #9414. After #9410-#9413 landed, the only still-unique content here was router/helpers_test.go; that file is brought forward in #9414.

PR #9416, opened 2026-05-10
shell: expose retention flags on mq.topic.configure

pmiriyev:
## Summary

The `ConfigureTopicRequest` proto already carries a `TopicRetention` message (`retention_seconds`, `enabled`), and both `GetTopicConfigurationResponse` and the admin UI surface it. But the `mq.topic.configure` shell command currently sends `Retention: nil` unconditionally, so CLI / IaC users have no way to set retention without going through the admin UI.

This patch wires three flags into the existing command:

```
mq.topic.configure -namespace <ns> -topic <t> \
    -retention 168h -retentionEnabled
```

| Flag | Type | Notes |
|---|---|---|
| `-retention` | duration | Go duration string (`168h`, `30m`). Mutually exclusive with `-retentionSeconds`. |
| `-retentionSeconds` | int64 | Raw seconds, useful when scripting. |
| `-retentionEnabled` | bool | Toggles retention enforcement. |

## Behavior / compatibility

- **No flag → unchanged.** The command detects whether any retention flag was actually provided via `flag.FlagSet.Visit`, so the request omits `Retention` (sends `nil`) when the user doesn't touch the new flags. Existing scripts that only care about `-partitionCount` keep their prior semantics: server-side retention state is left alone.
- **Any retention flag → request carries both fields.** Combines into one `TopicRetention{seconds, enabled}` so operators that pass only one of the two still get a complete proto.
- **`-retention` + `-retentionSeconds` → hard error.** Ambiguous, rejected up front.
- **Negative values → hard error.**

No proto / server changes required — purely a CLI binding for state the broker already accepts.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./weed/shell/...` clean
- [x] `gofmt -l` clean
- [x] `go test -count=1 -short ./weed/shell/...` passes
- [ ] Manual: `mq.topic.configure -namespace default -topic foo -retention 1h -retentionEnabled` and verify `mq.topic.describe` (or admin UI) shows the retention.
- [ ] Manual: re-run without retention flags — server state for `foo` unchanged.

Happy to add a unit test for the flag-detection logic if useful — wasn't sure whether the existing `weed/shell` tests cover broker-touching commands.

## Summary by CodeRabbit

* **New Features**
  * Topic retention configuration added to the topic configure command with multiple CLI options and expanded help.
* **Bug Fixes**
  * Validation prevents conflicting or negative retention inputs.
  * Broker now detects retention changes independently, avoids unnecessary updates, and preserves existing retention when retention is omitted.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9416)

### DIFF
--- a/weed/mq/broker/broker_grpc_configure.go
+++ b/weed/mq/broker/broker_grpc_configure.go
@@ -63,25 +63,44 @@ func (b *MessageQueueBroker) ConfigureTopic(ctx context.Context, request *mq_pb.
 			schemaChanged = true
 		}
 
-		if !schemaChanged {
+		// Check if retention needs to be updated. Callers that do not want to
+		// touch retention send Retention=nil; we only consider it a change
+		// when the request supplies a non-nil Retention that differs from
+		// the stored value.
+		retentionChanged := request.Retention != nil && !proto.Equal(request.Retention, resp.Retention)
+
+		if !schemaChanged && !retentionChanged {
 			glog.V(0).Infof("existing topic partitions %d: %+v", len(resp.BrokerPartitionAssignments), resp.BrokerPartitionAssignments)
 			return resp, nil
 		}
 
-		// Update schema in existing configuration
-		resp.MessageRecordType = request.MessageRecordType
-		resp.KeyColumns = request.KeyColumns
-		resp.SchemaFormat = request.SchemaFormat
+		if schemaChanged {
+			resp.MessageRecordType = request.MessageRecordType
+			resp.KeyColumns = request.KeyColumns
+			resp.SchemaFormat = request.SchemaFormat
+		}
+		if retentionChanged {
+			resp.Retention = request.Retention
+		}
 
 		if err := b.fca.SaveTopicConfToFiler(t, resp); err != nil {
-			return nil, fmt.Errorf("update topic schemas: %w", err)
+			return nil, fmt.Errorf("update topic conf: %w", err)
 		}
 
 		// Invalidate topic cache since we just updated the topic
 		b.invalidateTopicCache(t)
 
-		glog.V(0).Infof("updated schemas for topic %s", request.Topic)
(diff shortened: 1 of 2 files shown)


pmiriyev:
Pushed a follow-up commit (`443f1208d`) addressing all three review points; force-push includes a rebase onto current `master` so the diff stays clean.

**1. Server-side: detect retention changes in early-return.** Previously `ConfigureTopic` would return without persisting when partition count + schema were unchanged, even if the request supplied a different `Retention`. The early-return branch now checks both schema and retention, persists either or both when they differ, and the log line indicates which fields actually changed.

**2. Server-side: preserve existing retention in the fresh-allocation path.** Capture `prev := resp.Retention` before the `resp = &mq_pb.ConfigureTopicResponse{}` reset, then write `request.Retention` only when it's non-nil; otherwise carry `prev` forward. So partition-count changes (or any path that bypasses the early-return branch) no longer accidentally clear an existing retention configuration when the request omits `Retention`.

**3. CLI: `-retention` / `-retentionSeconds` mutex via `Visit`.** Switched from value-based (`!= 0`) to `flag.FlagSet.Visit`, so an explicit `-retention=0 -retentionSeconds=N` is also rejected.

**4. CLI: partial-flag merge against current state.** When any retention flag is set, the CLI fetches `GetTopicConfiguration` first and fills in the fields the user didn't explicitly provide. So `-retentionEnabled` alone no longer zeros the existing duration, and `-retention 24h` alone preserves the existing `Enabled` flag. When the topic doesn't exist yet, the `GetTopicConfiguration` error is treated as "no current state" and we proceed with just the user-supplied values.

`go build` / `go vet` / `gofmt` / `go test ./weed/shell/... ./weed/mq/broker/...` all clean.

chrislusf:
@pmiriyev how are you using this now?

pmiriyev:
> @pmiriyev how are you using this now?

admin ui, this flags needed for scripts and iac to bypass admin ui and api calls

Issue #6541, opened 2025-02-11
Performance Issue: Inefficient Part Copying in CopyObjectPartHandler - Resource Intensive Upload Instead of Referencing


MatinZare:
Description:

The current implementation of CopyObjectPartHandler in SeaweedFS performs a full copy and re-upload of object parts during multipart uploads. This is significantly slower and more resource-intensive than an approach that would simply create references to the source parts. This inefficiency leads to substantial performance degradation, high RAM consumption, and system overload, particularly when handling large files.

This issue severely impacts integration with applications like Harbor and Artifactory that use SeaweedFS as an S3-compatible storage backend. While these applications manage the upload of individual parts themselves, they then use a multipart upload request with CopyObjectPart to assemble the final object. This results in SeaweedFS reading and writing each part as if it were the entire image layer during the copy part phase, creating an intense amount of read and write operations. For instance, when assembling large image layers (e.g., 6GB) with multiple parts (e.g., 200MB each), the s3 (filer) becomes slow, and consumes excessive RAM, and in my Kubernetes environment, this causes out-of-memory (OOM) errors or readiness probe failures where the filer returned HTTP 5xx errors during heavy load.

Proposed Solution:

The CopyObjectPartHandler should be refactored to create entries in the filer’s data store that reference the source object’s parts(chunks), instead of copying and re-uploading data. This would dramatically improve performance, reduce resource usage, and resolve the observed scalability problems.

As an additional improvement, adding bandwidth and other operation limits to the filer and S3 interface may be beneficial to prevent any single operation from overwhelming the system.

Reproducible Steps:

Attached is a Python script that simulates a multipart upload using CopyObjectPartHandler, reproducing the described performance issues. You can configure the script to test with a large file and various part sizes to observe the problem firsthand. (Note that the provided script simulates the behavior and will only exhibit the issues based on the system’s limitations unless the actual underlying conditions that cause OOM are present)

Impact:

This issue affects the usability and scalability of SeaweedFS, particularly in production environments where large file uploads and integrations with other tools are common. Addressing it will significantly enhance the performance and stability of SeaweedFS as an S3-compatible storage solution.

I am willing to contribute a pull request with the necessary code changes to implement the proposed solution.

Additional Information:

Environment:

- SeaweedFS Version: 3.80 & 3.63
- Deployment Details:
   - 1 master server
   - 2 volume servers (100GB total capacity)
   - 1 filer with the -s3 flag enabled
[...]

PR #9426, opened 2026-05-10
feat(volume): X-Seaweedfs-Original-Size hint skips redundant gunzip

chrislusf:
Stacked on top of #9425 (\`stream-gzip-passthrough\`). Refs #6541.

After #9424 and #9425 the chunk-copy heap profile shows one remaining hot site: the destination volume's \`parseUpload\` runs \`util.DecompressData\` on the forwarded gzipped bytes, just to learn the uncompressed length so it can record \`OriginalDataSize\` in the needle metadata. For a 6-way concurrent 64 MiB UploadPartCopy that decompress-and-discard pass dominated the remaining inuse_space (~297 MiB via \`bytes.Buffer.ReadFrom\` inside \`util.GunzipStream\`).

## What

Add an \`X-Seaweedfs-Original-Size\` header on the multipart part. The s3 chunk-copy fast path always knows the uncompressed size (it's \`chunk.Size\`); when it forwards gzipped bytes, it sets the header. The volume's \`parseUpload\` reads the header and uses the value directly, skipping the decompress-and-discard pass.

The skip is gated:

- Header must be present and parse to a positive int.
- No Content-MD5 verification can be requested — MD5 is computed against the *uncompressed* bytes the s3 client originally sent, so the verification path still needs a real decompress.

If either condition fails, the existing decompress path runs unchanged.

The header prefix is \`X-*\` (not \`Seaweed-*\`) so it doesn't get auto-stored as a needle pair by \`PairNamePrefix\`.

## Backward compatibility

- Old s3 servers don't set the header; volume decompresses as before.
- New s3 servers talking to old volumes: header is ignored; volume decompresses.
- Bad header values (non-numeric, garbage) fall back to the existing decompress path.

## End-to-end repro impact

Building on the post-#9425 baseline (512 MiB src, 6 parallel UploadPartCopy):

| metric                       | + #9425 only  | + this PR    |
|------------------------------|---------------|--------------|
| RSS, round 2                 | 1149 MiB      | **594 MiB**  |
| heap inuse_space, post-test  | 545 MiB       | **349 MiB**  |
| HeapSys (MemStats)           | 1.35 GiB      | 777 MiB      |
| TotalAlloc cumulative        | ~9 GiB        | 3.5 GiB      |

Total reduction from pre-#9420 baseline: **3134 → 594 MiB (-81%)**.

After this PR the only \`bytes.Buffer.ReadFrom\` allocations left in the heap profile come from \`operation.UploadReaderInChunks\` — the source PutObject path the repro itself uses to populate the test object, which is unrelated to chunk-copy.

## Test

\`weed/storage/needle/needle_parse_upload_size_hint_test.go\` exercises five matrix corners and bounds the allocation per case:

- \`hint+no-MD5\`: 755 KB allocated for a 4 MiB gzipped body (skip works)
- \`no-hint\`: 17 MB allocated (decompress runs)
- \`hint+MD5-on-part\`: 17 MB (decompress for MD5)
- \`hint+MD5-on-request\`: 17 MB (decompress for MD5)
- \`garbage-hint\`: 17 MB (fallback)

A regression that re-introduces the unconditional decompress fails the \`hint+no-MD5\` allocation bound.

## Summary by CodeRabbit

* **New Features**
  * Optimized handling of gzip-encoded uploads by honoring original-size hints when safe, avoiding unnecessary decompression and reducing memory/CPU for pre-compressed uploads.

* **Tests**
  * Added tests validating original-size hint parsing, behavior when checksums are present, and that decompression is skipped or performed as expected.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9426)

### DIFF
--- a/weed/storage/needle/needle.go
+++ b/weed/storage/needle/needle.go
@@ -16,6 +16,15 @@ import (
 const (
 	NeedleChecksumSize = 4
 	PairNamePrefix     = "Seaweed-"
+
+	// OriginalSizeHeader is set on multipart parts whose body is already
+	// gzip-encoded by an upstream component (e.g. the s3 chunk-copy fast
+	// path forwarding compressed bytes). It declares the uncompressed
+	// length so parseUpload can record OriginalDataSize without paying a
+	// decompress-then-discard pass just to learn the size — see #6541.
+	// The prefix is X-* (not Seaweed-*) so PairNamePrefix's auto-record
+	// of Seaweed-prefixed headers as needle pairs doesn't pick this up.
+	OriginalSizeHeader = "X-Seaweedfs-Original-Size"
 )
 
 /*
--- a/weed/storage/needle/needle_parse_upload.go
+++ b/weed/storage/needle/needle_parse_upload.go
@@ -56,6 +56,11 @@ type ParsedUpload struct {
 	IsChunkedFile    bool
 	UncompressedData []byte
 	ContentMd5       string
+	// originalSizeHint is the value parsed from OriginalSizeHeader on the
+	// multipart part. If positive and IsGzipped is true, ParseUpload skips
+	// the size-learning DecompressData pass — we already know the
+	// uncompressed length.
+	originalSizeHint int
 }
 
 func ParseUpload(r *http.Request, sizeLimit int64, bytesBuffer *bytes.Buffer) (pu *ParsedUpload, e error) {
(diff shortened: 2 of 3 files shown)


ndianabasi:
I have Seaweedfs notifications on and couldn't help but comment on this
one. There are several known vulnerabilities linked to using headers such
as `X-Seaweedfs-Original-Size` from untrusted clients. If the header is set
by untrusted clients, I recommend you have a look again into the security
implications of using it.

PR #9457, opened 2026-05-12
feat(s3/lifecycle): swap daily_run to engine hash APIs (Phase 4a)

chrislusf:
## Summary

- Replace dailyrun's local hash helpers (\`localReplayContentHash\`, \`localMaxEffectiveTTL\`) with the engine package's canonical \`engine.ReplayContentHash\` / \`engine.MaxEffectiveTTL\` that landed in the Phase 4 view surface (#9447).
- Add \`engine.PromotedHash\` to the cursor's recovery triggers — a partition flip between replay and walk now fires the rule-change branch alongside a \`RuleSetHash\` mismatch.
- Persist the real PromotedHash value on the cursor (was always zero in Phase 2).

## What this is and isn't

This is Phase 4a — engine API swap and PromotedHash infrastructure. The retentionWindow passed to PromotedHash is set to \`MaxEffectiveTTL\` for now, which by construction keeps the promoted set empty and leaves the new trigger dormant. Phase 4b will plumb the real meta-log retention boundary so true scan_only promotions are detected; until then the trigger is forward-compatible no-op.

Cursor on-disk schema is unchanged (PromotedHash was already a field, just written as the zero hash).

## Test plan

- \`go test ./weed/s3api/s3lifecycle/... ./weed/worker/...\` — green
- Existing dailyrun tests (cursor round-trip, drain, dispatch retry, unsupported-rule rejection) all pass against the new hash helpers

## Summary by CodeRabbit

* **Refactor**
  * Consolidated replay hashing and TTL handling into engine-level components, removing localized helper implementations.
  * Updated cursor persistence to track promoted hash alongside existing state for more accurate lifecycle continuity.
  * Streamlined event drain/control flow to ensure clean reader cancellation at run boundaries.

* **Tests**
  * Removed several unit tests covering the local hashing/TTL helpers.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9457)


chrislusf:
Going with **option 3** — document the intentional one-time rewind.

Pushed `65a1f79c3` with a comment in `runShard` explaining the trade. Rationale:

- **daily_replay is off by default** (operators have to flip the algorithm flag on PR #9446). The affected population is early adopters who opted in during the brief Phase 2 window — small.
- **The rewind is bounded to `runNow - maxTTL`**, not a time-zero re-scan. For a typical 30-day expiration this is one day's worth of duplicate meta-log scanning per shard, and `LifecycleDelete` is idempotent on already-deleted objects.
- **Self-healing on save**: the rule-change branch persists the new hash, and the next run resumes steady-state.

Options 1 and 2 trade a permanent legacy hash encoder (or migration shim) for one bounded re-scan. The comment in `runShard` now makes the trade explicit so a future reader doesn't hunt for the rewind cause.

PR #9459, opened 2026-05-12
feat(s3/lifecycle): Phase 4b — daily walker for recovery and steady state

chrislusf:
## Summary

Phase 4b completes the walker side of the daily-replay path. With this PR, \`daily_replay\` services every lifecycle rule kind — including \`ExpirationDate\`, \`ExpiredDeleteMarker\`, \`NewerNoncurrent\`, and replay-eligible rules promoted to \`scan_only\` by retention — rather than failing the run with \`UnsupportedRuleError\` and forcing operators back to the streaming algorithm.

## Behavior change

\`runShard\` now performs two walker invocations under \`cfg.Walker\` (supplied by the worker handler):

- **Recovery branch** (\`RuleSetHash\` or \`PromotedHash\` mismatch): walks \`engine.RecoveryView(snap)\` before rewinding the cursor, so already-due objects across the rewritten rule set fire immediately instead of waiting on the sliding meta-log replay.
- **Steady state** (no recovery, every daily run): walks \`snap.RulesForShard(shardID, retentionWindow)\`'s walk view. Empty walk view skips the call so replay-only deployments don't pay an O(N) bucket walk.

\`router.Route\` already gates replay on \`Mode == ModeEventDriven\`, so walker-bound and \`scan_only\`-promoted rules are silently dropped by replay and picked up exclusively by the walker — no double-dispatch.

## Commits

1. \`Config.RetentionWindow\` plumbing (\`engine.PromotedHash\` retention input).
2. \`WalkerDispatcher\` — \`bootstrap.Dispatcher\` over \`LifecycleClient\`; adds \`VersionID\` to \`bootstrap.Entry\`.
3. \`Config.Walker\` hook in \`runShard\`'s recovery branch.
4. \`WalkBuckets\` — composable per-shard driver with shard-filtered \`ListFunc\`.
5. \`FilerListFunc\` — paginated filer-backed \`ListFunc\`.
6. \`.versions/\` expansion in \`FilerListFunc\` (per-sibling state matching streaming bootstrap).
7. MPU init emission (\`.uploads/<id>\` → \`IsMPUInit\` + \`DestKey\`).
8. Handler wires the walker (\`FilerListFunc\` + \`WalkerDispatcher\` + \`WalkBuckets\` → \`Config.Walker\`).
9. Steady-state walker call + drops \`checkSnapshotForUnsupported\`, deletes \`replayability.go\`/\`_test.go\`.

## Test plan

- \`go test ./weed/s3api/s3lifecycle/... ./weed/worker/tasks/s3_lifecycle/...\` green.
- Component tests pin: dispatcher request shape, hash-mismatch walker invocation, shard filter incl. MPU \`DestKey\`, bucket-error isolation, filer listing recursion, versioned expansion (pointer / explicit-null-bare / newest-fallback / coincidental-folder / skipBare / delete-marker), MPU init with and without \`ExtMultipartObjectKey\`, kill-resume via \`start\`, attribute propagation.

## What's left for Phase 5

- Delete the streaming algorithm path (\`AlgorithmStreaming\`) and the scheduler-side bootstrap walker (now duplicated by \`FilerListFunc\`).


chrislusf:
Pushed `e396e20fa` addressing the review:

- **walker_dispatcher.go nil-response panic** (coderabbit): added a defensive `resp == nil` check between `LifecycleDelete` and the outcome switch, returning a wrapped error so the walk halts cleanly. Pinned by `TestWalkerDispatcher_NilResponseReturnsError`.

- **retentionWindow=maxTTL comment misleading during bootstrap** (gemini): the claim that `promoted` "stays empty" was only true once rules are active. Rewrote the comment to spell out the bootstrap case: `MaxEffectiveTTL` filters by `IsActive` while `PromotedHash` only filters `ModeDisabled`, so during the inactive window `promoted` is non-empty, and the first post-activation run hits the recovery branch — which is the intended bootstrap walk, not a bug.

- **"Phase 2: no walker on rule-change" header in runShard** (coderabbit nit): already rewritten in commit `01d0de6ed` (the steady-state-walker commit) before the review landed — the function header now documents both recovery and steady-state walker invocations explicitly.

- **pagination test** (coderabbit nit): not addressing in this PR — the paginated `listAll` is a verbatim port of the same-named helper in `scheduler/bootstrap.go` which is covered by `scheduler/bootstrap_test.go`'s pagination scenarios; duplicating the test surface here would be churn ahead of the Phase 5 deletion that removes the scheduler copy.

CI was already passing on the prior tip; this push should rerun green.

chrislusf:
Pushed \`468adfada\` addressing the stale-pointer review.

**Structural fix:** \`expandVersionsDir\` now tracks a \`pointerResolved\` flag and falls through to the no-pointer branch (including the explicit-null-bare check) when \`ExtLatestVersionIdKey\` is set but no sibling carries that id. The default \`latestPos = 0\` no longer carries the load implicitly.

**Behavior note:** today's no-pointer fallback also yields \`latestPos = 0\` (newest sibling), so the observable behavior is unchanged. The fix is structural — it protects against a future refinement to the no-pointer fallback (e.g. "newest non-deletemarker sibling") accidentally diverging from the stale-pointer case.

**Test pin:** \`TestFilerListFunc_VersionedExpansionStalePointerFallsBackToNewestSibling\` builds a snapshot whose \`.versions/\` pointer names a ghost id; verifies the newest real sibling becomes IsLatest.

The mirror copy in \`scheduler/bootstrap.go\` has the same shape and the same behavior; not touching it here because Phase 5 deletes that file along with the streaming algorithm.

PR #9480, opened 2026-05-13
fix(storage): prune partial EC shards when sibling disk has healthy .dat (#9478)

### LINKED ISSUE #9478: Volume inaccessible due to incomplete EC files, though volume_x.dat still exists
**Description:**
I am encountering an issue where a specific volume has become inaccessible. It appears that some Erasure Coding (EC) files were deleted or are incomplete in a previous version, even though the main `volume_x.dat` file still exists and is visible in the Volume Server UI.

**Steps to Reproduce:**
1. Check the Volume Server UI, where `volume_x.dat` is listed and visible.
2. Attempt to access data within this volume.
3. The access fails, likely due to the missing or incomplete EC files required for data reconstruction.

**Expected Behavior:**
The volume should either be accessible using the existing data/parity chunks, or the system should provide a way to handle this inconsistency (e.g., fallback to the .dat file if possible, or a tool to clean up the broken EC state).

**Actual Behavior:**
The volume remains inaccessible. The system seems to expect complete EC files for recovery, but they are missing or corrupted.

**Environment & Additional Context:**
- **SeaweedFS Version:** [4.23]
- **OS:** [ Linux ]
- **Setup:** [ 3 Master, 8 Volume Servers]

**All volumes:**
```
find /data/seaweedfs/ -name "data_122.*" -exec du -h {} \;

volume server 1:
2.0M    /data/seaweedfs/volume1/data_122.idx
11G     /data/seaweedfs/volume1/data_122.dat
4.0K    /data/seaweedfs/volume1/data_122.vif
1.8M    /data/seaweedfs/volume1/data_122.ldb

volume server 2:
11G     /data/mnt/sdi/data_122.dat
2.0M    /data/mnt/sdi/data_122.idx
4.0K    /data/mnt/sdi/data_122.vif
1.8M    /data/mnt/sdi/data_122.ldb

volume server 3:
0       /data/seaweedfs/volume1/data_122.dat
4.0K    /data/seaweedfs/volume1/data_122.vif

volume server 4:
1.1G    /data/mnt/sdf/data_122.ec01
2.0M    /data/mnt/sdf/data_122.ecx
0       /data/mnt/sdf/data_122.ecj
4.0K    /data/mnt/sdf/data_122.vif
11G     /data/mnt/sdd/data_122.dat
2.0M    /data/mnt/sdd/data_122.idx
4.0K    /data/mnt/sdd/data_122.vif
1.8M    /data/mnt/sdd/data_122.ldb

volume server 5:
0       /data/seaweedfs/volume1/data_122.dat
4.0K    /data/seaweedfs/volume1/data_122.vif
0       /data/seaweedfs/volume2/data_122.dat
4.0K    /data/seaweedfs/volume2/data_122.vif
```

**Question:**
Is there a recommended way to recover from this state? Can we force the system to read directly from the existing `volume_x.dat`, or is there a command to safely remove the incomplete EC metadata so the volume can be accessed or re-replicated?

chrislusf:
Fixes #9478.

## What goes wrong today

`handleFoundEcxFile` (and `validateEcVolume`) only check for `.dat` in the **same** `DiskLocation` as the EC shards. In a multi-disk volume server an interrupted EC encode can leave `.ec??` + `.ecx` on disk B while the source `.dat` still lives on disk A. The per-disk loader sees no `.dat` next to `.ecx`, classifies the layout as distributed EC ("no .dat, so this is normal"), and mounts the partial shards. The volume server then heartbeats **both** a regular replica AND an EC shard for the same vid, leaving the master with a contradictory view of one volume id.

This is exactly the issue 9478 reporter's volume server 4:

```
volume server 4:
1.1G    /data/mnt/sdf/data_122.ec01    <- partial EC, 1 of 14 shards
2.0M    /data/mnt/sdf/data_122.ecx
0       /data/mnt/sdf/data_122.ecj
4.0K    /data/mnt/sdf/data_122.vif
11G     /data/mnt/sdd/data_122.dat     <- healthy .dat on sibling disk
2.0M    /data/mnt/sdd/data_122.idx
4.0K    /data/mnt/sdd/data_122.vif
1.8M    /data/mnt/sdd/data_122.ldb
```

## Fix

Add a Store-level pass after per-disk loading and before the existing cross-disk reconcile. For every EC volume currently mounted with `shardCount < DataShardsCount`, check whether any sibling disk on the same store holds a `.dat` for the same `(collection, vid)`. If so, unload the EC volume from memory, delete its on-disk artefacts, and push a `DeletedEcShardsChan` message so master forgets the new-shard registration the per-disk pass already emitted on `NewEcShardsChan`, rather than waiting for the next periodic heartbeat.

The gate on `shardCount < DataShardsCount` preserves the rare "full local EC, `.dat` retained" layout when shards happen to live on a different disk than the `.dat` — the per-disk loader already keeps that configuration on a single disk, and pruning it here would be a behaviour regression. Distributed EC volumes (no `.dat` on any disk of this server) fall through unchanged because the `.dat` index never matches.

## Regression test

`weed/storage/store_ec_hybrid_repro_test.go` lays the issue 9478 filesystem state down on two `DiskLocation`s, drives the per-disk EC scan, then runs `pruneIncompleteEcWithSiblingDat` + `reconcileEcShardsAcrossDisks`, and asserts:

  - the partial shard / `.ecx` / `.ecj` are removed from disk
  - `EcShardCount()` on the EC disk goes back to 0
  - the healthy `.dat` on the sibling disk is untouched
  - a `DeletedEcShardsChan` message is queued for master

Failing first to confirm the reproduction, then passing after the fix:

```
$ go test ./weed/storage/ -run TestIssue9478 -v -count=1
=== RUN   TestIssue9478_PartialEcOnSiblingDiskOfHealthyDat
W ec volume 122 (collection="") on .../sdf has only 1 shards (need 10) while a healthy .dat exists on sibling disk .../sdd; cleaning up leftover EC files (issue 9478)
    store_ec_hybrid_repro_test.go:146: after Store-level cleanup: shardFileLeft=false ecxLeft=false ecjLeft=false inMemoryShards=0 deletedEcShardMessages=1
--- PASS: TestIssue9478_PartialEcOnSiblingDiskOfHealthyDat (0.01s)
```

The existing CI `Test` job (`.github/workflows/go.yml`) already runs `go test ./...` from `weed/`, so the new test runs automatically with no workflow changes.

## Recovery for an already-broken cluster

An operator on 4.23 just needs to restart any volume server that has the leftover EC files alongside a healthy `.dat` on a sibling disk. The startup pass scrubs the partial state and tells master to forget it; no shell command is needed. Zero-byte `.dat` shells on other servers (the reporter's servers 3 and 5) are untouched but harmless — they fail to load anyway (`volume_loading.go:152` "volume not initialized"), so they don't register with master, and `volume.fix.replication` will rebuild proper replicas onto them via the normal flow.

## Summary by CodeRabbit

* **Bug Fixes**
  * Store now promptly prunes incomplete erasure-coded shards when a healthy replica exists on another disk, removing leftover shard files and emitting deletion notices while preserving healthy data files.

* **Tests**
  * Added regression tests covering pruning behavior, zero-byte sibling cases, and distributed-EC layouts.

* **API**
  * Exposed access to the recorded source file size for erasure-coded volumes to improve shard-size handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9480)

### DIFF
--- a/seaweed-volume/src/storage/disk_location.rs
+++ b/seaweed-volume/src/storage/disk_location.rs
@@ -304,7 +304,11 @@ impl DiskLocation {
     }
 
     /// Remove all EC-related files for a volume.
-    fn remove_ec_volume_files(&self, collection: &str, vid: VolumeId) {
+    /// `pub(crate)` so the Store-level cross-disk passes in
+    /// `store_ec_reconcile.rs` can call it to scrub partial EC artefacts
+    /// when a healthy `.dat` for the same vid lives on a sibling disk
+    /// (seaweedfs/seaweedfs#9478).
+    pub(crate) fn remove_ec_volume_files(&self, collection: &str, vid: VolumeId) {
         let base = volume_file_name(&self.directory, collection, vid);
         let idx_base = volume_file_name(&self.idx_directory, collection, vid);
         const MAX_SHARD_COUNT: usize = 32;
--- a/seaweed-volume/src/storage/erasure_coding/ec_volume.rs
+++ b/seaweed-volume/src/storage/erasure_coding/ec_volume.rs
@@ -120,10 +120,15 @@ impl EcVolume {
             shards.push(None);
         }
 
-        // Read expire_at_sec and version from .vif if present (matches Go's MaybeLoadVolumeInfo).
-        // Prefer the data dir; fall back to the idx dir for the
-        // cross-disk reconcile case (#9212 / #9244).
-        let (expire_at_sec, vif_version) = {
+        // Read expire_at_sec, version, and dat_file_size from .vif if
+        // present (matches Go's MaybeLoadVolumeInfo). Prefer the data
+        // dir; fall back to the idx dir for the cross-disk reconcile
+        // case (#9212 / #9244). `dat_file_size` is the source .dat
+        // size at encode time, used both by `locate_ec_shard_needle`
+        // for shard-size math and by the Store-level prune in
+        // `store_ec_reconcile.rs` to verify a sibling-disk .dat is
+        // plausibly the encoding source (#9478).
+        let (expire_at_sec, vif_version, vif_dat_file_size) = {
             let vif_path = locate_vif_path(dir, dir_idx, collection, volume_id);
             if let Ok(vif_content) = std::fs::read_to_string(&vif_path) {
                 if let Ok(vif_info) =
@@ -134,12 +139,12 @@ impl EcVolume {
                     } else {
                         Version::current()
                     };
-                    (vif_info.expire_at_sec, ver)
+                    (vif_info.expire_at_sec, ver, vif_info.dat_file_size)
                 } else {
-                    (0, Version::current())
+                    (0, Version::current(), 0)
                 }
             } else {
-                (0, Version::current())
+                (0, Version::current(), 0)
             }
         };
 
@@ -150,7 +155,7 @@ impl EcVolume {
             dir_idx: dir_idx.to_string(),
             version: vif_version,
             shards,
-            dat_file_size: 0,
+            dat_file_size: vif_dat_file_size,
             data_shards,
             parity_shards,
             ecx_file: None,
--- a/seaweed-volume/src/storage/store.rs
+++ b/seaweed-volume/src/storage/store.rs
@@ -85,6 +85,17 @@ impl Store {
 
         self.locations.push(loc);
 
+        // First scrub partial EC artefacts left on one disk by an
+        // interrupted encode while the source .dat still lives on a
+        // sibling disk of the same store. The per-disk loader cannot
+        // see the sibling .dat and so loads the partial shards as if
+        // they were a distributed-EC layout, which makes the volume
+        // server heartbeat both a regular replica and an EC shard set
+        // for the same vid (seaweedfs/seaweedfs#9478). Running before
+        // the cross-disk reconcile keeps that pass from later
+        // re-loading shards we just cleaned up.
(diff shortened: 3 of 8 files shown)


chrislusf:
Two notes that came up in review.

**Why this lives on the volume server, not in `weed shell` or an EC worker.** The single-disk version of this exact cleanup already runs at startup via `handleFoundEcxFile` → `validateEcVolume` → `removeEcVolumeFiles` in `weed/storage/disk_location_ec.go`. Moving only the multi-disk variant out would mean the single-disk case auto-recovers on restart while the multi-disk case needs operator action — the two would drift, and operators would have to know which knob to pull. Startup-time cleanup also runs before master registration, so the master never sees the contradictory "regular replica + EC shard for the same vid" state. A `weed shell` command and a master-side detector are still useful complements ("fix without restart", "alert on hybrid state across the cluster") but they don't replace this.

**Rust port included.** `seaweed-volume/src/storage/disk_location.rs` `handle_found_ecx_file` has the same single-disk-only check. Added `prune_incomplete_ec_with_sibling_dat` to `store_ec_reconcile.rs` and calls before `reconcile_ec_shards_across_disks` in both `Store::add_location` and `Store::load_new_volumes`. The Rust heartbeat path diff-emits deletes from the next `ec_volumes` snapshot, so no explicit delete-channel push is needed there. Tests cover both the issue 9478 layout and a distributed-EC layout that must be left alone.

PR #9544, opened 2026-05-18
fix(ec_balance): skip volumes with pending erasure_coding tasks

chrislusf:
## Summary

- The ec_balance detector built its move list directly from topology with no check against in-flight EC encoding or repair work. A volume that EC encoding was still trying to complete (or retrying after a failed mount) could see its shards moved by ec_balance in between attempts.
- The race produces exactly the symptom operators see in multi-disk EC clusters: a failed encode of a volume leaves a partial shard layout, ec_balance moves a shard to a different node, the encode retries against the now-rearranged layout, and the final state is orphan + duplicate shards spread across nodes — even though no single component "fails."
- After ``collectECCollections``, filter each collection's volume id list against ``clusterInfo.ActiveTopology.HasTask(vid, TaskTypeErasureCoding)``. ``TaskTypeErasureCoding`` covers both initial encode and repair, so one filter closes both races. Phases 1-3 use the filtered map directly; phase 4 (global) already honors ``allowedVids``, so removing a vid from the map also removes it from ``allowedVids``.

## Test plan

- [x] ``go test ./weed/worker/tasks/ec_balance/...``
- [x] New ``TestDetection_ExcludesVolumesWithPendingErasureCodingTask`` — a pending encoding task on one of two imbalanced volumes; only the other one gets moves.
- [x] New ``TestDetection_NoExclusionWhenNoActiveTask`` — both volumes get moves when no encoding tasks are pending, regression guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Exclude volumes that have active or pending erasure-coding tasks from rebalancing detection, skip and log them, and avoid producing results when no eligible collections remain.
* **Tests**
  * Added regression tests covering exclusion behavior and ensuring detection still produces moves when no active tasks exist.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9544?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/weed/worker/tasks/ec_balance/detection.go
+++ b/weed/worker/tasks/ec_balance/detection.go
@@ -7,7 +7,6 @@ import (
 	"sort"
 	"time"
 
-	"github.com/seaweedfs/seaweedfs/weed/admin/topology"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/worker_pb"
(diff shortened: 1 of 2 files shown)


chrislusf:
Closing — on review, this PR is redundant with the existing scheduler design.

LaneDefault (vacuum, balance, ec_balance, erasure_coding, admin_script) already serialises detection-and-queuing of its job types under a single admin lock (`runLaneSchedulerIterationLocked`), and `MaintenanceQueue.activeTaskForVolume` blocks cross-type assignment per volume. The remaining cross-cycle window (failed encode → ec_balance runs in a later cycle while the operator hasn't intervened) is intentional behaviour, not a race that an in-detector `HasAnyTask` filter should paper over.

If a follow-up is needed, the right place is making encoding retries persistent in HasAnyTask-visible state, not the ec_balance detector.

Companion enterprise branch retired locally.

PR #9560, opened 2026-05-19
[CheckDisk]: implement disk health detection

MilanFun:
SeaweedFS currently lacks a built‑in mechanism for directly checking disk health or status.
The existing implementation relies on indirect metrics (statvfs), which was used only for obtaining disk capacity (total and available space). Moreover, even if a statvfs call returned an error, no corrective actions would have been taken — the system would simply ignore the error or proceed without any disk status check. As a result, statvfs cannot reliably detect degraded, stuck, or read‑only disks.

In our deployment, we observe a steady growth of goroutines stuck in the select state — they never exit or time out, leading to memory leaks, lock contention, and unstable volume operations.

This PR introduces a draft version that demonstrates the basic ability to check disk status. We have also added flapping protection to prevent rapid disk state toggling (offline → online → offline). This avoids volume rebalancing storms, excessive retries, and log spam caused by transient or intermittent disk issues. 

For now, the implementation uses indirect file system information, but we are actively researching alternative system calls (such as ioctl, fsync, open, or direct block device queries, /proc/pressure/io and etc) that can provide more reliable, low‑level disk status without relying on statvfs.

This Draft is a first step toward a robust disk health detection mechanism, which is critical for preventing traffic routing to unhealthy volume servers.

# How is the PR tested?
Environment setup:
- A standalone Master server was started (with replication factor 001 )
- A Filer server was launched
- Two Volume server instances were started

Test procedure:
- The mounted directory used by the Volume server for data storage was moved (mv) to another location, effectively taking it out of service (simulating a disk failure or removal)
- Observed behavior:
- The Master server successfully detected the disk failure using Heartbeat mechanism 
- The affected Volume was automatically removed from the cluster topology
- All files remained accessible and available to clients
- No data loss or service disruption occurred
- The system continued operating normally with the remaining healthy volume(s)

Result:
- The disk status detection worked as expected
- The cluster correctly rebalanced/responded to the topology change
- Data availability was maintained throughout the test

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added disk health monitoring with error detection and reporting
  * Added IO latency monitoring to identify disk performance degradation
  * New Prometheus metric for tracking disk error status

* **Bug Fixes**
  * Disks with errors are now properly excluded from cluster operations and heartbeat responses

* **Tests**
  * Added tests for disk status validation under success and failure conditions

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9560?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


MilanFun:
Add disk IO latency probing to detect degraded disks.

The existing statfs check can tell whether the filesystem responds, but it may not catch disks that are technically alive while IO latency is badly degraded. This adds a lightweight write+fsync probe and tracks latency using an EWMA-style baseline, inspired by the decayed-history approach used in Linux loadavg - [link](https://github.com/torvalds/linux/blob/master/kernel/sched/loadavg.c)

A disk is only marked unhealthy after several consecutive slow probes, so short IO spikes should not immediately remove it from the topology.

chrislusf:
A disk could be slow due to heavy load. It would only make the situation worse if removing them.

MilanFun:
That is true for short load spikes, which is why the check uses adaptive latency tracking and consecutive slow detections instead of reacting to a single slow probe.

The goal is to detect sustained degradation or unresponsive storage, not temporary load.

Also, if a disk is excluded, traffic will be redirected to other available volume servers, so the load is redistributed rather than amplified on the same disk.

We could also make this behavior configurable by adding a volume server startup option, so users can choose whether IO latency degradation should affect disk availability or only be reported as a warning metric.

MilanFun:
@chrislusf Hi, I hope you're well. Disk I/O probe now configurable and disabled by default. Please review this changes, it's extremely important for us!

chrislusf:
* The configurable parameters are too low level to understand. I would avoid exposing these to regular users. These may be easier to understand:
```
-volume.disk.io.probe=false
-volume.disk.io.timeout=2s

-volume.hdd.io.slow.latency=500ms
-volume.ssd.io.slow.latency=100ms
-volume.nvme.io.slow.latency=50ms

-volume.disk.io.slow.percent=20
-volume.disk.io.error.percent=10
-volume.disk.io.window=1m
-volume.disk.io.min.samples=10
```

* Dumping all the parameters to the command line is not user friendly. Maybe scaffold a `volume.toml` file, similar to `master.toml` from "weed scaffold -conf=master".
* Also, different disk types would have different definition of slowness.
```
[volume.disk.io]
timeout = 2s
```

MilanFun:
Hi @chrislusf,

I updated the disk health detection implementation to make it more conservative and safer for production use.

The main changes are:
- Disk IO probing is still disabled by default.
- IO latency probing is now disabled automatically when multiple volume directories are configured, because the current implementation is intended for a single checked mount only.
- Transient statfs failures preserve the last known good capacity until the disk is actually marked unhealthy.
- Recovery now requires repeated successful statfs checks to avoid flapping.
- IO probe file descriptors are no longer reused, so remount/rebind cases are not hidden by a stale fd.
- Timed-out IO probes no longer cause repeated probe goroutine creation.
- Unavailable disk locations are excluded from new volume allocation.

This feature is important for our deployment because we need the volume server to stop advertising unhealthy storage instead of continuing to route traffic to a broken or stuck disk.

chrislusf:
> Hi @chrislusf,
> 
> I updated the disk health detection implementation to make it more conservative and safer for production use.
> 
> The main changes are:
> 
> * Disk IO probing is still disabled by default.
> * IO latency probing is now disabled automatically when multiple volume directories are configured, because the current implementation is intended for a single checked mount only.
> * Transient statfs failures preserve the last known good capacity until the disk is actually marked unhealthy.
> * Recovery now requires repeated successful statfs checks to avoid flapping.
> * IO probe file descriptors are no longer reused, so remount/rebind cases are not hidden by a stale fd.
> * Timed-out IO probes no longer cause repeated probe goroutine creation.
> * Unavailable disk locations are excluded from new volume allocation.
> 
> This feature is important for our deployment because we need the volume server to stop advertising unhealthy storage instead of continuing to route traffic to a broken or stuck disk.

what is the plan for https://github.com/seaweedfs/seaweedfs/pull/9560#issuecomment-4538327341 ?

MilanFun:
> > Hi @chrislusf,
> > I updated the disk health detection implementation to make it more conservative and safer for production use.
> > The main changes are:
> > 
> > * Disk IO probing is still disabled by default.
> > * IO latency probing is now disabled automatically when multiple volume directories are configured, because the current implementation is intended for a single checked mount only.
> > * Transient statfs failures preserve the last known good capacity until the disk is actually marked unhealthy.
> > * Recovery now requires repeated successful statfs checks to avoid flapping.
> > * IO probe file descriptors are no longer reused, so remount/rebind cases are not hidden by a stale fd.
> > * Timed-out IO probes no longer cause repeated probe goroutine creation.
> > * Unavailable disk locations are excluded from new volume allocation.
> > 
> > This feature is important for our deployment because we need the volume server to stop advertising unhealthy storage instead of continuing to route traffic to a broken or stuck disk.
> 
> what is the plan for [#9560 (comment)](https://github.com/seaweedfs/seaweedfs/pull/9560#issuecomment-4538327341) ?

Thanks, I understand.

I can move part of the low-level tuning options into config, for example:

- disk.io.slow.percent
- disk.io.error.percent
- disk.io.window
- disk.io.min.samples
- disk.io.max.stat.failures
- disk.io.recovery.coef

and keep only the main/high-level options in CLI.

MilanFun:
Hi @chrislusf,

I’ve addressed the previous feedback and added volume.toml support for disk health monitoring configuration. Users can now control probe behavior and failure thresholds through configuration rather than startup flags.

The implementation keeps backward compatibility with existing deployments and preserves the current defaults when no settings are specified.

Could you please take another look when you have a chance?

chrislusf:
the CLI options can be removed.
The options in toml files can be overwritten via env variables.

MilanFun:
@chrislusf Thanks for the clarification. I reworked this

PR #9590, opened 2026-05-20
Samba-over-FUSE integration test and distributed-lock handoff fixes

chrislusf:
Adds an integration test that serves a SeaweedFS FUSE mount over SMB
(smbd) and drives it with smbclient, plus two fixes to the distributed
lock client found while building the locking coverage.

Commits:
- test(mount): Samba harness and battery. Functional checks (round-trips,
  directories, rename, large-file chunking, recursive upload,
  cross-protocol consistency, deletes) plus a second -dlm mount for
  locking (POSIX fcntl byte-range locks, distributed-lock write
  coordination, concurrent writers). Runs locally via test/samba/run.sh
  or in Docker via the compose file; wired into CI as
  samba-integration.yml.
- fix(cluster): release the distributed lock without racing the renewal
  goroutine. Stop() could read a renewToken that an in-flight renewal had
  just rotated, send a stale unlock, and leave the lock lingering until
  its TTL - stalling other mounts on the same file.
- fix(cluster): poll for lock acquisition at a steady cadence instead of
  util.RetryUntil's growing backoff, so a waiter picks up a freed lock
  promptly.

Known limitation pinned by the test: two cross-mount handoff checks fail
because the distributed lock is released on FUSE Release, which the
kernel can delay under contention. Once Release fires the lock now drops
within ~2ms, so the fixes hold where Release is prompt; bounding the hold
independent of Release is left as a follow-up.

## Summary by CodeRabbit

* **New Features**
  * Improved Samba/SMB locking and concurrent-file access reliability

* **Tests**
  * Comprehensive Samba integration test suite for SMB ops, locking/concurrency, cross-protocol checks, and large-file scenarios
  * Containerized end-to-end test runner and CI-enabled execution

* **Documentation**
  * New README documenting test scope, execution, and known caveats

* **Chores**
  * Added CI workflow and container orchestration for automated Samba integration runs

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9590?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/samba-integration.yml
+++ b/.github/workflows/samba-integration.yml
@@ -6,6 +6,7 @@ on:
     paths:
       - 'weed/mount/**'
       - 'weed/filer/**'
+      - 'weed/cluster/**'
       - 'test/samba/**'
       - '.github/workflows/samba-integration.yml'
   pull_request:
@@ -13,6 +14,7 @@ on:
     paths:
       - 'weed/mount/**'
       - 'weed/filer/**'
+      - 'weed/cluster/**'
       - 'test/samba/**'
       - '.github/workflows/samba-integration.yml'
   workflow_dispatch:
--- a/.github/workflows/samba-integration.yml
+++ b/.github/workflows/samba-integration.yml
@@ -1,0 +1,118 @@
+name: "Samba on FUSE Integration"
+
+on:
+  push:
+    branches: [ master, main ]
+    paths:
+      - 'weed/mount/**'
+      - 'weed/filer/**'
+      - 'test/samba/**'
+      - '.github/workflows/samba-integration.yml'
+  pull_request:
+    branches: [ master, main ]
+    paths:
+      - 'weed/mount/**'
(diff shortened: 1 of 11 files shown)


chrislusf:
Second review pass:

- `lock_client.go`: `StartLongLivedLock`'s renewal loop slept uninterruptibly between attempts (up to 2.5×lockTTL while unlocked), which could outlast `Stop()`'s bounded wait and break the shutdown sync. Now it sleeps on a timer with a select on `cancelCh`, so the goroutine exits immediately on `Stop()` and `renewalDone` closes before the timeout.

- Workflow build-input paths (`docker/**`, `go.mod`, `go.sum`): leaving these out for consistency with `pjdfstest.yml`, which has the identical build-from-`docker/` pattern and the same path set. This workflow already covers more (`weed/cluster/**` added). Worth doing repo-wide if desired, rather than diverging here.

PR #9627, opened 2026-05-23
filer: serialize same-path mutations with a local lock

chrislusf:
First of a three-part series that takes the S3 object-write path off the distributed lock and onto route-by-key + filer-local serialization.

## Problem

`Filer.CreateEntry` is an unlocked `FindEntry`-then-`InsertEntry`/`UpdateEntry`. Two concurrent creates to the same path race: `OExcl` (create-if-absent) can admit two creators, and there is no atomic check-then-act for conditional writes. This is exactly why the S3 path and `AppendToEntry` reach for the distributed lock — the filer can't serialize a key on its own.

## Change

Add a per-path exclusive lock to the `CreateEntry` gRPC handler using the existing `util.LockTable[util.FullPath]`, which already evicts idle keys so the table stays bounded. The read and the write now run inside one exclusive critical section on the filer.

This local lock is the authoritative serialization point only once a key's writes are routed to its owner filer (the follow-up gateway PR does that, using the lock-ring view from #9626). On its own, this PR already removes the single-filer check-then-act race.

`AppendToEntry` — itself a read-modify-write that previously borrowed the distributed lock — switches to the same per-path lock. There are no in-tree callers constructing `AppendToEntryRequest`; the operation now serializes within the owner filer, consistent with the route-by-key model.

## Test

`TestCreateEntryOExclSerialized` fires concurrent `OExcl` creates for one path through the real handler (with a widened check-then-act window) and asserts exactly one winner.

## Series

1. **this PR** — filer serializes same-path mutations locally
2. add a `Condition` field to `CreateEntryRequest` so the filer evaluates `If-Match`/`If-None-Match`/time conditions under the lock
3. S3 gateway routes non-versioned object writes to the owner filer and uses these instead of the distributed lock

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced serialization of concurrent file mutations on the same path to ensure atomic and consistent behavior for simultaneous file creation and append operations across the distributed file system.

* **Tests**
  * Added test coverage for exclusive file creation under concurrent access and improved test utilities for rename operation timing scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9627?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/weed/server/filer_grpc_server.go
+++ b/weed/server/filer_grpc_server.go
@@ -9,8 +9,6 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/seaweedfs/seaweedfs/weed/cluster"
-
 	"github.com/seaweedfs/seaweedfs/weed/filer"
 	"github.com/seaweedfs/seaweedfs/weed/glog"
 	"github.com/seaweedfs/seaweedfs/weed/operation"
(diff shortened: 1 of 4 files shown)


chrislusf:
Superseded by #9639 (merged), which is this same per-path lock re-cut onto master as part of the filer-primitive foundation.

PR #9628, opened 2026-05-23
filer: evaluate a write precondition in CreateEntry

chrislusf:
Second of the series taking the S3 object-write path off the distributed lock. Builds on #9627 (per-path serialization).

## Change

Add an optional `WriteCondition` to `CreateEntryRequest`. When set, the `CreateEntry` handler evaluates it against the current entry **while holding the per-path lock from #9627**, so the precondition check and the write are a single atomic critical section on the filer. A failed condition returns the new `FilerError PRECONDITION_FAILED`.

The condition is one primitive:

| Kind | Fails when |
|------|------------|
| `IF_NOT_EXISTS` | entry exists (`If-None-Match: *`) |
| `IF_EXISTS` | entry absent (`If-Match: *`) |
| `IF_ETAG_MATCH` | absent or etag != expected (`If-Match: <etag>`) |
| `IF_ETAG_NOT_MATCH` | present and etag == expected (`If-None-Match: <etag>`) |
| `IF_UNMODIFIED_SINCE` | present and mtime > bound |
| `IF_MODIFIED_SINCE` | present and mtime <= bound |

The caller reduces request semantics (e.g. RFC 7232 precedence among headers) to one primitive; the filer just compares. This keeps S3 logic out of the filer. ETag comparison mirrors the gateway's precedence — the stored Seaweed ETag extended attribute, then the Md5/chunk fallback — so conditional results match what the gateway computes today.

This is a no-CAS design: correctness comes from routing a key's writes to its owner filer and the per-path lock, not from store-level compare-and-swap (no store has it). The S3 gateway wiring is the next PR.

## Proto

Additive only: a new optional field on `CreateEntryRequest`, a new `WriteCondition` message, and a new `FilerError` value. No RPC added, so `filer_grpc.pb.go` is unchanged. Old clients omit the field and get current behavior.

## Test

- `TestWriteConditionSatisfied` — table-driven over every kind.
- `TestStoredEntryETag` — the gateway-matching ETag precedence.
- `TestCreateEntryConditionEnforced` — through the real handler: matching `If-Match` overwrites, non-matching returns `PRECONDITION_FAILED`.

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -222,6 +222,29 @@ message CreateEntryRequest {
     bool is_from_other_cluster = 4;
     repeated int32 signatures = 5;
     bool skip_check_parent_directory = 6;
+    // Optional precondition evaluated against the current entry atomically with
+    // the write, under the filer's per-path lock. The caller must route the
+    // key's writes to this entry's owner filer for the check to be authoritative.
+    WriteCondition condition = 7;
+}
+
+// WriteCondition is a single precondition the filer evaluates against the
+// existing entry before writing. The client reduces request semantics (e.g.
+// RFC 7232 precedence) to one primitive; the filer just compares. A failed
+// condition returns FilerError PRECONDITION_FAILED.
+message WriteCondition {
+    enum Kind {
+        NONE                = 0; // unconditional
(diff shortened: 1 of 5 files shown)


chrislusf:
Superseded by #9650 (merged) — the write precondition, re-cut onto master with the richer clause form.

PR #9634, opened 2026-05-23
s3: route suspended-versioning PutObject off the DLM

chrislusf:
Routes **suspended-versioning** PutObject off the distributed lock, completing the object-key unification: normal, suspended, and versioned writes to the same object now all resolve one owner and serialize on the object key.

> **Stacked PR** on the series (#9626–#9633). Merge after them.

## Why suspended fits the object key

A suspended-versioning PUT writes the "null" version to the **main object path** — a single-entry overwrite, exactly like a non-versioned PUT. The only extra work is rewriting `IsLatest=false` on existing versions, which is **best-effort** ("recoverable on the next list-versions resync"), so it doesn't belong inside the atomic boundary.

So suspended PUT routes on the object key like a normal PUT:
- `putSuspendedVersioningObject` calls `putToFiler` with **no** `afterCreate` hook, so the route-by-key path takes the main write to `owner(object)` with the precondition and skips the distributed lock;
- `updateIsLatestFlagsForSuspendedVersioning` runs best-effort **after** the write rather than as an in-lock hook.

`routedObjectOwner` now excludes only versioning-**enabled** buckets (whose writes go to `.versions`) instead of all versioning-configured buckets, so suspended and unversioned writes — both on the main object path — share the same object-key owner and lock. This closes the cross-state gap: a normal write and a versioned write to the same object can no longer land on different owners during a versioning-state change.

## Stays on the lock

Object-lock buckets, and suspended **DELETE** (delete null version + create a delete marker — genuinely multi-step across the object path and `.versions`).

## Tests

Full `weed/s3api` suite passes locally; suspended-versioning behavior is covered by the CI versioning suite.


chrislusf:
Subsumed by #9631: with the unified putToFiler (create routes to the owner + runs afterCreate), suspended PutObject already routes off the distributed lock; its IsLatest sweep runs as the best-effort afterCreate. The null-version pre-delete is a best-effort cleanup not worth routing separately.

PR #9636, opened 2026-05-23
s3: route non-conditional suspended DELETE off the DLM

chrislusf:
Closes the suspended-versioning race: now that suspended PUT routes on the object key (per-path entry lock), a suspended DELETE left on the distributed lock wouldn't serialize against it — the two locks are independent. This routes suspended DELETE on the object key too.

A suspended DELETE removes the null version (the **main object** entry) and adds a delete marker. `FinalizeVersionedWrite` gains an optional `delete_path` it deletes under the same object-key lock before flipping the pointer, so "delete null + add marker" is one atomic step on the owner. `createDeleteMarker` threads it through (empty for the versioning-enabled marker path; the null path for suspended).

Scope: routed only when there's **no `If-Match`** — that condition targets the main object, not the `.versions` pointer the op evaluates, so a conditional suspended delete stays on the lock path. Object-lock can't apply (it requires versioning enabled, never suspended).

Stacked on the routing series + the object-lock-gate PR; merge after them.

## Release Notes

* **New Features**
  * Added atomic finalization for versioned S3 writes with optional preconditions
  * Added write precondition support (If-Match, If-Not-Exists, If-Exists) for conditional object operations
  * Added direct object routing to bypass distributed locks on eligible writes, improving write performance
  * Added idle heartbeat capability for metadata subscriptions

* **Improvements**
  * Enhanced conditional request handling with improved ETag validation
  * Improved atomicity guarantees for concurrent entry mutations  
  * Better support for versioned multipart uploads and lifecycle operations

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9636?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


chrislusf:
Subsumed by #9633: the suspended DeleteObject path (delete null version + create delete marker) now routes off the distributed lock via createDeleteMarker's routed RECOMPUTE_LATEST finalize and the versioned/suspended branch in DeleteObjectHandler. No separate change needed.

PR #9637, opened 2026-05-23
s3: route versioned MPU-complete + lifecycle markers off the DLM

chrislusf:
Both of these flip the `.versions` latest pointer **off** the object-key lock, racing routed versioned/suspended writes (the per-path entry lock and the distributed lock don't serialize against each other).

- **CompleteMultipartUpload (versioning enabled)** now routes its latest-pointer update through `routedVersionedFinalize` on the object's owner, **carrying the precondition** so the filer evaluates it atomically under the per-path lock — same shape as the versioned `PutObject` path (`versionedAfterCreate`). The completion then takes **no distributed lock at all**; only object-lock buckets, suspended versioning, unsupported preconditions (time-based, ETag lists, weak ETags, both `If-Match`+`If-None-Match`), or a missing ring view fall back to the lock. (Suspended MPU only writes the main object entry — last-writer-wins, benign — and is unchanged; non-versioned MPU was already routed.)
- The **lifecycle expiration delete-marker** (enabled and suspended) resolves the object owner and routes its marker pointer flip the same way.

Both fall back to the lock-based `updateLatestVersionInDirectory` when the owner isn't known. Cold paths, but the pointer race is real, and the fix reuses the existing op.

Stacked on the routing series; merge after it.

## Summary by CodeRabbit

* **New Features**
  * S3 write preconditions for creates/deletes and atomic versioned-write finalization with precondition support; server now returns a machine-readable precondition-failed error.
  * Faster route-by-key object writes that bypass distributed locks when safe.
  * Metadata subscription can opt into idle heartbeats.

* **Performance Improvements**
  * Faster PUT/COPY/DELETE via direct routing to object owners.

* **Bug Fixes**
  * More consistent delete-marker/version-pointer handling across versioning states.

* **Tests**
  * Added unit tests covering routing, conditions, and versioned finalize behavior.


chrislusf:
Subsumed by #9632: completeMultipartUpload routing covers the versioned path too (routedVersionedFinalize for the pointer flip, routedMkFile for non-versioned/suspended object writes), all in one change to the shared completion function.

PR #9668, opened 2026-05-25
mount,filer: re-assert POSIX locks via keepalive (ownership migration + restart)

chrislusf:
Stacked on #9665 (mount) → #9666 (lease) → #9664 (RPC) → … . This is the final piece — it makes cross-mount lock reaping live end-to-end.

The filer side (#9666) reaps the locks of leased sessions that stop renewing. This adds the mount side that does the renewing:

- The mount tracks the inode keys it currently holds locks on (`posixKeepalive`), added on grant (both flock and fcntl) and dropped only on owner release (flush/release), not on a partial range unlock — a superset, so a still-held lock is never under-renewed.
- A background loop (`loopRenewPosixLeases`, started from `StartBackgroundTasks` only when `-posixLock` is set) sends `KEEP_ALIVE(Sid)` for each held key every 5s, reaching each key's owner filer via forwarding — within the filer's 15s TTL.

Net effect with the full stack: a live `-posixLock` mount keeps its lease and is never reaped; a crashed or partitioned one stops renewing and each owner filer reclaims its locks after the TTL.

Tests cover the keepalive tracker (add/remove across both namespaces, snapshot). The renewal RPC and reaping are covered by #9666's handler/manager tests; full two-mount integration (kill one mid-lock, assert the survivor acquires after expiry) lands as a cluster integration test next.

## Summary by CodeRabbit

* **New Features**
  * Added POSIX lock state tracking on mounts to preserve locks across ownership changes and server restarts.
  * Implemented periodic lease renewal mechanism to ensure locks remain valid during long-running operations and prevent unexpected expiration.

* **Tests**
  * Added comprehensive unit tests for lock reassertion behavior, lease renewal, and idempotency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9668?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/weed/filer/posixlock/manager.go
+++ b/weed/filer/posixlock/manager.go
@@ -87,6 +87,69 @@ func (m *Manager) TryLock(key string, lk Range) (Range, bool) {
 	return Range{}, true
 }
 
+// Track records a lock the server already granted, without arbitration. A mount
+// uses it to mirror its own held locks so it can re-assert them to the inode's
+// current owner filer after an ownership change or owner restart.
+func (m *Manager) Track(key string, lk Range) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	s, ok := m.byKey[key]
+	if !ok {
+		s = &Set{}
+		m.byKey[key] = s
+	}
+	s.Grant(lk)
+	m.index(lk.Sid, key)
+}
+
+// Snapshot returns a copy of the held locks per key. A mount calls it to drive
+// re-assertion keepalives; the filer never does.
+func (m *Manager) Snapshot() map[string][]Range {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	out := make(map[string][]Range, len(m.byKey))
+	for key, s := range m.byKey {
+		out[key] = append([]Range(nil), s.locks...)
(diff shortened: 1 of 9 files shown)


chrislusf:
Expanded the keepalive into a **re-assertion** so POSIX locks survive a filer-ring rebalance or an owner restart (the lock `Manager` is in-memory soft state; before this the new/restarted owner was blind to existing holders and would double-grant).

- The mount mirrors its own granted locks (`posixOwn`, a `posixlock.Manager`) and each keepalive tick re-asserts the held ranges per key to the key's current owner.
- The owner rebuilds that session's locks from the assertion (`Manager.Reassert`), arbitrating against other sessions so it never double-grants; a lock that lost a migration race is reported, not forced.
- A bare keepalive (no ranges) still just renews the lease.
- Carries `repeated PosixLockRange locks` on `PosixLockRequest` (message-only proto change); forwarding now propagates it.

This is phase 1 of the migration plan (the design I drafted, not checked in). Phase 2 (cooling-off dual-read to close the brief re-assertion window during a ring change) can follow.

PR #9742, opened 2026-05-30
fix(topology): recover heartbeat-fulled volumes once they shrink

chrislusf:
A volume removed from the writable list by the write-assign path stamps `fullSince` in `RecordAssign`, and `UpdateVolumeSize`'s recovery branch uses that marker to re-add the volume once its reported size decays back under the limit.

A volume removed by the heartbeat capacity path (`SetVolumeCapacityFull`) never stamped `fullSince`. So once the reported size dropped — after a vacuum, TTL expiry, or bulk deletes — the volume stayed out of writables forever, even though every heartbeat already carried the smaller size. The only way it came back was an explicit `VolumeMarkWritable`.

Stamp `fullSince` when the capacity path removes a volume, so the existing recovery branch fires on the next under-limit heartbeat. Oversized volumes still stay out, as before.

Added `TestSetVolumeCapacityFullStampsFullSinceAndRecovers` covering capacity-full removal, the no-recovery-before-delay window, and recovery after the delay.

## Summary by CodeRabbit

* **Bug Fixes**
  * Volumes marked full are now timestamped when removed from writable, enabling reliable automatic re-addition after capacity frees up and reducing manual intervention and downtime.

* **Tests**
  * Added unit test verifying the timed recovery path: stamping of full timestamps, delay-respected non-recovery, and eventual restoration of writable volume counts and tracking state.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9742?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/weed/topology/volume_layout.go
+++ b/weed/topology/volume_layout.go
@@ -108,11 +108,11 @@ type volumeSizeTracking struct {
 	reportedSize    uint64    // last heartbeat-reported size (dedup replicas)
 	compactRevision uint32    // detect compaction to reset instead of decay
 	lastUpdateTime  time.Time // dedup replicas within the same heartbeat cycle
-	fullSince       time.Time // non-zero while the volume is eagerly marked full by RecordAssign
+	fullSince       time.Time // non-zero while the volume is marked full (RecordAssign or capacity heartbeat)
 }
 
 // capacityRecoveryDelay is the minimum time a volume must stay out of the
-// writable list after being eagerly removed by RecordAssign before it can
+// writable list after being removed for capacity before it can
 // be considered for re-addition by heartbeat-driven decay. Combined with
 // the effectiveSize hysteresis band, this avoids bouncing the volume in
 // and out of writable within a single burst of assigns.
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review feedback (commit a1827f5f3, squashed into the single PR commit):

- **CodeRabbit (major) — gate `fullSince` on successful removal.** Fixed. Moved the stamp inside the `wasWritable` branch so it only fires when `removeFromWritable` actually removed the volume, keeping it paired with the caller's `activeVolumeCount -= 1` (which `Topology.SetVolumeCapacityFull` does only when `vl.SetVolumeCapacityFull` returns true). This matches `RecordAssign` and avoids a `+1` on recovery with no matching `-1` for capacity-full events on an already-non-writable volume (e.g. readonly-but-oversized, which `CollectDeadNodeAndFullVolumes` reports every cycle).

- **CodeRabbit (nitpick) — stale comments.** Updated the `fullSince` and `capacityRecoveryDelay` comments to say "RecordAssign or the capacity heartbeat" instead of RecordAssign only.

- **Gemini — initialize `sizeTracking` if nil.** Skipped intentionally. With the stamp now gated on `wasWritable`, the volume was in the writable set, which means it went through `RegisterVolume` and already has a `sizeTracking` entry — the nil branch is unreachable here. Force-creating a zero-value `volumeSizeTracking` would also start it with `effectiveSize=0`, corrupting the decay/weighted-pick accounting; `UpdateVolumeSize` already handles first-time initialization correctly.

PR #9746, opened 2026-05-30
perf(s3.iam.GetUser): Make the API default to the request username if not specified

### LINKED ISSUE #9745: [BUG] seaweedfs embedded s3.iam API does not set the own username if sending get-user without username
**Describe the bug**
Sending a request to the embedded S3 IAM API does not resolve the request username by default and leads to a 'user does not exist' error due to an empty string being passed to the handler.
See example:

```bash
$ aws --profile seaweedfs-testenv iam get-user

An error occurred (NoSuchEntity) when calling the GetUser operation: the user with name  cannot be found.
```
(That double space is indicative of the empty string, and is also there in the console.)

**System Setup**
Started via docker compose, see compose file [here](https://github.com/LightJack05/AWS-AccessKey-Operator/blob/ef40af1e52cdb73ba1003926e1d415324a11bbb8/devenv/docker-compose.yml)
- Command: `weed server -dir=/data -filer -s3 -s3.iam.readOnly=false`
- OS version
```
Kernel: Linux version 7.0.5-zen1-1-zen (linux-zen@archlinux) (gcc (GCC) 16.1.1 20260430, GNU ld (GNU Binutils) 2.46.0) #1 ZEN SMP PREEMPT_DYNAMIC Fri, 08 May 2026 09:29:10 +0000
OS is whatever the container image is based off. Same issue appears on bare Arch Linux system when built from source.
```
- output of `weed version`
``` 
 version 30GB 4.29 1355c7a10 linux amd64

For enterprise users, please visit https://seaweedfs.com for SeaweedFS Enterprise Edition,
which has a self-healing storage format with better data protection.
```
Also happens on fresh build of 4.29 from source.

- if using filer, show the content of `filer.toml`
```
[leveldb2]
enabled = true
dir = "/data/filerldb2"
```

**Expected behavior**
The IAM API should default to the user that made the request instead of returning a not found error.

**Screenshots**
N/A

**Additional context**

The handler is located under `weed/s3api/s3api_embedded_iam.go:352`. It receives an empty string, the loop falls through and returns a "Not Found" error with an empty string for the username.

The API action should be added to the case in `weed/s3api/s3api_embedded_iam.go:2857` to default to the username as per documented AWS API behavior for [GetUser](https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetUser.html).

I'll open a draft PR in a sec and link it here so you can take a look at my proposed fix. I'm unfamiliar with your tooling though, so you probably want someone with that knowledge check if there are new tests needed etc.

LightJack05:
This makes the Embedded S3 IAM API align with the documented behavior of the AWS IAM API as per AWS Docs: https://docs.aws.amazon.com/IAM/latest/APIReference/API_GetUser.html

BREAKING CHANGE: This changes the default behavior of the Embedded IAM API to use the username of the user holding the accesskey used to make the request in the GetUsername request handler.

Closes #9745 

See issue #9745 

The IAM API should return information on the current user if there is no username parameter specified in the request. 

Add the GetUser endpoint to the case executing `e.handleImplicitUsername(r, values)` in `weed/s3api/s3api_embedded_iam.go`

# How is the PR tested?
One manual test to validate the solution, otherwise not yet.

# Checks

# Checks for AI generated PRs

> [!NOTE]
> This is intended to be a reference for a proper PR, I'm not intending for this to be merged, unless you are just fine with it. 
> I wanted to do a bit more than just drop an issue, and this is probably the best way to provide a clear description of how this could be fixed.

## Bug Fixes

* **IAM GetUser Action**: Enhanced the GetUser action to automatically resolve the user based on the caller's signing access key when no explicit username is provided, ensuring consistency with AWS IAM behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/seaweedfs/seaweedfs/pull/9746?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/weed/s3api/s3api_embedded_iam.go
+++ b/weed/s3api/s3api_embedded_iam.go
@@ -2854,7 +2854,7 @@ func (e *EmbeddedIamApi) DoActions(w http.ResponseWriter, r *http.Request) {
 
 	// Handle implicit username for HTTP requests
 	switch r.Form.Get("Action") {
-	case "ListAccessKeys", "CreateAccessKey", "DeleteAccessKey", "UpdateAccessKey", "ListUserPolicies":
+	case "ListAccessKeys", "CreateAccessKey", "DeleteAccessKey", "UpdateAccessKey", "ListUserPolicies", "GetUser":
 		e.handleImplicitUsername(r, values)
 	case "CreateServiceAccount":
 		createdBy := s3_constants.GetIdentityNameFromContext(r)
--- a/weed/s3api/s3api_embedded_iam_test.go
+++ b/weed/s3api/s3api_embedded_iam_test.go
@@ -339,6 +339,46 @@ func TestEmbeddedIamGetUser(t *testing.T) {
 	assert.Equal(t, "TestUser", *out.GetUserResult.User.UserName)
 }
 
+// TestEmbeddedIamGetUserImplicitUsername verifies GetUser without a UserName defaults
+// to the user that signed the request rather than returning NoSuchEntity for an empty
+// username, matching documented AWS IAM behavior.
+func TestEmbeddedIamGetUserImplicitUsername(t *testing.T) {
+	const accessKey = UserAccessKeyPrefix + "TESTFAKEKEY000001"
+
+	api := NewEmbeddedIamApiForTest()
+	api.mockConfig = &iam_pb.S3ApiConfiguration{
+		Identities: []*iam_pb.Identity{
+			{
+				Name: "TestUser",
+				Credentials: []*iam_pb.Credential{
+					{AccessKey: accessKey, SecretKey: "testsecretfake"},
+				},
+			},
+		},
+	}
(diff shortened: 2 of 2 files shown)


chrislusf:
Pushed a regression test (cb46b3d) that drives `GetUser` with no `UserName` through `DoActions` and asserts it resolves to the signing user; `TestEmbeddedIamGetUserNotFound` still confirms an explicitly-named missing user errors.

On extending `handleImplicitUsername` to `ListAttachedUserPolicies`/`PutUserPolicy`/`GetUserPolicy`/`DeleteUserPolicy`/`ListUserTags`/`TagUser`/`UntagUser`: in AWS those all have `UserName` as Required. Only `GetUser` and the access-key actions document the implicit-username default, so keeping this scoped to `GetUser` matches AWS; broadening to the others would diverge from it. Leaving the suggestion out.

PR #9761, opened 2026-05-31
EC bitrot detection: per-shard checksum sidecars

chrislusf:
## EC bitrot detection via per-shard checksum sidecars

Erasure-coded volumes have a blind spot for silent disk corruption (bitrot):

- **Data shards** are read on the serving path, so corruption in regions that happen to be read is caught by the existing per-needle CRC. Padding and unread regions are not.
- **Parity shards are never read during normal serving** — only during reconstruction. `reedsolomon.Reconstruct` treats every present shard as authoritative; it fills missing shards but does **not** detect corruption in the shards it reads. So a parity shard can rot silently for months, and the day a data shard is lost, the corrupt parity is fed into Reed-Solomon and **silently produces wrong data**.

This PR adds an optional per-volume checksum sidecar so that corruption in any shard — including cold parity — is detectable and the rebuild path refuses to consume unverified bytes.

### How it works

- **Sidecar format** (`<base>.ecsum`): a `CRC32C` (Castagnoli, reusing the existing needle CRC) for every fixed-size block (default 16 MiB) of every shard, stored as a small protobuf payload behind a self-integrity header (`magic | format_version | payload_len | payload_crc32c`). ~11 KB for a 30 GB volume. Absent sidecar = feature off; old binaries simply ignore the file.
- **At encode:** the checksums are computed inline in the single encode pass (`WriteEcFilesWithContext` now returns the protection) and written alongside the shards; the sidecar travels with them via a new `copy_ecsum_file` copy flag.
- **Scrub:** `ec.scrub -mode checksum` (and the gRPC `CHECKSUM` mode) reads each local shard in blocks and compares to the sidecar — the only path that exercises cold parity shards. It is read-only. A flagged shard is arbitrated by Reed-Solomon (reconstruct it from the clean shards and compare) so a stale sidecar block can never cause a healthy shard to be reported corrupt; a wholesale mismatch is treated as a suspect sidecar, not mass corruption.
- **Rebuild:** present input shards are verified against the sidecar and corrupt ones are excluded from Reed-Solomon and regenerated (temp file + atomic rename), then re-verified. A malformed/corrupt sidecar makes the rebuild **fail closed** unless `-unsafeIgnoreSidecar`; any verification failure removes every generated output so no unverified bytes are published.
- **Backfill:** a volume rebuilt on a server that can reach all shards opportunistically gains a sidecar, so existing (pre-feature) volumes become protected over time.
- **Config:** `-ec.bitrotChecksum` (default on) and `-ec.bitrotBlockSizeMB` (default 16) on the volume server.

### Tests

New unit tests cover the sidecar format/self-integrity, the rolling block-CRC builder across boundaries, manifest validation, encode→verify roundtrip, rebuild verify-and-exclude (corrupt present shard regenerated correctly), fail-closed cleanup, and the backfill primitive. The existing EC encode/rebuild/roundtrip suite passes with the threaded changes.

### Notes

- Commits are split per logical change (proto, sidecar module, encode capture + rebuild verify, scrub, server handlers, config flags).
- 10+4 only; the sidecar records the shard config but OSS uses the standard ratio.

## Summary by CodeRabbit

* **New Features**
  * EC bitrot protection: generation-aware checksum sidecars with atomic save/load, validation, backfill and reconstruction; sidecars produced during encode and propagated on copy/transfer.
  * New CHECKSUM scrub mode to verify EC shards against sidecars.
  * CLI flags to enable/disable bitrot, set block-size, copy sidecars, and an unsafe rebuild override.

* **Bug Fixes**
  * Best-effort sidecar cleanup on delete/convert and safer rebuild behavior when sidecars are missing, stale, or corrupt.

* **Tests**
  * Extensive end-to-end tests covering sidecar round-trip, validation, corruption detection, backfill, rebuild paths, and cleanup.

### DIFF
--- a/weed/command/fix.go
+++ b/weed/command/fix.go
@@ -381,7 +381,7 @@ func doFixEcxFromShards(basePath, baseFileName, collection string, volumeId int6
 	if !dataComplete {
 		ctx := &erasure_coding.ECContext{DataShards: dataShards, ParityShards: parityShards}
 		glog.Infof("volume %d: %d/%d shards present; reconstructing missing shards (%s) before index rebuild", volumeId, presentCount, dataShards+parityShards, ctx.String())
-		if _, err := erasure_coding.RebuildEcFilesWithContext(base, ctx); err != nil {
+		if _, err := erasure_coding.RebuildEcFilesWithContext(base, ctx, false); err != nil {
 			fail(fmt.Errorf("volume %d: reconstruct missing shards from %d survivors: %w", volumeId, presentCount, err))
 			return
 		}
--- a/weed/command/fix.go
+++ b/weed/command/fix.go
@@ -382,7 +382,7 @@ func doFixEcxFromShards(basePath, baseFileName, collection string, volumeId int6
 	if !dataComplete {
 		ctx := &erasure_coding.ECContext{DataShards: dataShards, ParityShards: parityShards}
 		glog.Infof("volume %d: %d/%d shards present; reconstructing missing shards (%s) before index rebuild", volumeId, presentCount, dataShards+parityShards, ctx.String())
-		if _, err := erasure_coding.RebuildEcFilesWithContext(base, ctx, *fixEcUnsafeIgnoreSidecar); err != nil {
+		if _, err := erasure_coding.RebuildEcFiles(base, ctx, *fixEcUnsafeIgnoreSidecar); err != nil {
 			fail(fmt.Errorf("volume %d: reconstruct missing shards from %d survivors: %w", volumeId, presentCount, err))
 			return
 		}
--- a/weed/command/fix.go
+++ b/weed/command/fix.go
@@ -37,14 +37,15 @@ var cmdFix = &Command{
 }
 
 var (
-	fixVolumeCollection = cmdFix.Flag.String("collection", "", "an optional volume collection name, if specified only it will be processed")
-	fixVolumeId         = cmdFix.Flag.Int64("volumeId", 0, "an optional volume id, if not 0 (default) only it will be processed")
-	fixIncludeDeleted   = cmdFix.Flag.Bool("includeDeleted", true, "include deleted entries in the index file")
-	fixIgnoreError      = cmdFix.Flag.Bool("ignoreError", false, "an optional, if true will be processed despite errors")
-	fixRemoteFile       = cmdFix.Flag.Bool("remoteFile", false, "an optional, if true will not try to load the local .dat file, but only the remote file")
-	fixGenerateEcx      = cmdFix.Flag.Bool("ecx", false, "regenerate a lost EC index (.ecx) — and the .vif when missing — from the local .ec## shards (missing shards are reconstructed from parity when enough survive). Run with the volume server stopped.")
-	fixEcDataShards     = cmdFix.Flag.Int("ecDataShards", 0, "EC data shard count for -ecx (0 = read from .vif, otherwise default 10)")
-	fixEcParityShards   = cmdFix.Flag.Int("ecParityShards", 0, "EC parity shard count for -ecx (0 = read from .vif, infer from shard count, otherwise default 4)")
+	fixVolumeCollection      = cmdFix.Flag.String("collection", "", "an optional volume collection name, if specified only it will be processed")
+	fixVolumeId              = cmdFix.Flag.Int64("volumeId", 0, "an optional volume id, if not 0 (default) only it will be processed")
+	fixIncludeDeleted        = cmdFix.Flag.Bool("includeDeleted", true, "include deleted entries in the index file")
+	fixIgnoreError           = cmdFix.Flag.Bool("ignoreError", false, "an optional, if true will be processed despite errors")
+	fixRemoteFile            = cmdFix.Flag.Bool("remoteFile", false, "an optional, if true will not try to load the local .dat file, but only the remote file")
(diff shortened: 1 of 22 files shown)


chrislusf:
Pushed a round addressing the review feedback. Summary of what changed and why:

**Applied**

- `weed/command/volume.go` — `-ec.bitrotBlockSizeMB` is now validated in `[1, 1024]` *before* the `*1024*1024` multiply, so a pathological value can't overflow `int64` and sneak past the power-of-two check, and the block size can't collapse a sidecar to a handful of oversized blocks.
- `weed/worker/tasks/erasure_coding/ec_task.go` + `weed/storage/erasure_coding/shard_distribution.go` — the worker encode wrote the generation-0 `.ecsum` locally but never added it to `shardFiles`, so `DistributeEcShards` never shipped it and distributed holders came up unprotected. It's now appended to `shardFiles` and the `ecsum` shard type is mapped to its extension in the sender, so it travels with the shards.
- `weed/server/volume_grpc_erasure_coding.go` — sidecar removal is now gated on `existingShardCount == 0` alone instead of also requiring a stray `.ecx`. An orphaned sidecar whose shards are all gone now gets cleaned up even when no `.ecx` remains; `.ecx/.ecj/.vif` removal stays gated on `hasEcxFile`.
- `weed/server/volume_grpc_scrub.go` — `CHECKSUM` mode no longer folds `ChecksumScrub`'s blocks-scanned count into `TotalFiles` (which is a needle/file count).
- `weed/storage/erasure_coding/ec_test.go` — `removeGeneratedFiles` now cleans up the generated `.ecsum` sidecar.
- `weed/storage/erasure_coding/ec_bitrot.go` — `SaveBitrotSidecar` now rejects a payload larger than `MaxUint32` before the `uint32(len(payload))` cast, so the `payload_len` header field can't be silently truncated (addresses the CodeQL conversion finding).

**Analyzed, no change needed**

The `io.EOF` flags on `ec_volume_scrub.go` (`ChecksumScrub`, `rsConfirmsShardCorrupt`) are already correct: those reads go through `EcVolumeShard.ReadAt`, which clears `io.EOF` when the full buffer was filled (`if err == io.EOF && n == len(buf) { err = nil }`). A short read at EOF (`n < toRead`) is genuine corruption and is still flagged by the length check, so there is no false positive. `readFull` in `ec_bitrot.go` does the same on the raw `*os.File` (returns `nil` once `total == len(buf)`), so it's correct as written.

chrislusf:
Second round pushed, addressing the latest review:

**Applied**

- `weed/command/volume.go` — lowered the `-ec.bitrotBlockSizeMB` ceiling from 1024 to **64 MiB**. Good catch: the block size becomes the per-shard scratch buffer the scrub/backfill path allocates (`make([]byte, blockSize)`), so a 1 GiB value was a real memory hazard per concurrent worker.
- `weed/command/fix.go` — added **`-ecUnsafeIgnoreSidecar`** (default false) for `weed tool fix -ecx` and threaded it into `RebuildEcFilesWithContext`. Without it an operator hitting a stale/malformed `.ecsum` couldn't complete the reconstruction without manually deleting the sidecar.
- `weed/storage/erasure_coding/ec_bitrot.go` (CodeQL) — replaced the `uint64(len(payload)) > MaxUint32` check with a direct `len(payload) > maxBitrotPayloadSize` (1 GiB) int bound before the allocation, so the size is provably bounded; dropped the now-unused `math` import.
- `weed/storage/erasure_coding/ec_bitrot.go` (gemini) — inlined `f.ReadAt` with `io.EOF` handling in `verifyShardFileBlocks` and removed the `readFull` helper. A full read proceeds; a short read at EOF (truncated shard) still surfaces as an error, matching the prior behavior.

**Already addressed in the previous push (re-raises)** — `volume_grpc_scrub.go` discards `ChecksumScrub`'s block count from `TotalFiles`; `ec_test.go`'s `removeGeneratedFiles` calls `RemoveBitrotSidecars`; `ec_task.go` + `shard_distribution.go` ship `.ecsum` with the shards. All three are present in HEAD.

**No change, by design** — the remaining `io.EOF` flags on `ec_volume_scrub.go` are correct as written: those reads go through `EcVolumeShard.ReadAt`, which clears `io.EOF` when the buffer is fully filled, and a short read is still caught by the length check. (The `ec_bitrot.go` `readFull` path used the raw `*os.File`, which is why the simplification applies there but not here.)

The OSS-applicable changes are mirrored from the enterprise build and verified equivalent; local build, vet, and the `erasure_coding`/`command` tests pass.

chrislusf:
Addressed both review comments:

- **`ec_context.go` — exported mutable singleton.** `BackgroundECContext` is now a function returning a fresh `&ECContext{}` per call, so no caller can mutate a shared default or race on it. This also makes it a closer analog to `context.Background()`, which is itself a function. Call sites updated to `BackgroundECContext()`.
- **`ec_encoder.go` — fail closed on an unreadable `.vif`.** `RebuildEcFiles` now propagates the `MaybeLoadVolumeInfo` error instead of discarding it: a present-but-corrupt `.vif` returns an error rather than silently falling back to the default 10+4 layout (which would rebuild a custom-ratio volume with the wrong layout). A missing `.vif` still proceeds with the default, and passing an explicit `ctx` (e.g. `fix -ecx -ecDataShards/-ecParityShards`) overrides the `.vif` read entirely.

Both build/vet/test clean. The earlier CodeQL allocation-overflow alert is also cleared (0 open `ec_bitrot` alerts on the PR head).

Issue #9330, opened 2026-05-05
Large files get corrupted on copy to seaweed cluster


bahciem:
Large files (>10Gb) get corrupted on copy to seaweed cluster.
ver: 4.22

Steps to reproduce:
1. Just run simple cluster: master, volume, filer, mount (same for large multi-node cluster)
2. cp (or cat or rsync)  large file (20..200 Gb) from local disk to cluster
3. compare md5sum's, they will be different
4. weed filer.sync also reported errors:
(provider lookup failed: master volume lookup errors: volume XXX: volume id XXX not found)

Binary comparison shows that one or two 2Mb blocks filled by zeros on destination side.

Fast fix:
for files 20..100Gb:  weed mount -chunkSizeLimitMB=8
for files 100..200Gb:  weed mount -chunkSizeLimitMB=16

bahciem:
Seems workaround with chunkSizeLimitMB=8/16 does not fix active-active sync.
Source file on local disk successfully copied to cluster A, but it differs in size on cluster B.
With large files (>150Gb) I have sync error:
...
file_mode:420 crtime:1778051295 mime:"application/zip" inode:15839097307827841653 ctime:1778055324 mtime_ns:41689291 ctime_ns:41689291}} delete_chunks:true new_parent_path:"/aaa" signatures:-898230835 signatures:-1450875626} ts_ns:1778055324053499081: replicate /aaa/tasks.zip compare chunks error: fail to read manifest 526,0920f21991e4271f: http://x.x.x.x:8083/526,0920f21991e4271f?readDeleted=true: 404 Not Found: not found

bahciem:
Issue not fixed in 4.28, files ~150Gb synced with different size,
sync: active-active
log message:
seaweedfs-sync[1098821]: E0526 12:56:21.808190 filer_sync_jobs.go:260 process directory:"/data/ItemsMatcherMasks"  event_notification:{old_entry:{name:"tasks_4000000_5000000.zip"  chunks:{file_id:"617,0b91a023121b9206"  size:167772160000  fid:{volume_id:617  file_key:194093091  cookie:303796742}  is_chunk_manifest:true}....

chrislusf:
@bahciem   Two things I'd want to know to actually pin this down:

  1. The full log line — specifically everything after event_notification:{...} ... : <error>. The error string tells us if this is replicate manifest data chunks ..., resolve manifest ...,
  upload manifest data ..., read part ..., etc.
[...]

PR #9777, opened 2026-06-01
filer.sync: back off on transient upload errors

chrislusf:
Large-file sync (`weed filer.sync` of 100-200GB files) intermittently fails on the last few GB with the destination volume server returning:

    400 {"error":"read tcp ...: i/o timeout"}

That is the volume server's connection idle deadline (`-idleTimeout`, default 30s) firing while it reads the upload body: under load — especially a filer.sync co-located with the destination volume server — the body read stalls past the deadline.

`fetchAndWrite` already retried this, but only on the flat ~1s `RetryUntil` backoff, so it re-issued the upload against the still-overloaded destination almost immediately. This routes timeout / connection-reset / broken-pipe / `net.Error` timeouts through the same escalating 10s-2min backoff already used for interrupted reads, giving the destination room to recover. The volume server returns the timeout as a JSON error string, so the predicate matches on text as well as the `net.Error` interface.

Operational note: raising the destination volume `-idleTimeout` and/or lowering `-concurrency`/`-chunkConcurrency` also reduces how often the deadline is hit.

Ref: #9330

## Summary by CodeRabbit

* **Bug Fixes**
  * Broadened replication retry logic to detect and handle more transient network failures (timeouts, EOFs, connection resets, broken pipe, etc.)
  * Transfer status now shows wait duration and preserves received bytes during automatic retries

* **Tests**
  * Added tests validating transient network error classification and retry behavior

### DIFF
--- a/weed/replication/sink/filersink/fetch_write.go
+++ b/weed/replication/sink/filersink/fetch_write.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review: lower-case the error string before substring matching so capitalized variants (`I/O timeout`, `Connection reset by peer`, `Broken pipe`) are also caught, and added capitalized test cases.

PR #9778, opened 2026-06-01
filer.sync: repair a destination shorter than the source

chrislusf:
A file left truncated by a failed replication is never corrected, because `FilerSink.UpdateEntry` treats a newer destination mtime as last-writer-wins and skips the update. The trap (seen in #9330): the source kept the file's original mtime while the partial copy on the destination was written recently, so the destination looks "newer" forever. `filer.sync.verify` already flags exactly this as `late_updates_skip_likely` (B_NEWER).

When the mtime guard would skip but the destination is strictly shorter than the source (`filer.FileSize`), re-replicate the full source content and replace the chunk list instead of skipping. The same shorter-than-source bypass is added to `CreateEntry`'s skip. Only the previously-skipped path changes behavior; the normal catch-up path is untouched.

The skip/repair/normal decision is factored into `chooseUpdateAction` and unit-tested.

Note: the repair fires on the next metadata event for the path, so an already-truncated file under a stable source still needs the source to re-emit (touch / re-copy). This stops the trap going forward and repairs on the next change.

Ref: #9330

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file replication to better detect and repair truncated or incomplete copies encountered during sync; now avoids skipping necessary overwrites when destination is newer but smaller.
  * Enhanced conflict resolution for files with mismatched timestamps and sizes to ensure genuine updates are applied reliably.

* **Tests**
  * Added targeted tests covering replication decision logic to validate normal, skip, and repair behaviors.

### DIFF
--- a/weed/replication/sink/filersink/filer_sink.go
+++ b/weed/replication/sink/filersink/filer_sink.go
@@ -184,8 +184,14 @@ func (fs *FilerSink) CreateEntry(key string, entry *filer_pb.Entry, signatures [
 				return nil
 			}
 			if resp.Entry.Attributes != nil && resp.Entry.Attributes.Mtime >= entry.Attributes.Mtime {
-				glog.V(3).Infof("skip overwriting %s", key)
-				return nil
+				if filer.FileSize(resp.Entry) >= filer.FileSize(entry) {
+					glog.V(3).Infof("skip overwriting %s", key)
+					return nil
+				}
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review: `updateSkip` now returns early instead of re-writing the unchanged entry back to the filer, and added a nil existing-entry case to the decision test.

PR #9817, opened 2026-06-04
s3: keep dynamic IAM live when -iam.config is set

chrislusf:
`-iam.config` was reused as the static `-config` identity file, which set `useStaticConfig=true`. That makes `onIamConfigChange` skip the filer metadata-subscription reload, so identities and policies created at runtime via the IAM gRPC API never took effect. Enabling advanced IAM (OIDC/STS) and managing identities dynamically were therefore mutually exclusive.

An advanced IAM config carries no inline identities (it parses with `DiscardUnknown`), so gating `useStaticConfig` on whether identities were actually loaded keeps an OIDC/STS-only deployment's dynamic credential store live. A `-config` identity file still freezes its identities as before. Live reloads re-sync policies into the advanced IAM engine, so dynamically created identities and policies authorize correctly.

Added a test covering both paths: an `-iam.config`-only setup keeps reloading filer-backed identities; a config file with inline identities stays static.

## Summary by CodeRabbit

* **Bug Fixes**
  * Static-identity mode now activates only when inline static identities are present; configuration reloads now respect static vs dynamic identities and avoid freezing dynamic ones.
  * Reload success now emits a verbose log including the loaded identity count.
* **Tests**
  * Added tests validating IAM configs with and without inline static identities and reload behavior that marks new static identities while keeping dynamic ones writable.

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -267,8 +267,6 @@ func NewIdentityAccessManagementWithStore(option *S3ApiServerOption, filerClient
 		if err := iam.loadS3ApiConfigurationFromFile(startConfigFile); err != nil {
 			glog.Fatalf("fail to load config file %s: %v", startConfigFile, err)
 		}
-
-		iam.markStaticIdentities()
 	}
 
 	// Always try to load/merge config from credential manager (filer/db)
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed the reload feedback: `markStaticIdentities` now runs inside `loadS3ApiConfigurationFromFile`, so a `grace.OnReload` of `-config` marks reloaded identities static too, not just the set present at startup. It's additive and scoped to the file's identities, so a reload protects newly added ones without freezing dynamic (filer-managed) identities. Added `TestReloadStaticConfigMarksNewIdentitiesWithoutFreezingDynamic` as a regression test.

chrislusf:
Pushed two follow-ups:

- **CredentialManager sync (line 501):** done — `loadS3ApiConfigurationFromFile` now calls `updateCredentialManagerStaticIdentities()` right after marking, so reloaded static identities are pushed into the credential manager for every caller. Dropped the now-redundant call in the `grace.OnReload` hook.
- **markStaticIdentities on reload (gemini):** addressed in the earlier commit — the marking lives in `loadS3ApiConfigurationFromFile`, so a `-config` reload protects newly added identities, additively and scoped to the file (without freezing dynamic filer identities).

**Deferring line 500** (reload not applying edits/deletes to *existing* static identities): this is pre-existing and unchanged by this PR — the merge path has always skipped names already in `staticIdentityNames` on reload, so editing/removing an existing `-config` identity required a restart before this change too. A correct fix means making a file reload authoritative for its static identities (apply edits, drop removed ones) while preserving dynamic filer identities and staying atomic against concurrent auth requests; that touches the core merge path and is better as its own PR with focused tests. This change is a strict improvement on the reload path and doesn't regress edits/deletes.

PR #9822, opened 2026-06-04
s3: return BucketAlreadyOwnedByYou when recreating your own bucket

### LINKED ISSUE #9820: Incorrect error raised when bucket already exists by same account
**Summary**
When creating a bucket, there are two possible exceptions ` BucketAlreadyExists` and ` BucketAlreadyOwnedByYou` according to the S3 documentation. `BucketAlreadyOwnedByYou` is specifically for buckets you own.
Seaweed will always return `BucketAlreadyExists`, regardless of whether the user owns the bucket or not.

**Consequence**
Most applications should be designed to fail with "BucketAlreadyExists" because that error indicates that bucket name is already in use by a different application/user. This is a clear sign you either need a new unique bucket name or you have the wrong credentials.

Whereas with the `BucketAlreadyOwnedByYou` is an error that lets you move forward knowing that it is your bucket, it just existed already (such as if you are restarting a container which creates the bucket)

**Reproduce**
* Create a bucket twice in a row using the same boto3 client

```py
import logging
import boto3

logger = logging.getLogger(__name__)
client = boto3.client("s3", **config)
bucket="mm.artifacts"
logger.debug("Trying to create bucket")
client.create_bucket(Bucket=bucket, ObjectLockEnabledForBucket=True)
logger.debug("Bucket created successfully")
try:
    logger.debug("Trying to recreate bucket with same client")
    client.create_bucket(Bucket=bucket, ObjectLockEnabledForBucket=True)
except ClientError as e:
    error_code = getattr(e, "response", {}).get("Error", {}).get("Code")
    logger.error("Error creating bucket. error_code=%s", error_code)
    raise
```

app log
```
DEBUG  [aws] Trying to create bucket
DEBUG  [urllib3.connectionpool] Starting new HTTP connection (1): seaweed:8333
DEBUG  [urllib3.connectionpool] http://seaweed:8333 "PUT /mm.artifacts HTTP/1.1" 200 0
DEBUG  [aws] Bucket created successfully
DEBUG  [aws] Trying to recreate bucket with same client
DEBUG  [urllib3.connectionpool] http://seaweed:8333 "PUT /mm.artifacts HTTP/1.1" 409 434
ERROR  [aws] Error creating bucket. error_code=BucketAlreadyExists
```

chrislusf:
Fixes #9820

## Problem

`CreateBucket` always returned `BucketAlreadyExists` when a bucket already existed, regardless of who owned it. AWS distinguishes the two:

- `BucketAlreadyExists` — the name is taken in the shared namespace (someone else owns it). The caller needs a different name or different credentials.
- `BucketAlreadyOwnedByYou` — you already own this bucket; the create is just idempotent (e.g. a container that creates its bucket on every startup).

Collapsing both into `BucketAlreadyExists` broke that idempotency signal for clients.

## Fix

Recreating a bucket you own now returns `BucketAlreadyOwnedByYou`, unless the request conflicts with the existing bucket:

- a different Object Lock setting, or
- an ACL on either the request or the existing bucket.

In those conflict cases it still returns `BucketAlreadyExists`, matching the ceph s3-tests `test_bucket_recreate_overwrite_acl` / `test_bucket_recreate_new_acl` (while `test_bucket_create_exists` now gets the correct `BucketAlreadyOwnedByYou`).

To detect the existing-ACL case, a requested non-default canned/grant ACL is now persisted on bucket creation instead of being silently dropped (a `private` canned ACL is the default and is not stored).

## Reproduce (before this change)

```py
client.create_bucket(Bucket="mm.artifacts", ObjectLockEnabledForBucket=True)
client.create_bucket(Bucket="mm.artifacts", ObjectLockEnabledForBucket=True)  # -> 409 BucketAlreadyExists
```

After the change the second call returns `409 BucketAlreadyOwnedByYou`.

## Verified

Ran a local server through every relevant scenario; all match AWS / s3-tests expectations:

| scenario | result |
| --- | --- |
| plain create, plain recreate (same owner) | `BucketAlreadyOwnedByYou` |
| recreate by a different owner | `BucketAlreadyExists` |
| create with ACL, recreate plain | `BucketAlreadyExists` |
| create plain, recreate with ACL | `BucketAlreadyExists` |
| object-lock create + recreate (issue) | `BucketAlreadyOwnedByYou` |
| plain create, recreate with object-lock | `BucketAlreadyExists` |
| explicit `private` create, plain recreate | `BucketAlreadyOwnedByYou` |

Added a unit test for the new ACL-header detection helper; `go test ./s3api/` passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket-creation conflicts to correctly distinguish ownership vs. existing-bucket errors when object-lock or ACLs differ.

* **New Features**
  * Non-default ACLs provided at bucket creation are detected and persisted so subsequent ACL reads reflect requested grants.
  * Bucket owner canonical ID is now recorded and used when returning ACLs.

* **Tests**
  * Added tests validating detection of explicit (non-default) ACL requests from incoming headers.

### DIFF
--- a/weed/s3api/s3api_acl_helper.go
+++ b/weed/s3api/s3api_acl_helper.go
@@ -308,6 +308,45 @@ func ValidateAndTransferGrants(accountManager AccountManager, grants []*s3.Grant
 	return result, s3err.ErrNone
 }
 
+// buildAccessControlList converts stored ACP grants into the XML response form.
+// When no grants are stored it falls back to a single full-control grant for the
+// owner, matching AWS's default private ACL.
+func buildAccessControlList(accountManager AccountManager, grants []*s3.Grant, ownerId, ownerDisplayName string) AccessControlList {
+	if len(grants) == 0 {
+		return AccessControlList{Grant: []Grant{{
+			Grantee: Grantee{
+				ID:          ownerId,
+				DisplayName: ownerDisplayName,
+				Type:        "CanonicalUser",
+				XMLXSI:      "CanonicalUser",
+				XMLNS:       "http://www.w3.org/2001/XMLSchema-instance",
+			},
+			Permission: Permission(s3_constants.PermissionFullControl),
+		}}}
+	}
+
+	var acl AccessControlList
+	for _, grant := range grants {
+		localGrant := Grant{Permission: Permission(*grant.Permission)}
+		if grant.Grantee != nil {
+			localGrant.Grantee = Grantee{
+				Type:   *grant.Grantee.Type,
+				XMLXSI: *grant.Grantee.Type,
+				XMLNS:  "http://www.w3.org/2001/XMLSchema-instance",
+			}
+			if grant.Grantee.ID != nil {
+				localGrant.Grantee.ID = *grant.Grantee.ID
+				localGrant.Grantee.DisplayName = accountManager.GetAccountNameById(*grant.Grantee.ID)
+			}
+			if grant.Grantee.URI != nil {
+				localGrant.Grantee.URI = *grant.Grantee.URI
+			}
+		}
+		acl.Grant = append(acl.Grant, localGrant)
+	}
+	return acl
+}
+
 // GetAcpGrants return grants parsed from entry
 func GetAcpGrants(entryExtended map[string][]byte) []*s3.Grant {
 	acpBytes, ok := entryExtended[s3_constants.ExtAmzAclKey]
--- a/weed/s3api/s3api_bucket_handlers.go
+++ b/weed/s3api/s3api_bucket_handlers.go
@@ -165,6 +165,29 @@ func (s3a *S3ApiServer) PutBucketHandler(w http.ResponseWriter, r *http.Request)
 	// Get authenticated identity from context (secure, cannot be spoofed)
 	currentIdentityId := s3_constants.GetIdentityNameFromContext(r)
 
+	// Parse any requested bucket ACL (canned ACL or grant headers) up front so it
+	// can be validated, persisted on creation, and factored into the already-exists
+	// response. A "private" canned ACL is the default and counts as no explicit ACL.
+	requestHasACL := hasExplicitBucketACL(r)
+	var aclGrantsBytes []byte
+	if requestHasACL {
+		accountId := getAccountId(r)
+		_, grants, errCode := ParseAndValidateAclHeaders(r, s3a.iam, "", accountId, accountId, false)
+		if errCode != s3err.ErrNone {
+			s3err.WriteErrorResponse(w, r, errCode)
+			return
+		}
+		if len(grants) > 0 {
+			grantsBytes, err := json.Marshal(grants)
+			if err != nil {
+				glog.Errorf("PutBucketHandler: marshal ACL grants for %s: %v", bucket, err)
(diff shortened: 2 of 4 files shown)


chrislusf:
Addressed the review feedback in three separate commits:

**1. `0a7bf53` — don't treat an unreadable config as idempotent recreation** (CodeRabbit / Gemini)
When the bucket already exists but `getBucketConfig` fails, the handler no longer falls through to `BucketAlreadyOwnedByYou`. It now surfaces the read error, so a backend failure can't mask a conflicting recreate (Object Lock / ACL unknown).

**2. `f1b7c39` — `GetBucketAcl` now returns the stored ACL**
`GetBucketAclHandler` decoded nothing and always returned the owner's default full-control grant, so a bucket created with a canned ACL (or one set via `PutBucketAcl`) never read back correctly. It now decodes the stored grants, sharing one grants-to-XML helper with `GetObjectAclHandler` (no parallel copy). The shared helper also emits each grantee's real `xsi:type` (e.g. `Group` for `public-read`) instead of a hardcoded `CanonicalUser`, which fixes group-grant round-tripping for both bucket and object ACLs.

**3. `93baefa` — concurrent-create race resolves the right error**
The mkdir-race fallback returned a flat `BucketAlreadyExists`, bypassing the same-owner idempotency check. Both the pre-check and the race fallback now route through a single `existingBucketError` helper, so a same-owner recreate that loses the race still gets `BucketAlreadyOwnedByYou`.

Verified against a local server: the full error-code matrix (create_exists, recreate_*_acl, nonowner, object-lock match/conflict) plus `GetBucketAcl`/`GetObjectAcl` round-trips for `public-read` (canned-on-create, `PutBucketAcl`, `PutObjectAcl`) all behave as expected; `go test ./s3api/` passes.

chrislusf:
Addressed the new `GetBucketAcl` owner finding in two commits:

**`7096335` — record the bucket owner's account id at creation**
`setBucketOwner` only stored the creating identity *name* (`AmzIdentityId`), so the canonical account id wasn't available at read time. It now also stores `ExtAmzOwnerKey` (the same field `PutBucketAcl` writes), so the owner can be reported independently of the reader.

**`c737e99` — report the persisted owner from `GetBucketAcl`**
`GetBucketAclHandler` now uses the bucket's persisted owner (`bucketConfig.Owner`) for both the `Owner` element and the default-grant fallback, dropping back to the caller header only when no owner was recorded.

One deliberate deviation from the suggested patch: I did **not** fall back to `entry.Extended[AmzIdentityId]`. That field holds the identity *name* (e.g. `s3_tests`), whereas ACL `Owner.ID` must be the canonical *account id* (e.g. `testid`); using it would have changed self-read owners from `testid` to `s3_tests` and broken existing ACL expectations. Persisting the account id at creation (commit above) gives the correct value for every bucket instead.

Verified locally:
- admin cross-account read of a bucket created by `s3_tests` now returns `Owner.ID = testid` (the creator), not the admin;
- owner self-reads still return `testid`;
- `public-read` ACL + Owner round-trip correctly;
- full error-code matrix (create_exists / recreate_*_acl / nonowner / object-lock match+conflict) unchanged; `go test ./s3api/` passes.

PR #9826, opened 2026-06-04
security: hot-reload JWT signing keys on SIGHUP

### LINKED ISSUE #9823: JWT signing key is loaded once at startup, never reloaded — recovery from key mismatch requires process restart
## Summary

If a `weed` process (volume / master / filer / s3) reads the JWT signing key at startup, then the key changes (via Secret rotation, divergent reads, or any other means) — the process never picks up the new value. Authentication failures (`401: wrong jwt`) persist until manual restart of the affected processes.

## Reproduction (observed in production)

A SeaweedFS cluster (master + volume + filer + s3 in Kubernetes, JWT signing key delivered via a mounted `Secret`) backing Harbor's registry storage entered a stuck state after a brief Kubernetes apiserver disruption. After things came back, the cluster did **not** self-recover.

Observed errors:

**volume server (`seaweedfs-volume-1`):**
```
error JSON response status 401: wrong jwt
response method:POST URL:/235,xxxxxxxxxx with httpStatus:401 and JSON:{"error":"wrong jwt"}
```

**s3 server:**
```
upload_content.go:214 uploading 2 to http://seaweedfs-volume-1.seaweedfs-volume.<ns>:8080/235,xxxxxxxxxx:
  unmarshalled error: wrong jwt

s3api_object_handlers_copy.go:570 CopyObjectPartHandler copy chunks error: chunk 3:
  upload chunk range data: upload chunk: unmarshalled error: wrong jwt

upload_chunked.go:252 chunked upload failed: assign volume: all filers failed,
  last error: assign volume: rpc error: code = Unknown
  desc = failed to find writable volumes for collection:bucket-<uuid> replication:001 ttl:
  error: No writable volumes
```

**Harbor (S3 frontend caller):**
```
HTTP 500
{"errors":[{"code":"UNKNOWN","message":"unknown error",
  "detail":{"DriverName":"s3aws","Enclosed":{"RequestFailure":{}}}}]}
```

**Effect on consumer (in this case a `kaniko` image builder pushing to Harbor whose registry storage backend is SeaweedFS-S3):**
```
error checking push permissions ... POST https://<harbor>/v2/<project>/<repo>/blobs/uploads/:
  UNKNOWN: unknown error; map[DriverName:s3aws Enclosed:map[RequestFailure:map[]]]
```

`kubectl delete pod -l app.kubernetes.io/component=volume` resolved the situation immediately. New volume processes re-paired with master cleanly.

## Root cause analysis (commit `87b71029f`, tag 4.05)

1. JWT signing key is read once in server constructors, stored in `security.Guard` struct:
   - `weed/server/volume_server.go:81-121`
   - `weed/server/master_server.go:93-160`
   - `weed/server/filer_server.go:120-186`
   - `weed/security/guard.go:45-60`
2. SIGHUP `Reload()` methods reload `security.toml` into viper but **do not refresh `SigningKey` / `ReadSigningKey` on the existing `Guard`**:
   - `weed/server/volume_server.go:167-173`
   - `weed/server/master_server.go:431-439`
3. HTTP client side is guarded by `sync.Once` — JWT key loaded exactly once per process lifetime:
   - `weed/util/http/http_global_client_util.go:29-39`
4. No `fsnotify` / `viper.WatchConfig` / `OnConfigChange` anywhere in the tree.

## Expected behaviour

JWT signing key should be hot-reloadable. Minimal acceptable behaviour: SIGHUP path (which already exists for whitelist refresh) refreshes the in-memory `SigningKey` / `ReadSigningKey` so operators can rotate keys without process restart.

## Workaround

Restart affected `weed` processes after any event that could cause JWT key drift between master / volume / filer / s3.

## Related (different root causes)

Existing `wrong jwt` issues (#1399, #1808, #6220, #6003) cover clock skew, large-upload timeouts, and missing bearer tokens on filer→volume streams. None of them addresses the key-mismatch recovery path described here.

chrislusf:
Fixes #9823.

## Problem

JWT signing keys are read once in the server constructors and stored on the in-memory `security.Guard`. The SIGHUP `Reload()` paths refresh the IP whitelist but never the signing keys, and the global chunk-read JWT is cached behind a `sync.Once`. So after a key rotation (mounted Secret update, divergent reads, apiserver blip) the stale key sticks and every request keeps failing `401: wrong jwt` until the affected `weed` process is restarted.

Two extra gaps surfaced while fixing this:
- `MasterServer.Reload()` existed but was never registered with `grace.OnReload`, so the master ignored SIGHUP entirely (whitelist included).
- The Rust volume server's SIGHUP handler had the identical bug: it refreshed the whitelist but not the signing keys.

## Change

- Add `Guard.UpdateSigningKeys` and call it from the master / volume / filer reload paths and the s3 reload hook, next to the existing whitelist refresh.
- Make the global chunk-read JWT (`jwt.signing.read.key`) reloadable via an atomic pointer swap instead of a one-shot `sync.Once`.
- Register `MasterServer.Reload` with `grace.OnReload`.
- Mirror the same refresh in the Rust volume server (`Guard::update_signing_keys` + SIGHUP handler).

Operators can now rotate keys and recover with `kill -HUP` instead of restarting every process.

## Tests

- Go: `TestUpdateSigningKeysRotates`, `TestUpdateSigningKeysTogglesWriteActive` — a token minted with the rotated key verifies and the old key is rejected; the write-active flag tracks key presence.
- Rust: `test_update_signing_keys_rotates`, `test_update_signing_keys_toggles_write_active`.

## Summary by CodeRabbit

* **New Features**
  * Master, volume, filer and S3 servers now pick up updated JWT signing keys and expiration settings on config reload (no restart).
  * HTTP-layer JWT read configuration is reloaded with security changes.
  * Write authorization now toggles dynamically when signing keys or whitelist settings change.

* **Tests**
  * Added tests for signing-key rotation, concurrent rotations, token validation with new/old keys, and write-auth state transitions.

### DIFF
--- a/seaweed-volume/src/main.rs
+++ b/seaweed-volume/src/main.rs
@@ -545,6 +545,12 @@ async fn run(
                     whitelist.extend(sec.guard_white_list.iter().cloned());
                     let mut guard = state_reload.guard.write().unwrap();
                     guard.update_whitelist(&whitelist);
+                    guard.update_signing_keys(
+                        SigningKey(sec.jwt_signing_key),
+                        sec.jwt_signing_expires,
+                        SigningKey(sec.jwt_read_signing_key),
+                        sec.jwt_read_signing_expires,
+                    );
                 }
 
                 // Trigger heartbeat to report new volumes
--- a/seaweed-volume/src/security.rs
+++ b/seaweed-volume/src/security.rs
@@ -172,6 +172,26 @@ impl Guard {
         self.is_write_active = !is_empty_whitelist || !self.signing_key.is_empty();
     }
 
+    /// Refresh the JWT signing keys and their expirations in place so operators
+    /// can rotate keys via SIGHUP without restarting the process. Mirrors Go's
+    /// `Guard::UpdateSigningKeys`: it swaps the fields and recomputes
+    /// `is_write_active`.
+    pub fn update_signing_keys(
+        &mut self,
+        signing_key: SigningKey,
+        expires_after_sec: i64,
+        read_signing_key: SigningKey,
+        read_expires_after_sec: i64,
+    ) {
+        self.signing_key = signing_key;
+        self.expires_after_sec = expires_after_sec;
+        self.read_signing_key = read_signing_key;
+        self.read_expires_after_sec = read_expires_after_sec;
+
+        let is_empty_whitelist = self.whitelist_ips.is_empty() && self.whitelist_cidrs.is_empty();
+        self.is_write_active = !is_empty_whitelist || !self.signing_key.is_empty();
+    }
+
     /// Check if a remote IP is in the whitelist.
     /// Returns true if write security is inactive (no whitelist and no signing key),
     /// if the whitelist is empty, or if the IP matches.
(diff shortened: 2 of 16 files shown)


chrislusf:
Addressed the data-race feedback in 4a4ba6f.

`Guard.UpdateSigningKeys` no longer mutates the `[]byte` fields in place. The four signing-key fields now live in an immutable `signingConfig` snapshot behind `atomic.Pointer`; a rotation swaps the whole pointer, so a concurrent reader sees either the old keys or the new ones — never a torn slice header. Reads go through new `SigningKey()` / `ExpiresAfterSec()` / `ReadSigningKey()` / `ReadExpiresAfterSec()` accessors (lock-free, so the hot assign/auth paths are unaffected). Added `TestUpdateSigningKeysConcurrent`, which rotates while readers mint and verify tokens and passes under `go test -race`.

The Rust guard was already safe: every read and the SIGHUP write go through the shared `RwLock<Guard>`.

chrislusf:
Addressed all three in a42fd38.

- **guard.go (auth/whitelist sync boundary):** moved *all* hot-reloadable state — keys, expirations, whitelist maps, and the `isWriteActive`/`isEmptyWhiteList` flags — into a single immutable `guardState` swapped behind one `atomic.Pointer`. `WhiteList()`/`IsWhiteListed()` and the key accessors all read from one `Load()`, so a SIGHUP can no longer expose a mix of old/new activation bits and whitelist contents. The `Update*` methods take a small mutex to serialize their read-modify-write; readers stay lock-free.
- **filer + volume JWT checks (double-load):** each branch now reads the key once into a local and branches on it, so a reload landing mid-check can't take the allow-fast-path after auth was enabled or verify against a different key than the branch condition saw.

`TestUpdateSigningKeysConcurrent` now also rotates the whitelist and probes `IsWhiteListed` concurrently; the package passes under `go test -race`.

Issue #9830, opened 2026-06-05
ip.bind does not actually guarantee seaweedfs to use the ip


wzrdtales:
The behavior of ip.bind currently only guarantees the binding. However, it should as well cover **outgoing** requests. An outgoing request will just pick an arbitrary IP, which may not even be able to reach the target.

PR #9834, opened 2026-06-05
ip.bind: bind outbound connections to the configured address

chrislusf:
Addresses #9830.

`-ip.bind` only controlled which address the listeners bound to. Connections this process *initiates* (gRPC to masters/volumes/filers, HTTP fetches between components, etc.) still let the kernel pick a source IP, which on a multi-homed host may be one that can't reach the target.

This mirrors the bind address into a process-global outbound source address and applies it to outbound TCP dials:

- gRPC: a context dialer with `LocalAddr` set, installed only when a bind address is configured (default deployments keep gRPC's stock dialer).
- HTTP: `DialContext` on the per-client transports and on `http.DefaultTransport`. The setting is read per dial, so it applies no matter when it's configured relative to client construction.

Each daemon registers its resolved bind address at startup (`master`, `volume`, `filer`, `s3`, `sftp`, `webdav`), and `weed server` sets it once up front so every embedded component inherits it before any of them dials.

Loopback targets (`localhost` / `127.0.0.0/8` / `::1`) and unix sockets keep the OS-chosen source: a routable bind IP can't originate a packet on the loopback interface, so same-host traffic (including the local gRPC unix-socket fast path) is unaffected. A wildcard bind (`0.0.0.0` / `::`) or empty value leaves the source address to the OS, preserving current behavior.

Unit tests cover the parse/clear logic and the per-dial bind decision (remote vs loopback, tcp vs unix/udp).

## Summary by CodeRabbit

* **New Features**
  * Servers can bind outbound network connections to a configured local IP, making source IPs consistent across HTTP, gRPC, and other protocols.
  * Once a non-empty outbound IP is set at startup, it is locked (first-call wins) and later changes are ignored.

* **Tests**
  * Added comprehensive tests covering outbound binding behavior, IP validation, family mismatches, and loopback handling.

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -335,6 +335,7 @@ func (fo *FilerOptions) startFiler() {
 	if *fo.bindIp == "" {
 		*fo.bindIp = *fo.ip
 	}
+	util.SetOutboundLocalIP(*fo.bindIp)
 	if *fo.allowedOrigins == "" {
 		*fo.allowedOrigins = "*"
 	}
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -188,6 +188,7 @@ func startMaster(masterOption MasterOptions, masterWhiteList []string) {
 	if *masterOption.ipBind == "" {
 		*masterOption.ipBind = *masterOption.ip
 	}
+	util.SetOutboundLocalIP(*masterOption.ipBind)
 
 	myMasterAddress, peers := checkPeers(*masterOption.ip, *masterOption.port, *masterOption.portGrpc, *masterOption.peers)
 
--- a/weed/command/s3.go
+++ b/weed/command/s3.go
@@ -324,6 +324,7 @@ func (s3opt *S3Options) startS3Server() bool {
 	if *s3opt.bindIp == "" {
 		*s3opt.bindIp = "0.0.0.0"
 	}
+	util.SetOutboundLocalIP(*s3opt.bindIp)
 
 	defaultFileMode, fileModeErr := s3opt.parseDefaultFileMode()
 	if fileModeErr != nil {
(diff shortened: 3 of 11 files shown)


chrislusf:
Addressed both review findings in 7bacec3:

- Address-family mismatch (Gemini): `outboundLocalAddrForDial` now skips source binding when the target is a literal IP of a different family than the bind address, so an IPv4 bind dialing an IPv6 literal (or vice versa) falls back to the OS-chosen source instead of failing the dial. Hostnames still bind optimistically; the dialer matches the source family among resolved candidates.
- Clobbering the process-global source IP (CodeRabbit): `SetOutboundLocalIP` is now first-write-wins. `weed server` applies the top-level `-ip.bind` before its embedded components start, so a component's own `-x.ip.bind` (e.g. an explicit `-s3.ip.bind=0.0.0.0`) can no longer overwrite the address the other components already dial from. Standalone daemons are unaffected since each makes the single authoritative call.

Both paths have unit tests.

wzrdtales:
tested this @chrislusf the s3 server still tried to reach out with a foreign address do i need to specify more than `-ip.bind=ip` ?

wzrdtales:
```
E0607 10:15:14.685949 s3api_object_handlers.go:878 GetObjectHandler: failed to stream registry/docker/registry/v2/blobs/sha256/57/5761ee581278864f629140e3a7a5b1ed64f4400c7239f7333a75de07f9e93e19/data from volume servers: Get "
http://volume:8080/227787,0b658ae1652dcd42?readDeleted=true": dial tcp correctlocalip:0->volume:8080: connect: connection refused
```

the log is actually weird. it shows the local ip of the node, but port 0, but still got connection refused, which starts as soon as the other ips get added

Issue #9816, opened 2026-06-03
Volume grow stuck / doesn't complete


kklingenberg:
Hello @chrislusf !

After experiencing a forceful shut down, we have an issue with a seamingly locked cluster that won't grow automatically or manually, without any error message or indication. Grow operations simply don't complete. We can read data just fine, however.

**Setup:**

- 5 virtual machines:
  - 1 for the single master and single filer, with S3 gateway
  - 4 for the volume servers, each with 3 volume servers assigned
  - Each volume server has a single SSD of about 100TB; Total capacity should be about 1.2PB
- Version: `30GB 4.23 73fc9e3833cfff2c259e0239ed6c0444e1c1890a linux amd64`
- No replication (`000`)
- No EC volumes
- `max=0`

**Cluster status:**

```text
cluster:
	id:       topo
	status:   unlocked
	nodes:    12
	topology: 1 DC, 12 disks on 1 rack

volumes:
	total:    14,604 volumes, 2 collections
	max size: 32 GB
	regular:  14,604/41,460 volumes on 14,605 replicas, 14,514 writable (99.37%), 91 read-only (0.62%)
	EC:       0 EC volumes on 0 shards (0 shards/volume)

storage:
	total:           459 TB (459 TB raw, 100%)
	regular volumes: 459 TB (459 TB raw, 100%)
	EC volumes:      0 B (0 B raw, 0%)
```

**Volumes:**

[volumes.txt](https://github.com/user-attachments/files/28570354/volumes.txt)

**Further context:**

We had a seamless operation until our VMs were forecefully shut down because of an issue with our hypervisor. The clients were populating an S3 bucket and had reached about 418TB of used capacity. After having restarted everything however, no more writes were allowed and the master started emitting these:

```text
master_grpc_server_assign.go:163 assign count:1  collection:"eodata"  expected_data_size:8388608 {"collection":"eodata","replication":{},"ttl":{"Count":0,"Unit":0},"version":3}: failed to find writable volumes for collection:eodata replication:000 ttl: error: No writable volumes
```

Before those, there's this record of an attempt to grow the collection:

```text
master_grpc_server_volume.go:153 volume grow &{Option:{"collection":"eodata","replication":{},"ttl":{"Count":0,"Unit":0},"version":3} Count:0 Force:false Reason:grpc assign}
```

[...]

PR #9839, opened 2026-06-05
master: migrate legacy raft identity to hashicorp raft, deprecate legacy raft

chrislusf:
## What

Make the legacy `seaweedfs/raft` -> hashicorp raft switch lossless, and deprecate the legacy engine.

### 1. Carry cluster identity into hashicorp raft

The master raft FSM holds only `{MaxVolumeId, TopologyId}`, and `MaxVolumeId` is rebuilt from volume heartbeats — so switching engines is "discard old state, rebuild." The one thing that is *not* recoverable is `TopologyId` (cluster identity). Previously the hashicorp server only recovered `TopologyId` from a hashicorp-format snapshot, so a master switching from legacy generated a brand-new identity.

On first start on hashicorp raft, `migrateLegacyRaftStateIfNeeded`:
- checks a marker in the hashicorp stable store; if present, it is already migrated and nothing happens;
- otherwise, if a legacy `snapshot/` exists, imports `TopologyId` + `MaxVolumeId` from it (seeded before raft starts, so `ensureTopologyId` won't generate a fresh id);
- marks migrated either way, so the import runs at most once and a fresh install never re-scans.

Safe across paths: stale legacy files next to live hashicorp state can't override it (raft replay re-asserts the real id, and `MaxVolumeId` only rises).

### 2. Deprecate legacy seaweedfs/raft

Startup warning when the legacy engine is selected, plus a deprecation note in `-raftHashicorp` help.

The default is intentionally **unchanged** in this PR: flipping it would silently turn a multi-master rolling upgrade into a mixed, incompatible-protocol cluster mid-restart. Operators opt in deliberately; the identity migration carries state across. The default flip can follow once this is proven.

## Commits
- `master: carry cluster identity from legacy raft into hashicorp raft`
- `master: deprecate legacy seaweedfs/raft`

## Tests
`raft_migration_test.go` covers import-once (marker gates re-import), no-legacy-just-marks, and the legacy-state existence check.

## Context
Surfaced by #9816, where a single-master legacy raft loop wedged after a hard power-off and stalled all volume growth via `NextVolumeId`. This makes hashicorp raft a safe, identity-preserving destination.

## Summary by CodeRabbit

* **New Features**
  * Added automatic state migration to seamlessly upgrade cluster configurations from legacy Raft to Hashicorp Raft during node startup.

* **Documentation**
  * Updated all CLI flag help text to recommend Hashicorp Raft and clearly indicate the legacy Raft implementation as deprecated.

* **Deprecation Notice**
  * Added warning message at startup when clusters are running with the deprecated legacy Raft implementation.

### DIFF
--- a/weed/command/master.go
+++ b/weed/command/master.go
@@ -103,7 +103,7 @@ func init() {
 	m.raftResumeState = cmdMaster.Flag.Bool("resumeState", true, "resume previous state on start master server")
 	m.heartbeatInterval = cmdMaster.Flag.Duration("heartbeatInterval", 300*time.Millisecond, "heartbeat interval of master servers, and will be randomly multiplied by [1, 1.25)")
 	m.electionTimeout = cmdMaster.Flag.Duration("electionTimeout", 10*time.Second, "election timeout of master servers")
-	m.raftHashicorp = cmdMaster.Flag.Bool("raftHashicorp", false, "use hashicorp raft")
+	m.raftHashicorp = cmdMaster.Flag.Bool("raftHashicorp", false, "use hashicorp raft (recommended; legacy seaweedfs/raft is deprecated)")
 	m.raftBootstrap = cmdMaster.Flag.Bool("raftBootstrap", false, "Whether to bootstrap the Raft cluster")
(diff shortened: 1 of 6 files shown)


chrislusf:
Closing since upgrading existing clusters with legacy raft to hashicorp raft would require a stop-all-masters operation, which may not be feasible for all use cases.

PR #9841, opened 2026-06-06
s3: rescan .versions when the cached latest pointer is missing on a list

chrislusf:
Companion to the RECOMPUTE_LATEST notify fix, on the read side.

ListObjectsV2 (what `mc du` / `mc ls` use) resolves a versioned object's current version from the latest-version pointer cached on the `.versions` directory entry, for a single-scan listing. When that pointer is absent on the filer serving the list - e.g. a multi-filer cluster where the pointer was written on the key's owner filer and hasn't converged - `getLatestVersionEntryFromDirectoryEntry` returned an error and the object was silently dropped from the listing, so the bucket looked far smaller than it is.

Fall back to a read-only rescan of `.versions/` to pick the newest version when the cached pointer is missing. The version files replicate even when the pointer doesn't, so the rescan resolves the current version locally. This is the listing-side counterpart to the read path's `recoverLatestVersionWithoutPointer`; the pagination/select loop is factored into a shared `scanLatestVersionEntry`.

Read-only on purpose: a list can touch many objects, so it does not persist a pointer (no write amplification). Overhead is zero when the pointer is present (fast path unchanged) and ~one local `.versions/` list per pointer-missing object otherwise.

Verified on a 3-filer cluster: source 122 objects, versioned bucket `mc du` 122/168 MiB on all endpoints (was 32/4 MiB). s3api unit tests pass; single-filer versioning tests unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved object version listing reliability by adding a scan-based recovery when cached version pointers or metadata are stale, missing, or partial.
  * Prevented hard failures during listings by falling back to rescans instead of erroring when latest-version fields are absent.
  * Better detection and handling of delete markers and scenarios with no remaining versions to ensure consistent list responses.

### DIFF
--- a/weed/s3api/s3api_object_versioning.go
+++ b/weed/s3api/s3api_object_versioning.go
@@ -1829,6 +1829,35 @@ func selectLatestVersion(entries []*filer_pb.Entry) (latestEntry *filer_pb.Entry
 	return
 }
 
+// scanLatestVersionEntry paginates a .versions/ directory and returns the
+// chronologically newest version entry (including delete markers; see
+// selectLatestVersion). A single-shot list would miss the true latest when
+// old-format (raw timestamp) version ids spill past one page, since filesystem
+// order is lexicographic-ascending = oldest-first for that format. latestEntry
+// is nil when the directory holds no version entries.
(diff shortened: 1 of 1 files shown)


chrislusf:
Good catch - copied the Extended map before stamping the version id instead of mutating the scanned entry's map.

PR #9853, opened 2026-06-07
operation: index VidCache by map instead of slice

### LINKED ISSUE #9850: [Bug]: Volume server use huge memory,  when update from 3.93 to 4.31
VidCache causes multi-GB memory explosion on clusters with large volume IDs (OOM kill)

Version 4.31

Summary

Volume servers consume 8-17 GB of heap memory on clusters where volume IDs have grown into the hundreds of millions. This causes cascading OOM kills when multiple volume server processes share the same host.

Symptoms

1. Volume server RSS grows to 9-17 GB (expected: 200-500 MB)
2. /proc/<pid>/status shows RssAnon ≈ 9.6 GB (pure heap, not file cache)
3. Restarting one volume server triggers other volume servers on the same host to spike in memory and get OOM-killed
4. Problem did not exist on version 3.93 with identical data and workload

pprof Evidence
heap profile: 123: 4400791440 [362: 12187969728] @ heap/1048576

[heap.txt](https://github.com/user-attachments/files/28682836/heap.txt)

# Largest live allocations:
1: 2409938944 [1: 2409938944] @ ... operation.(*VidCache).Set+0x2a8
1: 1927938048 [1: 1927938048] @ ... operation.(*VidCache).Set+0x2a8
1: 1542340608 [1: 1542340608] @ ... operation.(*VidCache).Set+0x2a8
1: 1233862656 [1: 1233862656] @ ... operation.(*VidCache).Set+0x2a8
...
(~10 entries totaling several GB, all from VidCache.Set)

# runtime.MemStats
# HeapAlloc = 17461814712
# HeapInuse = 17469333504
Full call path:

VolumeServer.PostHandler
  → topology.ReplicatedWrite
    → topology.GetWritableRemoteReplications
      → operation.LookupVolumeId
        → operation.LookupVolumeIds
          → VidCache.Set    ← allocations here
Environment

- ~76 volumes per disk, 36 disks per server, 64 GB total RAM
- Volume IDs auto-assigned by master over years: max volume ID ≈ 166,000,000
- Replication between volume servers on the same host
- Index mode: leveldb with -index.leveldbTimeout=1

Likely Cause

VidCache in weed/operation/lookup_vid_cache.go uses a []VidInfo slice indexed directly by numeric volume ID. When Set("166117011", ...) is called, the slice grows to 166M+ entries. Each VidInfo is ~48 bytes, so a single Set() allocates ~8 GB for a mostly-empty array.

This was latent before 4.03 because commit ae7333d28 fixed an inverted condition (Error != "" → Error == ""). Before the fix, successful lookups were never cached, so Set() was never called with valid volume IDs on the replication write path.

Workaround

Roll back to 3.93

chrislusf:
## Problem

`VidCache.cache` is a `[]VidInfo` indexed directly by volume id. `Set` grows the slice until its length reaches the id, appending a zeroed `VidInfo{}` for every index below it:

```go
if id > len(vc.cache) {
    for i := id - len(vc.cache); i > 0; i-- {
        vc.cache = append(vc.cache, VidInfo{})
    }
}
```

On clusters that have churned up to high volume ids (deletes push ids up while the active count stays modest), caching a single volume with a large id allocates an entry for every id below it. The cache sits on the hot write-replication path (`PostHandler -> ReplicatedWrite -> GetWritableRemoteReplications -> LookupVolumeIds -> vc.Set`), so volume servers pay this on every write.

Reproduced locally — caching exactly one volume with id `32,000,000`:

```
backing slice length = 32000000 entries
VidInfo size         = 48 bytes
heap growth          = 1470.8 MB     (from a single cached volume)
```

That matches the heap profile in the issue (~1.5GB resident, plus geometric realloc churn from the append-doubling loop as it grows).

## Fix

Use `map[uint32]VidInfo`. Memory now scales with the number of volumes actually cached rather than the largest id seen, and the append-grow loop and its reallocation churn are gone. `Get`/`Set` behavior is otherwise unchanged.

Added a regression test asserting a single large id produces exactly one cache entry.

Fixes #9850

## Summary by CodeRabbit

* **Bug Fixes**
  * Cache storage reworked for more efficient handling of large volume IDs and improved expiration checks to avoid stale results.
  * Invalid or non-numeric volume IDs now produce parsing errors instead of being treated as missing entries.

* **Tests**
  * Added coverage for caching behavior with very large and out-of-range volume identifiers.

### DIFF
--- a/weed/operation/lookup_vid_cache.go
+++ b/weed/operation/lookup_vid_cache.go
@@ -17,43 +17,46 @@ type VidInfo struct {
 }
 type VidCache struct {
 	sync.RWMutex
-	cache []VidInfo
+	cache map[uint32]VidInfo
 }
 
 func (vc *VidCache) Get(vid string) ([]Location, error) {
-	id, err := strconv.Atoi(vid)
+	id, err := strconv.ParseUint(vid, 10, 32)
 	if err != nil {
 		glog.V(1).Infof("Unknown volume id %s", vid)
 		return nil, err
 	}
+	if id == 0 {
+		return nil, ErrorNotFound
+	}
 	vc.RLock()
 	defer vc.RUnlock()
-	if 0 < id && id <= len(vc.cache) {
-		if vc.cache[id-1].Locations == nil {
-			return nil, errors.New("not set")
-		}
-		if vc.cache[id-1].NextRefreshTime.Before(time.Now()) {
-			return nil, errors.New("expired")
-		}
-		return vc.cache[id-1].Locations, nil
+	info, found := vc.cache[uint32(id)]
+	if !found || info.Locations == nil {
+		return nil, ErrorNotFound
 	}
-	return nil, ErrorNotFound
+	if info.NextRefreshTime.Before(time.Now()) {
+		return nil, errors.New("expired")
+	}
+	return info.Locations, nil
 }
+
 func (vc *VidCache) Set(vid string, locations []Location, duration time.Duration) {
-	id, err := strconv.Atoi(vid)
+	id, err := strconv.ParseUint(vid, 10, 32)
 	if err != nil {
 		glog.V(1).Infof("Unknown volume id %s", vid)
 		return
 	}
+	if id == 0 {
+		return
+	}
 	vc.Lock()
 	defer vc.Unlock()
-	if id > len(vc.cache) {
-		for i := id - len(vc.cache); i > 0; i-- {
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the out-of-range id nitpick: `Get`/`Set` now parse with `strconv.ParseUint(vid, 10, 32)`, so ids past the uint32 volume-id range are rejected rather than silently wrapping to key 0. Added that boundary case ("4294967296") to the test.

PR #9855, opened 2026-06-07
ec placement: spread EC shards evenly across machines, not onto the lowest-id one

chrislusf:
Follow-up to #9854. On a cluster of co-located volume servers, the first machine (lowest id) accumulated far more EC shards than the rest — e.g. on six machines the first sat at ~1.7x the others after many encodes.

Cause: EC encode places every volume against one shared topology snapshot (it reserves the shards it assigns, so later volumes see reduced capacity), but the within-rack node selection ranked candidates only by *this volume's* shard count and broke ties by sorted id. Since each volume starts with zero of its own shards everywhere, the first shard of every volume went to the lowest-id machine, and the accumulated load was never considered.

Fix: rank eligible nodes by the machine's shards of this volume, then the machine's free capacity, then the node's shards of this volume, then the node's free capacity. Free capacity reflects load already placed, so ties steer toward the least-loaded machine rather than the lowest id, keeping total EC shards even.

`TestPlaceDistributesEvenlyAcrossManyVolumes` places 120 volumes against a shared snapshot across six two-server machines and asserts the per-machine shard totals stay even (it fails on the old sorted-id tie-break: 360/360/240/240/240/240).

## Summary by CodeRabbit

* **Improvements**
  * Enhanced placement logic to prefer machines and nodes with fewer shards and greater free capacity, improving evenness of shard distribution across hosts.

* **Tests**
  * Added tests ensuring even shard distribution across many volumes and that rebalancing converges to a tight per-host spread over multiple rounds.

### DIFF
--- a/weed/storage/erasure_coding/ecbalancer/balancer.go
+++ b/weed/storage/erasure_coding/ecbalancer/balancer.go
@@ -1120,8 +1120,7 @@ func countShardsByHost(vk volKey, nodes map[string]*Node) map[string]int {
 	return m
 }
 
-// freeSlotsByHost sums each machine's free EC shard slots, for steering placement
-// toward less-loaded machines.
+// freeSlotsByHost sums each machine's free EC shard slots.
 func freeSlotsByHost(nodes map[string]*Node) map[string]int {
 	m := make(map[string]int)
 	for _, node := range nodes {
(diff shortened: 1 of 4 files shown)


chrislusf:
Addressed both in dc0be0343: `perMachine` now seeds every host (so a fully drained machine still counts), and the convergence loop fails explicitly if it doesn't reach a no-op round before the 200-round cap.

PR #9876, opened 2026-06-08
feat(admin): export full cluster volume list as JSON

chrislusf:
Adds an **Export All (JSON)** button on the admin UI's Cluster Volumes page. It pulls the whole cluster's volume list from the master in a single `VolumeList` call and streams it back as a downloadable JSON report, a superset of the `volume.list` shell command.

The existing per-page CSV button (now labeled **Export Page**) is unchanged; it still scrapes the rendered table for the current page.

### Endpoint
`GET /api/volumes/export[?collection=<name>]` (authenticated read API). Streams indented JSON with `Content-Disposition: attachment; filename=seaweedfs-volumes-<timestamp>.json`. The button carries the page's active collection filter.

### Report shape
Hierarchy-preserving (data center -> rack -> node -> disk -> volumes / EC shards), split per physical disk like `volume.list`, with cluster `totals` and a `duplicate_volume_ids` block.

Fields beyond the table columns / `volume.list` output:
- per volume: `garbage_ratio`, `fullness_ratio` (vs size limit), human-readable `modified_at`, `compact_revision`, `remote_storage_name/key`, `disk_type/disk_id`
- per disk: `max/active/free/remote_volume_count`
- per EC shard: per-shard sizes, `file_count`, `delete_count`
- cluster-wide cross-collection duplicate-volume-id scan

Decoded `replica_placement` and `ttl` reuse `storage.NewVolumeInfo`, same as the shell command. No proto, master, or volume-server changes.

## Summary by CodeRabbit

* **New Features**
  * Added full-cluster JSON export via a new API at /api/volumes/export (downloads an indented .json file).
  * UI now offers both “Export Page” and “Export All (JSON)” buttons for page-only or full-cluster exports.
  * Exports include optional collection filtering, per-disk volume/EC-shard details, duplicate-volume detection, and derived metrics (garbage/fullness ratios, totals).

### DIFF
--- a/weed/admin/dash/volume_export.go
+++ b/weed/admin/dash/volume_export.go
@@ -1,0 +1,340 @@
+package dash
+
+import (
+	"context"
+	"sort"
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
+	"github.com/seaweedfs/seaweedfs/weed/storage"
+	"github.com/seaweedfs/seaweedfs/weed/storage/erasure_coding"
+)
+
+// VolumeListExport is a full-cluster volume report, downloadable from the admin
+// UI as JSON. It mirrors the topology the `volume.list` shell command walks
+// (data center -> rack -> node -> disk -> volumes / EC shards) and adds derived
+// fields the shell command does not print: garbage and fullness ratios, a
+// human-readable modified time, remote-tiering keys, per-disk capacity counts,
+// and the cluster-wide duplicate-volume-id scan.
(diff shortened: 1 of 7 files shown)


chrislusf:
Addressed: `ExportClusterVolumeList` now takes a `context.Context` and passes it straight to `client.VolumeList`; the handler hands in `r.Context()`, so the export cancels when the client disconnects. Left out an arbitrary fixed timeout since the request context already bounds the work and a magic value isn't a pattern here.

Issue #9866, opened 2026-06-08
Remote metadata is not synced properly 


Aison0:
Hello,
I have two seaweedfs clusters, one remote with a very large storage and one as local store (as cache). On the local store I mounted a bucket from the remote store with the commands

```
remote.configure -name=phoenix -type=s3 -s3.access_key=xxx -s3.secret_key=yyy -s3.endpoint=http://mys3seaweedfiler:8080
remote.mount -dir=/buckets/clips -remote=phoenix/clips
```

After syncing the metadata from the remote bucket, the files and directories are shown, but the directory dates differ, also some of the posix permissions are different. I took two screenshots of the file view of both admin ui's.

Remote server:
<img width="1592" height="892" alt="Image" src="https://github.com/user-attachments/assets/e897d96f-2759-4ce4-a5f4-bfd75f410118" />

Local server:
<img width="1592" height="888" alt="Image" src="https://github.com/user-attachments/assets/3f9b148c-7e62-4987-b1df-502f0ad3b275" />

There seems to be a problem with transfering the data also. When I download a remote file over the admin ui from the local cluster, the size matches, but the content is NULL.

version 30GB 4.32 78da9572aee7589f773ca09a987b494888829ba3 linux amd64

chrislusf:
https://github.com/seaweedfs/seaweedfs/pull/9879 should address this

PR #9879, opened 2026-06-08
fix(remote): correct content and permissions when syncing/caching remote objects

chrislusf:
Addresses #9866 (mounting one SeaweedFS cluster's bucket onto another via `remote.mount`). Two distinct problems.

## 1. Cached content is NULL while the size matches

The on-demand cache fetch never validated that the bytes read from the remote matched the requested size:

- The S3 client (`ReadFileWithConcurrency`) pre-allocates `make([]byte, size)`, wraps it in `aws.NewWriteAtBuffer`, and discards the downloaded byte count. The AWS SDK's `io.Copy` doesn't check it received the full requested range, so a short-but-well-formed response (stale listing size, truncated/flaky response) leaves the buffer **zero-padded** and returns `nil` error. Azure's client has the identical pre-sized-buffer / discarded-count bug.
- `FetchAndWriteNeedle` then records the chunk with the *requested* size, so the file looks complete.

Result: a cached file whose size matches but whose tail is NULL. Worse, the entry is marked cached and the remote-only flag cleared, so the corruption is sticky and never re-fetches.

Reproduced with a proxy that reports size 3600000 in the listing but delivers only 100 bytes on the ranged GET: before, HTTP 200 with 3599900 NULL bytes, persisted as cached; after, 503 + Retry-After, the entry stays remote-only, and the log shows `short read ... got 100 bytes, want 3600000`. Healthy full reads still cache correctly.

Fix: check the downloaded byte count against the requested size in the S3 and Azure clients, plus a backend-agnostic `len(data) == req.Size` guard in `FetchAndWriteNeedle`.

## 2. Permissions diverge from the source

Remote object listings carry no POSIX mode, so synced entries were created with a hardcoded `0644`. Against a SeaweedFS remote — whose S3 layer writes objects as `0660` and auto-creates directories as `0771` (`0660|0111`) — the mounted copy ended up `0644`/`0755`, visibly different from the source.

Fix: default to the S3 modes (files `0660`, directories `0771`) in `pullMetadata` and `remote.cache`. The filer derives parent-dir modes from the child as `fileMode|0111`, so fixing the file default brings the directories into line too. Verified the mounted copy now matches the source (`0660` files, `0771` dirs).

Directory **mtimes** still reflect sync time: S3 listings don't enumerate directories, so the remote's directory timestamps aren't retrievable over the S3 protocol.

## Summary by CodeRabbit

* **Bug Fixes**
  * Remote reads (Azure, S3, and remote fetch) now validate the exact number of bytes returned and error on short reads to avoid treating zero-padded data as valid.
  * Remote cache and metadata sync now set file and directory permissions to S3-style defaults (files 0660, directories 0771) instead of a fixed permissive mode.

### DIFF
--- a/weed/remote_storage/azure/azure_storage_client.go
+++ b/weed/remote_storage/azure/azure_storage_client.go
@@ -324,7 +324,7 @@ func (az *azureRemoteStorageClient) ReadFileWithConcurrency(loc *remote_pb.Remot
 	}
 
 	data = make([]byte, size)
-	_, err = blobClient.DownloadBuffer(context.Background(), data, &blob.DownloadBufferOptions{
+	n, err := blobClient.DownloadBuffer(context.Background(), data, &blob.DownloadBufferOptions{
 		Range: blob.HTTPRange{
 			Offset: offset,
 			Count:  size,
@@ -334,6 +334,11 @@ func (az *azureRemoteStorageClient) ReadFileWithConcurrency(loc *remote_pb.Remot
 	})
 	if err != nil {
 		return nil, fmt.Errorf("failed to download file %s%s: %w", loc.Bucket, loc.Path, err)
+	}
+	// Pre-sized buffer: a short read stays zero-padded. Reject it rather than
+	// cache corrupt content.
+	if n != size {
+		return nil, fmt.Errorf("short read from %s%s at offset %d: got %d bytes, want %d", loc.Bucket, loc.Path, offset, n, size)
 	}
 
 	return data, nil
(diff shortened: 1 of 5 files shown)


Aison0:
Thx for the very fast fix!

> Directory mtimes still reflect sync time: S3 listings don't enumerate directories, so the remote's directory timestamps aren't retrievable over the S3 protocol.

(I guess it's the wrong place to ask) But instead of using some 3rd-party protocol like S3 for remote accessing another seaweed cluster, is there some better suitable, maybe bulit-in protocol, to achive the same?

In the end, I want a central cluster to which several branch offices are connected. The branch office servers should only replicate/cache the data from the main cluster that is used at their location. I don't want a 1:1 replication of all data. Therefore, I was thinking of setting up a central cluster and also a local cluster at each branch office, which would then connect to the central cluster via remote mount.

Issue #9883, opened 2026-06-09
Bug: filer.backup stalls forever on a superseded/deleted source chunk


kisow:
## Summary
`weed filer.backup`, when resuming from a checkpoint, stops making any progress and never advances past a single metadata event. 

The destination stops receiving updates. Restarting resumes at the same offset and stalls again. 

It happens when the source has rapid create/rename (e.g. Git over a FUSE mount) that is replayed from an older checkpoint.

## Version

* Source Filer : 4.25
* Target Filer : 4.25
* filer.backup : 4.25, 4.32

## Steps
```
weed filer.backup -filer=<source-filer> -filerPath=/buckets -doDeleteFiles
```

* replication.toml
```
[sink.filer]
enabled = true
grpcAddress = "<target-filer>"
directory = "/buckets"
replication = "002"
collection = "data"
ttlSec = 0
is_incremental = false
```

## Result
The same log lines repeat every few seconds indefinitely; the offset never advances:

```
I0608 17:43:22.604200 filer_sink.go:248 lookup /buckets/data/hub/git/vis-git-test.git/config.lock: filer: no entry is found in filer store 
I0608 17:43:22.614445 fetch_write.go:343 permanent size mismatch replicating 100,0c6da995cbf07f for /buckets/data/hub/git/vis-git-test.git/config: chunk size mismatch: read 100,0c6da995cbf07f got 0 bytes, source metadata says 107
E0608 17:43:22.614478 filer_sink.go:198 refuse to replicate entry with corrupt chunk /buckets/data/hub/git/vis-git-test.git/config: copy 100,0c6da995cbf07f: chunk size mismatch: read 100,0c6da995cbf07f got 0 bytes, source metadata says 107
[...]

PR #9886, opened 2026-06-09
fix(filer.backup): skip replay events whose source chunk was superseded or deleted

kisow:

* #9883

`filer.backup` (and the shared filer sink used by `filer.sync`) can get permanently stuck on a single metadata event and stop all replication.

**When the backup replays the source metadata log from a checkpoint, an event may reference a chunk whose needle was overwritten or deleted and garbage-collected on the source volume.** Fetching it now returns 0 bytes, correctly flagged as a permanent `errChunkSizeMismatch` — but the sink returned that error to the subscription, so the same offset was retried forever and replication never advanced. Restarting resumes at the same offset and wedges again.

Let the sink **skip** such an event, but only when it can prove the event is stale; otherwise keep refusing so genuine loss of a still-live file halts loudly instead of being dropped silently.

- `onCorruptChunk` centralizes the three `errChunkSizeMismatch` sites (`CreateEntry`, `UpdateEntry` repair/default) into one decision.
- `sourceSupersedes` re-reads the entry's current state on the source: **gone** (`ErrNotFound`, including wrapped / gRPC-string forms) or a **strictly-newer mtime** than the replayed version ⇒ skip; any other lookup error or a same/older version ⇒ keep (refuse).
- `getEntryMtimeNs` compares mtime at nanosecond precision so versions rewritten within the same second are ordered correctly.
- `targetPathToSourcePath` reports "unmappable" for incremental sinks (their date-prefixed keys cannot be reversed to a real source path), so a mis-mapped not-found never skips a live entry.

**Skipping is lossless because metadata events are full-entry snapshots, not deltas: a later event re-carries the current chunk set, and a delete event reconciles a removed file.** Only the provably-superseded case is auto-skipped; genuine loss of a live chunk still halts replication by design.

# How is the PR tested?

Unit tests in `weed/replication/sink/filersink` (`go test ./weed/replication/sink/filersink/`):

- `TestSourceSupersedes` — not-found (sentinel / wrapped / gRPC string) and nil entry → skip; network error → keep; source newer → skip; same/older → keep.
- `TestGetEntryMtimeNs` — nanosecond precision, same-second ordering, nil safety.
- `TestOnCorruptChunkRefusesWhenSupersessionUnconfirmed` — never skip silently when supersession is unconfirmed.
- `TestTargetPathToSourcePath` — incremental sink is reported unmappable.

Also verified end to end on a green→blue `filer.backup`: the run that previously wedged for 45+ minutes now skips the stale events (663 skips, 0 refuses), passes the stuck offset, and drains the backlog.

We also confirmed that, in the actual filer.backup production environment, processing continued successfully beyond that checkpoint.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved stale-source detection and retry behavior using nanosecond mtime propagation and a refined supersession check to avoid replaying older data.
  * Switched to nanosecond-precision timestamp comparisons for ordering and conflict resolution.
  * Corrupt-chunk handling now skips truly stale replays or surfaces genuine upload errors via a dedicated handler.
  * Incremental-mode sink target paths are treated as unmappable to avoid false deletions.

* **Tests**
  * Added tests for ns timestamp ordering, corrupt-chunk handling, source-supersession logic (including epoch mtime), and incremental-path mapping.

### DIFF
--- a/weed/replication/sink/filersink/fetch_write.go
+++ b/weed/replication/sink/filersink/fetch_write.go
@@ -381,10 +381,8 @@ func (fs *FilerSink) fetchAndWrite(sourceChunk *filer_pb.FileChunk, path string,
 
 const maxTransientBackoff = 2 * time.Minute
 
-// nextTransientBackoff returns the next backoff duration for a transient
-// network failure. It starts at 10s, doubles each time, and caps at 2 minutes
-// so an overloaded destination can recover instead of being hammered every
-// second by the surrounding RetryUntil loop.
+// nextTransientBackoff escalates 10s -> doubling -> 2m cap so an overloaded
+// destination can recover instead of being hammered by the RetryUntil loop.
 func nextTransientBackoff(current time.Duration) time.Duration {
 	if current < 10*time.Second {
 		return 10 * time.Second
@@ -403,12 +401,10 @@ func isEofError(err error) bool {
 	return errors.Is(err, io.ErrUnexpectedEOF) || errors.Is(err, io.EOF)
 }
 
-// isRetryableNetworkError reports whether err is a transient network failure
-// worth a backoff-and-retry: an interrupted read (EOF), a timeout (e.g. the
-// destination volume server hitting its idle deadline while reading a large
-// upload body under load), or a reset/broken connection. The volume server
-// returns the timeout as a JSON error string, so match on text in addition to
-// the net.Error interface.
+// isRetryableNetworkError reports whether err is a transient network failure worth
+// a backoff-and-retry: EOF, timeout (e.g. the destination's idle deadline under
+// load), or a reset/broken connection. The volume server returns the timeout as a
+// JSON string, so match on text as well as the net.Error interface.
 func isRetryableNetworkError(err error) bool {
 	if err == nil {
 		return false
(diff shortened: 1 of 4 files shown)


chrislusf:
Looked into the rename (P1) and Windows-colon (P2) findings. Both are real but root in pre-existing bugs on shared code paths, so I'm keeping this PR scoped to the supersession-skip and tracking them separately.

1. Rename application: the filer sink's `UpdateEntry` sends the old name to the filer `UpdateEntry` RPC, which updates in place and can't move entries, so renames are mishandled independent of this change. This PR's `onCorruptChunk` makes the chunk-error skip *conditional* (skip only when the source is provably gone/newer, otherwise return the error and halt) — strictly safer than the previous unconditional swallow, so it doesn't worsen the rename path. The supersession lookup itself already uses the new path (filer.sync passes unescaped keys to `UpdateEntry`, so that reverse-map is lossless). Tracked in #9891.

2. escapeKey reversibility: `filer_sync.go` runs every sink key through `escapeKey`, so on Windows a colon path is already mis-replicated (`/a:b` -> `/ab`) before the supersession check ever runs — the sink only sees the already-stripped key and can't recover it. The clean fix is to scope `escapeKey` to `localsink` in `filer_sync.go`; filer.backup is unaffected (it never escapes). Tracked in #9892.

kisow:
In fact, I also encountered an issue with Rename Application. However, in that case, the chunk was present and the root cause of the bug was different, so I planned to handle it in a separate PR. (The PR is kept as minimal as possible.)

Nice, this should solve everything in one shot. 👍

PR #9894, opened 2026-06-09
fix(filer.sync): scope filesystem key sanitization to the local sink

### LINKED ISSUE #9892: filer.sync escapeKey strips colons for all sink types, mis-replicating colon paths on Windows
`command/filer_sync.go`'s `destKey` runs every sink key through `escapeKey`, and the Windows build of `escapeKey` removes colons. Only `localsink` (writing to a local Windows filesystem, where colons are invalid in filenames) needs that. For `filersink`/`s3sink`/etc. a colon is a valid name byte, so on Windows a source path `/a:b` is replicated as `/ab` — a silent name divergence from the source.

Beyond the fidelity loss, this breaks the reverse mapping that the stale-chunk skip in #9886 relies on: `filersink.targetPathToSourcePath` maps the escaped sink key `/target/ab` back to `/source/ab`, which returns `ErrNotFound`, so a live corrupt event can be silently skipped instead of halting. Narrow trigger: Windows + filer.sync + a colon in the path + chunk corruption + a still-live source entry. (filer.backup is unaffected — it never escapes keys.)

Fix: scope `escapeKey` to `localsink` only (escaping is a property of the destination filesystem, not the sync), or thread the source path into the sink so no lossy reverse map is needed.

chrislusf:
Fixes #9892

`destKey` ran every sink key through `escapeKey`, whose Windows build strips colons. Colons are illegal in NTFS filenames so the **local** sink needs that, but s3/filer/azure/gcs/b2 accept colons as ordinary key bytes — stripping them silently diverged the destination key (a source `a:b` replicated as `ab`). The escaping was also inconsistent within a run: the in-place-update branch built keys that bypassed `escapeKey`, so on Windows an update wrote an escaped create key but deleted the unescaped old key.

Move the sanitization out of the command layer into the local sink, behind a Windows build tag, applied at every entry point (`CreateEntry`/`UpdateEntry`/`DeleteEntry`). `LocalIncSink` inherits it via embedding.

- Non-local sinks now receive the raw key (colons preserved).
- Non-Windows builds are byte-identical to before for every sink (`sanitizeFsKey` is identity).
- A leading drive-letter colon (`C:\...`) is preserved so absolute target paths stay valid — the old `escapeKey` stripped it, corrupting absolute Windows paths.

Verified host build/test/gofmt and `GOOS=windows` build/vet (which compiles the Windows-tagged source and tests).

## Summary by CodeRabbit

* **Bug Fixes**
  * Sync destination keys preserve colon characters for non-local sinks.
  * Local filesystem sinks now sanitize keys on Windows (preserving drive-letter prefixes) to avoid illegal path characters; non-Windows behavior is unchanged.

* **Tests**
  * Added platform-specific tests verifying key sanitization and that destination keys preserve colons where expected.


chrislusf:
Addressed the review nitpicks (pushed fe8db6286):

- **Incremental `destKey` coverage**: `TestDestKeyPreservesColonForNonLocalSink` now has a non-incremental and an incremental subtest; the incremental case asserts the colon-bearing name survives the date-partitioned key (without pinning the timezone-dependent date).
- **Windows local-sink coverage**: `TestCreateEntryWritesSanitizedPath` now also exercises `UpdateEntry` and `DeleteEntry`, confirming all three entry points sanitize the key, not just `CreateEntry`.

PR #9556, opened 2026-05-19
feat(filer): add atime to FuseAttributes + TouchAccessTime RPC

petedodd-pd:
# What problem are we solving?

SeaweedFS filer entries currently have no access-time tracking. FUSE clients see `atime` aliased to `mtime` from a bounded, in-memory overlay (`weed/mount/weedfs_attr.go` `atimeMap`), and other surfaces (S3, WebDAV, the filer HTTP API, `weed shell fs.meta.cat`) have no notion of last-read time at all. Downstream tooling — read-LRU cache eviction for remote-mounted buckets, hot-data detection, audit — has no signal it can rely on.

Concrete use case: capacity-driven LRU eviction of cached objects for remote-mounted buckets. Today only last-cached time (`RemoteEntry.LastLocalSyncTsNs`) is available; we need last-read time.

# How are we solving the problem?

This PR is the foundational, behaviour-preserving step. It adds the on-the-wire shape required by follow-ups, with zero behavioural change:

- `FuseAttributes` gains `int64 atime = 22;` and `int32 atime_ns = 23;` in `weed/pb/filer.proto` (and the Java client proto kept in lock step).
- A new `SeaweedFiler.TouchAccessTime` RPC and its `TouchAccessTimeRequest`/`TouchAccessTimeResponse` messages are declared, ready for the policy/handler PR that follows.
- The in-memory `filer.Attr` struct grows an `Atime time.Time` field and the codec helpers (`EntryAttributeToPb`, `EntryAttributeToExistingPb`, `PbToEntryAttribute`) round-trip it. Zero atime decodes as Mtime so legacy entries report a sane value and on-disk records remain wire-compatible.
- `Filer.CreateEntry` and `Filer.UpdateEntry` stamp `Atime = Mtime` (or `Crtime` if Mtime is zero) when callers do not set it explicitly — POSIX semantics where a freshly written file has `atime == mtime`.

No storage migration is required: every filer backend (`abstract_sql`, `mysql{,2}`, `postgres{,2}`, `cassandra{,2}`, `ydb`, `mongodb`, `redis{,2,3}`, `leveldb{,2,3}`, `rocksdb`, `etcd`, `tikv`, `tarantool`, `hbase`, `arangodb`, `foundationdb`, `elastic`) stores attributes inside an opaque `meta` protobuf blob, so the additive proto fields are picked up automatically.

Follow-ups (planned as separate, minimal PRs):

[...]

PR #9898, opened 2026-06-09
feat(filer): add AtimePolicy (off/relatime/strict) primitive

petedodd-pd:

Filers fronting remote/cloud storage need to evict their local cache by
*least-recently-accessed* (true LRU) rather than least-recently-synced, but the filer
does not track access time. This is the second slice of that work, following the merged
proto/codec foundation in #9556.

This PR adds **only** the policy primitive that decides *whether and when* an entry's
access time should be persisted. There is **no read-path wiring yet** — kept deliberately
small and self-contained so it can be reviewed in isolation. The server handler and CLI
flags that consume it follow in a stacked PR.

A new `filer.AtimePolicy` type modeled on POSIX mount options:

- `off` (default) — never track; no behaviour changes unless an operator opts in.
- `relatime` — persist only when the candidate atime is older than a threshold
  (default 24h) or older than the entry's mtime/ctime, debouncing frequent reads.
- `strict` — persist every forward-moving access.

An optional path-prefix allow-list (`AppliesToPath`) scopes tracking to specific
subtrees, matched at path-component boundaries so `/buckets/cache` matches
`/buckets/cache/x` but not `/buckets/cache_backup/x`. Helper parsers (`ParseAtimeMode`,
`ParsePathPrefixes`) support command-line plumbing in the follow-up PR.

Pure logic; no other package depends on it yet.

# How is the PR tested?

Unit tests in `weed/filer/atime_policy_test.go` cover the relatime predicate, strict
mode, default-off behaviour, path-prefix matching (including component-boundary and
trailing-slash cases), and the parse helpers.

`go test ./weed/filer/ -run 'AtimePolicy|ParseAtimeMode|PathPrefixes' -count=1`

# Checks

- [x] All AI code review comments have been addressed.

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added configurable atime update policy with modes: **off**, **relatime**, and **strict**.
  * Policies can be optionally scoped to specific path prefixes and include a configurable relatime threshold (with a sensible default when unset/invalid).
  * Added validation for atime mode and normalized parsing for path-prefix lists.
* **Tests**
  * Added unit tests for mode-specific update decisions, threshold debouncing behavior, parsing/validation, and path-prefix matching edge cases (including boundary-aware prefix checks).

### DIFF
--- a/weed/filer/atime_policy.go
+++ b/weed/filer/atime_policy.go
@@ -27,6 +27,9 @@ type AtimePolicy struct {
 	PathPrefixes []string
 }
 
+// NewAtimePolicy builds a policy from a mode string, relatime threshold and
+// optional path-prefix list. A non-positive threshold falls back to
+// DefaultRelatimeThreshold.
 func NewAtimePolicy(mode string, threshold time.Duration, pathPrefixes []string) (*AtimePolicy, error) {
 	parsed, err := ParseAtimeMode(mode)
 	if err != nil {
@@ -42,6 +45,8 @@ func NewAtimePolicy(mode string, threshold time.Duration, pathPrefixes []string)
 	}, nil
 }
 
+// ParseAtimeMode normalises a mode string (case- and whitespace-insensitive),
+// treating the empty string as AtimeModeOff and rejecting anything unknown.
 func ParseAtimeMode(s string) (AtimeMode, error) {
 	switch AtimeMode(strings.ToLower(strings.TrimSpace(s))) {
 	case AtimeModeOff, "":
@@ -70,6 +75,10 @@ func normalisePathPrefixes(in []string) []string {
 	out := make([]string, 0, len(in))
 	for _, p := range in {
 		p = strings.TrimSpace(p)
+		// Treat "/a/b" and "/a/b/" as the same subtree so a prefix also
+		// matches the directory itself. A bare "/" collapses to "" and is
+		// dropped, which correctly means "no restriction".
+		p = strings.TrimRight(p, "/")
 		if p != "" {
 			out = append(out, p)
 		}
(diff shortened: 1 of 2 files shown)


petedodd-pd:
Context and roadmap for this series: #9899. This is step 2a (the standalone policy type); the server handler + flags that consume it are in #9583 (kept in draft until this merges).

petedodd-pd:
Thanks for the review. Addressed in fa909618a:

- **`ShouldUpdate` relatime threshold (@coderabbitai, @gemini-code-assist):** now falls back to `DefaultRelatimeThreshold` when `RelatimeThreshold <= 0`, mirroring `NewAtimePolicy`'s normalization. This is defensive hardening of the exported method — no in-tree caller constructs a relatime policy with a zero threshold today (the only struct literal is mode-off), but a struct-literal/future caller shouldn't be able to silently degrade relatime to strict.
- **Trailing-slash path prefixes (@gemini-code-assist):** `normalisePathPrefixes` now trims trailing slashes, so `/a/b` and `/a/b/` name the same subtree and a prefix matches the directory itself; a bare `/` collapses to no restriction.
- Added regression tests for both, plus doc comments on the exported `NewAtimePolicy`/`ParseAtimeMode`/`ShouldUpdate`.

I skipped the CodeRabbit docstring-coverage 80% threshold deliberately — it's the bot's own config rather than a repo merge gate, and chasing it would bloat this small primitive. Added doc comments to the exported symbols, which is the idiomatic ask.

petedodd-pd:
@chrislusf gentle bump on this

chrislusf:
hi, the PR code is clean, but it storing atime inside the metadata turns every read into a metadata write.  

Access time is cache-eviction telemetry, not filesystem metadata. It can be managed elsewhere.

PR #9958, opened 2026-06-14
fix(ec): recover EC shards with the volume's own ratio, not the build default

chrislusf:
## Problem

`recoverOneRemoteEcShardInterval` (the read-time EC shard reconstruction path) built its Reed-Solomon decoder with the hardcoded build default — `reedsolomon.New(DataShardsCount, ParityShardsCount)` (10+4) — and counted shard sufficiency / iterated the shard set against the `DataShardsCount`/`TotalShardsCount` constants. `cachedLookupEcShardLocations` likewise gated on `DataShardsCount`.

For a **custom-ratio** volume (an enterprise feature, e.g. 9+3) this reconstructs with the *wrong matrix* and **corrupts the recovered bytes**, and the lookup can wrongly reject a degraded-but-recoverable custom-ratio read (e.g. a 9+3 volume with exactly its 9 data shards present, judged `< 10 required`).

## Fix

Use the volume's **own** EC ratio — `ecVolume.ECContext`, loaded from its `.vif` `EcShardConfig` at mount — for the Reed-Solomon encoder, the shard-iteration bound (`ECContext.Total()`), and the data-shard sufficiency checks (`ECContext.DataShards`), with a default-context fallback if `ECContext` is somehow nil.

In **OSS** the ratio is always 10+4, so this is a behavioral no-op. It brings the Go volume server in line with the **Rust** volume server, which already reconstructs with the volume's ratio (`ecv.data_shards`/`parity_shards`). When merged into the enterprise fork it fixes the custom-ratio read-recovery corruption (this is the data-plane half of the enterprise custom-ratio work; the EC-balance planner side remains a separate enterprise change).

## Test

`TestNewEcVolumeLoadsCustomRatio` pins that a 9+3 `.vif` loads `ECContext` as 9+3 — the ratio source the recover path now uses. `go test ./weed/storage/ ./weed/storage/erasure_coding/` is green (the existing 10+4 EC roundtrip tests confirm no regression).

## Summary by CodeRabbit

* **Tests**
  * Added coverage to verify custom erasure coding settings are loaded correctly when mounting an erasure-coded volume.

* **Bug Fixes**
  * Improved erasure coding shard recovery to respect the volume’s configured shard ratio, including reconstruction thresholds and validation.
  * Updated related shard-location caching and recovery diagnostics to align with the configured erasure coding layout.
  * Fixed concurrent recovery flag handling to ensure accurate detection of deleted/invalid shards.

### DIFF
--- a/weed/storage/erasure_coding/ec_volume_test.go
+++ b/weed/storage/erasure_coding/ec_volume_test.go
@@ -119,3 +119,39 @@ func TestNewEcVolumeDoesNotWriteStubVif(t *testing.T) {
 		t.Fatalf("expected default EC context, got %+v", ev.ECContext)
 	}
 }
+
+// TestNewEcVolumeLoadsCustomRatio pins that a volume's own EC ratio is loaded from
+// its .vif into ECContext, so the read/recover/decode paths reconstruct with the
+// matrix that produced the shards rather than the build default. (Custom ratios are
+// an enterprise feature; in OSS the .vif always records 10+4.)
+func TestNewEcVolumeLoadsCustomRatio(t *testing.T) {
+	dir := t.TempDir()
+	const vid = needle.VolumeId(125)
+	base := EcShardFileName("", dir, int(vid))
+
+	if err := os.WriteFile(base+".ecx", nil, 0o644); err != nil {
+		t.Fatalf("write .ecx: %v", err)
+	}
+	if err := volume_info.SaveVolumeInfo(base+".vif", &volume_server_pb.VolumeInfo{
+		Version: uint32(needle.Version3),
+		EcShardConfig: &volume_server_pb.EcShardConfig{
+			DataShards:   9,
+			ParityShards: 3,
+		},
+	}); err != nil {
+		t.Fatalf("save .vif: %v", err)
+	}
+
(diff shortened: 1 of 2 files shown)


chrislusf:
Re: the custom-ratio degraded-read regression (CodeRabbit nitpick) — acknowledged as a known coverage gap rather than added here, for a few reasons:

- `recoverOneRemoteEcShardInterval` and `cachedLookupEcShardLocations` are gRPC-bound (remote shard reads via `readRemoteEcShardInterval`, and a master `LookupEcVolume` RPC). A real degraded-read regression needs a multi-node EC cluster.
- OSS can't *natively* produce a custom-ratio (9+3) volume — the OSS encode path is always 10+4 — so such a test would have to hand-plant custom-ratio shards + `.vif` across multiple nodes. That's disproportionate for a change that is a behavioral no-op in OSS.
- The existing 10+4 EC roundtrip tests guard the default path against regression; `TestNewEcVolumeLoadsCustomRatio` pins the ratio source (`.vif` → `ECContext`) the recover path now consumes; and the custom-ratio reconstruction is exercised in the enterprise fork where 9+3 volumes exist (this PR is the data-plane half that flows there via the master→enterprise merge).

PR #9968, opened 2026-06-15
admin: add connected Mount Clients page and dashboard section

chrislusf:
## What

Adds a **Cluster → Mount Clients** page to the admin UI listing the FUSE/VFS mounts currently connected to the cluster, plus a **Mount Clients** section on the main dashboard showing the live count.

## Why

The filer already knows exactly which clients are subscribed to its metadata stream (FUSE mounts, S3, filer.sync, peer filers, …), but that registry only tracked `clientId -> epoch` internally and wasn't exposed anywhere. Operators had no way to see connected mounts from the admin UI.

## How

**Filer (data + gRPC)**
- Enrich the in-memory subscriber registry with client name / type / address / path-prefix / connected-time. It's populated in `addClient` and cleared in `deleteClient`, so it reflects *currently connected* clients only.
- New `ListMetadataSubscribers` filer gRPC, with an optional client-type filter. Returns the calling filer's connected subscribers (each filer only knows its own).

**Admin**
- `GetMountClients()` fans out to every discovered filer and aggregates, filtering to mount client types: `mount` (Go `weed mount`) and `sw-vfs` (Rust VFS). Unreachable filers are skipped.
- New page (route `/cluster/mount-clients`, handler, `mount_clients.templ`) with a sortable table (address, type, path prefix, client id, connected-since, filer) and CSV export.
- Sidebar entry under Cluster, and a Mount Clients section on the dashboard with the live count + a link to the page.

## Notes
- The subscribe hot path is unchanged; the registry write happens only on connect/disconnect (already-locked paths).
- `filer.proto` regenerated with the pinned toolchain (protoc v6.33.4, protoc-gen-go v1.36.6, protoc-gen-go-grpc v1.6.2); Java proto copy synced. templ regenerated with the pinned v0.3.1001.
- `go build ./...`, `go vet` (changed pkgs), and `weed/server` + `weed/admin/...` tests pass.

## Release Notes

* **New Features**
  * Added a “Mount Clients” card to the admin dashboard showing the total number of connected FUSE/VFS mount clients.
  * Added a new “Mount Clients” page under the Cluster section with a table of connected clients (address, type, path prefix, client ID, connected-since, and filer address) and a last-updated timestamp.
  * Added an **Export** button to download the visible mount clients table as `mount-clients.csv`.


chrislusf:
Addressed the review feedback in 7ee727869:

- **Parallelized the filer fan-out with a per-call timeout** — `GetMountClients` now queries all filers concurrently, each `ListMetadataSubscribers` call bounded by a 5s context timeout, so a slow or unreachable filer can no longer stall the admin dashboard.
- **Nil-guard on the subscriber registry** — `fs.subscribers` is defensively initialized before the first write, preventing a nil-map panic if a `FilerServer` is constructed outside `NewFilerServer` (e.g. in tests).
- **Robust CSV export** — the Mount Clients export now builds a `Blob` with proper quote-escaping instead of a `data:` URI, so paths containing `#`, `?`, `%` or commas export correctly.

chrislusf:
The inline suggestions here (parallelize the filer fan-out + per-call timeout, add `sync` import, nil-guard `fs.subscribers`, and the Blob-based CSV export) were already addressed in 7ee727869 — those threads are against the original commit. Current branch state:
- `GetMountClients` queries filers concurrently, each under a 5s context timeout.
- `fs.subscribers` is nil-guarded before first write.
- CSV export uses a `Blob` + `URL.createObjectURL` with quote-escaping.

Issue #9967, opened 2026-06-14
Grafana Dashboard feedback


m-sementsov:
Good afternoon Chris!

I've seen your latest commit upstream for a new dashboard.
I've also made one myself some while ago.
I do think there's something that could be used from mine and pushed into public one.
Do give criticism about it, I might update it a bit to your feedback.

[dashboard.json](https://github.com/user-attachments/files/28936442/dashboard.json)

And yes, I know, there're some system-metrics that are not emitted by Seaweedfs, but rather with basic Telegraf plugins.

My metrics are pulled via vmagent, like this

```
    - job_name: 'seaweedfs-infra'
      scheme: http
      static_configs:
      - targets:
        {% for host in groups['seaweedfs_storage_infra'] -%}
        - '{{ hostvars[host].inventory_hostname }}:9191' #seaweedfs-master
        - '{{ hostvars[host].inventory_hostname }}:9391'  #seaweedfs-s3
        {% if host in groups['seaweedfs_master'] -%}
        - '{{ hostvars[host].inventory_hostname }}:9291'  #seaweedfs-filer
        {% endif -%}
        {% for port in range(9091, 9099) -%}  #seaweedfs-volume-servers
        - '{{ hostvars[host].inventory_hostname }}:{{ port }}'
        {% endfor %}
        {%- endfor %}
```

You may close the issue if don't feel like it 😄

chrislusf:
Thanks for sharing this! The per-bucket S3 breakdowns were the obvious win - the dashboard only sliced by type, but the metrics already carry a bucket label, so I pulled those in: per-bucket request rate, response codes, and p95 latency, plus a bucket filter and a volume slot utilization panel. PR #9969.

Left out the Telegraf/node_exporter panels (CPU, disk I/O, network) since the bundled dashboard assumes only SeaweedFS scrape targets - those are better paired with node_exporter on your end.

PR #9969, opened 2026-06-15
metrics: add per-bucket S3 panels and volume slot utilization to dashboard

chrislusf:
The S3 Gateway and S3 Buckets rows sliced S3 requests only by `type`, never by bucket — even though `SeaweedFS_s3_request_total` (`{type,code,bucket}`) and both the `request_seconds` / `time_to_first_byte_millisecond` histograms (`{type,bucket}`) already carry a `bucket` label. Per-bucket request rate, error codes, and latency are some of the most-asked-for operational views, so surface them.

Additive only (79 → 84 panels), no existing panel touched:

**New `Bucket` template variable** — `label_values(SeaweedFS_s3_request_total, bucket)`, multi-select, defaults to All. Mirrors the existing `cluster` variable.

**S3 Buckets row** — four per-bucket panels:
- S3 Requests/sec by Bucket
- S3 Response Codes by Bucket (by status code)
- S3 Request Duration p95 by Bucket
- S3 Time-to-First-Byte p95 by Bucket

**Volume Servers row** — Volume Slots Utilization: `volumes / max_volumes * 100` per instance (0-100%).

Same uid/title, so it upgrades in place. Scoped to the bundled `other/metrics/grafana_seaweedfs.json` source of truth.

From dashboard feedback in #9967. The Telegraf/node_exporter panels suggested there (CPU, disk I/O, network) are intentionally left out — they depend on external exporters and don't belong in a dashboard that assumes only SeaweedFS scrape targets.

## Summary by CodeRabbit

* **New Features**
  * Added **Volume Slots Utilization** timeseries panel to visualize per-instance volume slot usage as a percentage.
  * Added four new **S3 Buckets** per-bucket panels: **requests/sec**, **response codes**, **request duration (p95)**, and **time-to-first-byte (p95)**.
  * Added a multi-select **Bucket** template variable (with **All** option) to filter S3 metrics by bucket.

### DIFF
--- a/other/metrics/grafana_seaweedfs.json
+++ b/other/metrics/grafana_seaweedfs.json
@@ -3688,6 +3688,104 @@
             }
           ],
           "pluginVersion": "10.3.1"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "title": "Volume Slots Utilization",
+          "description": "Used volume slots as a percentage of max volumes, per volume server.",
+          "type": "timeseries",
+          "id": 205,
+          "gridPos": {
+            "h": 8,
+            "w": 12,
+            "x": 0,
+            "y": 94
+          },
+          "fieldConfig": {
+            "defaults": {
+              "color": {
+                "mode": "palette-classic"
+              },
+              "custom": {
+                "axisBorderShow": false,
+                "axisCenteredZero": false,
+                "axisColorMode": "text",
+                "axisLabel": "",
+                "axisPlacement": "auto",
+                "barAlignment": 0,
+                "drawStyle": "line",
+                "fillOpacity": 10,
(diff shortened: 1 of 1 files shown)


chrislusf:
Addressed both findings in cc42b95e4:

- **Bucket variable scoping** — switched the source to `label_values(SeaweedFS_s3_bucket_size_bytes{cluster=~"$cluster"}, bucket)`. That gauge is emitted by the periodic collector for every bucket (empty ones are set to 0 in `collectAndUpdateBucketSizeMetrics`), so the dropdown now lists all buckets in the selected cluster, including idle ones, rather than only those that received requests. This also resolves the cluster-scoping point.
- **Divide-by-zero** — wrapped the slot-utilization denominator in `clamp_min(..., 1)` so the panel can't return +Inf/NaN when `max_volumes` is briefly zero or absent.

PR #9970, opened 2026-06-15
s3: honor canned ACL header in PutObjectAcl regardless of request body

chrislusf:
## Problem

`TestVersionedObjectAcl` (S3 versioning suite) started failing: a canned `PutObjectAcl` (`ACL: private`) on a versioned object returns **400 InvalidRequest** —

```
PutObjectAcl … StatusCode: 400 … api error InvalidRequest: Invalid Request
```

## Root cause

`ExtractAcl` (`weed/s3api/s3api_acl_helper.go`) chose between an XML `AccessControlPolicy` body and ACL **headers** by checking whether the request had a body:

```go
if r.Body != nil && r.Body != http.NoBody { /* XML-parse body */ } else { /* parse headers */ }
```

A canned ACL is specified entirely via the `x-amz-acl` header and carries no ACP body. But:
- On the server, `r.Body` is never `http.NoBody`, and SigV4 auth re-buffers the body into a `NopCloser` — so the body branch is effectively always taken.
- Newer `aws-sdk-go-v2` (default request checksums, bumped in #9871) attaches a body even to header-only requests — an empty payload, or one wrapped in **aws-chunked** framing with a checksum trailer.
- `PutObjectAcl`/`PutBucketAcl` never call `getRequestDataReader`, so the body is **not de-chunked**.

So `ExtractAcl` tried to XML-parse an empty / aws-chunked-framed body and returned `ErrInvalidRequest`.

## Fix

Honor the canned ACL header (`x-amz-acl`) before inspecting the body, and drain the body on all paths. When no canned header is present, the XML `AccessControlPolicy` body path is unchanged.

## Tests

- `TestExtractAclCannedAclWithNonEmptyBody` — canned `private` + a non-XML (aws-chunked-framed) body now returns `ErrNone` with the expected grant. **Fails on the old code** (returns `ErrInvalidRequest`), passes with the fix.
- `TestExtractAclNonXmlBodyWithoutCannedHeader` — guards that the body is still parsed (and a non-XML body rejected) when no canned header is present.
- Full `go test ./weed/s3api/` passes.

This is the regression behind the S3 Versioning integration test failures seen on recent PRs/CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 ACL handling to correctly prioritize canned-ACL headers over request body content, ensuring proper compliance with S3 API specifications.

* **Tests**
  * Added comprehensive test coverage for S3 ACL extraction with canned-ACL headers and various request body formats to validate correct behavior.

### DIFF
--- a/weed/s3api/s3api_acl_helper.go
+++ b/weed/s3api/s3api_acl_helper.go
@@ -22,9 +22,23 @@ type AccountManager interface {
 
 // ExtractAcl extracts the acl from the request body, or from the header if request body is empty
 func ExtractAcl(r *http.Request, accountManager AccountManager, ownership, bucketOwnerId, ownerId, accountId string) (grants []*s3.Grant, errCode s3err.ErrorCode) {
-	if r.Body != nil && r.Body != http.NoBody {
+	if r.Body != nil {
 		defer util_http.CloseRequest(r)
+	}
 
+	// A canned ACL is specified via the x-amz-acl header and carries no ACP
+	// body. Clients such as the AWS SDK (with default request checksums) may
+	// still attach a body to the request — an empty payload, or one wrapped in
+	// aws-chunked framing with a checksum trailer. PutObjectAcl/PutBucketAcl do
+	// not de-chunk the body, so honor the canned header first instead of
+	// mis-parsing that body as an AccessControlPolicy and rejecting it with
+	// InvalidRequest.
+	if r.Header.Get(s3_constants.AmzCannedAcl) != "" {
+		_, grants, errCode = ParseAndValidateAclHeadersOrElseDefault(r, accountManager, ownership, bucketOwnerId, accountId, true)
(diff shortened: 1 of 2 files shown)


chrislusf:
Closing — my root-cause analysis here was wrong. I verified with a RoundTripper + httptest that the AWS SDK sends the canned PutObjectAcl with an **empty body** (Content-Length 0), which arrives server-side as `http.NoBody`, so ExtractAcl already takes the canned-header path — the request body was never the problem.

The actual 400 comes from downstream account resolution: server logs show `invalid canonical grantee! account id[some_admin_user] is not exists`. Root cause is #9962 synthesizing a distinct account for account-less identities without registering it in the account lookup. The correct fix is #9971.

The canned-header-priority change here was harmless but unnecessary, so I'm closing rather than merging a PR with a misleading rationale.

PR #9971, opened 2026-06-15
s3: register account-less identities' synthesized account so ACL/owner ids resolve

chrislusf:
## Problem

`TestVersionedObjectAcl` regressed: a canned `PutObjectAcl` (`ACL: private`) returns **400 InvalidRequest**. Server logs:

```
bucket_metadata.go:137  owner[id=some_admin_user] is invalid, bucket: test-versioned-acl
s3api_acl_helper.go:283  invalid canonical grantee! account id[some_admin_user] is not exists
```

## Root cause (trigger: #9962)

#9962 ("give account-less identities a distinct owner instead of admin") added `accountForUnscopedIdentity`, so an identity configured without an `account` block now owns resources under a distinct account id derived from its name (e.g. `some_admin_user`) instead of collapsing into the shared admin account.

But that synthesized account is assigned to the identity **without being registered in the `accounts` (id → account) map** that `GetAccountNameById` consults. So:

- `ValidateAndTransferGrants` rejects a canonical-user grant to that id (`account id … is not exists` → `ErrInvalidRequest`). A canned `private` ACL grants to the caller's *own* account, so any account-less identity's `PutObjectAcl` fails.
- bucket/object owner display is dropped (`owner … is invalid`).

This is unrelated to the request body / SDK checksum behavior — the empty body arrives as `http.NoBody` and the canned-ACL header path is taken correctly; the failure is downstream in account resolution.

## Fix

Register the synthesized account in the `accounts` map at config load (both `ReplaceS3ApiConfiguration` and `MergeS3ApiConfiguration`), so its id resolves to a display name for ACL grantee validation and owner display.

## Tests

- `TestUnscopedIdentityAccountResolvesByName` — after loading a config with an account-less identity, `GetAccountNameById(id)` returns its display name. **Fails on master** (empty), passes with the fix.
- #9962's existing tests (`TestAccountForUnscopedIdentity`, `TestUnscopedIdentitiesGetDistinctAccounts`, `TestCheckAccessByOwnershipDeniesNonOwner`) still pass.
- Full `go test ./weed/s3api/` passes.

Fixes the S3 Versioning suite failure (`TestVersionedObjectAcl`) seen on recent PRs/CI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed account-name lookups for identities without explicitly configured accounts by ensuring synthesized accounts are registered during both full configuration replacement and incremental merges.
  * Ensured that when an explicit account is configured for an id that could otherwise be synthesized, the configured account’s display details are preserved and reused.

* **Tests**
  * Added regression coverage for unscoped identity name resolution by id and for reuse of configured accounts instead of synthesized ones.

### DIFF
--- a/weed/s3api/auth_account_collapse_test.go
+++ b/weed/s3api/auth_account_collapse_test.go
@@ -96,3 +96,36 @@ func TestUnscopedIdentityAccountResolvesByName(t *testing.T) {
 	assert.Equal(t, "alice", iam.GetAccountNameById("alice"),
 		"account-less identity id must resolve to a display name for ACL/owner validation")
 }
+
+// When an account is explicitly configured with the same id an account-less
+// identity would synthesize, the identity must reuse that configured account so
+// its custom display name/email are preserved.
+func TestUnscopedIdentityReusesConfiguredAccount(t *testing.T) {
+	resetMemoryStore()
+
+	config := `{
+  "accounts": [
+    {"id": "alice", "displayName": "Alice Smith", "emailAddress": "alice@example.com"}
+  ],
+  "identities": [
+    {"name": "alice", "credentials": [{"accessKey": "alice_ak", "secretKey": "alice_sk"}], "actions": ["Read"]}
+  ]
+}`
+	tmp, err := os.CreateTemp("", "s3-config-*.json")
+	require.NoError(t, err)
+	defer os.Remove(tmp.Name())
+	_, err = tmp.WriteString(config)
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed in 8c6e4f227: when an account is explicitly configured with the same id an account-less identity would synthesize, the identity now reuses that configured account (preserving its display name/email) instead of the synthesized one. Added `TestUnscopedIdentityReusesConfiguredAccount` to cover it.

PR #9982, opened 2026-06-16
fix(shell): return fs.verify topology errors

7y-9:

`fs.verify` ignored topology collection failures. When `collectVolumeIds()` failed, the command returned `parseErr` instead of the actual topology error, so master `VolumeList` failures could be hidden from callers.

Return the real error from `collectVolumeIds()`.

A regression test was added with a fake master/filer setup where `VolumeList` returns an error, then verifies `fs.verify` propagates that error.

# How is the PR tested?

```bash
go test ./weed/shell -run TestFsVerifyPropagatesCollectVolumeIdsError -count=1
```

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed fs.verify to correctly report volume collection errors instead of masking them with earlier parse errors, improving error diagnostics for users troubleshooting initialization failures.


chrislusf:
remove the tests

PR #9998, opened 2026-06-17
mq(kafka): don't drop an existing topic when auto-create races

chrislusf:
## Problem

The Kafka E2E test `TestOffsetManagement/ConsumerGroupResumption` flakes at the Phase-1 produce:

```
write messages: [3] Unknown Topic Or Partition
```

It fails immediately (0.00s, i.e. at the Metadata lookup, before any produce) for a topic that was created ~27s earlier and whose config reads back fine from the filer. The preceding subtest spends ~26s in consumer-group churn, so the broker/filer is under load at that moment.

## Root cause

`SeaweedMQHandler.TopicExists` can return a **false-negative** for a topic that genuinely exists: the broker's `TopicExists` returns `Exists:false` (no error) on a transient filer read, and the gateway's `TopicExists` returns `false` on any gRPC blip — both indistinguishable from "doesn't exist".

When that happens, the metadata/produce handler auto-creates the topic, `checkTopicInFiler` finds it, `CreateTopicWithSchemas` returns `"already exists"`, and the handler treated that error as a failure and **dropped the topic from the Metadata response**. kafka-go then sees no partition for the topic and surfaces `UNKNOWN_TOPIC_OR_PARTITION`.

So a topic that definitely exists gets reported as missing, purely because auto-create's "already exists" was treated as fatal.

## Fix

- Return a sentinel `integration.ErrTopicAlreadyExists` from `CreateTopicWithSchemas` / `CreateTopicWithRecordType` (and the test mock).
- Add `Handler.ensureTopicExists`, which treats `ErrTopicAlreadyExists` as confirmation the topic exists, and folds in the duplicated `TopicExists → invalidate → recheck → create` logic shared by every metadata handler (v0–v8) and both produce paths (net −75 lines).
- v2+ produce now auto-creates too (was deliberately not), consistent with the `auto.create.topics.enable=true` behavior the rest of the gateway already simulates.
- Regression test `topic_autocreate_test.go` covers the false-negative-then-already-exists case, a genuine create failure, and the present-topic path.

`go build ./...`, `go vet`, and the full `weed/mq/kafka/...` suite pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Kafka topic auto-creation to handle concurrent “topic already exists” outcomes without failing the request.
  * Centralized topic existence checks for Metadata and Produce so cache and auto-create decisions are consistent, including avoiding stale “missing topic” results after creation.
  * Updated topic-creation errors to use a detectable “already exists” sentinel for more reliable error handling.

* **Tests**
  * Added new unit tests covering auto-create success on “already exists,” genuine failure behavior, and the no-create-needed path when topics are already present.

### DIFF
--- a/weed/mq/kafka/gateway/test_mock_handler.go
+++ b/weed/mq/kafka/gateway/test_mock_handler.go
@@ -62,7 +62,7 @@ func (m *mockSeaweedMQHandler) CreateTopic(topic string, partitions int32) error
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	if _, exists := m.topics[topic]; exists {
-		return fmt.Errorf("topic already exists")
+		return fmt.Errorf("%s: %w", topic, integration.ErrTopicAlreadyExists)
 	}
 	m.topics[topic] = &integration.KafkaTopicInfo{
 		Name:       topic,
@@ -75,7 +75,7 @@ func (m *mockSeaweedMQHandler) CreateTopicWithSchemas(name string, partitions in
 	m.mu.Lock()
 	defer m.mu.Unlock()
 	if _, exists := m.topics[name]; exists {
-		return fmt.Errorf("topic already exists")
+		return fmt.Errorf("%s: %w", name, integration.ErrTopicAlreadyExists)
 	}
 	m.topics[name] = &integration.KafkaTopicInfo{
 		Name:       name,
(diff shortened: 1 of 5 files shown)


chrislusf:
Addressed both review points in the force-push:
- `ensureTopicExists` now invalidates the existence cache when create reports `ErrTopicAlreadyExists`, so the stale negative from the recheck isn't left behind for direct `TopicExists` callers.
- Metadata v5-v8 no longer advertises a system topic when auto-create genuinely fails; `ensureTopicExists` already tolerates the create race, so a false result means real absence.

PR #10024, opened 2026-06-20
mount: auto-enable Linux FUSE passthrough for read-only files

chrislusf:
## What

On a **read-only** open, when the kernel supports `FUSE_PASSTHROUGH` (Linux ≥ 6.9, privileged mount), `weed mount` materializes the file into a local backing file and registers it with the kernel via `RegisterBackingFd`, so the kernel serves reads and `mmap` **directly from the backing file**, bypassing the mount process for the data path.

**No opt-in flag.** Kernel/privilege *support* is auto-detected: the first `RegisterBackingFd` call probes it and latches off on failure, so unsupported or unprivileged mounts transparently use the normal read path. The only knob is a size cap, `-fusePassthroughMaxMB` (default 256; `0` disables).

## How

- On a read-only `Open`, `tryEnablePassthrough` materializes the file into the cache dir (reusing the normal read path, so bytes are identical), registers the fd, and replies with `BackingID` + `FOPEN_PASSTHROUGH` (clearing `FOPEN_KEEP_CACHE`).
- The backing is shared per inode (one handle per inode) and unregistered when the last open is released.
- `MaxStackDepth=1` is set (when the cap is > 0) so the kernel accepts backing-file registration.
- Linux-only; a no-op stub on other platforms. The data path is unchanged when disabled/unsupported.

## Tradeoff (please review)

A passthrough handle reads a **point-in-time snapshot** taken at open — concurrent writes (this mount or another) are not reflected while the inode stays open. This is the right behavior for immutable / read-mostly data (the common object-storage pattern), and it's why it's gated to read-only opens and a size cap. Set `-fusePassthroughMaxMB=0` to disable if strict read-while-write coherence is required.

## Testing

Mechanism verified on Linux 6.12 (on the enterprise build of this same code):
- `FUSE_INIT` negotiates `PASSTHROUGH` + `MaxStack 1`; `RegisterBackingFd` succeeds; the `OPEN` reply carries `backing=N PASSTHROUGH`.
- md5 through the passthrough mount matches the reference (data integrity).
- 3 overlapping readers of one inode share a single backing (one `BACKING_OPEN` → one `BACKING_CLOSE` at last release); open/close balanced, no fd leak.
- Benchmark (256 MB): warm random `mmap` ≈ memory speed (~1.3 µs/fault) vs ~140 µs/fault on the normal FUSE read path.

This PR builds clean for darwin + linux and passes the existing mount tests.

https://claude.ai/code/session_011sdkuCoRghMokGdHJAGLZL

## Release Notes

* **New Features**
  * Added Linux FUSE passthrough for **read-only** opens to improve performance on compatible systems.
  * Added `-fusePassthroughMaxMB` flag (default **256 MB**) to limit passthrough to files up to the specified size (set to `0` to disable).
  * Automatically enables when supported and applies snapshot behavior while a file is open; falls back gracefully when unsupported and avoids repeated failed attempts.

### DIFF
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -74,6 +74,7 @@ type MountOptions struct {
 	cacheSymlink            *bool
 	fuseMaxBackground       *int
 	fuseCongestionThreshold *int
+	fusePassthroughMaxMB    *int64
 
 	// macOS-specific FUSE options
 	novncache *bool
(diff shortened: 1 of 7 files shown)


chrislusf:
Thanks — addressed both in 8a2608f:

1. **Teardown lock ordering:** moved `fh.teardownPassthrough()` ahead of the `fhLockTable` lock in `ReleaseHandle`. (For the record: the materialize path calls `readDataByFileHandle`, which takes the *dirtyPages* read lock — not `fhLockTable` — so the exact `passthroughMu`→`fhLockTable` inversion described isn't actually reachable. But `teardownPassthrough` is fully synchronized by `passthroughMu` and needs no other lock, so moving it out of the `fhLockTable` critical section is cleaner and removes any future coupling.)

2. **Buffer reuse:** `materializeFile` now reuses `wfs.copyBufferPool` instead of allocating a fresh buffer per materialization.

PR #10035, opened 2026-06-22
Volume balancing by step

m-sementsov:

In case of large disbalance between nodes - balancing may take a long time.
During this time the topology may change due to volume.growth
It might be beneficial to have -volumesPerStep flag for volume.balance to make the command only execute 5-10 volume movements.

Add a volumesPerStep flag.

**There is a catch that volume.fix.replication volumesPerStep-flag keeps the command running, but my implementation stops/breaks the execution entirely.
It might be more reasonable to name the flag "volumesPerExec"
Or rework it, to keep working while rescanning topo after each step**

# How is the PR tested?

I've deployed my version of code and it works. The command executes 2 balances and finishes.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Updated `volume.balance` with a `-volumesPerExec` flag to cap how many volume moves can succeed per command execution.
* **Bug Fixes**
  * Added validation to reject negative `-volumesPerExec` values and enforce the move limit consistently across balancing loops.
* **Tests**
  * Added a regression test ensuring balancing stops after the configured move count and that the involved node ends with the expected volume state.

### DIFF
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -56,18 +56,20 @@ func (c *commandVolumeBalance) Help() string {
 	    * Use exact match: volume.balance -collection="^mybucket$"
 	    * Match multiple buckets: volume.balance -collection="bucket.*"
 	    * Match all user collections: volume.balance -collection="user-.*"
-	
-	The -volumesPerStep parameter limits the maximum number of volume moves in one run.
-	
+
+	The -volumesPerStep parameter limits the maximum number of volume moves in one command execution.
+	If unset - the command will try to balance all volumes at once.
+	It might be beneficial to set it if your cluster has lots of blocks growing and topology changes faster than balancing can occur.
+
 	Algorithm:
-	
+
 	For each type of volume server (different max volume count limit){
 		for each collection {
 			balanceWritableVolumes()
 			balanceReadOnlyVolumes()
 		}
 	}
-	
+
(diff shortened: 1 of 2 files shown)


chrislusf:
Pushed a few fixes to the branch (maintainer edit):

- Restored the build: `balanceSelectedVolume` became a method, but `command_volume_balance_test.go` still called the old free function, so the test package failed to compile (the Go Vet / Test job failures). Updated the call site.
- Added `TestBalanceVolumesPerStep` covering the cap (one move with `-volumesPerStep=1`).
- Rejected negative `-volumesPerStep` with a fail-fast error, per the CodeRabbit comment.
- Fixed help wording ("blocks" -> "volumes").

m-sementsov:
Do you think that the parameter should be volumesPerExec?
Currently this parameter behaves differently from volumesPerStep of volume.fix.replication
There may be confusion about their purpose despite being similarly named

chrislusf:
right. it is confusing. And this is limited due to the algorithm. The admin worker is more advanced and I would recommend migrate to that.

m-sementsov:
I've redone the parameter name (to volumesPerExec) so that it is distinct from volume.fix.replication.

I haven't tried admin+workers setup yet, so I'm not that familiar with their drawbacks and considerations.

IMO there should be an option to do this via shell in case of problems with workers.

PR #10043, opened 2026-06-22
fix(shell): exclude failed EC shard copies from rebuild recoverability gate

qzhello:
## Problem
In apply mode, `prepareDataToRecover` incremented the remote-shard counter
before issuing the copy RPC, so a failed `VolumeEcShardsCopy` was still
counted toward the `DataShardsCount` recoverability gate. The gate could pass
with fewer real shards than required, deferring the failure to the deeper
reconstruct step and printing an inflated count in the "not enough shards" error.

## Fix
Count a remote shard only after a successful copy (apply mode), or when
planning (dry-run). Rename `wouldCopy` → `recoverableRemoteShards` for clarity.

## Test
Added `TestPrepareDataToRecover_ApplyCopyFailureDoesNotCountAsRecoverable`,
which fails before the fix and passes after: 9 local + 1 remote shard whose
copy fails must leave 9 usable shards and fail the gate with "9 shards are not
enough", with the failed copy not recorded as deletable.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery validation to count only successfully-copied remote shards when determining whether recovery is possible.
  * Refined apply-mode behavior so failed shard copy operations no longer make recovery appear feasible.
* **Tests**
  * Added coverage for scenarios where copying fails, ensuring such failures don’t contribute to recoverability accounting.


chrislusf:
Thanks — the gate fix is correct. I pushed a small follow-up so a copy that fails *after* the gate already counted earlier successful copies doesn't strand those copied working shards on the rebuilder.

Going forward, EC maintenance is better driven through the admin worker than the interactive `ec.rebuild` shell command. EC encode and ec_balance already run there as worker tasks with detection, scheduling, concurrency limits, and progress/retry; shard recovery is still shell-only. Recommendation is to run rebuilds through the admin worker, and ideally this recoverability/copy/cleanup logic eventually lives in a worker task rather than being extended in the shell command.

PR #10061, opened 2026-06-23
fix(storage): add nil check in MountVolume and remove dead code in vo…

sshhan:
The user is showing me the SeaweedFS PR template and wants me to fill it out. Let me fill it in based on the changes we've made.

fix(storage): add nil check in MountVolume and remove dead code in volume_write

1. `MountVolume` in store.go calls `findVolume()` without a nil check,
   then dereferences `v.diskId`. If `findVolume` returns nil (e.g. the
   volume was destroyed between `LoadVolume` and `findVolume`), this panics.

2. `deleteNeedle2` in volume_write.go has a hardcoded `fsync = false`,
   making the async-append else branch permanently dead code.

3. `writeNeedle2` in volume_write.go has a redundant `else { ... }` block
   — the `if` branch already returns unconditionally.

- store.go: Add a nil check after `findVolume()`, returning a descriptive
  error instead of panicking.

- volume_write.go: Simplify `deleteNeedle2` to directly call `syncDelete`,
  removing the dead `fsync` variable and unreachable else branch.

- volume_write.go: Flatten `writeNeedle2` by removing the unnecessary
  `else` wrapper, reducing indentation (Go idiom: avoid else when if
  returns).

All function signatures unchanged. Zero caller impact.

# How is the PR tested?

- Lint passes with zero errors on both files.
- Existing unit tests (`volume_vacuum_test`, `volume_vacuum_crash_safe_test`,
  `volume_read_test`) continue to call `deleteNeedle2` with the same
  signature — all test call sites are unaffected.
- The nil-check path in `MountVolume` is defensive; no known reproducer
  for the race condition, but the check cannot regress normal behaviour.

# Checks

# Checks for AI generated PRs

## Release Notes

* **Bug Fixes**
  * Fixed a potential crash in volume mounting by adding validation to ensure volumes are properly accessible after loading.

* **Refactor**
  * Streamlined internal storage write and delete operations for improved efficiency.

### DIFF
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -734,6 +734,9 @@ func (s *Store) MountVolume(i needle.VolumeId) error {
 		if found := location.LoadVolume(uint32(diskId), i, s.NeedleMapKind); found == true {
 			glog.V(0).Infof("mount volume %d", i)
 			v := s.findVolume(i)
+			if v == nil {
+				return fmt.Errorf("volume %d not found after LoadVolume", i)
+			}
 			v.diskId = uint32(diskId) // Set disk ID when mounting
 			s.NewVolumesChan <- &master_pb.VolumeShortInformationMessage{
 				Id:               uint32(v.Id),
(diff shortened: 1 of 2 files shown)


chrislusf:
> The user is showing me the SeaweedFS PR template and wants me to fill it out. Let me fill it in based on the changes we've made.

Please review your PR first.

PR #10062, opened 2026-06-23
worker: don't leak task goroutines on forced shutdown

chrislusf:
## Problem

`Worker.Stop()` drains in-flight tasks for up to 30s, then sends `ActionStop` and the manager loop exits. A task still running past that deadline keeps reporting through the unbuffered `w.cmds` channel after the loop is gone:

- `completeTask()` -> `getAdmin()` (send + response read)
- the `ActionIncTaskComplete` / `ActionIncTaskFail` send
- the deferred `removeTask()` -> `ActionRemoveTask` send + `getTaskLoad()`

With no receiver left, the send blocks, and for the getters the `<-respCh` read blocks too. The task goroutine leaks until process exit.

Buffering `cmds` does not fix this: `getAdmin()` and `getTaskLoad()` need a *response* from the loop, so a buffered send still leaves the response read blocked forever, and `completeTask()` is on the completion path.

## Fix

Close a `done` channel when the manager loop exits, and route the task-goroutine interactions through a small `dispatch()` helper that selects on `done`. Once the loop is gone these abort and return zero values instead of blocking:

- `dispatch()` returns false; `getAdmin()` returns nil; `getTaskLoad()` / `removeTask()` return 0.

Because `getAdmin()` can now return nil during shutdown, its double-call sites (`completeTask`, `sendHeartbeat`, `requestTasks`, the connection monitor, and the two progress-report sites in `executeTask`) are collapsed to a single nil-checked call so a forced shutdown can't turn into a nil deref.

The actor model and the unbuffered channel are kept as-is; only the post-shutdown abort path is added.

## Test

`TestManagerLoopSendersAbortAfterStop` stops a worker and asserts every interaction a lingering task goroutine has with the manager loop returns promptly (a 5s guard catches the leak) and yields zero values. Passes under `-race`. Fails on the previous code, where the calls block.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved worker shutdown reliability by ensuring command senders abort safely once shutdown begins, preventing goroutine deadlocks.
  * Made admin/task reporting and task completion/progress handling resilient when admin is unavailable, avoiding nil-related failures during shutdown.

* **Tests**
  * Added a concurrency regression test that confirms multiple worker interactions return promptly after `Stop()`, and verifies expected post-shutdown return values.

### DIFF
--- a/weed/worker/worker.go
+++ b/weed/worker/worker.go
@@ -26,6 +26,7 @@ type Worker struct {
 	config         *types.WorkerConfig
 	registry       *tasks.TaskRegistry
 	cmds           chan workerCommand
+	done           chan struct{} // closed when managerLoop exits, so senders stop blocking
 	state          *workerState
 	taskLogHandler *tasks.TaskLogHandler
 }
@@ -193,6 +194,7 @@ func NewWorker(config *types.WorkerConfig) (*Worker, error) {
 		registry:       registry,
 		taskLogHandler: taskLogHandler,
 		cmds:           make(chan workerCommand),
+		done:           make(chan struct{}),
 	}
 
 	glog.V(1).Infof("Worker created with %d registered task types", len(registry.GetAll()))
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review feedback in 236618e:

- **Stop() ActionStop send (CodeRabbit):** guarded with a select on `w.done`, so a second `Stop()` (e.g. an admin-shutdown timer racing an explicit Stop) aborts instead of blocking on a send with no receiver. Stop is now idempotent.
- **setTask / handleTaskCancellation (Gemini):** routed through `dispatch` so they abort after the loop exits; setTask returns a "worker is stopping" error.
- **Test:** added the `removeTask` zero-value assertion plus coverage for setTask, handleTaskCancellation, and a repeat Stop().

Left the startup-only getters/setters (getStopChan, getHbTick, etc.) as raw sends - they only run during Start, before any stop, so they are not reachable after the loop exits.

PR #10068, opened 2026-06-23
iceberg: support table rename

chrislusf:
Add the Iceberg REST catalog table-rename endpoint (POST /v1/tables/rename and the /v1/{prefix} variant). Rename is a catalog-only pointer move within the same table bucket: the destination entry inherits the source's MetadataLocation, so metadata.json and data files stay put. The destination is written before the source is deleted so a mid-rename failure never loses the table.

## Summary by CodeRabbit

* **New Features**
  * Added authenticated Iceberg REST table rename endpoints (`POST /v1/tables/rename`, `POST /v1/{prefix}/tables/rename`); validates input and returns `204` on success.
  * Added S3 Tables catalog-only rename within the same bucket, returning the destination `TableARN` while preserving the original `MetadataLocation`.
* **Bug Fixes**
  * `GetTable` now returns `404 NoSuchTable` when table metadata attributes are missing.
  * `DeleteTable` now purges derived table data when the metadata location maps elsewhere in the same bucket.
* **Tests**
  * Added REST rename validation tests (expects `400` for malformed/incomplete requests).
  * Added end-to-end S3 Tables rename tests covering success and key failure scenarios.

### DIFF
--- a/weed/s3api/iceberg/handlers_table.go
+++ b/weed/s3api/iceberg/handlers_table.go
@@ -531,6 +531,63 @@ func (s *Server) handleDropTable(w http.ResponseWriter, r *http.Request) {
 	w.WriteHeader(http.StatusNoContent)
 }
 
+// handleRenameTable moves a table's catalog pointer to a new namespace/name.
+func (s *Server) handleRenameTable(w http.ResponseWriter, r *http.Request) {
+	var req RenameTableRequest
+	if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+		writeError(w, http.StatusBadRequest, "BadRequestException", "Invalid request body")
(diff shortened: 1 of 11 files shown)


chrislusf:
Known limitation: dropping a renamed table removes the catalog entry at the new name but leaves the original data directory orphaned, since DropTable deletes by catalog path rather than the table's actual location. A location-aware drop is a follow-up.

chrislusf:
Resolved the drop-leak noted above: DropTable now purges the table's data at its actual location (derived from MetadataLocation), so dropping a renamed table no longer orphans the original data directory. Verified end-to-end (create, rename, drop -> source data dir is gone); normal drops are unchanged.

PR #10070, opened 2026-06-23
s3: keep a file promoted to a directory retrievable as an object

### LINKED ISSUE #10063: Inconsistent behavior when a path transitions from file to directory: filer vs S3 gateway
**Environment**

SeaweedFS version: 4.35 (commit eb939761665ca0ecead5026b02663531ea9fbf9e)
Filer store: Cassandra (cassandra2)
Access tested via: Filer HTTP API (port 8888) and S3 gateway (port 8333)

**Summary**
When a path that exists as a file is followed by a write to a child path under it (forcing it to become a directory), the original file behaves differently depending on the access method. Via the filer HTTP API the original file is immediately inaccessible, but via the S3 gateway the old object remains retrievable at the parent path — until the parent "directory" is deleted, after which the old data is permanently unretrievable.
Steps to Reproduce
1. Create test as a file:
curl -F file=@abc1 "http://<filer>:8888/foldernew/test"
→ {"name":"abc1","size":75}
At this point foldernew/test is a 75-byte file (confirmed in Filer UI — shows as a file with size).
2. Write a child object under test, forcing it to become a directory:
curl -F file=@abc1 "http://<filer>:8888/foldernew/test/abc1"
→ {"name":"abc1","size":75}
After this, foldernew/test becomes a directory (confirmed in Filer UI — now shows as a folder icon, no size).

**Observed Behavior**
**Filer HTTP API:**

After step 2, foldernew/test is a directory.
The original file content written in step 1 is no longer retrievable via the filer path foldernew/test.

**S3 gateway (same underlying operations):**

After step 2, GET foldernew/test still returns the original file content from step 1, even though test is also functioning as a directory containing abc1.
If the foldernew/test "directory" is then deleted, the original file data becomes permanently unretrievable.

**Questions / Expected Behavior**

Is a path simultaneously serving as both a file (retrievable object at foldernew/test) and a directory (containing foldernew/test/abc1) intended behavior on the S3 gateway, or should the file-to-directory transition be rejected / handled consistently with the filer?
The filer API and S3 gateway diverge on the same sequence of operations — the filer drops access to the old file immediately, while S3 retains it until directory deletion. Is this divergence expected?
When the directory is deleted and the shadowed file becomes unretrievable, is the underlying data chunk orphaned (reclaimable by vacuum/fsck), or is it leaked?

chrislusf:
## Problem

Writing a child key under an existing object (`PUT foldernew/test` then `PUT foldernew/test/abc1`) promotes the file to a directory. The promoted entry keeps its chunks/content, but the S3 gateway handled it inconsistently:

- full `GET` served the object, but ranged `GET` and `HEAD` returned 404
- `LIST` hid it whenever its `Mime` was empty
- the empty-folder cleaner could later delete it once its children were gone, permanently losing the data

The root cause is that "a directory that is also an S3 object" was recognized only by a non-empty `Mime`. The filer strips the default `application/octet-stream` content-type to empty, so a promoted file (especially one created via the filer API) often has no `Mime` and was not recognized as the object it still is.

## Changes

- `filer`: `IsDirectoryKeyObject` (both the `filer_pb.Entry` and the filer-level helper) now also treats a directory carrying chunks/content as a key object, not only one with a `Mime`. This restores LIST visibility, delete-demotion, and lifecycle handling, and stops the empty-folder cleaner from reclaiming the object.
- `s3`: ranged `GET` rejects only zero-size directories, so a promoted object streams range requests.
- `s3`: `HEAD` 404s a directory only when it has no data, so the object is retrievable while empty/implicit directories still fall back to LIST (preserves the s3fs/PyArrow behavior).

## Behavior after

A path that is both an object and a prefix now coexists like S3's flat keyspace: the object lists, HEADs, and GETs (full and ranged), and `DELETE` of the object reclaims only its chunks while keeping the children. The object's data is no longer destroyed by background cleanup. Chunks stay in the canonical `Chunks` field, so fsck/replication continue to see them.

## Notes

- The filer HTTP API still presents the path as a directory listing — that is the hierarchical filesystem surface and is left as-is.
- `LIST` with a delimiter rolls the key into its common prefix rather than also emitting it as a key; this is pre-existing directory-object behavior, not changed here.

Fixes #10063

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed data loss issue for directories with content during cleanup operations.
  * Improved S3 API to properly retrieve directories containing data, enabling access to promoted file directories.

* **Tests**
  * Added test coverage for directory content detection logic.

### DIFF
--- a/weed/filer/filer.go
+++ b/weed/filer/filer.go
@@ -616,5 +616,6 @@ func (f *Filer) IsDirectoryKeyObject(ctx context.Context, p util.FullPath) (bool
 	if entry == nil {
 		return false, nil
 	}
-	return entry.IsDirectory() && entry.Mime != "", nil
+	// Mirror filer_pb.Entry.IsDirectoryKeyObject so the cleaner keeps a promoted file's data.
+	return entry.IsDirectory() && (entry.Mime != "" || len(entry.GetChunks()) > 0 || len(entry.Content) > 0 || entry.IsInRemoteOnly()), nil
 }
--- a/weed/pb/filer_pb/filer_pb_helper.go
+++ b/weed/pb/filer_pb/filer_pb_helper.go
@@ -23,7 +23,10 @@ func (entry *Entry) IsInRemoteOnly() bool {
 }
 
 func (entry *Entry) IsDirectoryKeyObject() bool {
-	return entry.IsDirectory && entry.Attributes != nil && entry.Attributes.Mime != ""
+	// Also true for a file promoted to a directory by a child write, which keeps its
+	// chunks/content, or its remote entry when the file was tiered to remote storage.
+	return entry.IsDirectory &&
+		((entry.Attributes != nil && entry.Attributes.Mime != "") || len(entry.GetChunks()) > 0 || len(entry.GetContent()) > 0 || entry.IsInRemoteOnly())
 }
 
 func (entry *Entry) GetExpiryTime() (expiryTime int64) {
--- a/weed/pb/filer_pb/filer_pb_helper_test.go
+++ b/weed/pb/filer_pb/filer_pb_helper_test.go
@@ -1,0 +1,32 @@
+package filer_pb
+
+import (
+	"testing"
+)
+
+func TestIsDirectoryKeyObject(t *testing.T) {
+	chunk := []*FileChunk{{FileId: "1,01", Size: 75}}
+
+	cases := []struct {
+		name string
+		e    *Entry
+		want bool
+	}{
+		{"plain directory", &Entry{IsDirectory: true, Attributes: &FuseAttributes{}}, false},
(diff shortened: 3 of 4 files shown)


chrislusf:
Good catch on the remote-tiering case. Updated both `IsDirectoryKeyObject` checks to also count a remote-only entry via `IsInRemoteOnly()`, so a remote-tiered file promoted to a directory is preserved by the cleaner and stays listable. Used `IsInRemoteOnly()` rather than a bare `RemoteEntry != nil` so a remote-mount directory (which carries a RemoteEntry with RemoteSize 0) is not mistaken for an object. Added unit cases for both the remote-tiered (true) and mount-dir (false) outcomes.

PR #10074, opened 2026-06-24
volume: overlap the local needle write with the replica fan-out

chrislusf:
**Draft / RFC — needs validation on a real multi-host cluster before merge.**

## What

`ReplicatedWrite` did the local disk write and then the replica fan-out
serially, so a replicated write cost `local + replicas`. The needle is already
fully buffered in memory, so this runs the local write and the replica fan-out
concurrently and acks the client only after both complete.

To keep it race-free, the fields the fan-out reads (TTL, last-modified, flags,
name, mime, pairs, data) are snapshotted up front, because the local
`WriteVolumeNeedle` stamps the volume TTL and mutates needle flags. After that
the fan-out touches only locals plus `n.Data` (read-only on both sides).

## Why

Where the local disk and the replica network are separate resources, write
latency drops from `local + replicas` to `max(local, replicas)`.

## Correctness

- Validated under `-race`: replicas byte-identical for 1 KB–20 MB writes, no
  data races in the volume write path.
- Replica-down still fails the write (unchanged semantics).
- `go vet` clean; topology/operation/needle unit tests pass.

## The honest caveat (why this is a draft)

On a single machine where all volume servers share one disk, overlapping the
two writes **doubles the simultaneous disk I/O and regresses** — a loopback A/B
(replication=001) showed:

| workload | serial | overlap |
|---|---|---|
| 4 MB × 1 | 232 MB/s | 249 |
| 4 MB × 16 | 700 | 340 |
| 32 MB × 1 | 323 | 152 |
| 32 MB × 16 | 422 | 65 (p50 5.8 s) |

That contention is a single-disk artifact. The intended win (`max` vs `sum`)
only appears when replicas sit on separate disks/hosts with real network
latency, which a loopback benchmark structurally can't show. Holding as a draft
until someone measures it on a distributed cluster; if it doesn't pay off there
either, this should be dropped.

One behavior change worth a look: the local write no longer strictly precedes
replication, so a local failure can leave a (valid, retry-superseded) needle on
a replica. The client ack still waits for both, so read-after-write is
preserved.

### DIFF
--- a/weed/topology/store_replicate.go
+++ b/weed/topology/store_replicate.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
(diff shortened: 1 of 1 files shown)


chrislusf:
Addressed the review:
- The local-write goroutine now assigns a local `localIsUnchanged` and the function returns it after `wg.Wait()`, instead of writing the named return from the goroutine.
- On the TTL suggestion: `TTL.String()` already guards a nil receiver (returns `""`), so there is no panic. I kept the `n.Ttl == EMPTY_TTL` check as-is because it mirrors `writeNeedle2`'s own stamping — normalizing a nil TTL to `EMPTY_TTL` and applying `vol.Ttl` here would make the replica's stamped TTL diverge from the local needle, which `writeNeedle2` leaves nil.

PR #10076, opened 2026-06-24
filer: optional client-side replication fan-out

chrislusf:
**Draft / RFC — opt-in (`WEED_FILER_CLIENT_SIDE_REPLICATION=true`), default off. Performance needs validation on a real multi-host cluster.**

## What

Today the filer uploads a chunk to **one** volume server, and that volume server relays the copies to the other replica holders (server-driven replication). This adds a `type=replicate` fan-out path on the filer: it uploads the chunk to **every** holder in parallel and writes each with `type=replicate`, so no volume relays.

The assign result already lists every holder (`AssignResult.Replicas`), so the filer addresses them directly. Gated behind `WEED_FILER_CLIENT_SIDE_REPLICATION`; unset keeps the existing server-driven path byte-for-byte.

- `assignNewFileInfoReplicas` — assign + one `?type=replicate` URL per holder.
- `doUploadToReplicas` — upload the chunk to all holders concurrently; ack only if all succeed (all-or-nothing, matching server-driven).
- `dataToChunkWithSSE` branches on the flag.

## Why it might help

It removes the primary volume server's **receive-then-resend relay** and writes all copies from the source at once. Under concurrent load that relay is a per-primary bottleneck, so spreading writes evenly across holders should help most there.

## Correctness — validated

On a 3-volume `replication=001` cluster (writes via the S3 gateway):
- Each chunk lands on 2 of 3 volumes (file counts confirm the fan-out), reads return correct data.
- **Killed one of three volumes → all objects still readable** (every chunk survived on another holder). Replication is real and durable.

## Performance — inconclusive on a single machine

A/B (flag off vs on), 3 volumes + filer + s3 + load generator all co-located on one RAM disk, ran twice:

| 32 MB ×16 | run 1 | run 2 |
|---|---|---|
| server-driven | 387 MB/s | 404 |
| client-fanout | 427 (**+10%**) | 250 (**−38%**) |

The two runs **disagree by ~50 points** — the result is noise-dominated, because every volume shares one box's storage/CPU and the A/B halves run sequentially (load drifts between them). I can't get a reliable read here.

The intended win (relay elimination, even load spread, parallel-from-source) only shows cleanly when holders are on **separate hosts/disks** — and it comes with a real tradeoff: the filer now sends the data N× from its own NIC, so a filer-uplink-constrained deployment could regress. Holding as a draft until measured on a distributed cluster.

### DIFF
--- a/weed/server/filer_server_handlers_write.go
+++ b/weed/server/filer_server_handlers_write.go
@@ -77,6 +77,53 @@ func (fs *FilerServer) assignNewFileInfo(ctx context.Context, so *operation.Stor
 	return
 }
 
+// assignNewFileInfoReplicas assigns a file id and returns one upload URL per
+// replica holder (the assigned volume plus assignResult.Replicas). Each URL
+// carries type=replicate so the receiving volume writes locally and does not
+// fan out again — the filer performs the replication itself by uploading to
+// every holder. Used by the client-side replication write path.
+func (fs *FilerServer) assignNewFileInfoReplicas(ctx context.Context, so *operation.StorageOption, expectedDataSize uint64) (fileId string, urls []string, auth security.EncodedJwt, err error) {
+
+	stats.FilerHandlerCounter.WithLabelValues(stats.ChunkAssign).Inc()
+	start := time.Now()
+	defer func() {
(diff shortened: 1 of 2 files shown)


chrislusf:
Superseded by the s3-gateway fan-out PR: the fan-out belongs in the S3 gateway's chunked uploader, not the filer write path (which S3 writes don't use).

PR #10078, opened 2026-06-24
s3: replicate by fanning out from the gateway to every holder

chrislusf:
**Draft / RFC.** Supersedes #10076 (which put the fan-out in the filer write path — a path S3 writes never use, since the gateway runs its own chunked uploader).

## What

S3 writes go to the S3 gateway, and the gateway uploads each chunk directly to a volume server, which then **relays** the copies to the other replica holders. The gateway now uploads each chunk to **every** holder in parallel (`type=replicate`), so no volume relays.

- `AssignVolumeResponse` gains `repeated Location replicas` (forwarded from the master assign, which already knows the holders).
- The filer's `AssignVolume` populates it; the s3api `assignFunc` captures it; `operation.UploadReaderInChunks` fans out whenever a chunk has more than one holder.
- Cipher uploads keep the server-driven path (per-call encryption would diverge the replicas).
- A failed holder upload cancels the remaining ones.

## Why

It removes the primary volume server's **receive-then-resend relay** — the gateway writes all copies from the source at once, so a write costs `max(holder uploads)` instead of `upload + relay`.

## Correctness — validated

3-volume `replication=001`, writes via the S3 gateway:
- The fan-out runs (trace shows each chunk written to its 2 holder volumes directly).
- Write+read round-trips match MD5; killing one of three volumes loses nothing.
- `go vet` clean, `operation` tests pass.

## Performance

A/B (fan-out vs master), **fresh wiped cluster per measurement** to avoid accumulation noise, gateway + filer + 3 volumes + load gen co-located on one RAM disk, 2 passes:

| workload | pass 1 | pass 2 |
|---|---|---|
| 4 MB ×1 | +8% | +23% |
| 32 MB ×1 | **+20%** | **+20%** |
| 4 MB ×16 | −2% | −1% |
| 32 MB ×16 | **−20%** | **−15%** |

A consistent **tradeoff** on a single shared disk: fan-out wins single-stream writes (32 MB ×1 +20% both passes, lower p50) but loses large concurrent writes (32 MB ×16 ~−17% both passes). The regression is expected on this setup — at 16 concurrent 32 MB writes the gateway pushes ~2× the simultaneous write streams onto the **one shared disk**. In a real deployment the two copies land on **separate disks/hosts**, where they don't contend and the relay-elimination win should hold; that's the case this single box can't measure. Needs a distributed cluster before leaving draft.

## Note
`filer.pb.go` is regenerated; the large diff is the re-encoded descriptor (the only semantic change is the new `replicas` field). Worth regenerating with the repo's canonical protoc before merge.

## Summary by CodeRabbit

* **New Features**
  * Chunked uploads can now fan out to replica locations (when available) for improved resilience.
  * Volume assignment responses now expose replica locations in addition to the primary location.
* **Bug Fixes**
  * Replicated chunk uploads run concurrently, return the first successful result, and perform best-effort cleanup when fan-out partially fails.
* **Tests**
  * Added coverage to verify replica rollback behavior on partial fan-out failure.


chrislusf:
Addressed the review:
- Removed the `WEED_UPLOAD_REPLICA_FANOUT` env var entirely — fan-out is now the default — so the repeated `os.Getenv` / atomic-caching concern no longer applies.
- `uploadChunkToHolders` now cancels the remaining holder uploads on the first failure via a cancelable context.

chrislusf:
Addressed the remaining fan-out cleanup finding (failed copies invisible to cleanup). On a partial fan-out failure, `uploadChunkToHolders` now deletes the chunk from the holders that already succeeded (`DELETE ?type=replicate`, local-only) before returning the error, so a half-replicated write leaves no orphaned needle rather than relying on the caller to surface it. The earlier cancelable-context change only stopped in-flight uploads; this covers the copies that landed before the cancel. Added a unit test asserting the succeeded holder receives the cleanup delete and the failed holder does not.

chrislusf:
Made the rollback test deterministic instead of timing-dependent: the failing holder now blocks on a channel until the good holder has stored the chunk, so the cancel can't fire before the successful upload completes. The good copy is therefore always what gets rolled back. Verified with `-race -count=5`.

PR #10089, opened 2026-06-24
mount: skip redundant permission checks under default_permissions

chrislusf:
Follow-up to #10008. That PR cached the supplementary-group lookup; this removes the work that needed caching in the first place.

With `default_permissions` (the mount default, `-defaultPermissions=true`) the kernel enforces unix permission bits from the getattr/lookup attributes before it ever calls `Open`, `Create`, or `Mknod`. The mount was still re-checking in `AcquireHandle` and `createRegularFile`, duplicating the kernel's check and keeping the group lookup on the per-file hot path (a long copy re-pays a synchronous NSS/LDAP round-trip each time the cache TTL expires).

Gate those checks on `default_permissions` being off, so a non-root copy does no permission work on open/create — identical to root. `createRegularFile` also drops its per-file parent-directory load. With `default_permissions` off the mount stays the sole enforcer, so the full check still runs.

`TestAcquireHandleHonorsDefaultPermissions` covers both directions: skipped when on, EACCES when off.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mount permission handling to better align with the kernel’s `default_permissions` behavior when enabled.
  * Reduced redundant permission checks during open and create flows, minimizing incorrect access denials.
* **Tests**
  * Added automated coverage to confirm expected access results for owner-only files with `default_permissions` on vs off.
  * Added a test ensuring file creation under a missing parent directory is correctly rejected.

### DIFF
--- a/weed/command/mount_std.go
+++ b/weed/command/mount_std.go
@@ -359,6 +359,7 @@ func RunMount(option *MountOptions, umask os.FileMode) bool {
 		Cipher:                      cipher,
 		UidGidMapper:                uidGidMapper,
 		IncludeSystemEntries:        *option.includeSystemEntries,
+		DefaultPermissions:          *option.defaultPermissions,
 		DisableXAttr:                *option.disableXAttr,
 		IsMacOs:                     runtime.GOOS == "darwin",
 		MetadataFlushSeconds:        *option.metadataFlushSeconds,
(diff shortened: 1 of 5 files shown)


chrislusf:
Fixed: `createRegularFile` now always loads the parent to validate it exists (the create RPC sets `SkipCheckParentDirectory`, so it is the only parent guard); `default_permissions` gates only the mode-bit check. Added a test.

PR #10093, opened 2026-06-24
s3tables: allow hyphens in namespace and table names

chrislusf:
Iceberg REST clients routinely use hyphenated namespace and table names (`my-table`, `daily-events`, `rest-integration-test`). The S3 Tables charset only allowed `a-z`, `0-9`, `_`, so the Iceberg REST catalog rejected those names with a 400 — not conformant for those clients.

Accept `-` as an interior character in `validateNamespacePart` / `validateTableName`. Names must still start (and namespaces still end) with a letter or digit, and the `.`/`..`/`/`/`aws`-prefix guards are unchanged, so this is a permissive superset of the AWS S3 Tables charset — it never breaks an AWS-valid name.

With this, the full apache/iceberg-go REST integration suite passes against SeaweedFS using its original (hyphenated) names, unpatched: CreateNamespace, CreateTable, CreateView, LoadNamespaceProps, MultiTableCommit, UpdateNamespaceProps, UpdateView, WriteCommitTable.

## Summary by CodeRabbit

* **New Features**
  * Expanded S3 Tables namespace and table-name validation to allow hyphens (`-`) in valid positions.
* **Tests**
  * Added coverage for hyphenated identifiers, including ARN parsing and REST request validation, and ensured invalid hyphen placements (such as leading hyphens) are rejected with clear errors.

### DIFF
--- a/weed/s3api/s3api_tables_rest_validation_test.go
+++ b/weed/s3api/s3api_tables_rest_validation_test.go
@@ -21,7 +21,7 @@ func TestBuildListTablesRequestRejectsInvalidNamespaceQuery(t *testing.T) {
 		namespace string
 	}{
 		{name: "uppercase", namespace: "InvalidNamespace"},
-		{name: "hyphen", namespace: "invalid-ns"},
+		{name: "leading hyphen", namespace: "-invalid"},
 		{name: "slash", namespace: "a/b"},
 	}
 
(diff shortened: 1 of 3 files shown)


chrislusf:
Good catch on the ARN patterns — addressed in 46353755c: widened `tableNamespacePatternStr`/`tableNamePatternStr` to include `-` so `parseTableFromARN` accepts hyphenated names, and added `TestParseTableFromARNAllowsHyphens`. Also retargeted the existing reject-the-hyphen REST tests at a still-invalid leading-hyphen name, since interior hyphens are now valid.

PR #10096, opened 2026-06-24
fix(s3api): apply static config file updates on reload

chrislusf:
Addresses https://github.com/seaweedfs/seaweedfs/discussions/10079.

Editing an existing identity's `secretKey` in the `-config` file and sending `SIGHUP` no longer applied the change. The reload path (`grace.OnReload` → `loadS3ApiConfigurationFromFile`) flows through `MergeS3ApiConfiguration`, which skips identities marked static so that dynamic admin/filer updates can't clobber them. That skip also blocked the authoritative config file from updating its own identities.

Thread a `fromStaticFile` flag from the file-load path into the merge: a reload of the config file overwrites its static identities (so an operator can rotate a static `secretKey` by editing the file and reloading), while dynamic admin/filer updates still leave static identities immutable. The protection added in #7989 is preserved — the file stays the single source of truth for static identities.

Added `TestReloadStaticConfigUpdatesExistingSecretKey`; verified it fails before the change and passes after.

Out of scope: removing an identity from the file on reload still doesn't evict it from memory (pre-existing merge behavior — merge never prunes identities absent from the incoming config).

## Summary by CodeRabbit

* **Bug Fixes**
  * Reloading the S3 IAM config file now refreshes existing static identities with the updated credentials from the file.
  * Static identities stay marked as static after reload, and updates to related service-account credentials under a static parent are reapplied correctly.
* **Tests**
  * Added unit tests to verify that rotated static identity secret keys (including those for service accounts) are reflected after config reload.

### DIFF
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -503,7 +503,9 @@ func (iam *IdentityAccessManagement) loadS3ApiConfigurationFromFile(fileName str
 		glog.Warningf("KMS initialization failed: %v", err)
 	}
 
-	config, err := iam.loadS3ApiConfigurationFromBytes(content)
+	// fromStaticFile=true: this file is the source of truth for its static
+	// identities, so a reload overwrites them (e.g. a rotated secretKey).
+	config, err := iam.loadS3ApiConfigurationFromBytes(content, true)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


chrislusf:
Thanks for the reviews. Addressed in the latest push:

**CodeRabbit (atomicity of static marking).** The security-relevant part is the immutability guard in `RemoveIdentity`, which checks `identity.IsStatic`. Since the reload now rebuilds static identities as fresh objects, the merge sets `IsStatic = fromStaticFile` on them directly, so a published static identity is never observable as removable in the window before `markStaticIdentities` runs. The remaining `staticIdentityNames` map population still happens just after publish, but that gap is pre-existing and only matters for a concurrent admin upsert of a brand-new name landing mid-SIGHUP; the removal guard itself is now fully covered.

**CodeRabbit (service-account test).** Added `TestReloadStaticConfigUpdatesServiceAccountSecret`, which rotates a service-account secret under a static parent and asserts it's reapplied on reload.

**Gemini (data race on `parentIdent.Credentials` append).** In every current merge caller the parent is present in `config.Identities` (full credential-manager config, or the full static file), so the service-account loop's `parentIdent` is the freshly built identity from the identities loop above — not yet published, so concurrent readers can't hold it and there's no shared mutation. The clone-under-lock would only be needed if a caller passed service accounts without their parent identities, which none do today; not introduced by this PR, so leaving it to keep the change scoped.

PR #10098, opened 2026-06-24
fix(replication): add JWT auth header for chunk reads via filer proxy

MorezMartin:
## What

When the source filer has `jwt.filer_signing.key` configured, `filer.sync` and `filer.backup` now include a valid Bearer JWT token in HTTP GET requests made through the filer proxy (`-a.filerProxy` / `-filerProxy`).

## Why

Before this change, `FilerSource.ReadPart()` passed an empty JWT string (`""`) when proxying chunk data:

```go
filename, header, resp, err = downloadFn("http://"+fs.address+"/?proxyChunkId="+fileId, "", offset)
```

This caused **401 Unauthorized** errors when the destination filer had JWT authentication enabled, because the HTTP GET had no `Authorization: Bearer <JWT>` header. The S3 API already generates JWT tokens for filer reads, but replication didn't — making filer.sync unusable behind JWT-authenticated filers with `-a.filerProxy`.

This matches issue [#3352](https://github.com/seaweedfs/seaweedfs/issues/3352).

### `weed/replication/source/filer_source.go`
- Add `signingKey` and `signingExpiresSec` fields to `FilerSource` struct
- Add `SetJwtSigningKey(key, expiresSec)` setter method
- Modify `ReadPart()` to generate JWT via `GenJwtForFilerServer()` and pass it to `downloadFn()` when `proxyByFiler` is true

### `weed/command/filer_sync.go`
- After creating `FilerSource`, configure JWT signing key from `jwt.filer_signing.read.key` and `jwt.filer_signing.read.expires_after_seconds` (default 60s)

### `weed/command/filer_backup.go`
- Same JWT configuration for `filer.backup`

## Testing
- Build passes clean: `go build ./weed/replication/source/ ./weed/command/`
- All existing replication tests pass: `go test ./weed/replication/`
- Backward compatible: when `signingKey` is empty, the code behaves exactly as before (empty JWT, no auth header)

## Summary by CodeRabbit

* **New Features**
  * Backup and sync now use authenticated filer-proxy chunk reads by generating and attaching JWTs when signing keys are configured.

* **Bug Fixes**
  * Improved reliability for backup/replication when accessing chunk data via the filer proxy.
  * Added a default JWT read token expiry (60 seconds) and fallback key behavior to keep proxy read access working smoothly.

### DIFF
--- a/weed/command/filer_backup.go
+++ b/weed/command/filer_backup.go
@@ -167,6 +167,14 @@ func doFilerBackup(grpcDialOption grpc.DialOption, backupOption *FilerBackupOpti
 		*backupOption.proxyByFiler); err != nil {
 		return fmt.Errorf("filersource initialization failed: %v", err)
 	}
+	// Configure JWT signing key for authenticated chunk reads via filer proxy.
+	// Falls back to jwt.filer_signing.key if read.key is not set.
+	util.GetViper().SetDefault("jwt.filer_signing.read.expires_after_seconds", 60)
+	filerSource.SetJwtSigningKeys(
+		util.GetViper().GetString("jwt.filer_signing.read.key"),
+		util.GetViper().GetString("jwt.filer_signing.key"),
+		util.GetViper().GetInt("jwt.filer_signing.read.expires_after_seconds"),
+	)
 
 	if err := repl_util.InitializeSSEForReplication(filerSource); err != nil {
 		return fmt.Errorf("SSE initialization failed: %v", err)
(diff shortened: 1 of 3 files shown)


chrislusf:
Thanks for digging into this. The proxy chunk read isn't gated by a filer token — the filer's `?proxyChunkId=` path returns before its JWT check and just forwards to the volume server, which is what enforces read signing and needs a fileId-scoped `GenJwtForVolumeServer` token (volume key), not `GenJwtForFilerServer` (filer key). Fixed at the filer in #10100, which covers filer.sync, filer.backup, filerProxy mounts, the MQ broker and the upload gateway in one place. Closing in favor of that.

MorezMartin:
Thanks for your excellent work

PR #10105, opened 2026-06-25
Add performance CI (profiling, throughput, S3 read/write)

chrislusf:
Adds a performance workflow with three parallel jobs:

- profile: starts the server under write load and captures CPU + heap pprof, uploaded as artifacts with a top-functions summary.
- throughput: runs `weed benchmark` (native volume path) plus the Go micro-benchmarks.
- s3: concurrent S3 GET/PUT against the gateway, reporting requests/sec, transfer rate, and latency percentiles.

Runs on push to master and manual dispatch (tunable profiling duration, object count, size, concurrency); not on every PR.

The S3 job uses a small self-contained load tool under `test/s3/benchmark`, built on the aws-sdk-go-v2 client the S3 tests already use, so no extra benchmark binary is required.

## Summary by CodeRabbit

* **New Features**
  * Added a performance workflow for automated and manual benchmark runs.
  * Added S3 benchmark support with throughput and latency reporting.
  * Added profiling output collection and summarized benchmark results in run summaries.
* **Bug Fixes**
  * Improved run handling by waiting for services to be ready before starting tests.
  * Added clearer failure logs and uploaded artifacts for easier troubleshooting.

### DIFF
--- a/.github/workflows/performance.yml
+++ b/.github/workflows/performance.yml
@@ -1,0 +1,286 @@
+name: "Performance"
+
+on:
+  push:
+    branches: [ master ]
+    paths:
+      - '**/*.go'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/performance.yml'
+  workflow_dispatch:
+    inputs:
+      profile_duration:
+        description: "CPU profiling duration in seconds"
+        required: false
+        default: "30"
+        type: string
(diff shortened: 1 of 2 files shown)


chrislusf:
Applied the suggestions in the load tool: atomic.Int64 counters, slices.Sort, and full-precision ms conversion.

PR #10116, opened 2026-06-26
fix(filer.sync.verify): byte-order re-sort listings for pre-4.32 filers

kisow:

`filer.sync.verify` walks two filers' directory listings in lockstep (sorted-merge). Its only precondition is that both sides return entries in the **same order**: equal names pair up, and `eA.Name < eB.Name` is just the tie-breaker for a name present on one side only.

This held while both filers ran the same version, but it breaks when the two clusters being compared run **different versions**. Since #9824 (4.32) the filer forces `ORDER BY name COLLATE "C"` (byte order) regardless of the SQL `name` column collation, whereas a pre-4.32 filer whose `filemeta.name` column uses a locale collation lists in case-insensitive order. The two sides then return the **same name set in a different order**, the merge desyncs, and entries are reported as spurious `MISSING` / `ONLY_IN_B` (the recursive counter then walks whole subtrees). A real 4.25 → 4.34 comparison produced ~24k false `MISSING` with no actual data loss.

4.32+ already lists in byte order, so we **align the older side to that same order**: query each filer's version via `GetFilerConfiguration` (`MajorVersion`/`MinorVersion`, set since 3.71), and for any side older than 4.32 buffer its directory entries and re-sort them client-side in byte order before the merge. 4.32+ sides keep streaming; a mixed pair buffers only the older side. Detection failure sorts defensively. Automatic — no new flags, and identical behavior when both sides are 4.32+.

Implementation: an `entrySource` interface backs both the existing `entryStream` (streaming) and a new `sortedEntrySource` (buffer + `sort.Slice` by `Name`, using the same `<` the merge uses); `versionListsByteOrdered` / `filerNeedsByteOrderSort` decide per side. Additive, touching only `weed/command/filer_sync_verify.go` (+ tests).

# How is the PR tested?

- Unit tests: version boundary (4.31 → re-sort; 4.32 / 4.34 / 5.0 → no; 3.99 / 0.0 → re-sort) and a byte-order skew reproduction — same set in locale vs byte order yields false `MISSING` + `ONLY_IN_B` without re-sort, zero diffs with re-sort on both or only the older side.
- `go build ./...`, `go test ./weed/command/`, and `gofmt` clean.
- Live 4.25 → 4.34 run: the ~24k false `MISSING` collapse to the small set of genuine differences.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved sync verification for directory listings by accounting for filer version–dependent ordering, reducing false mismatch reports when entries are returned in different orders.
  * Added safer fallback behavior: automatically sorts entries when version/order assumptions can’t be confirmed.
  * Updated verification output to display the detected versions per side and whether re-sorting is enabled.

* **Tests**
  * Added regression tests covering byte-order skew scenarios, version-based ordering behavior, and concurrent verification listing behavior.

### DIFF
--- a/weed/command/filer_sync_verify.go
+++ b/weed/command/filer_sync_verify.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"sort"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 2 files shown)


chrislusf:
Thanks @kisow — solid diagnosis. Went with the always-sort variant in #10117 (merged 7c3c5ed2a): it buffers to re-sort anyway, so sorting both sides unconditionally drops the version detection, the `GetFilerConfiguration` round-trip, and the 4.32 constant. Closing as superseded — thanks for chasing down the root cause.

PR #10125, opened 2026-06-28
s3: support AWS object form for bucket policy Principal, add NotPrincipal

chrislusf:
## Problem

Bucket policy statements only accepted a bare string or array of strings for the `Principal` element, so the AWS-documented object form was rejected by `PutBucketPolicy`:

```json
"Principal": { "AWS": "arn:aws:iam::123456789012:root" }
"Principal": { "AWS": ["arn:aws:iam::123456789012:root", "999999999999"] }
```

This is the shape emitted by the AWS SDKs and Terraform (e.g. `Principal = { AWS = ["*"] }` for a public-read bucket), so policies written to the documented format failed to apply.

## Change

- New `PolicyPrincipal` type parses all three shapes:
  - bare `"*"` and a bare `["arn", ...]` array (the latter kept for backward compatibility)
  - the object form keyed by `AWS`, `Service`, `Federated`, or `CanonicalUser`, each value being a string **or** an array
- Keyed values are flattened into the match set, so `{"AWS":["*"]}` grants public access exactly like `"*"`.
- The original JSON is preserved and re-emitted on marshal, so `PutBucketPolicy` → `GetBucketPolicy` returns the **exact shape submitted**. This avoids the normalization drift that breaks Terraform/Ansible idempotency.
- Added **`NotPrincipal`** support (a statement applies to every principal except those named), wired through compilation and both policy evaluators, plus validation that rejects a statement specifying both `Principal` and `NotPrincipal` (per AWS).

Behavior follows the AWS *Principal* element reference; the keyed-object handling mirrors MinIO (`AWS` set) and Ceph RGW (full `AWS`/`Federated`/`Service`/`CanonicalUser`).

## Tests

`weed/s3api/policy_engine/principal_test.go`:
- every `Principal` shape + invalid inputs (`123`, `{}`, `{"AWS":123}`)
- faithful round-trip (idempotency)
- end-to-end parse → compile → evaluate for `{"AWS":["*"]}` public read and a named principal
- `NotPrincipal` deny-except semantics through both evaluators
- both-`Principal`-and-`NotPrincipal` rejection

`go test ./weed/s3api/... ./weed/iam/...` passes; `go vet` clean.

## Summary by CodeRabbit

* **New Features**
  * Added full support for exclusion-based principals (NotPrincipal) during policy evaluation, including dynamically generated matches.
  * Enhanced principal handling so policies can be parsed from and evaluated against multiple common JSON principal formats.

* **Bug Fixes**
  * Reject policies that specify both Principal and NotPrincipal in the same statement.

* **Tests**
  * Expanded regression coverage for principal parsing/round-tripping and for NotPrincipal behavior across both runtime and compiled policy evaluation paths.

### DIFF
--- a/weed/s3api/policy_engine/engine.go
+++ b/weed/s3api/policy_engine/engine.go
@@ -218,6 +218,17 @@ func (engine *PolicyEngine) evaluateStatement(stmt *CompiledStatement, args *Pol
 		}
 	}
 
+	// Check NotPrincipal (statement applies to everyone EXCEPT these principals)
+	if len(stmt.NotPrincipalPatterns) > 0 || len(stmt.DynamicNotPrincipalPatterns) > 0 {
+		matchedNotPrincipal := engine.matchesPatterns(stmt.NotPrincipalPatterns, args.Principal)
+		if !matchedNotPrincipal {
+			matchedNotPrincipal = engine.matchesDynamicPatterns(stmt.DynamicNotPrincipalPatterns, args.Principal, args)
+		}
+		if matchedNotPrincipal {
+			return false
+		}
+	}
+
 	// Check conditions
 	if len(stmt.Statement.Condition) > 0 {
 		condCtx := args.Conditions
--- a/weed/s3api/policy_engine/principal_test.go
+++ b/weed/s3api/policy_engine/principal_test.go
@@ -41,11 +41,48 @@ func TestPolicyPrincipalUnmarshal(t *testing.T) {
 }
 
 func TestPolicyPrincipalUnmarshalInvalid(t *testing.T) {
-	for _, in := range []string{`123`, `{}`, `{"AWS":123}`} {
+	// Includes object-form abuses: unknown key, empty array, empty string -- all
+	// of which must error rather than compile to zero matchers (which would let
+	// the match-all fallback silently make an Allow statement public).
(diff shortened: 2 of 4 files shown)


chrislusf:
Addressed the review feedback in `2f7728c68`:

- **Validate the Principal object form** (CodeRabbit): unsupported keys are rejected (only `AWS`/`Service`/`Federated`/`CanonicalUser`) and empty values are rejected, so a form like `{"AWS":[]}` or `{"AWS":""}` now errors instead of compiling to zero matchers and silently falling back to match-all.
- **Reject `Principal`+`NotPrincipal` by field presence** (CodeRabbit), not by flattened length, so a present-but-empty field is still caught.
- **Honor dynamic (policy-variable) `NotPrincipal`/`Principal` patterns in the compiled evaluator** (Gemini + CodeRabbit): a `NotPrincipal` made only of variables (e.g. `${aws:username}`) was previously treated as absent in the `CompiledStatement` path, bypassing the exclusion. The engine (pattern) path already handled this; the compiled path now does too, for both `Principal` and `NotPrincipal`.

Added regression tests for the object-form validation and the dynamic-`NotPrincipal` case. `go test ./weed/s3api/... ./weed/iam/...` passes; `go vet` clean.

The two red checks (Spark Iceberg / RisingWave catalog integration) are unrelated to this change — they exercise Iceberg REST-catalog operations over S3, not bucket-policy parsing — and have been re-triggered by the new push.

PR #10156, opened 2026-06-30
s3: invalidate stale reader cache locations on chunk read failure

MilanFun:

S3 GET can fail for an existing object when the optimized read path keeps stale volume locations in `ReaderCache`.

After a volume is moved or its holders change, S3 API may still try to read a chunk from an old volume server. The old server can return `404 Not Found` / `volume is not local`, even though the chunk exists on the current holders. In this case the client receives a failed or partial read, and restarting S3 API only helps because it clears the local lookup cache.

Add self-healing to the `ReaderCache` chunk read path.

When a chunk fetch fails, `ReaderCache` now invalidates the cached volume locations for that fileId, performs a fresh lookup, and retries the read once if the resolved locations changed. This mirrors the recovery behavior that already existed in the older streaming path.

The change also evicts completed failed downloaders from `ReaderCache`, so later reads do not keep returning a cached fetch error.

# How is the PR tested?

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved chunk read reliability by retrying downloads after refreshing file location information when prior fetches fail.
  * Added cache-invalidation driven recovery for stale locations, including robust handling of downloaders that complete with errors.
  * Prevents reuse of bad in-progress download paths by removing failed downloaders; improves stream retry consistency with fresh URLs.

* **Tests**
  * Added coverage for retry-after-cache-refresh and recovery after repeated fetch failures.
  * Updated concurrency tests to more precisely verify single-flight download deduplication.

### DIFF
--- a/weed/filer/filechunk_group.go
+++ b/weed/filer/filechunk_group.go
@@ -41,7 +41,7 @@ func NewChunkGroup(lookupFn wdclient.LookupFileIdFunctionType, chunkCache chunk_
 	group := &ChunkGroup{
 		lookupFn:          lookupFn,
 		sections:          make(map[SectionIndex]*FileChunkSection),
-		readerCache:       NewReaderCache(readerCacheLimit, chunkCache, lookupFn),
+		readerCache:       NewReaderCache(readerCacheLimit, chunkCache, lookupFn, nil),
 		concurrentReaders: concurrentReaders,
 	}
 
--- a/weed/filer/reader_at_test.go
+++ b/weed/filer/reader_at_test.go
@@ -77,7 +77,7 @@ func TestReaderAt(t *testing.T) {
 	readerAt := &ChunkReadAt{
 		chunkViews:    ViewFromVisibleIntervals(visibles, 0, math.MaxInt64),
 		fileSize:      10,
-		readerCache:   NewReaderCache(3, &mockChunkCache{}, nil),
+		readerCache:   NewReaderCache(3, &mockChunkCache{}, nil, nil),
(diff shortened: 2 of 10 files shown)


chrislusf:
Pushed cleanups and went through the bot review:

- **gemini — atomic global for `readerCacheFetchChunkData`:** removed the package-global test seam entirely; it's now a per-instance `fetchChunkDataFn` field set in `NewReaderCache` (tests set it on the cache). With no shared mutable global the race those comments flagged can't happen; `go test -race` is clean.
- **coderabbit — logging full chunk URLs:** the self-heal retry now logs the location count instead of the URL list, in the shared `retryFetchWithFreshLocations`.
- **coderabbit — gate invalidation to stale-location errors:** keeping it firing on any fetch error, to match the existing streaming self-heal. `urlSlicesEqual` already skips the refetch when locations are unchanged, so a transient failure only costs one invalidate + re-lookup. A `404` / `volume is not local` string predicate would be brittle, so leaving it out.
- **coderabbit — thread an invalidator into ChunkGroup/parquet/query:** out of scope here (this is the S3 read path); those callers pass `nil` and the retry is a graceful no-op. Mount can be wired separately if wanted.

Also factored the three copies of the invalidate -> re-lookup -> retry logic into one `retryFetchWithFreshLocations`, made the invalidator an explicit constructor argument, and dropped a redundant `completedTimeNew` store.

MilanFun:
@chrislusf Thanks a lot!

PR #10162, opened 2026-06-30
volume.balance: gate on real physical disk usage (fixes #10160)

### LINKED ISSUE #10160: volume.balance does not balance actual data distribution across volume servers
In my installation, volume servers use around 1 TB disks, but `max_volume_count` is configured too high for that disk size.

This is not an ideal setup. It can lead to many small volumes being placed on the same server, and it can distort how the balancer sees capacity.

However, even with this configuration, the expected balancing behavior still does not happen.

## Expected behavior

I expected `volume.balance` to help even out data distribution across volume servers and reduce pressure on the most physically full disks.

In other words, if some volume servers are much more filled than others, the balancer should select those servers as sources and move data to less utilized servers.

## Actual behavior

In my installation, the dry-run plan generated by `volume.balance` does not select the most physically full volume servers as the main sources.

Instead, it can move many volumes from less physically utilized servers and place them all onto a single target server.

In the observed case, the selected target server was `volume-1`, which was also one of the most physically loaded servers, around `99%` disk usage.

Example observed plan:

```text
total planned moves: 2088

source -> target:
volume-10 -> volume-1   1272 moves
volume-9  -> volume-1    345 moves
volume-11 -> volume-1    294 moves
volume-6  -> volume-1     90 moves
volume-8  -> volume-1     83 moves
volume-0  -> volume-1      4 moves
```

At the same time, the most physically full servers are barely selected as sources, or are not selected at all. Even worse, in this installation, the plan selected an already physically full server as the target for thousands of moves.

## Why this is a problem

Yes, the current `max_volume_count` value is too high for 1 TB disks, and this can make the situation worse.

But from my perspective as an operator, the main goal of running `volume.balance` is to improve the actual distribution of data and reduce disk pressure on overloaded volume servers.

This does not happen in this case.

The current balancing decision appears to be based mostly on an internal volume-density metric derived from volume sizes and `max_volume_count`, rather than actual physical disk utilization.

As a result, a volume server can be almost full on disk, but still not be treated as overloaded by `volume.balance`.

In my case, `volume-1` was already around `99%` used, but the generated plan still selected it as the only target for all planned moves.

## Expected improvement

It would be helpful if `volume.balance` considered the real state of the volume server, not only logical volume density.

For example, the balancing score could take into account:

- actual disk usage;
- available disk space;
- `VolumeCount`;
- `MaxVolumeCount`;
- `FreeVolumeCount`.

Even if `max_volume_count` is configured too high, the balancer should still try to move data in a way that improves actual data distribution across volume servers, instead of only optimizing the internal volume-density metric.

chrislusf:
## Problem (#10160)

`volume.balance` ranks servers by slot density (used volumes / `maxVolumeCount`). When `maxVolumeCount` is configured higher than the disk can physically hold, a near-full server's density looks tiny, so the planner treats it as the emptiest node and picks it as the move *target* — draining less-full servers onto it. In the reported case a server at ~99% disk usage became the sole target for ~2000 moves.

Root cause: the master topology carried only volume-slot counts, no physical disk bytes, so the balancer literally could not see real disk fullness. The same blind spot exists in the automatic maintenance balance worker.

## Change

Plumb real per-disk filesystem capacity end-to-end and use it to guard both balancers.

- **Proto**: `Heartbeat` carries `disk_total_bytes`/`disk_free_bytes` per disk type; `DiskInfo` carries them per disk.
- **Volume server** (Go + Rust `seaweed-volume`): caches the statfs total/free from the existing disk-space probe and reports it in the heartbeat.
- **Master topology**: ingests the bytes into `DiskUsageCounts` (delta-applied, same machinery as `maxVolumeCount`) and exposes them in `DiskInfo`.
- **Shell `volume.balance`**: a physical-fullness **gate** (`-maxDiskUsagePercent`, default 90) skips any move target whose disk is at/above that mark. Judged **per server against its own disk**, so heterogeneous disk sizes are fine. Slot density stays the in-bounds ranking objective. Servers not reporting bytes fall back to today's behavior. The earlier `-byDiskUsage` ranking mode is kept and is orthogonal.
- **Maintenance balance worker** (`weed/worker/tasks/balance`): the same gate — the greedy least-utilized-destination pick and its score-based fallback both skip servers whose disk is at/above 90% used; a full server can still be a source; when every destination is full, no tasks are created.
- **Admin dashboard**: prefers the real reported disk capacity instead of the slot-based `maxVolumeCount * volumeSizeLimit` estimate, which overstated capacity in exactly this misconfiguration.

## Tests

- Shell: `TestBalanceSkipsPhysicallyFullTarget` (gate on → no pile-on; off → bug reproduces; no-bytes → fallback) and `TestBalanceGateSteersToEmptierDisk`.
- Worker: `TestDetection_SkipsPhysicallyFullDestination` (steers to the empty disk, never the full one) and `TestDetection_NoDestinationWhenAllDisksFull` (no tasks when all destinations are full).
- Rust `test_build_heartbeat_reports_disk_bytes`: heartbeat carries the reported capacity.
- Full Go suite for the touched packages and the Rust `seaweed-volume` heartbeat/disk_location tests pass.

Within a single shell run the gate also tracks projected free space as volumes land, so it won't pile a run's worth of moves onto one initially-empty disk.

Note: the gate is on by default at 90% (`-maxDiskUsagePercent`), so healthy clusters are unaffected and over-full disks stop receiving data. The worker uses the same 90% default.

## Summary by CodeRabbit

* **New Features**
  * Added `volume.balance` targeting by actual physical disk usage (`-byDiskUsage`) plus a disk fullness safety gate (`-maxDiskUsagePercent`).
  * Heartbeats and topology now expose per-disk-type physical total/free byte metrics when available.
* **Bug Fixes**
  * Balancing now avoids move destinations predicted to be physically too full, and keeps disk free-space simulation consistent within a run.
* **Tests**
  * Added regression coverage for disk-usage targeting, fullness skipping, and destination steering.
* **Documentation**
  * Updated help/flags to reflect the new balancing options.

### DIFF
--- a/seaweed-volume/proto/master.proto
+++ b/seaweed-volume/proto/master.proto
@@ -99,6 +99,10 @@ message Heartbeat {
   volume_server_pb.VolumeServerState state = 23;
 
   repeated DiskTag disk_tags = 24;
+
+  // physical disk capacity per disk type, in bytes, from the underlying filesystem
+  map<string, uint64> disk_total_bytes = 25;
+  map<string, uint64> disk_free_bytes = 26;
 }
 
 message HeartbeatResponse {
@@ -331,6 +335,9 @@ message DiskInfo {
   // including disks with no volumes or EC shards; recovers empty disks that
   // carry no per-volume/per-shard records.
   map<uint32, int64> max_volume_count_by_disk = 11;
+  // physical disk capacity in bytes, from the underlying filesystem (0 if unknown)
+  uint64 disk_total_bytes = 12;
+  uint64 disk_free_bytes = 13;
 }
 message DataNodeInfo {
   string id = 1;
--- a/seaweed-volume/src/server/heartbeat.rs
+++ b/seaweed-volume/src/server/heartbeat.rs
@@ -802,6 +802,8 @@ fn build_heartbeat_with_ec_status(
     let mut volumes = Vec::new();
     let mut max_file_key = NeedleId(0);
     let mut max_volume_counts: HashMap<String, u32> = HashMap::new();
+    let mut disk_total_bytes: HashMap<String, u64> = HashMap::new();
+    let mut disk_free_bytes: HashMap<String, u64> = HashMap::new();
 
     // Collect per-collection disk size and read-only counts for metrics
     let mut disk_sizes: HashMap<String, (u64, u64)> = HashMap::new(); // (normal, deleted)
@@ -827,8 +829,12 @@ fn build_heartbeat_with_ec_status(
         if effective_max_count < 0 {
             effective_max_count = 0;
         }
-        *max_volume_counts.entry(disk_type_str).or_insert(0) += effective_max_count as u32;
+        *max_volume_counts.entry(disk_type_str.clone()).or_insert(0) += effective_max_count as u32;
         disk_max_by_id[disk_id] = effective_max_count;
+        *disk_total_bytes.entry(disk_type_str.clone()).or_insert(0) +=
+            loc.disk_total_bytes.load(Ordering::Relaxed);
+        *disk_free_bytes.entry(disk_type_str).or_insert(0) +=
+            loc.disk_free_bytes.load(Ordering::Relaxed);
 
         let mut delete_vids = Vec::new();
         for (_, vol) in loc.iter_volumes() {
@@ -962,6 +968,8 @@ fn build_heartbeat_with_ec_status(
         has_no_volumes,
         has_no_ec_shards,
         max_volume_counts,
+        disk_total_bytes,
+        disk_free_bytes,
         grpc_port: config.grpc_port as u32,
         location_uuids,
         disk_tags,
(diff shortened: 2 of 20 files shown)


chrislusf:
Rebased onto master after #10166 merged, and addressed review feedback.

**Rebase:** #10166 took `DiskInfo` field 11 (`max_volume_count_by_disk`), so the physical-byte fields moved to `disk_total_bytes = 12` / `disk_free_bytes = 13`; `store.go`, `data_node.go`, `master_helper.go`, and the Rust `heartbeat.rs` were merged with #10166's per-disk reporting (they're complementary — #10166 adds per-disk max counts + empty-disk recovery, this adds physical bytes). Regenerated `master.pb.go`; Go + Rust build and tests pass. No behavioral change was needed for #10166 — empty-disk reporting just gives the gate a more accurate topology.

**Review comments:**
- nil `DiskInfo` guards added in `capacityByActualDataUsage`, `diskBytes`, and `hasFreeVolumeSlot`.
- High-water cutoff now `>=` and the `90` default is a single constant — both live in the shared `balancer.DiskTooFullAfter` / `balancer.DefaultMaxDiskUsagePercent`, used by the shell and worker.
- `AdjustDiskUsageBytes`: a `0` report now clears previously stored bytes (0 means "not reported" for bytes, unlike maxVolumeCount), so a server that stops reporting falls back to slot-only instead of trusting stale capacity.
- Worker within-cycle accounting: each planned move's bytes are charged to its destination so the gate sees a target fill up across a detection cycle (the shell equivalent already did this via `adjustAfterMove`).
- Filesystem double-count (multiple `-dir` on one mount): documented. The gate is ratio-based (`used% = (total-free)/total`), and proportional double-counting cancels in the ratio, so gating stays correct; only absolute capacity over-reports. Reporting bytes per physical disk (mirroring `max_volume_count_by_disk`) is the exact fix and is a clean follow-up.

PR #10167, opened 2026-07-01
ec: skip physically near-full disks when placing EC shards

chrislusf:
Stacked on #10162 (needs the `DiskInfo.disk_total_bytes/disk_free_bytes` fields it adds). Will retarget to master once #10162 merges.

## Problem

EC placement had the same physical-disk blind spot #10162 fixes for volume balancing: every EC destination path scores candidates by free EC shard slots (derived from `MaxVolumeCount − VolumeCount`) and shard counts, and none reads real disk bytes. A disk that is physically full but still shows free EC slots keeps being chosen.

It's arguably worse for EC: EC shard files (`.ec00`–`.ec13`, `.ecx`) are counted by the statfs `DiskFreeBytes` but **not** by any slot/`VolumeInfos` accounting — so the slot math is precisely the metric that cannot see EC fullness.

## Change

Treat a disk at/above 90% physical usage (`balancer.DefaultMaxDiskUsagePercent`, the shared gate from #10162) as having **zero free EC slots** at snapshot-build time. Every existing `freeSlots > 0` placement predicate then excludes it, with no change to the planner model. Applied in all three snapshot builders:

- shell `countFreeShardSlots` (covers `ec.balance` and `ec.encode`),
- shared `ecbalancer.FromActiveTopology` (worker EC encode/repair `Place`),
- worker `ec_balance` `buildBalancerTopology`.

Judged per node/disk-type against its own disk. Servers not reporting disk bytes fall back to slot-only behavior. `ec.rebuild` recovery placement is intentionally **not** gated, so shard recovery can still complete onto fuller disks (durability first).

## Tests

- `TestFromActiveTopologySkipsPhysicallyFullDisk` — a 96%-full disk contributes 0 free slots to the shared snapshot; a 10%-full one keeps its slots.
- `TestCountFreeShardSlotsPhysicalDiskGate` — shell path returns 0 for a full disk, positive for an empty one, and falls back to slot math when bytes are unreported.
- Full `ecbalancer`, `ec_balance`, and `erasure_coding` suites pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved erasure-coding balancing to avoid placing data on disks that are physically near full, even if they still appear to have logical shard slots available.
  * Updated capacity calculations so near-full disks are treated as ineligible targets, helping keep balancing decisions aligned with real disk space.
  * Preserved normal behavior for disks with unavailable byte information, ensuring older or incomplete reports still work as expected.

### DIFF
--- a/weed/shell/command_ec_common.go
+++ b/weed/shell/command_ec_common.go
@@ -19,6 +19,7 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 	"github.com/seaweedfs/seaweedfs/weed/storage/super_block"
 	"github.com/seaweedfs/seaweedfs/weed/storage/types"
+	"github.com/seaweedfs/seaweedfs/weed/topology/balancer"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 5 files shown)


chrislusf:
Rebased onto master now that #10162 merged (diff is just the EC gate), and addressed the review comment: on a node with mixed disk types where one type is physically full, the node's free EC slots are now spread over only the **non-full** disks (dividing by the non-full disk count), so usable disks aren't diluted by the full type's disks (which still get 0). The total is preserved.

PR #10169, opened 2026-07-01
balance: share replica-placement logic between shell and worker

chrislusf:
Stacked on #10162 (uses the shared `weed/topology/balancer` package it introduces). Retargets to master once #10162 merges.

This is **Stage 1** of consolidating the two volume-balancing implementations (shell `volume.balance` and the maintenance balance worker), which had grown parallel copies of the same logic.

## Change

The replica-placement rule — data-center / rack / same-node limits plus best-effort host anti-affinity — existed three times:
- the shell's `satisfyReplicaPlacement` / `isGoodMove`, used by `volume.balance`, `volume.fix.replication`, and `volume.tier.move`;
- a line-for-line port in the maintenance balance worker (`weed/worker/tasks/balance/replica_placement.go`).

Move the canonical logic into `weed/topology/balancer` on a shared `Location` type. The shell and worker keep thin adapters that convert their own location representation and call the shared functions.

Behavior is unchanged: the shared `IsGoodMove` keeps the shell's explicit reject-move-to-self guard (which the worker's port lacked but never exercised), and all four replica test suites pass — `TestIsGoodMove`, `TestSatisfyReplicaPlacement*`, the worker's `TestIsGoodMove_*`, and `tier.move`/`fix.replication`.

## Next (Stage 2, separate)

The greedy engine itself is still duplicated. Per maintainer decision, the shared planner will standardize on the shell's bytes-aware density metric (the worker currently uses volume-count utilization, which #10160 showed is fooled by an over-set maxVolumeCount) — a pure `Plan(snapshot) []Move` both callers adapt to, with the worker's detection tests re-validated against density. That lands as its own PR.

## Summary by CodeRabbit

* **New Features**
  * Improved replica-balancing decisions when moving or fixing volumes, with smarter handling of placement constraints across data centers, racks, nodes, and hosts.

* **Bug Fixes**
  * Reduced the chance of creating invalid replica moves that would violate anti-affinity or leave volumes incorrectly placed.
  * Better preserves safe replica distribution during over-replicated cleanup and repair operations.

### DIFF
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -659,43 +659,29 @@ func moveVolume(commandEnv *CommandEnv, v *master_pb.VolumeInformationMessage, f
 	return nil
 }
 
+// toBalancerLocation converts a shell replica location to the shared placement
+// abstraction, resolving the physical host for machine anti-affinity.
+func toBalancerLocation(loc *location) balancer.Location {
+	return balancer.Location{
+		DataCenter: loc.dc,
+		Rack:       loc.rack,
+		NodeID:     loc.dataNode.Id,
+		Host:       pb.NewServerAddressFromDataNode(loc.dataNode).ToHost(),
+	}
+}
+
 func isGoodMove(placement *super_block.ReplicaPlacement, existingReplicas []*VolumeReplica, sourceNode, targetNode *Node) bool {
-	for _, replica := range existingReplicas {
-		if replica.location.dataNode.Id == targetNode.info.Id &&
-			replica.location.rack == targetNode.rack &&
-			replica.location.dc == targetNode.dc {
-			// never move to existing nodes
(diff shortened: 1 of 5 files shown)


chrislusf:
Rebased onto master now that #10162 merged (diff is just the replica-placement extraction), and addressed both review comments:
- `IsGoodMove` now removes only the single replica being moved off the source node (guarded by `!sourceFound`), so an over-replicated node's other replica stays in the after-move count. Added `TestIsGoodMove_OverReplicatedSourceKeepsOtherReplica`, which is rejected-with / allowed-without the fix.
- The worker adapter converts `types.ReplicaLocation` → `balancer.Location` field-by-field (`toBalancerLocation`) instead of a whole-struct cast, so a future field change is a compile error at the mapping rather than a silent mismatch.

PR #10174, opened 2026-07-01
balance: extract the bytes-aware density metric to a shared package

chrislusf:
Part of consolidating the two volume balancers (shell `volume.balance` + maintenance worker) onto one bytes-aware density metric. The gate and replica placement were already shared (#10162, #10169); this factors out the density metric itself.

## Change

Move the shell's density math — used-volume-equivalents over free capacity — into `weed/topology/balancer`:
- `VolumeDensity(maxVolumeCount, volumeBytes, volumeSizeLimit)` → `(capacity, usedVolumes)`
- `DensityRatio(capacity, usedVolumes)` — a server's load (empty servers get a small positive weight so they sort below loaded ones)
- `DensityNextRatio(...)` — the load after one more volume, for the overshoot/convergence guard
- `UsedVolumeEquivalents(...)`

The shell's `capacityByMinVolumeDensity`, `capacityByActualDataUsage`, and the `localVolumeDensity*Ratio` methods now call these. Shell behavior is unchanged — all `volume.balance` tests pass as-is.

## Why a shared home now

It sits alongside the balancing primitives already in this package (`DiskTooFullAfter`, `Location`, `IsGoodMove`), and gives the maintenance worker a single density definition to adopt. The worker adoption is a follow-up: its ranking and convergence are currently volume-*count* based, and switching to bytes-aware density means summing per-server `FullnessRatio` (or plumbing the volume-size limit) and re-validating its detection tests. That's a behavior change worth its own PR — and lower priority, since the disk-fullness gate already prevents the over-configured-`maxVolumeCount` pathology in the worker regardless of the ranking metric.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume balancing so servers are ranked more consistently during placement and rebalancing.
  * Fixed edge cases for empty or zero-capacity servers, reducing unexpected balancing behavior.
  * Made density-based selection more accurate when estimating available space and next placement decisions.

* **Tests**
  * Added coverage for density and ratio calculations, including empty, loaded, and zero-capacity scenarios.

### DIFF
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -23,10 +23,7 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 )
 
-const (
-	thresholdVolumeSize      = 1.01
-	countZeroSelectedVolumes = 0.5
-)
+const thresholdVolumeSize = 1.01
 
 func init() {
 	Commands = append(Commands, &commandVolumeBalance{})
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed the negative-capacity ranking issue: `DensityRatio` and `DensityNextRatio` now return `+Inf` for any non-positive capacity (a full server, or an overfull one whose data exceeds its configured slots so capacity goes negative), so it sorts as the **fullest** — a move source, never the emptiest target. Added tests for negative capacity and an ascending-sort ordering check (`TestDensityRatio_OverfullSortsFullest`); shell balance tests still pass unchanged (the shell also filters `capacity > 0` before ranking, but this hardens the shared helper for any consumer and for a server that crosses into overfull mid-run).

Issue #9872, opened 2026-06-08
Seaweed doesn't recover/clean up after EC encoding failures


plisandro:
We've recently hit the same issue originally reported by @hanxiantao in https://github.com/seaweedfs/seaweedfs/issues/6445, which got closed last year 😢  It manifested as sporadic read/delete failures on EC volumes due to mismatches between EC shard contents and their indexes:

```
...
1780307163159	2026-06-01T09:46:03.159Z	I0601 09:46:03.159734 common.go:152 error JSON response status 500: Deletion Failed: ec volume 1766 needle 77313fccf4d3241ed4ac71cc Size:-1408379405, DataSize:0, Name:, Mime: Compressed:false offset 100878845512 size 4194405: size mismatch: entry not found: offset 100878845512 found id 37373331336663636634643332343165 size -1408379405, expected size 4194405
1780307163159	2026-06-01T09:46:03.159Z	I0601 09:46:03.159809 common.go:109 response method:DELETE URL:/1766,2bb60ee200000000 with httpStatus:500 and JSON:{"error":"Deletion Failed: ec volume 1766 needle 77313fccf4d3241ed4ac71cc Size:-1408379405, DataSize:0, Name:, Mime: Compressed:false offset 100878845512 size 4194405: size mismatch: entry not found: offset 100878845512 found id 37373331336663636634643332343165 size -1408379405, expected size 4194405"}
1780314734869	2026-06-01T11:52:14.869Z	I0601 11:52:14.869040 common.go:152 error JSON response status 500: Deletion Failed: ec volume 1752 needle a5ca5097309d774e47157f78 Size:-1327273323, DataSize:0, Name:, Mime: Compressed:false offset 39949334504 size 1560: size mismatch: entry not found: offset 39949334504 found id 61356361353039373330396437373465 size -1327273323, expected size 1560
1780314734869	2026-06-01T11:52:14.869Z	I0601 11:52:14.869653 common.go:109 response method:DELETE URL:/1752,2b8a6cad00000000 with httpStatus:500 and JSON:{"error":"Deletion Failed: ec volume 1752 needle a5ca5097309d774e47157f78 Size:-1327273323, DataSize:0, Name:, Mime: Compressed:false offset 39949334504 size 1560: size mismatch: entry not found: offset 39949334504 found id 61356361353039373330396437373465 size -1327273323, expected size 1560"}
...
```

[...]

PR #10204, opened 2026-07-02
weed shell: add ec.shard.unmount command

chrislusf:
Adds `ec.shard.unmount` to the weed shell, so broken or over-replicated EC shards can be unmounted (and optionally deleted) without stopping volume servers and deleting files by hand. Helps with #9872.

Alternative to #10201, couldn't push to that branch.

- default is unmount, `--delete` also removes the shard files
- shard IDs as `2`, or `3@host:port` to pick one copy (same form the listing prints)
- prints the live topology, then dry-runs unless `--apply`

## Summary by CodeRabbit

* **New Features**
  * Added a new CLI command for unmounting erasure-coded shards from a volume.
  * Supports selecting one or more shard IDs, optional node-specific targeting, dry-run preview, forced execution, and optional shard-file deletion.
  * Includes safeguards for invalid, ambiguous, or missing shard selections and validates actions against the current cluster topology.

* **Tests**
  * Added coverage for shard input parsing, including qualified shard targets, invalid formats, and out-of-range values.
  * Added a round-trip check to confirm shard targets can be converted to text and parsed back correctly.

### DIFF
--- a/weed/shell/command_ec_shard_unmount.go
+++ b/weed/shell/command_ec_shard_unmount.go
@@ -1,0 +1,276 @@
+package shell
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"io"
+	"math"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/operation"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review notes:

- `Do` now calls `handleHelpRequest` so `-h`/`--help` prints the command help instead of a parse error.
- Explicit `missing shardId` error when `--shardId` is omitted, instead of the confusing `empty shard ID in ""`.
- Shard ID is bounds-checked as an int before narrowing to uint32, so an out-of-range value (e.g. 4294967301) can't wrap into a valid-looking shard ID. Added a test for it.

PR #10197, opened 2026-07-02
feat(k8s): add HTTP + gRPC Ingress templates and Traefik TCP routing for filer

MorezMartin:
## Summary

Add HTTP and gRPC Ingress templates for the SeaweedFS filer component, along with a Traefik IngressRouteTCP for mTLS-enabled gRPC access. Also restructures the Helm values `ingress` field into an `ingresses` sub-structure (http/grpc/tcp) for cleaner configuration.

Additionally includes a fix for the filer replication source to include the JWT auth header when reading chunks through the filer proxy.

## Problem Solved

This PR addresses the lack of proper gRPC ingress support for the SeaweedFS filer Helm chart. Without these changes:

1. **No gRPC ingress template** — the chart only had a single HTTP ingress, making it impossible to properly expose the filer's gRPC port (18888) behind Kubernetes ingress controllers like Traefik.
2. **No Traefik TCP passthrough** — gRPC requires TCP-level passthrough, not HTTP/2 termination, which is only possible with a `IngressRouteTCP` custom resource.
3. **Replication breaks via filer proxy** — when the filer proxies chunk reads (for replication/sync), it needs to authenticate to the volume server via JWT. Without the JWT auth header in `filer_source.go`, the replication sink cannot read chunks through the filer proxy, causing `403` or `401` errors.

### 1. HTTP + gRPC Ingress templates (`templates/filer/filer-ingress.yaml`)
- Replaces the old single `ingress` field with a structured `ingresses.http` and `ingresses.grpc` sub-structure
- Both ingresses expose the filer service via HTTP (port 8888) and gRPC (port 18888)
- gRPC ingress uses `protocol: TCP` for gRPC compatibility
- Configurable via $.Values.filer.ingresses.http and $.Values.filer.ingresses.grpc

### 2. Traefik TCP IngressRoute (`templates/filer/filer-ingress-tcp.yaml`)
- New file: `IngressRouteTCP` for mTLS gRPC passthrough
- Uses Traefik's `entryPoints.filer-grpc` for TCP passthrough
- Supports TLS passthrough for end-to-end encryption
- Configurable via $.Values.filer.ingresses.tcp

### 3. Values restructure (`values.yaml`)
- `filer.ingress` → `filer.ingresses.http` / `grpc` / `tcp`
- All 3 sub-keys support: enabled, annotations, host, path, pathType

### 4. JWT auth fix (`weed/replication/source/filer_source.go`)
- Adds `BearerPrefix` constant for Authorization header
[...]

PR #10205, opened 2026-07-02
feat(k8s): filer HTTP + gRPC ingress for the Helm chart

chrislusf:
Adds gRPC ingress support to the filer Helm chart on the standard `Ingress` kind, so it stays consistent with the chart's other ingresses (master, volume, s3) and needs no controller-specific CRD:

- HTTP Ingress + a separate gRPC Ingress (nginx `backend-protocol: GRPC`), driven off a new `filer.ingresses.{http,grpc}` structure
- For end-to-end mTLS gRPC (TLS not terminated at the edge), the gRPC ingress documents the nginx `ssl-passthrough` annotation — SNI-based L4 passthrough, no CRD required

Focused re-scope of #10197 (templates + values by @MorezMartin, preserved via cherry-pick). Dropped from that PR: the `filer_source.go` replication JWT change (`SetJwtSigningKeys` had no caller, and the `?proxyChunkId=` path doesn't check an inbound JWT) and unrelated values.yaml whitespace churn. The original's Traefik `IngressRouteTCP` was replaced with the portable `ssl-passthrough` approach to keep the chart controller-agnostic.

Verified with `helm template` (http/grpc; standalone + all-in-one) and `helm lint`.

## Summary by CodeRabbit

* **New Features**
  * Added separate Filer ingress settings for HTTP and gRPC, allowing independent exposure and configuration.
  * You can now enable/config HTTP and gRPC routes separately (host, path, TLS, and annotations).

* **Changes**
  * Updated the chart configuration structure to use distinct `http` and `grpc` ingress sections.
  * HTTP and gRPC ingresses are created as separate resources with distinct names (HTTP UI vs gRPC endpoint).

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer/filer-ingress.yaml
+++ b/k8s/charts/seaweedfs/templates/filer/filer-ingress.yaml
@@ -1,6 +1,8 @@
-{{- /* Filer ingress works for both normal mode (filer.enabled) and all-in-one mode (allInOne.enabled) */}}
+{{- /* Filer ingresses work for both normal mode (filer.enabled) and all-in-one mode (allInOne.enabled) */}}
 {{- $filerEnabled := or .Values.filer.enabled .Values.allInOne.enabled }}
-{{- if and $filerEnabled .Values.filer.ingress.enabled }}
+
+{{- /* HTTP Ingress */}}
+{{- if and $filerEnabled .Values.filer.ingresses.http.enabled }}
 {{- /* Determine service name based on deployment mode */}}
 {{- $serviceName := ternary (include "seaweedfs.componentName" (list . "all-in-one")) (include "seaweedfs.componentName" (list . "filer")) .Values.allInOne.enabled }}
 {{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion }}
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed in 98f317e and 5640cad3:

- `ingressClassName` now renders only when a class is set — an empty value was opting the Ingress out of the cluster's default IngressClass. Added the `kubernetes.io/ingress.class` fallback for k8s <1.18 and version-gated `pathType`, so the filer ingress now matches the `volume`/`admin` templates.
- `host` is quoted, so wildcard hosts like `*.example.com` stay valid YAML.
- gRPC ingress path defaults to `/` with `pathType: Prefix`; gRPC methods are called at `/<package>.<Service>/<Method>`, which the HTTP UI's regex path never matches.

The `IngressRouteTCP` route-dedup nit is moot — that Traefik CRD was dropped. The mTLS-passthrough case is now documented via the nginx `ssl-passthrough` annotation on the gRPC ingress, keeping the whole chart on the standard `Ingress` kind with no controller-specific CRD.

Verified with `helm template` (empty/set className, wildcard host, k8s 1.16 and 1.19+, standalone + all-in-one) and `helm lint`.

MorezMartin:
Thanks @chrislusf 
Problematic for traefik users because Ingress only route HTTP / HTTPS, tls passthrough need IngressRouteTCP, apparently nginx has a workaround but traefik has not.
So not controller agnostic to my opinion because mTLS will be broken using traefik.
It is possible but redundant and heavy tu use https with ServersTransport. I can work on a PR if needed.
What is the way ?
Thanks in advance for your answer,
Sincerely yours,
Martin

PR #10208, opened 2026-07-02
fix(fuse-tests): don't declare the FUSE mount ready before it is mounted

chrislusf:
## What

`waitForMount` probed the mount point with stat+ReadDir — checks a bare, not-yet-mounted local directory also passes — so `Setup` could return before the `weed mount` process finished mounting. It now additionally requires the mount point's device ID to differ from its parent directory's, which confirms a filesystem is actually mounted there, and keeps the ReadDir probe as a liveness check.

## Why

A FUSE integration CI run caught the long-standing `TestConcurrentFileOperations/ConcurrentReadWrite` ENOENT flake with the diagnostics from 6cbcdf488 in place, and the captured timeline shows the race directly (cold runner, slow mount startup):

- 19:18:21.15 — first two subtests run and "pass" (writing to the local disk underneath the mount point)
- 19:18:21.17 — ConcurrentReadWrite creates its file the same way; last write ~21.37
- 19:18:21.383 — `mount_std.go: mounted` (per the mount's own glog timestamps) — the mount activates and shadows everything written so far
- 19:18:21.43 — AssertFileExists goes through FUSE: ENOENT through 1.5s of retries, ReadDir of the mount root returns 0 entries, and the lookupEntry diagnostic never fired — the mount never saw any of those files

This also explains the original flake signature: stat returning ENOENT for a path all writers had just succeeded against.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mount readiness checks to ensure FUSE mounts are fully active before tests proceed.
  * Reduced false-positive readiness detection by verifying the mount point is a real filesystem mount, not just an accessible directory.

### DIFF
--- a/test/fuse_integration/framework_test.go
+++ b/test/fuse_integration/framework_test.go
@@ -333,11 +333,19 @@ func (f *FuseTestFramework) waitForService(addr string, timeout time.Duration) e
 	return fmt.Errorf("service at %s not ready within timeout", addr)
 }
 
-// waitForMount waits for the FUSE mount to be ready
+// waitForMount waits for the FUSE mount to be ready. A stat/ReadDir probe
+// alone is not enough: the bare mount point directory passes both before the
+// mount process finishes starting, letting tests race ahead and write to the
+// local disk underneath the mount. The mount point's device ID differing from
+// its parent's confirms a filesystem is actually mounted there.
 func (f *FuseTestFramework) waitForMount(timeout time.Duration) error {
+	parentDev, err := deviceID(filepath.Dir(f.mountPoint))
+	if err != nil {
+		return fmt.Errorf("stat mount point parent: %v", err)
+	}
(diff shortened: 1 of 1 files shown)


chrislusf:
Follow-up: running the enterprise CI with this fix exposed two pre-existing bugs the old probe had been masking (tests were silently running against the local disk before the mount came up):

1. TestWriteBufferCap passed `-logtostderr=true -v=2` in MountOptions — glog globals the `weed mount` subcommand flag set rejects, so its mount had always died with a usage error. Dropped in 1fe294f.
2. Test teardowns took ~28s each: `util.Conn`'s activity-timeout extension clobbers the deadline `net/http` sets in `abortPendingRead`, wedging keep-alive connections and making the volume server's graceful HTTP drain wait out its full 30s StopTimeout on every shutdown. Fixed separately in #10212.

With both, the full suite passes on Linux with real mounts throughout.

Issue #9619, opened 2026-05-22
volume id not found yet is present in admin ui and volume list


jrcichra:
I just upgraded from 4.17 to 4.27/4.28 and noticed some files were 404'ing that weren't before:

```
weed-exodus-0 weed E0522 03:56:01.024983 s3api_object_handlers.go:874 GetObjectHandler: failed to stream drr/metadata/Mon Feb  2 15:36:31 EST 2026_20260202_160137_83_conf_0.06.flac.json from volume servers: operation LookupFileId 18994,3c466b38d85377 failed, err: volume 18994 not found for fileId 18994,3c466b38d85377: provider lookup failed: all 3 filer(s) failed with non-retryable error: filer.LookupVolume failed: rpc error: code = Unknown desc = provider lookup failed: master volume lookup errors: volume 18994: volume id 18994 not found
weed-exodus-0 weed volume 18994 not found
weed-exodus-0 weed volume 18994 not found
```

But yet this volume is clearly known by the system:

<img width="2104" height="818" alt="Image" src="https://github.com/user-attachments/assets/97b8de2d-6f43-4180-89ef-4444c8aa8bff" />

Also in the shell:
```
> volume.list -volumeId 18994
Topology volumeSizeLimit:1024 MB hdd(volume:14774/16386 active:14767 free:1500 remote:0)
  DataCenter church hdd(volume:14774/16386 active:14767 free:1500 remote:0)
    Rack server-room-1 hdd(volume:14774/16386 active:14767 free:1500 remote:0)
      DataNode weed-exodus-0.weed-master.seaweedfs.svc.cluster.local:8080 hdd(volume:7766/7809 active:7763 free:0 remote:0)
        Disk hdd(volume:7766/7809 active:7763 free:0 remote:0) id:0
          volume Id:18994, Size:1077696856, ReplicaPlacement:001, Collection:drr, Version:3, Ttl:, FileCount:7997, DeleteCount:60, DeletedByteCount:1037402, ReadOnly:false, ModifiedAtSecond:1778815572
        Disk hdd {Size:1077696856 FileCount:7997 DeletedFileCount:60 DeletedBytes:1037402}
      DataNode weed-exodus-0.weed-master.seaweedfs.svc.cluster.local:8080 {Size:1077696856 FileCount:7997 DeletedFileCount:60 DeletedBytes:1037402}
[...]

PR #10226, opened 2026-07-03
master: repair the lookup index after a vacuum that raced a disconnect

chrislusf:
A volume present on a data node but missing from the master `vid2location` shows in volume.list/admin UI yet fails LookupVolume with "volume id not found" — the split #9620 self-heals on the next full heartbeat. A disconnect during a long vacuum can re-open that split, and two repair paths missed it:

- `SetVolumeAvailable` (vacuum commit) dereferenced `vid2location[vid]` with no nil check, so committing a single-replica volume the index had lost panicked the master instead of re-registering it.
- `VolumeMarkReadonly` (the maintenance worker marking a volume writable after commit) only touched nodes already in the index, so it could not re-register a dropped volume and left it "not found" until the next heartbeat healed it.

Both now re-create the entry from the node that still holds the volume. Added `TestSetVolumeAvailableRepairsMissingVolume`, which panics without the first fix.

Relates to #9619.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume availability recovery to repair index/location splits after temporary disconnects and long vacuum/maintenance scenarios.
  * Prevented nil dereferences when marking a volume available if its in-memory location data was missing.
  * Improved readonly/writable state handling so updates remain consistent even when the target node isn’t found during the request.
* **Tests**
  * Added coverage for the “missing volume” recovery flow to help prevent regressions.

### DIFF
--- a/weed/server/master_grpc_server_volume.go
+++ b/weed/server/master_grpc_server_volume.go
@@ -15,6 +15,7 @@ import (
 	"github.com/seaweedfs/raft"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/security"
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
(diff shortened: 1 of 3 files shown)


chrislusf:

- SetVolumeAvailable now seeds size tracking when it re-creates a dropped entry, so assigns are counted immediately instead of only after the next heartbeat (was the untracked-write/overfill gap Gemini flagged).
- Extracted the get-or-create into `getOrCreateLocationList` (and `initSizeTracking`), shared by RegisterVolume and SetVolumeAvailable.
- Added a `glog.V(0)` line when the mark-writable path re-registers a missing volume, so the rare repair is visible in production logs.
- Test now also asserts size tracking is seeded after the repair.

PR #10229, opened 2026-07-04
volume: bound intra-cluster HTTP so an unresponsive peer can't hang reads and writes

chrislusf:
Follow-up to the second scenario in discussion #10206: after a stopped volume server is **started back up**, an active FUSE mount's write and read both stall.

## Root cause

The shared intra-cluster HTTP transport (`weed/util/http/client`) had a 10s **dial** timeout but **no response timeout**. When a volume server is TCP-reachable but not answering, the client blocks forever waiting for the response header. That happens exactly during a restart:

- the volume server accepts connections while still loading its volumes, but doesn't serve yet, and
- in Docker Swarm the server returns on a new container IP while clients still hold a keep-alive connection to the old, now-blackholed IP.

A chunk **read** never fell over to the healthy replica, and a **replicated write** (primary → peer) never failed, so the client's upload hung too. The earlier scenario-1 fix only covered hard failures (DNS / connection refused), which surface as immediate errors; "connected but silent" does not.

## Fix

Add `ResponseHeaderTimeout` (30s) and `IdleConnTimeout` (90s) to the transport. Intra-cluster peers answer headers within milliseconds, so this only ever fires on a genuinely stuck peer: the read then fails over to another replica, the replicated write fails fast for the client to retry, and stale pooled sockets to a departed server are evicted. The timeout is a ceiling, not added latency — a request completes the moment the peer becomes responsive again.

## Reproduced and verified

With a 1-master / 3-volume (replication 001) / filer / 2-FUSE-mount cluster, freezing a volume server with `SIGSTOP` (TCP-alive but unresponsive — the restart/loading state):

- **before**: read of a chunk on the frozen server and a replicated write to it both hung indefinitely.
- **after**: the read fails over to the healthy replica and returns correct data; the replicated write returns a prompt error to retry. In a live mount write+read with a 10s freeze mid-stream, the one in-flight operation recovered the instant the server responded again and the file was byte-identical across both mounts.

## Summary by CodeRabbit

* **Bug Fixes**
  * HTTP clients now enforce explicit timeouts when waiting for response headers, avoiding hangs from unresponsive peers.
  * Idle HTTP connections are now cleaned up more consistently after the configured idle period (including TLS clients).
* **New Features**
  * Added a new HTTP client option to configure the response-header timeout.
* **Tests**
  * Expanded coverage for default timeout behavior, option overrides, and timeout failure handling for both standard and TLS HTTP clients.

### DIFF
--- a/weed/util/http/client/http_client.go
+++ b/weed/util/http/client/http_client.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"google.golang.org/grpc/credentials/tls/certprovider"
 
@@ -20,6 +21,17 @@ import (
 
 var (
 	loadSecurityConfigOnce sync.Once
+)
+
+// Intra-cluster peers (volume/filer/master) answer headers within milliseconds.
+// A peer that is TCP-reachable but not answering -- a volume server still loading
+// after a restart, or a stale keep-alive to a container that returned on a new IP
+// -- otherwise blocks a chunk read or a replicated write forever. The response
+// timeout bounds that wait so the read fails over to another replica and the
+// write fails fast to retry; the idle timeout evicts sockets to a departed server.
+const (
+	responseHeaderTimeout = 30 * time.Second
+	idleConnTimeout       = 90 * time.Second
 )
 
 type HTTPClient struct {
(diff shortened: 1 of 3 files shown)


chrislusf:
Made the defaults `const` and added a `WithResponseHeaderTimeout` option so the test configures the client instead of mutating a package global. Tests pass under `-race`.

chrislusf:
Tightened the unresponsive-peer test to assert the error is a `net.Error` timeout, and added the same coverage over the `NewHttpClientWithTLS` path.

PR #10249, opened 2026-07-06
filer: require JWT authorization on TUS upload endpoints

chrislusf:
## Summary

The filer's TUS resumable-upload endpoints weren't covered by the JWT check the rest of the filer API goes through. `filerHandler` and `readonlyFilerHandler` run every request through `maybeCheckJwtAuthorization` before dispatching, but the TUS routes were registered only behind `filerGuard.WhiteList(...)`, which is a pass-through for the filer (the guard is built with an empty whitelist). So with `jwt.filer_signing.key` configured, normal `PUT`/`POST`/`DELETE` went through the JWT check while the TUS routes did not.

This brings the TUS endpoints under the same JWT enforcement and the same read-only / WORM handling as the normal filer write path.

## Changes

- Run the filer JWT check in `tusHandler` before routing. `HEAD` uses the read key; `POST`/`PATCH`/`DELETE` use the write key; `OPTIONS` stays open for capability discovery, matching the main filer handler.
- Scope prefix-restricted tokens against the resolved target path (request URL minus the `/.tus` prefix, leading slash preserved) at session creation.
- Route TUS session creation and completion through the same read-only / WORM checks the normal write path applies, and surface those rejections with the same status codes (507 / 403).

## Summary by CodeRabbit

* **New Features**
  * TUS uploads now require JWT authorization for non-`OPTIONS` requests, improving access control during resumable uploads.
  * Upload completion now respects read-only and WORM protections more consistently.

* **Bug Fixes**
  * Fixed TUS upload handling to return clearer authorization and completion errors instead of generic failures.
  * Improved method- and path-based JWT checks for TUS requests, including better handling of scoped upload paths.

### DIFF
--- a/weed/server/filer_server_handlers.go
+++ b/weed/server/filer_server_handlers.go
@@ -216,6 +216,16 @@ func (fs *FilerServer) maybeCheckJwtAuthorization(r *http.Request, isWrite bool)
 		return true
 	}
 
+	return fs.checkJwtAuthorization(r, isWrite, jwtScopedRequestPaths(r))
+}
+
+// checkJwtAuthorization verifies the request carries a valid filer JWT for the
+// requested access level and, for prefix-restricted tokens, that every path in
+// scopedPaths falls within the token's AllowedPrefixes. Callers whose write
+// target is not r.URL.Path — the TUS handler, whose URL points at the /.tus
+// route — pass the resolved filer path(s) here instead.
+func (fs *FilerServer) checkJwtAuthorization(r *http.Request, isWrite bool, scopedPaths []string) bool {
+
 	var signingKey security.SigningKey
 
 	if isWrite {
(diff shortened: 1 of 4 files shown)


chrislusf:

- **Leading slash on the resolved target** (gemini): added `tusTargetPath`, used in both `checkTusJwtAuthorization` and `tusCreateHandler`, so the target keeps a leading slash and matches stored absolute paths / `AllowedPrefixes` even if `TusBasePath` were set with a trailing slash. (In practice it's normalized in `filer_server.go`, but this makes the prefix scoping robust regardless.)
- **Error status mapping** (coderabbit): `completeTusUpload` read-only/WORM rejections now surface as 507 and 403 via `writeTusCompleteError`, instead of a generic 500. Added an `ErrWormEnforced` sentinel so the handler can `errors.Is` it.
- **Early read-only rejection**: `tusCreateHandler` now rejects a read-only prefix up front, before creating a session or uploading chunks, so a rejected upload leaves no orphaned chunks.
- **Tests**: added method-restricted (`AllowedMethods`) cases to the TUS auth table.

PR #10254, opened 2026-07-07
restore: weed nfs gateway (NFSv3) with opt-in inode index scoping

Aelindra:
## What

Restores the `weed nfs` (NFSv3 gateway, built on `github.com/willscott/go-nfs`) that was removed in #9724, with an **opt-in inode index scoping** flag to address the index bloat, plus fixes that make the gateway fully functional out-of-the-box.

## Why #9724 removed it

`weed nfs` needs an inode-to-path reverse-lookup to decode fixed-length NFS file handles. The original wrote one KV row per filer entry on every create/update/delete, regardless of NFS usage — reaching ~1.1B rows / 30GB on S3-heavy deployments. Since the index only benefits NFS, it was removed.

## Changes (3 commits)

**1. Restore the gateway** from 4.29, adapted to current APIs (WithGrpcClient ctx arg, NewReaderCache invalidator, UploadWithRetry callback removal). No change to NFS protocol logic or filehandle format.

**2. Opt-in index scoping** — filer flag `-nfs.inodeIndexPrefixes=/exports`:
- Only entries under the configured prefixes get an index row
- Default empty = index disabled = zero KV rows = identical to current master
- Directory-boundary matching (`/exports` matches `/exports/x` not `/exportsx`)

**3. Bug fixes from real kernel-client testing**:
- `lookupGeneration` fell back to initial generation when index absent (mount no longer hard-fails)
- `mountproto` hint corrected to `udp` (MOUNT responder is UDP-only)
- **`weed nfs` now self-manages its index via the filer KV RPC** — it writes the inode→path row itself on create, so the gateway is fully functional with default filer config (no `-nfs.inodeIndexPrefixes` needed for a working export). This is additive: when the filer flag IS set, both writers produce byte-identical records and converge.

## Verification

**Unit**: `go build ./weed/...` passes (CGO_ENABLED=0); `go test ./weed/server/nfs/...` 64+2 tests pass; `go test ./weed/filer/ -run TestInodeIndex` both scoping tests pass.

**Real kernel-client E2E** (privileged container, Linux 5.10 NFSv3 client, live single-node cluster):

Default filer config (no index flag):
- `mount -t nfs -o nfsvers=3,nolock,noacl,port=2049,mountport=2049,proto=tcp,mountproto=udp` — mount OK exit 0
- Write + read back — content matches, VERIFY OK
- mkdir + nested write — OK
- `dmesg | grep nfs` — zero kernel errors

With `-nfs.inodeIndexPrefixes=/exports`:
- All operations pass — no regression

The 4.29-era kernel-client interop bugs do NOT reproduce.

## Non-goals

NFSv3 + MOUNTv3 only; no NLM/NSM (mount with `-o nolock`); AUTH_SYS only; no pNFS.

## For maintainers

Default config is fully functional. The opt-in flag adds filer-side index writes for deployments wanting broader index coverage; nfs self-manages index for its own export subtree regardless. Index bloat is bounded by NFS export volume, not total cluster objects.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -81,6 +81,7 @@ require (
 	github.com/tylertreat/BoomFilters v0.0.0-20210315201527-1a82519a3e43
 	github.com/valyala/bytebufferpool v1.0.0
 	github.com/viant/ptrie v1.0.1
+	github.com/willscott/go-nfs v0.0.4
 	github.com/xdg-go/pbkdf2 v1.0.0 // indirect
 	github.com/xdg-go/scram v1.2.0
 	github.com/xdg-go/stringprep v1.0.4 // indirect
@@ -163,7 +164,11 @@ require (
 	google.golang.org/grpc/security/advancedtls v1.0.0
 )
 
-require github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
+require (
+	github.com/k0kubun/colorstring v0.0.0-20150214042306-9440f1994b88 // indirect
+	github.com/rasky/go-xdr v0.0.0-20170124162913-1a41d1a06c93 // indirect
+	github.com/willscott/go-nfs-client v0.0.0-20251022144359-801f10d98886 // indirect
+)
 
 require (
 	atomicgo.dev/cursor v0.2.0 // indirect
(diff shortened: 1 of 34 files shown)


chrislusf:
Thanks for putting this together — most of the restore comes back byte-identical to the pre-removal tree, which made it straightforward to review.

The one part I want to land differently is the `-nfs.inodeIndexPrefixes` filer flag: a process-global flag has to be kept identical across every filer in a cluster, and a missed one silently drops index rows. #10277 adds the same scoping through filer.conf instead — an `inode_index` location rule plus shell `nfs.enable` / `nfs.disable` that backfill and clean up the index — so filers pick it up without restarts and there is no flag to coordinate.

Once that lands, the gateway restore can drop the flag and the self-managed index writes, and check filer.conf at startup instead. I'll follow up on a rebased branch keeping your commits.

Aelindra:
Hi @chrislusf, thanks for the review and for opening #10277 — the `filer.conf` + `nfs.enable`/`nfs.disable` approach is clearly the right call for cluster-wide consistency.

Happy to drop the flag and the self-managed index writes. Since the filer-side `inode_index` mechanism lands via #10277, I'll leave my branch as-is so your rebase stays clean — please go ahead and carry the gateway restore on top of #10277. Happy to help if you need anything from my side.

Issue #10257, opened 2026-07-08
[help] Regarding Nginx Basic Auth issues~~


MamyAKKK:
Does the gRPC library of [seaweedfs-client] in java support the Basic Auth setting of Nginx?
I have a seaweedfs service that is placed behind Nginx, and Nginx has Basic Auth settings configured.
So......What should I do, please~

chrislusf:
Not out of the box before, but I just added it in #10258. The Java client now sends an HTTP Basic Auth header on both the filer gRPC channel and the volume HTTP requests, so it can go through an Nginx `auth_basic` proxy.

Set it in `security.toml`:

```toml
[basic_auth]
username = "myuser"
password = "mypass"
```

or in code before you create the client:

```java
FilerSecurityContext.setBasicAuth("myuser", "mypass");
```

One caveat: a volume upload already uses the Authorization header for its JWT, and Basic Auth and Bearer can't share the same header. So put Nginx in front of the filer, and access the volume servers directly (or with the volume JWT disabled).

MamyAKKK:
> Not out of the box before, but I just added it in [#10258](https://github.com/seaweedfs/seaweedfs/pull/10258). The Java client now sends an HTTP Basic Auth header on both the filer gRPC channel and the volume HTTP requests, so it can go through an Nginx `auth_basic` proxy.
> 
> Set it in `security.toml`:
> 
> [basic_auth]
> username = "myuser"
> password = "mypass"
> or in code before you create the client:
> 
> FilerSecurityContext.setBasicAuth("myuser", "mypass");
> One caveat: a volume upload already uses the Authorization header for its JWT, and Basic Auth and Bearer can't share the same header. So put Nginx in front of the filer, and access the volume servers directly (or with the volume JWT disabled).

------
This is incredibly useful to me. Thank you very much~

PR #10258, opened 2026-07-08
java: HTTP Basic Auth for filer behind an Nginx reverse proxy

chrislusf:
The Java client couldn't reach a filer fronted by Nginx with `auth_basic`, since it never sent an Authorization header on the gRPC channel or on the volume HTTP requests.

Add optional HTTP Basic Auth:
- gRPC: a `BasicAuthInterceptor` puts `Authorization: Basic <base64(user:pass)>` on every filer call.
- HTTP: the same header goes on the volume/proxy reads and writes.

Configure it in `security.toml`:

```toml
[basic_auth]
username = "myuser"
password = "mypass"
```

or programmatically before constructing the client:

```java
FilerSecurityContext.setBasicAuth("myuser", "mypass");
```

Nothing is sent when unset, so existing setups are untouched.

Caveat: a volume upload already uses the Authorization header for its JWT, and the two can't share it. When the volume JWT is present it keeps the header, so front the filer (gRPC + metadata/proxy) with Basic Auth and access volumes directly or with the volume JWT disabled.

Addresses #10257.

## Summary by CodeRabbit

* **New Features**
  * Added optional Basic Auth support for client requests, configurable via settings or a programmatic API.
  * Outgoing gRPC and relevant HTTP requests can now automatically include an `Authorization` header when Basic Auth is enabled.
  * Added a one-time warning when Basic Auth would be sent over an unencrypted connection.

* **Bug Fixes**
  * Improved authentication selection so Bearer auth still takes precedence, while Basic Auth is used when Bearer credentials are absent.

* **Tests**
  * Added tests covering enable/disable behavior, Basic Auth encoding, and header injection into outgoing requests.

### DIFF
--- a/other/java/client/src/main/java/seaweedfs/client/BasicAuthInterceptor.java
+++ b/other/java/client/src/main/java/seaweedfs/client/BasicAuthInterceptor.java
@@ -1,0 +1,35 @@
+package seaweedfs.client;
+
+import io.grpc.CallOptions;
+import io.grpc.Channel;
+import io.grpc.ClientCall;
+import io.grpc.ClientInterceptor;
+import io.grpc.ForwardingClientCall;
+import io.grpc.Metadata;
+import io.grpc.MethodDescriptor;
+
+// Adds an HTTP Basic Auth "Authorization" header to every gRPC call so the filer can be
+// reached through an Nginx reverse proxy with auth_basic enabled.
+public class BasicAuthInterceptor implements ClientInterceptor {
+
+    private static final Metadata.Key<String> AUTHORIZATION_KEY =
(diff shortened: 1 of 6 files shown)


chrislusf:

- Precompute the Base64 `Authorization` value once (in `computeBasicAuthHeaderValue`, set from `security.toml` and `setBasicAuth`) instead of re-encoding on every chunk read/write. Dropped the now-unused `basicAuthUsername`/`basicAuthPassword` fields.
- Log a one-time warning when Basic Auth is used on a plaintext gRPC channel, since the credentials would travel in cleartext there.
- Extended `BasicAuthTest` to cover the empty-username disabled case.

Left the PR description as-is (no test checklist) to match the repo's usual style.

PR #10263, opened 2026-07-08
iam: surface OIDC groups and roles into the STS session request context for resource-policy ABAC

chriswydra:

OIDC `groups` and `roles` claims are surfaced only into role **trust** policies (assume-time), never into **resource/permission** policies (request-time). So a single STS role cannot scope access by the caller's group membership or roles - group/role-based ABAC in a resource policy silently denies.

Root cause: in `oidc_provider.go`, both `groups` and `roles` are in the `processedClaims` exclusion set, so they go into `ExternalIdentity.Groups` / are dropped, but **not** into `Attributes`. `AssumeRoleWithWebIdentity` builds the session `RequestContext` from `Attributes` (+ a few scalars), so neither reaches the request context. Single-valued claims like `preferred_username` work as `jwt:preferred_username`; `groups`/`roles` match nothing.

The evaluator itself is fine - `EvaluateStringCondition` already handles multi-valued (`[]interface{}` / `[]string`) context keys and `ForAnyValue:`/`ForAllValues:`. The arrays just never arrive.

Surface `ExternalIdentity.Groups` and the token's `roles` claim as `[]string` in the session `RequestContext` in `AssumeRoleWithWebIdentity` (via a new `ExternalIdentity.Roles` field). The S3 middleware already exposes unprefixed claims additionally as `jwt:<key>`, so a resource policy can then do:

```json
"Condition": { "ForAnyValue:StringEquals": { "jwt:roles": "team-a-rw" } }
```
or the equivalent on `jwt:groups`.

Minimal, additive; only affects the advanced-IAM/STS path. `groups`/`roles` were already extracted, so no new parsing.

# How is the PR tested?

End-to-end against a live Keycloak (OIDC device-flow) + `weed server -s3 -s3.iam.config`:

- One role; permission policy with per-bucket statements each gated on `ForAnyValue:StringEquals` on `jwt:roles` (and separately validated on `jwt:groups`).
- A user holding the realm role / group for a bucket gets read/write on it from a single AssumeRoleWithWebIdentity; a user without it is denied - and a user in two groups/roles gets both buckets in one session.

Before this change the claims were absent from the request context and every such request was denied; after, access matches the caller's groups/roles. Trust-policy `oidc:groups` conditions and single-valued `jwt:` conditions are unaffected.

# Checks
- [ ] I have added unit tests if possible. (The request-context assembly is inline in `AssumeRoleWithWebIdentity`; kept the change minimal rather than refactoring for a unit seam - happy to extract a helper + test if preferred.)
- [x] The PR is kept as minimum as possible.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved web identity login session claim handling by carrying federated group memberships into request-time policy evaluation.
  * Added support for role claims from the authenticated token, making role data available for permission checks (not just trust-policy/logging).
  * As a result, session context is more complete, improving access evaluation accuracy.


chrislusf:
Added an integration test for this, but GitHub does not allow maintainer pushes to organization-owned forks, so I could not push it onto this branch directly. Please cherry-pick:

```
git fetch https://github.com/seaweedfs/seaweedfs.git pr-10263-abac-test
git cherry-pick c1ac4a311d8298d15db1def2d4af187c30e1f81d
```

Commit: https://github.com/seaweedfs/seaweedfs/commit/c1ac4a311d8298d15db1def2d4af187c30e1f81d

It drives the full path: AssumeRoleWithWebIdentity embeds the claims in the session JWT, AuthenticateJWT restores them, and AuthorizeAction evaluates ForAnyValue:StringEquals on jwt:groups / jwt:roles per bucket, with a second test pinning the []interface{} round-trip shape in the session request context. The mock OIDC provider now surfaces roles and carries token claims through, matching the real provider. The tests fail without the sts_service.go change.

Issue #10272, opened 2026-07-08
SeaweedFS Feature Request: Per-Volume Parallelism in `volume.fix.replication` for Large-Capacity HDD Deployments


qqren:
# SeaweedFS Feature Request: Per-Volume Parallelism in `volume.fix.replication` for Large-Capacity HDD Deployments

> Draft GitHub issue targeting https://github.com/seaweedfs/seaweedfs
> Related: closed PR #6789 (partial parallelization), stalled PR #6422 (volume-level parallelization attempt)

---

## Title

`volume.fix.replication`: add per-volume parallelism for faster recovery on large-capacity HDD deployments

---

### What problem are we solving?

On a modern volume server with large-capacity HDDs (18 TB / 22 TB / 26 TB per disk, JBOD of 12+ disks), a single node routinely hosts **hundreds of 32 GB volumes**. When one such node goes down and cannot be brought back quickly, every volume that lived on it becomes under-replicated. `volume.fix.replication` is the operator-facing tool used to recover from this state, but today its recovery throughput is bounded by **serial per-volume replication**, which creates an unacceptably long service-degraded window on HDD-heavy clusters.

#### Current code path (verified against `master`)

`weed/shell/command_volume_fix_replication.go`:

```go
// fixUnderReplicatedVolumes -- L349
for _, vid := range volumeIds {
    for i := 0; i < retryCount+1; i++ {
        if copied, err = c.fixOneUnderReplicatedVolume(...); err == nil {
            ...
            break
        }
    }
}
```

Inside `fixOneUnderReplicatedVolume` → `replicateVolumeToServer` (`weed/shell/command_volume_move.go` L250) is a **blocking gRPC `VolumeCopy` stream** that pulls the entire `.dat` + `.idx` from the source volume server before returning. So each iteration of the outer `for` blocks until one full 32 GB volume has been streamed to a new server.

PR #6789 (merged 2025-07) introduced `maxParallelization`, but it only parallelizes the **three top-level phases** (fix-under-replicated / delete-over-replicated / delete-misplaced) via `ewg.Add(...)`. The **inside** of `fixUnderReplicatedVolumes` was left serial:

```go
ewg.Add(func() error {
    fixedVolumeReplicas, err = c.fixUnderReplicatedVolumes(...)  // single goroutine drives the whole slice
    return err
})
```

[...]

PR #10275, opened 2026-07-08
volume.fix.replication: parallelize under-replicated volume copies

chrislusf:
## What

Fixing under-replicated volumes now copies up to `-maxParallelization` volumes concurrently (the flag from #6789, which so far only parallelized the three top-level phases). Addresses #10272.

## How the concurrent target-selection problem is handled

#6422 stalled because concurrent fixes would all pick the same emptiest server. Two mechanisms address it:

- **Atomic reservation**: destination selection, sorting, and free-slot accounting run under one scheduler mutex, so every selection sees prior reservations. A failed copy returns its reserved slot, so retries don't drain the free-slot accounting.
- **Per-server copy cap**: reserving one slot doesn't stop a large empty server from staying "emptiest", so a slot-accounting fix alone still funnels every concurrent copy to one destination. `-maxParallelizationPerServer` (default 1) caps in-flight copies per destination; a capped server is skipped for the next-best candidate, which spreads placement, and when *every* eligible destination is at the cap the fix waits for a slot instead of failing.

## Notes

- `-maxParallelization=1` runs tasks inline and preserves the exact previous serial behavior.
- Simulation mode (no `-apply`) reserves slots exactly as before and never waits on the copy cap.
- Recovery of a failed node with hundreds of volumes drops from hours (serial full-volume copies) to roughly serial-time ÷ min(maxParallelization, eligible servers × per-server cap).

Tests cover reservation spread, cap blocking/wake-up, failed-copy slot return, and a parallel simulation run that exactly fills a constrained topology; `go test -race ./weed/shell/` passes.

## Summary by CodeRabbit

* **New Features**
  * Added a `-maxParallelizationPerServer` option to limit concurrent volume copy work per destination server.
  * Updated replication-fix behavior to better distribute under-replication fixes across eligible servers while honoring per-server copy caps.

* **Bug Fixes**
  * Prevents overload by enforcing per-destination-server capacity limits for concurrent copies.
  * Improved backpressure so additional work waits when all eligible destinations are at capacity, then resumes when slots free up.

* **Tests**
  * Added unit coverage for reservation spreading, blocking/wake-up behavior, and parallel under-replication fixing accounting.

### DIFF
--- a/weed/shell/command_volume_fix_replication.go
+++ b/weed/shell/command_volume_fix_replication.go
@@ -351,6 +351,11 @@ func (c *commandVolumeFixReplication) fixUnderReplicatedVolumes(commandEnv *Comm
 		volumeIds = volumeIds[0:volumesPerStep]
 	}
 
+	// own a private copy of the locations list: the scheduler re-sorts it on
+	// every reservation, and the caller's slice is shared with the concurrent
+	// delete phases
+	allLocations = slices.Clone(allLocations)
+
 	scheduler := newVolumeCopyScheduler(maxParallelizationPerServer)
 	var fixedVolumesMu sync.Mutex
 	ewg := NewErrorWaitGroup(maxParallelization)
--- a/weed/shell/command_volume_fix_replication.go
+++ b/weed/shell/command_volume_fix_replication.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"slices"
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed review feedback in c397fe6:
- `fixUnderReplicatedVolumes` now clones `allLocations` before the fan-out, so the scheduler owns the ordering it re-sorts. (Every read/sort already went through the scheduler mutex and the delete phases never read the slice, but exclusive ownership makes that a non-question.)
- Made the test's per-iteration `location` copy explicit (`loc := loc`); the map intentionally holds copies whose shared `dataNode` pointers carry the accounting.

PR #10276, opened 2026-07-08
topology/balancer: share replica selection between shell and workers

chrislusf:
## What

Moves the replica-selection logic out of `weed/shell` into `weed/topology/balancer`, following the earlier `SatisfyReplicaPlacement` extraction:

- `PickOneReplicaToCopyFrom` — most recently modified source
- `PickOneReplicaToDelete` — over-replication trim with the writable-survivor guard
- `PickOneMisplacedVolume` — relocation pick (still acts on read-only replicas)
- `IsMisplaced`
- `SatisfyReplicaCurrentLocation` — the under-replicated-by-domain check

Selection functions take a shared `balancer.Replica` shape (location + size/mtime/compact-revision/read-only) and return an index into the caller's slice, so the shell keeps its own richer replica type and maintenance workers can adapt from ActiveTopology without duplicating the safety rules.

## Why

Groundwork for a maintenance worker that repairs under-replicated volumes automatically: the worker needs the same source/trim/misplaced selection semantics as `volume.fix.replication`, and these rules (especially the writable-survivor guard) should have exactly one implementation.

Pure refactor, no behavior change; the existing shell tests cover the moved logic through the thin adapters. `go test -race ./weed/shell/ ./weed/topology/balancer/` passes.

Note: originally stacked on #10275, now rebased directly onto master and independent of it.

## Summary by CodeRabbit

* **New Features**
  * Improved replica placement handling for volumes during replication and balancing.
  * Added smarter selection of which replica to copy from, remove, or recreate when fixing placement issues.

* **Bug Fixes**
  * Better respects rack and data center spread rules when checking whether replicas are correctly placed.
  * Improves trimming decisions so over-replicated volumes keep a valid, well-placed set of replicas.
  * Makes misplaced-volume recovery more reliable and consistent.

### DIFF
--- a/weed/shell/command_volume_fix_replication.go
+++ b/weed/shell/command_volume_fix_replication.go
@@ -499,24 +499,11 @@ func keepDataNodesSorted(dataNodes []location, diskType types.DiskType) {
 }
 
 func satisfyReplicaCurrentLocation(replicaPlacement *super_block.ReplicaPlacement, replicas []*VolumeReplica) bool {
-	existingDataCenters, existingRacks, _ := countReplicas(replicas)
-
-	if replicaPlacement.DiffDataCenterCount+1 > len(existingDataCenters) {
-		return false
+	locs := make([]balancer.Location, len(replicas))
+	for i, r := range replicas {
+		locs[i] = toBalancerLocation(r.location)
 	}
-	if replicaPlacement.DiffRackCount+1 > len(existingRacks) {
-		return false
-	}
-	if replicaPlacement.SameRackCount > 0 {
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed review feedback in 7a8b932 (rebased on the updated #10275):
- `PickOneReplicaToDelete` filters writable replicas and early-returns before any placement checks, restoring the original short-circuit; placement is then evaluated only for writable candidates.
- Skipped the fully allocation-free rewrite: the index slices are tiny and this runs once per over-replicated volume in an operator command, not on a hot path.

PR #10277, opened 2026-07-08
filer: opt-in inode->path index with shell nfs.enable / nfs.disable

chrislusf:
Groundwork for bringing back the `weed nfs` gateway. The gateway resolves NFS filehandles by reversing an inode to its path, which the deterministic hash inode cannot do, so the filer needs a reverse index — but nobody should pay for it unless they export something.

- filer.conf `PathConf` gains an `inode_index` option. The store wrapper writes inode->path rows only for entries under a marked location; with no rule (the default) no index work happens at all. Every filer picks the rule up through the normal filer.conf reload, so there is no per-process flag to keep in sync across a cluster.
- `nfs.enable -path=/exports -apply` marks the path and backfills rows for the entries already under it, assigning inodes to entries that predate filer-assigned inodes, so pre-existing files resolve by filehandle immediately. Re-running is idempotent.
- `nfs.disable -path=/exports -apply` removes the mark and deletes the rows, so turning NFS off leaves no index data behind.

Verified against a live server: enable backfills exactly the in-scope entries, files created afterwards get rows through the filer-side hook, out-of-scope paths get nothing, and disable clears every row.

The gateway restore itself comes as a follow-up on top of this.

https://claude.ai/code/session_01AQDAwEYZsGsmGZKo3myZxB

## Summary by CodeRabbit

* **New Features**
  * Added support for maintaining path-to-inode mappings for NFS-backed directories.
  * Introduced commands to enable or disable NFS export support for a path, with optional preview mode before applying changes.

* **Bug Fixes**
  * Improved handling of renamed, deleted, and hard-linked files so path mappings stay accurate.
  * Added safer recursive cleanup for directory removals and better support for existing records during upgrades.

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -732,6 +732,7 @@ message FilerConf {
         bool worm = 14;
         uint64 worm_grace_period_seconds = 15;
         uint64 worm_retention_time_seconds = 16;
+        bool inode_index = 17;
     }
     repeated PathConf locations = 2;
 }
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -527,6 +527,7 @@ message AssignVolumeResponse {
     string replication = 7;
     string error = 8;
     Location location = 9;
+    repeated Location replicas = 10;
 }
 
 message LookupVolumeRequest {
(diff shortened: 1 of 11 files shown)


chrislusf:
This may work, but the additional inode->path index overhead is still unnecessary. Not planning to support this in the long term.

The enterprise version uses kernel module VFS, which can avoid this persistent inode problem from NFS.

PR #10342, opened 2026-07-16
fix(volume): reject overflowing needle ID deltas

7y-9:

The Go volume server accepts an optional decimal `_delta` suffix when parsing a file ID.

`Needle.ParsePath` added this delta directly to the parsed `uint64` needle ID. If the delta exceeded the remaining `uint64` capacity, the addition wrapped around. For example, adding `1` to needle ID `ffffffffffffffff` produced needle ID `0` instead of returning an error.

The HTTP delete handler and the gRPC `BatchDelete` and `Query` handlers also ignored errors returned by `ParsePath`, allowing invalid file IDs to continue into later request processing.

This change is limited to rejecting invalid file IDs in these Go volume server paths. Valid file IDs and deltas are unchanged.

Before adding the delta, `Needle.ParsePath` now compares it with the remaining `uint64` capacity and returns an error when the addition would overflow.

The existing `ParsePath` error is also handled by:

- the HTTP delete handler, which returns `400 Bad Request`
- gRPC `BatchDelete`, which records a `400 Bad Request` result for that file ID
- gRPC `Query`, which returns the parsing error

`BatchDelete` now validates the file ID before looking up the EC volume.

Regression tests cover the overflow check and the three request handlers.

# How is the PR tested?

```text
go test ./weed/storage/needle -run '^TestNeedleParsePathRejectsDeltaOverflow$' -count=1
go test ./weed/server -run '^(TestDeleteHandlerRejectsOverflowingNeedleDelta|TestBatchDeleteRejectsOverflowingNeedleDelta|TestQueryRejectsOverflowingNeedleDelta)$' -count=1
go test ./weed/server ./weed/storage/needle -count=1
git diff --check
```

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Rejects file identifiers with overflowing path delta values.
  * Delete requests now return `400 Bad Request` with an explicit JSON error for invalid file identifiers.
  * Batch deletes continue processing remaining file IDs and report invalid entries individually.
  * Queries now fail fast with an error when a file identifier can’t be parsed; batch-cookie mismatch no longer stops processing early.
* **Tests**
  * Added unit coverage for delta overflow handling across delete, batch delete, and query flows.

### DIFF
--- a/seaweed-volume/src/storage/needle/needle.rs
+++ b/seaweed-volume/src/storage/needle/needle.rs
@@ -712,7 +712,7 @@ fn format_needle_id_cookie(key: NeedleId, cookie: Cookie) -> String {
 /// Parse "needle_id_cookie_hex" or "needle_id_cookie_hex_delta" into (NeedleId, Cookie).
 /// Matches Go's ParsePath + ParseNeedleIdCookie: supports an optional `_delta` suffix
 /// where delta is a decimal number added to the NeedleId (used for sub-file addressing).
-/// Rejects strings that are too short or too long.
+/// Rejects strings that are too short or too long, and deltas that overflow the needle id.
 pub fn parse_needle_id_cookie(s: &str) -> Result<(NeedleId, Cookie), String> {
     // Go ParsePath: check for "_" suffix containing a decimal delta
     let (hex_part, delta) = if let Some(underscore_pos) = s.rfind('_') {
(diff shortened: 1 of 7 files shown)


chrislusf:
Pushed three follow-ups: the overflow error prints the needle id in hex to match fid formatting, batch delete continues past a cookie mismatch instead of silently dropping the rest of the batch, and the Rust volume server rejects the same overflow (it was wrapping) with a matching test.

PR #10353, opened 2026-07-16
shell: show who holds the cluster lock

chrislusf:
A blocked `lock` command sat silent — the per-second "already locked by" refusals only surfaced at `-v=2` — and `cluster.status` printed the shell's own lock state as "unlocked" even while another client held the cluster lock. Confusing enough that an operator with a stale day-old shell session holding the lock had no way to see why every new `lock` hung.

The master now answers `GetAdminLockStatus`, so:

- `lock` prints `waiting for lock held by <client>: <message>` before blocking, and `lock acquired` once the wait ends
- `cluster.status` shows `status: LOCKED by <client> (<message>)` for the cluster-wide holder, not just the local session

Both degrade silently against masters without the RPC: no waiting line, local-only status, same blocking behavior as today. Verified against a 4.39 master.

Exercising this end to end surfaced two latent bugs in the lock client, fixed in separate commits: an unlock racing the in-flight lease renewal re-creates the lock on the master (sometimes with a blank client name) and the stale-token release is then ignored, leaving a ghost lock until expiry; and the renewal goroutine never cleared its running flag on error, so later locks in the same process were never renewed and silently expired after ten seconds.

## Summary by CodeRabbit

- **New Features**
  - Added an admin RPC to query cluster-wide shell lock status (holder and optional message).
  - Expanded `cluster.status` output to show lock state, holder, and message.
  - Enhanced the lock command to indicate when it’s waiting and when the lock is acquired.

- **Bug Fixes**
  - Refused admin lock releases when not the cluster leader, and ensured release only occurs with a matching token.
  - Improved exclusive lock renewal/release synchronization and safer local lock-state handling with bounded RPC timeouts.

- **Tests**
  - Updated cluster status command tests to cover the new lock output formatting.


chrislusf:
Applied the timeout points: each lease/renew/release/status attempt now carries its own short context, so an unresponsive master can no longer hang unlock behind the renewal mutex or stall the prompt.

WithClient's first parameter is streamingMode, not leader selection — the master client already follows the leader — so it stays false. The real gap behind the release comment was master-side: ReleaseAdminToken answered success on followers; it now refuses like LeaseAdminToken does.

chrislusf:
Took the token race: release now captures the token it started with and compares on clear, so a RequestLock racing a slow release (the admin presence lock can do this on shutdown) keeps its fresh token and the release can never carry it.

Declined the rest. Moving the RPC to a goroutine changes unlock from released-when-returned to released-eventually, the same kind of semantic drift just reverted here; the stall it worries about predates this PR, since unlock has always waited on that call. Swallowing the release error is deliberate: a failed release is covered by lease expiry, and retrying only delays unlock for no benefit.

PR #10367, opened 2026-07-20
fix(s3): fall back to remote on local chunk 404 for remote-mounted objects

### LINKED ISSUE #10366: S3 read of remote-mounted object fails with 404 when a local chunk needle is missing instead of re-pulling from remote
## Problem

When a SeaweedFS **remote-storage mount** is used as a read-through cache over an object store (e.g. GCS/S3), an S3 `GetObject` can fail with a `404 Not Found` even though the object is fully intact in the remote store.

## Setup

We run a remote-storage mount as a read-through cache:

- Objects are dual-populated: the application writes to **both** the backing object store and SeaweedFS.
- A local eviction/uncache process later clears local chunks and stamps a remote-only `RemoteEntry` on the filer entry, so subsequent reads re-pull from the remote.

The **remote-only re-pull path works correctly**: when a filer entry has no local chunks and a `RemoteEntry` (`IsInRemoteOnly()` is true), `streamFromVolumeServers` calls the stream-through cache (`cacheRemoteObjectForStreamingWithShortTimeout` → `CacheRemoteObjectToLocalCluster`) and serves the object.

## The bug

Sometimes a filer entry **still has local chunks** (so it is *not* remote-only) but the referenced volume needle has been **reclaimed** by vacuum-compaction or deletion. The S3 GET then reads a dead needle, receives a 404, retries the same dead needle for the duration of the retry window, and finally errors the request — even though the object is intact in the remote store.

Because the entry still has (stale) chunks, `IsInRemoteOnly()` returns `false`, so the code never falls back to the remote stream-through cache.

### Observed log evidence (generalized; placeholders for internal values)

```
stream.go: read chunk <VID>,<NEEDLE> failed, invalidating cache and retrying: <VOLUME_URL>/<VID>,<NEEDLE>?readDeleted=true: 404 Not Found
reader_at.go: fetching chunk {FileId:<VID>,<NEEDLE> ... ChunkSize:4096}: 404
s3api_object_handlers.go: streamFromVolumeServers: streamFn failed after writing 0 bytes: ...404
s3api_object_handlers.go: GetObjectHandler: failed to stream <bucket>/<key> from volume servers: ...404
```

## Root cause

`streamFromVolumeServers` only re-pulls from the remote when `entry.IsInRemoteOnly()` is true (entry has zero chunks). When a remote-backed entry retains **stale** local chunks pointing at a reclaimed needle, the read fails with 404 and there is no fallback to the remote copy.

## Impact

Transient `GetObject` failures (surfaced as 404/500 to S3 clients) for objects that are fully present in the remote store, whenever a local needle backing a cached-but-not-uncached entry is lost (vacuum-compaction/deletion) before the entry has been converted to remote-only.

## Proposed fix

In the S3 GET path, when a local chunk read fails with **not-found** (needle missing), *before any response bytes have been written*, and the entry is **remote-backed** (under a remote-storage mount):

1. Drop the stale local chunks so the entry becomes remote-only again (this also makes future reads self-heal through the existing, well-tested remote-only path).
2. Re-pull the object from the remote using the existing mechanism (`CacheRemoteObjectToLocalCluster` / `cacheRemoteObjectForStreamingWithShortTimeout`).
3. Stream the requested range from the freshly cached chunks.

The fallback is guarded to only trigger for entries with a remote backing, so behavior for non-remote buckets is unchanged.

A draft PR implementing this is linked below.

petedodd-pd:
## Motivation

Fixes #10366.

When a remote-storage mount is used as a read-through cache over an object store, an S3 `GetObject` can fail with `404 Not Found` even though the object is intact in the remote store.

Objects are dual-populated (app writes to both the object store and SeaweedFS), and a local eviction process later clears local chunks and stamps a remote-only `RemoteEntry` so reads re-pull. That remote-only re-pull path works. The bug is the case where a filer entry **still has local chunks** (so `IsInRemoteOnly()` is false) but the referenced volume needle has been reclaimed by vacuum-compaction/deletion — the chunk read 404s and there is no fallback to the remote copy.

Production log evidence (generalized):

```
stream.go: read chunk <VID>,<NEEDLE> failed, invalidating cache and retrying: ...?readDeleted=true: 404 Not Found
reader_at.go: fetching chunk {FileId:<VID>,<NEEDLE> ...}: 404
s3api_object_handlers.go: streamFromVolumeServers: streamFn failed after writing 0 bytes: ...404
s3api_object_handlers.go: GetObjectHandler: failed to stream <bucket>/<key> from volume servers: ...404
```

## Overview

`streamFromVolumeServers` now detects a not-found chunk read (via `errors.Is(err, util_http.ErrNotFound)`), and — only when no response bytes have been written yet and the entry is remote-backed — recovers instead of failing:

1. **Drop the stale chunks** (`dropLocalChunks`) so the filer entry becomes remote-only again. This is required because `CacheRemoteObjectToLocalCluster` short-circuits as "already cached" while chunks are present, and it also makes future reads self-heal through the existing, well-tested remote-only path.
2. **Re-pull from the remote** via the existing `cacheRemoteObjectForStreamingWithShortTimeout` → `CacheRemoteObjectToLocalCluster` mechanism.
3. **Stream** the requested range from the freshly cached chunks (`streamEntryRangeToResponse`, mirroring the existing reader setup).

The recovery is guarded on `entry.RemoteEntry` being present, so non-remote buckets are completely unaffected. Since `Content-Length` has already been sent and no bytes were written, streaming exactly the requested range still yields a valid response.

All new code lives in `weed/s3api/s3api_object_handlers.go`.

## Status: Draft / WIP

- The SSE-encrypted read path (`streamFromVolumeServersWithSSE`) is not yet covered — only the plain `streamFromVolumeServers` path (where the production 404 occurred).
- Only a guard-level unit test is included (`TestRecoverFromStaleRemoteChunksGuards`), asserting the recovery is skipped for non-remote entries and canceled requests without touching the filer. **TODO:** a full integration test with fake volume servers exercising the drop-chunks → re-pull → stream flow end to end.

## Testing

- `go build ./...` passes.
- `go vet ./weed/s3api/` reports only pre-existing warnings in unrelated test files.
- `go test ./weed/s3api/ -run 'TestRecoverFromStaleRemoteChunksGuards|TestCacheRemoteObjectForStreaming'` passes.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved S3 object downloads by automatically recovering when remote-mounted reads detect stale local chunk references.
  - Recovery re-reruns the originally requested byte range and avoids returning not-found when zero bytes were written initially.
  - Added safeguards to skip recovery for local objects and for canceled requests, and ensured egress traffic accounting continues during recovery.
- **New Features**
  - Added a new Prometheus metric to track stale remote-chunk recovery outcomes (recovered vs. failed) and included it in bucket metric cleanup.
- **Tests**
  - Added coverage for guard behavior to ensure recovery doesn’t dial or cache when it must be skipped.

### DIFF
--- a/weed/s3api/s3api_object_handlers.go
+++ b/weed/s3api/s3api_object_handlers.go
@@ -1154,6 +1154,18 @@ func (s3a *S3ApiServer) streamFromVolumeServers(w http.ResponseWriter, r *http.R
 		BucketTrafficSent(cw.written, r)
 	}
 	if err != nil {
+		// A remote-mounted entry can still reference local chunks whose backing
+		// needle has been reclaimed (vacuum-compaction / deletion), so the
+		// volume-server read returns 404 even though the object is intact in the
+		// remote store. Because the entry still has chunks, IsInRemoteOnly() is
+		// false and the normal remote re-pull above never runs. When nothing has
+		// been written yet, drop the dead chunks and re-pull from the remote
+		// rather than failing the GET.
+		if cw.written == 0 && errors.Is(err, util_http.ErrNotFound) &&
+			s3a.recoverFromStaleRemoteChunks(w, r, entry, bucket, object, versionId, offset, size, totalSize) {
+			glog.V(1).Infof("streamFromVolumeServers: recovered %s/%s from remote after stale-chunk 404", bucket, object)
+			return nil
+		}
 		switch {
 		case isCanceledStreamingError(err):
 			// Client disconnected mid-stream (e.g. Nginx upstream timeout, browser cancel) - expected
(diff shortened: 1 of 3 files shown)


petedodd-pd:
Investigation update: the root cause is now pinned in #10366 — a write-path last-write-wins race in the filer's `UpdateEntry` on the chunk list (`validateUpdateEntryPreconditions` guards `Extended` but not chunks), reproduced deterministically and via a real eviction path against v4.38.

That reframes this PR as **defense-in-depth**: it self-heals a read when an entry has been stranded, which is valuable regardless. But the primary **correctness** fix is on the write path — extending the existing `ExpectedExtended` precondition in `UpdateEntry` to cover the chunk list, so a stale chunk-preserving RMW can't resurrect a needle already queued for deletion.

Suggest keeping this PR as the tolerance layer and adding a companion change for the `UpdateEntry` chunk CAS. Happy to put that up.

PR #10368, opened 2026-07-20
fix(filer): opt-in chunk-list optimistic concurrency in UpdateEntry (RFC)

petedodd-pd:
> **RFC / DRAFT** — opening this for design feedback before hardening. The proto field shape, error/retry semantics, and which call sites should populate the precondition are open questions (see below). Companion to the read-side defense-in-depth in #10367; both address #10366.

## Motivation

A filer entry can end up referencing a volume needle that has been deleted and vacuumed, so reads 404 even though the object is intact in a remote-storage mount. The root cause is a last-write-wins race in the filer's `UpdateEntry`.

`UpdateEntry` (`weed/server/filer_grpc_server.go`) is a read-modify-write:

1. `FindEntry` reads the current stored entry.
2. `validateUpdateEntryPreconditions(entry, req.ExpectedExtended)` — guards **only** the `Extended` map, **not** the chunk list.
3. `cleanupChunks(entry, req.Entry)` diffs stored-vs-incoming chunks to compute garbage needles.
4. the incoming entry is written, then `DeleteChunksNotRecursive(garbage)` fire-and-forget queues the garbage for deletion.

Because the precondition ignores the chunk list, two concurrent RMWs on one entry are last-write-wins on chunks. The stranding interleaving:

1. entry `chunks=[F]`, needle F live.
2. **R** (a chunk-preserving RMW — S3 tag/attr/copy, FUSE setattr, another controller) reads the entry and prepares a metadata-only change that keeps `chunks=[F]`.
3. **D** (eviction/uncache) commits `UpdateEntry chunks=[]`: garbage = `[F]`, F is enqueued for deletion.
4. **R** commits its stale snapshot last: stored is now `[]`, so its garbage diff is empty, and it writes `chunks=[F]` — **resurrecting the reference to F**, which is already queued for deletion.
5. the deletion queue tombstones F; vacuum reclaims it → the entry references a dead needle. `IsInRemoteOnly()` is false (the entry still has chunks) so there is no remote re-pull → **404**.

This has been reproduced deterministically and via a real eviction/uncache path in a testcontainers harness against a released build.

## Overview

Add **opt-in optimistic concurrency on the chunk list** to `UpdateEntry`, mirroring the existing `ExpectedExtended` mechanism:

- **Proto** (`weed/pb/filer.proto`): a new optional `repeated string expected_chunks = 6` on `UpdateEntryRequest`. It carries the client's expected current chunk set as fid strings (`FileChunk.GetFileIdString()`). A fid-string list was chosen as the lightest faithful representation: needle liveness — the thing this bug is about — is keyed exactly by fid `(volume, key, cookie)`, and fid strings are already the identity used throughout chunk diffing (`DoMinusChunks`, `SeparateGarbageChunks`). Regenerated with `make gen` in `weed/pb` (protoc + go + go-grpc + vtproto).
- **Handler**: `validateUpdateEntryPreconditions` now also takes `expectedChunks` and delegates to a new `validateExpectedChunks`. When `expected_chunks` is non-empty, the stored entry's current chunk set is compared (order-independent, multiset by fid) against it, and a mismatch returns a retryable `codes.FailedPrecondition`. This makes R's stale chunk-preserving write fail cleanly so the caller can re-read the emptied state instead of resurrecting F.
- **Backward compatible / opt-in**: empty or nil `expected_chunks` ⇒ no check ⇒ current behavior is preserved exactly. No existing call site changes behavior until it starts populating the field.

### Related: `CreateEntry`

`CreateEntry` shares the same `cleanupChunks` → `DeleteChunksNotRecursive` pattern and the same class of race on overwrite. It is intentionally **not** changed here to keep this PR focused; covering it is a candidate follow-up (see open questions).

## Test

`weed/server/filer_grpc_server_chunk_cas_test.go`:

- `TestValidateExpectedChunks` — unit coverage of the precondition: nil/empty disables, exact match and reordering pass, and every mutation (stored emptied, chunk added/removed/replaced, duplicate-count mismatch) is rejected with `FailedPrecondition`.
- `TestUpdateEntryChunkCASPreventsStrand` — drives the real `UpdateEntry` gRPC handler through the #10366 interleaving:
  - **without** the precondition, R's stale write resurrects `chunks=[F]` (demonstrates the current buggy behavior);
  - **with** `expected_chunks=[F]` against the concurrently-emptied entry, R's write is rejected and the stored entry correctly stays `chunks=[]` (no resurrected reference — the strand does not occur);
  - a fresh RMW whose precondition still matches is applied normally.

`go build ./...`, `go vet`, and `go test ./weed/server/ ./weed/pb/filer_pb/` all pass.

## Open questions for maintainers

1. **Field shape / name.** Is `repeated string expected_chunks` (fid strings) the representation you'd want, or would you prefer a compact fingerprint (e.g. a hash of the sorted fid set) or a structured `repeated FileId`? Fid strings are simple and debuggable but grow with chunk count for large/manifest objects.
2. **Should `CreateEntry` get the same precondition?** It has the same overwrite race. Happy to extend it here if the shape is agreed, or split it into a follow-up.
3. **Which upstream RMW call sites should populate the precondition?** Candidates: the S3 API metadata/tagging/copy paths, FUSE `setattr`, and other filer-client read-modify-write callers. This PR only adds the mechanism; wiring the producers is deliberately deferred pending direction.
4. **Error type / retry semantics.** `FAILED_PRECONDITION` mirrors the existing `ExpectedExtended` behavior and signals "re-read and retry". Is that the contract you want, or should this be a distinct/typed error (or `ABORTED`, which gRPC conventionally marks retryable) so callers can auto-retry without conflating it with object-lock / If-Match precondition failures?

## Summary by CodeRabbit

* **New Features**
  * Added an optional optimistic concurrency guard for file chunk updates via a new “expected chunk set” parameter.
  * When provided, updates are rejected unless the stored chunk set matches the expected list exactly.
  * If not provided, behavior remains backward compatible.
* **Bug Fixes**
  * Prevents concurrent chunk changes from being reverted by stale read-modify-write updates.
* **Tests**
  * Added coverage for chunk-set validation and end-to-end update behavior, including CAS enabled vs disabled scenarios.

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -447,6 +447,13 @@ message UpdateEntryRequest {
     bool is_from_other_cluster = 3;
     repeated int32 signatures = 4;
     map<string, bytes> expected_extended = 5;
+    // Optimistic concurrency on the chunk list. When non-empty, the filer
+    // rejects the update with FAILED_PRECONDITION unless the stored entry's
+    // current chunk set (by fid string) exactly matches this set. Empty/nil
+    // disables the check (backward compatible). Prevents a stale
+    // chunk-preserving read-modify-write from resurrecting a needle that a
+    // concurrent update already queued for deletion.
+    repeated string expected_chunks = 6;
 }
 message UpdateEntryResponse {
     SubscribeMetadataResponse metadata_event = 1;
--- a/weed/pb/filer.proto
+++ b/weed/pb/filer.proto
@@ -447,6 +447,13 @@ message UpdateEntryRequest {
     bool is_from_other_cluster = 3;
     repeated int32 signatures = 4;
     map<string, bytes> expected_extended = 5;
+    // Optimistic concurrency on the chunk list. When non-empty, the filer
+    // rejects the update with FAILED_PRECONDITION unless the stored entry's
+    // current chunk set (by fid string) exactly matches this set. Empty/nil
+    // disables the check (backward compatible). Prevents a stale
+    // chunk-preserving read-modify-write from resurrecting a needle that a
+    // concurrent update already queued for deletion.
+    repeated string expected_chunks = 6;
 }
 message UpdateEntryResponse {
     SubscribeMetadataResponse metadata_event = 1;
--- a/weed/pb/filer_pb/filer.pb.go
+++ b/weed/pb/filer_pb/filer.pb.go
@@ -2310,8 +2310,15 @@ type UpdateEntryRequest struct {
 	IsFromOtherCluster bool                   `protobuf:"varint,3,opt,name=is_from_other_cluster,json=isFromOtherCluster,proto3" json:"is_from_other_cluster,omitempty"`
 	Signatures         []int32                `protobuf:"varint,4,rep,packed,name=signatures,proto3" json:"signatures,omitempty"`
 	ExpectedExtended   map[string][]byte      `protobuf:"bytes,5,rep,name=expected_extended,json=expectedExtended,proto3" json:"expected_extended,omitempty" protobuf_key:"bytes,1,opt,name=key" protobuf_val:"bytes,2,opt,name=value"`
-	unknownFields      protoimpl.UnknownFields
-	sizeCache          protoimpl.SizeCache
+	// Optimistic concurrency on the chunk list. When non-empty, the filer
+	// rejects the update with FAILED_PRECONDITION unless the stored entry's
+	// current chunk set (by fid string) exactly matches this set. Empty/nil
+	// disables the check (backward compatible). Prevents a stale
+	// chunk-preserving read-modify-write from resurrecting a needle that a
+	// concurrent update already queued for deletion.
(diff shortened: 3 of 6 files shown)


chrislusf:
Thanks — the diagnosis, the stranding interleaving writeup, and the test approach here were exactly right. This landed via #10382 in a different shape: rather than a dedicated expected_chunks field, the check became an IF_CHUNKS_EQUAL clause on the existing WriteCondition DSL, and UpdateEntry (plus DeleteEntry/AppendToEntry) now takes the per-path lock so the precondition and the write are atomic. Clause presence also makes "expect no chunks" expressible, covering the reverse race. Your strand test survives essentially as written.

On the open questions: FailedPrecondition stayed as the error code, and wiring call sites is still ahead — metadata-only paths are better served by ObjectTransaction PATCH_EXTENDED (server-side merge under the lock, no precondition needed), while full-entry writers like setattr will use the clause. Closing in favor of #10382.

PR #10383, opened 2026-07-21
S3: fold large chunk lists into manifest chunks on the direct write path

### LINKED ISSUE #10379: S3 API: Large files missing manifest chunk generation (lost after direct-to-volume-server change)
## Problem                                                                                                                                  
                                                                                                                                              
When S3 API switched from `S3 → Filer → Volume Server` to direct `S3 → Volume Server` path (commit ca84a8a71), the `MaybeManifestize` call  was lost.                                                                                                                                   
                                                                                                                                              
**Affected paths:**                                                                                                                         
- `S3 PutObject` (`s3api_object_handlers_put.go`)                                                                                           
- `S3 CompleteMultipartUpload` (`s3api/filer_multipart.go`)                                                                                 
                                                                                                                                              
**Expected:** Files with >10,000 chunks should generate manifest chunks for hierarchical indexing (every 10,000 chunks → 1 manifest chunk). 
                                                                                                                                              
**Actual:** Large files (e.g., 900GB / 120k chunks) stored with all chunks flat, no manifest generated.                                     
                                                                                                                                              
**Impact:**                                                                                                                                 
  - Metadata overhead: all chunk references stored flat                                                                                       
  - Performance degradation on large files                                                                                                    
  - Inconsistent with mount/WebDAV paths which still call MaybeManifestize

chrislusf:
Fixes #10379.

The gateway uploads chunks itself and hands the filer a fully prepared entry. On the routed write path (ObjectTransaction) the filer stores that entry as-is, so a large PutObject or CompleteMultipartUpload persisted its whole flat chunk list — the 900GB case in the issue keeps 120k chunk references in one entry. (The non-routed CreateEntry RPC still folds server-side, which is why only routed writes were affected.) Manifestize on the gateway before the entry is written, the same way mount, WebDAV, and filer.copy prepare theirs. Part uploads stay flat, since completion rebases their offsets.

Multipart part boundaries also record byte offsets now: the stored chunk indexes stop matching the entry once the list is folded, and partNumber reads plus GetObjectAttributes prefer the offsets. Legacy index-only records keep working, with bounds checks instead of a possible panic.

Copy paths resolve a manifested source into data chunks before their per-chunk copy loops — copying a manifest chunk raw would store its blob as object data still pointing at the source's chunks — and a large copied list is folded again on the destination. Completion likewise resolves manifest chunks a part entry may carry (the filer folds an oversized UploadPartCopy range) before rebasing part offsets; without that, such a completion produced an unreadable object.

Reproduced and verified end to end with ManifestBatch temporarily lowered to 100 on a local `weed mini`: before, a 1GB PutObject and a 128-part multipart upload both persisted 128 flat chunks; after, both fold to one manifest chunk plus the remainder, and full GET, ranged GET across the manifest span, partNumber GETs, GetObjectAttributes part sizes, CopyObject, UploadPartCopy, and delete all check out byte-identical.

## Summary by CodeRabbit

* **New Features**
  * Improved multipart upload handling with reliable byte-offset tracking for part boundaries.
  * Added support for manifest-based chunk storage across uploads and object copies.
  * Improved multipart object attributes and part sizing, including compatibility with legacy metadata.

* **Bug Fixes**
  * Improved copying of objects and parts containing manifest chunks.
  * Added safer handling for invalid multipart boundaries and chunk ranges.
  * Improved cleanup of incomplete uploads and failed object creations.

### DIFF
--- a/weed/s3api/filer_multipart.go
+++ b/weed/s3api/filer_multipart.go
@@ -205,6 +205,11 @@ type multipartPartBoundary struct {
 	StartChunk int    `json:"start"`
 	EndChunk   int    `json:"end"`
 	ETag       string `json:"etag"`
+	// Byte offsets of the part within the object. Readers prefer these over
+	// the chunk indexes above, which stop matching the entry's chunk list
+	// once large completions fold it into manifest chunks.
+	StartOffset int64 `json:"startOffset,omitempty"`
+	EndOffset   int64 `json:"endOffset,omitempty"` // exclusive
 }
 
 type multipartSSES3Info struct {
@@ -492,7 +497,17 @@ func (s3a *S3ApiServer) prepareMultipartCompletionState(r *http.Request, input *
 				continue
 			}
 
+			// A part entry can itself carry manifest chunks (the filer folds an
+			// oversized part, e.g. a large UploadPartCopy range). Resolve them
+			// first: the rebase below shifts chunk offsets, which a manifest
+			// chunk cannot express.
+			if flattenErr := s3a.flattenManifestChunks(r.Context(), entry); flattenErr != nil {
+				glog.Errorf("completeMultipartUpload %s %s part %d resolve manifest chunks: %v", *input.Bucket, *input.UploadId, partNumber, flattenErr)
+				return nil, nil, s3err.ErrInternalError
+			}
+
 			partStartChunk := len(finalParts)
+			partStartOffset := offset
 			partETag := getEtagFromEntry(entry)
 
 			for _, chunk := range entry.GetChunks() {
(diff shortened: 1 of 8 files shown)


chrislusf:
Pushed a follow-up covering manifest blob lifecycle: a failed partial fold now deletes the batches it already saved; a failed completion deletes the manifests it created (skipped only when a failed rollback left the version entry holding them); and the part-manifest blobs superseded by flattening are deleted once completion commits, since the upload-dir removal is metadata-only.

Two adjacent shared-chunk hazards became sharper with manifests in play and are fixed alongside: the version-file rollback used to delete its data — which is the still-registered parts' data, so a retried completion produced a dangling object — and the idempotent-replay cleanup data-deleted leftover parts whose chunks the live object references. Both are metadata-only now.

Verified live (threshold lowered to 100 locally): an oversized UploadPartCopy part folded by the filer completes correctly, the superseded part-manifest blob 404s on the volume server after completion, and both the completed object and the copy source still read back byte-identical.

PR #10395, opened 2026-07-22
mount: refresh open file handles from the metadata event entry

### LINKED ISSUE #10394: [mount] Metadata invalidation can leave an open FileHandle permanently stale
### Description

A transient failure while refreshing an already-open `FileHandle` from a metadata subscription event can permanently pin that handle to an old filer entry.

The metadata event is consumed and the subscription cursor advances even when the handle refresh fails. Subsequent `stat` and `read` calls continue to use the stale `fh.Entry`, with no retry until another metadata event arrives, the file is reopened, or the mount restarts.

I reproduced this on current upstream master at `6c4eb95a3`.

### Relevant code path

The mount's metadata invalidation callback finds an open handle and performs a second lookup:

```go
if fh, found := wfs.fhMap.FindFileHandle(inode); found {
    newEntry, status := wfs.maybeLoadEntry(filePath)
    if status == fuse.OK {
        if fh.GetEntry().GetEntry() != newEntry {
            fh.SetEntry(newEntry)
        }
    }
}
```

If `maybeLoadEntry()` returns a transient error, the callback keeps the old entry and records no pending refresh. A successful lookup can also still return stale cached metadata.

The subscription event already carries `OldEntry` and `NewEntry`, so this extra lookup creates a failure/staleness window after the authoritative event has arrived.

### Deterministic reproduction

1. Start one filer setup with two FUSE mounts, a writer mount and a reader mount.
2. The writer creates a file, writes 88 bytes, and syncs it.
3. The reader opens the file, reads the 88-byte prefix, and keeps the same file descriptor open.
4. Arm a one-shot fault on the reader mount so that `maybeLoadEntry(path)` returns `fuse.EIO` when the final metadata event refreshes the open handle.
5. The writer appends data and closes the file. The final filer entry is 180,020 bytes.
6. The reader receives the event, but its open handle remains at 88 bytes. `fstat` reports 88 and reading the same descriptor reaches EOF at byte 88.
7. Emit another same-path metadata event from the healthy writer mount, for example a same-mode `chmod`. The same reader descriptor and inode immediately refresh to 180,020 bytes and the full SHA-256 matches.

This demonstrates that the first event was consumed, the open-handle refresh failed once, and a second event is required to recover.

Control tests also show:

- A normal subscription interruption followed by replay catches the open handle up correctly when no refresh failure is injected.
- If the old reader handle is closed, a one-shot `maybeLoadEntry()` failure makes the first reopen return EIO; the next reopen retries and reads the final file. It does not pin an old entry.

### Expected behavior

A transient secondary lookup failure must not leave an open file handle permanently stale after the metadata event has been consumed.

### Suggested fix

For a same-path update, propagate the event's `NewEntry` through the invalidation and update the open `FileHandle` directly instead of calling `maybeLoadEntry()` again.

The entry should be cloned before storing it in the handle, and filer UID/GID mapping should still be applied. Rename and delete semantics should be preserved:

- same-path update: destination path + `NewEntry`
- rename old path: old path + `OldEntry`
- rename destination: new path + `NewEntry`
- delete: old path + `OldEntry`

Calling `fh.SetEntry()` then rebuilds the chunk group/file size and invalidates the handle's chunk-offset cache.

### Impact

An open reader can indefinitely observe an obsolete file size and chunk list even though the filer entry and other mounts are correct. A metadata mutation issued through the stale mount may also write the stale full entry, including its chunk list, back to the filer.

chrislusf:
Fixes #10394

The invalidation callback looked the path up again after a subscription event touched an open handle. A transient lookup failure, or a stale cached result, left the handle pinned to its old entry with no retry: the cursor had already advanced, so only another event for the same path would recover it. An open reader then kept observing the obsolete size and chunk list indefinitely, and a metadata mutation through that handle could write the stale entry back to the filer.

Apply the entry the event itself carries instead. Each invalidation now attaches the entry that lives at the invalidated path — NewEntry for in-place updates, rename destinations, and creates; nil when the path was vacated — and the handle refresh clones it, applies the filer-to-local uid/gid mapping, and sets it directly, with no filer round-trip. A nil entry keeps the last entry so unlinked-but-open reads still work. Self-originated events cannot clobber local write state through this path: the local apply seeds the dedup ring, so the subscription's redelivery is dropped before invalidation.

Reproduced in a test that delivers an update event to a mount whose filer address has no listener, so any secondary lookup fails like a transient error: before, the open handle stayed at the old 88-byte entry; after, it refreshes to the event's 180,020-byte entry. A second test pins the per-event-shape invalidation payloads for update, rename, create, and delete.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved metadata synchronization by propagating filer event timestamps through cache invalidations, ensuring correct ordering across create, update, rename/move, and delete (including vacated paths).
  - Open file handles now refresh safely from metadata subscription updates using timestamp-based deduplication, preventing older queued events from overwriting newer local state; buffered directory-build events are revalidated on completion.
- **Tests**
  - Expanded test coverage to verify authoritative invalidation entries, correct rename vacate behavior, delete handling, and open-handle ordering across queued, buffered, and ack/no-ack scenarios.

### DIFF
--- a/weed/mount/weedfs.go
+++ b/weed/mount/weedfs.go
@@ -686,6 +686,27 @@ func (wfs *WFS) latestKnownFilerTsNs() int64 {
 	return wfs.metaCache.LatestEventTsNs()
 }
 
+// filerBarrierTsNs returns a timestamp at or below the filer's current log
+// position. Metadata events are stamped with the filer clock, and a self-ping
+// reads that same clock, so unlike latestKnownFilerTsNs this also covers
+// events already committed but not yet delivered to the subscription. Call
+// before the RPC whose result the barrier fences. Best-effort: one bounded
+// attempt, falling back to the newest delivered event — no WithFilerClient
+// retry waves for an optimization.
+func (wfs *WFS) filerBarrierTsNs() int64 {
+	baseline := wfs.latestKnownFilerTsNs()
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
+	defer cancel()
+	_ = pb.WithGrpcClient(ctx, false, wfs.signature, func(conn *grpc.ClientConn) error {
+		resp, err := filer_pb.NewSeaweedFilerClient(conn).Ping(ctx, &filer_pb.PingRequest{})
+		if err == nil && resp.StartTimeNs > baseline {
+			baseline = resp.StartTimeNs
+		}
+		return err
+	}, wfs.getCurrentFiler().ToGrpcAddress(), false, wfs.option.GrpcDialOption)
+	return baseline
+}
+
 // invalidateOpenFileHandle refreshes an open file handle from a metadata
 // subscription event. No filer lookup happens here: it can fail transiently,
 // and since the subscription cursor has already advanced past the event, the
--- a/weed/mount/weedfs.go
+++ b/weed/mount/weedfs.go
@@ -701,9 +701,12 @@ func (wfs *WFS) invalidateOpenFileHandle(filePath util.FullPath, entry *filer_pb
 	// snapshot by now: a local flush can install newer state while the event
 	// sits in the queue, and the flush's own event is dedup-suppressed, so a
 	// rollback would never heal. The apply loop has already ordered this event
-	// and any later state into the local store, so prefer the store's entry;
-	// it misses only for read-through directories, where the event entry is
-	// the freshest ordered information available.
+	// and any later state into the local store, so prefer the store's entry.
+	// The store misses for read-through directories and TTL-expired entries;
+	// falling back to the event entry there can still roll back a racing
+	// local flush (neither write reaches the store in a read-through
+	// directory), but it is the best ordered information available without a
+	// filer round-trip.
 	if localEntry, findErr := wfs.metaCache.FindEntry(context.Background(), filePath); findErr == nil && localEntry != nil {
 		fh.SetEntry(localEntry.ToProtoEntry())
 		return
--- a/weed/mount/weedfs.go
+++ b/weed/mount/weedfs.go
@@ -302,8 +302,8 @@ func NewSeaweedFileSystem(option *Option) *WFS {
 			wfs.inodeToPath.MarkChildrenCached(path)
 		}, func(path util.FullPath) bool {
 			return wfs.inodeToPath.IsChildrenCached(path)
-		}, func(filePath util.FullPath, entry *filer_pb.Entry) {
-			wfs.invalidateOpenFileHandle(filePath, entry)
+		}, func(filePath util.FullPath, entry *filer_pb.Entry, eventTsNs int64) {
+			wfs.invalidateOpenFileHandle(filePath, entry, eventTsNs)
 		}, func(dirPath util.FullPath) {
 			if wfs.inodeToPath.RecordDirectoryUpdate(dirPath, time.Now(), wfs.dirHotWindow, wfs.dirHotThreshold) {
 				wfs.markDirectoryReadThrough(dirPath)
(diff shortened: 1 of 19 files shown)


weed/mount/weedfs.go
```diff
@@ -694,6 +676,51 @@ func (wfs *WFS) lookupEntry(fullpath util.FullPath) (*filer.Entry, fuse.Status)
...
+	fh.dirtyPages = newPageWriter(fh, wfs.option.ChunkSizeLimit)
+
+	// Invalidations apply asynchronously, so the event entry may be a stale
+	// snapshot by now: a local flush can install newer state while the event
+	// sits in the queue, and the flush's own event is dedup-suppressed, so a
+	// rollback would never heal. The apply loop has already ordered this event
+	// and any later state into the local store, so prefer the store's entry;
+	// it misses only for read-through directories, where the event entry is
+	// the freshest ordered information available.
+	if localEntry, findErr := wfs.metaCache.FindEntry(context.Background(), filePath); findErr == nil && localEntry != nil {
```
chrislusf: During a directory build this premise doesn't hold. An event touching a building dir is buffered — its store write is deferred to `completeDirectoryBuildNow` — while its invalidation is enqueued immediately (`applyMetadataSideEffectsSkippingBuildingDirs`). If the in-progress listing has already inserted this path, `FindEntry` here hits the listing-snapshot entry, which is *older* than the event (that's exactly why the event was buffered rather than snapshot-skipped), and pins the handle to pre-event state. The completion replay then applies buffered events with `MetadataResponseApplyOptions{}` — no `InvalidateEntries` — so no re-refresh ever arrives. With the first commit alone this case was handled correctly (the event entry was applied directly), so this is a regression for the build window.

Suggestion: replay buffered events in `completeDirectoryBuildNow` with `InvalidateEntries: true`. By then the store is ordered, so the re-invalidation lands the final state; events skipped by the snapshot filter need no re-invalidation since the listing state is at least as new.

weed/mount/weedfs.go
```diff
@@ -694,6 +676,51 @@ func (wfs *WFS) lookupEntry(fullpath util.FullPath) (*filer.Entry, fuse.Status)
...
+	fh.dirtyPages.Destroy()
+	fh.dirtyPages = newPageWriter(fh, wfs.option.ChunkSizeLimit)
+
+	// Invalidations apply asynchronously, so the event entry may be a stale
+	// snapshot by now: a local flush can install newer state while the event
+	// sits in the queue, and the flush's own event is dedup-suppressed, so a
+	// rollback would never heal. The apply loop has already ordered this event
+	// and any later state into the local store, so prefer the store's entry;
+	// it misses only for read-through directories, where the event entry is
+	// the freshest ordered information available.
```
chrislusf: Two nits on this sentence. In an uncached directory the local flush's apply doesn't write the store either (`atomicUpdateEntryFromFilerLocked` gates inserts on `allowUncachedInsert || isCachedFn`), so on this miss the fallback still applies the queued snapshot and the rollback this commit fixes remains possible there — equally unhealable, since the flush's echo is dedup-suppressed. "Freshest ordered information available" is true of the store's ordering, but the handle itself may hold newer locally-flushed state. Probably fine to accept (it needs a local flush racing a remote write on the same open file in an uncached dir), but worth stating honestly here; a follow-up could stamp the fh with the ts of its last local apply and skip this fallback SetEntry when the handle is newer.

Also, `FindEntry` misses for TTL-expired entries too, not only read-through directories.

chrislusf (commented):
Reviewed the store-resolution commit — the queued-rollback fix and its mechanism look right (FindEntry is local-only and applies the uid/gid mapping, so both paths are consistent, and Shutdown drains the worker before taking the cache lock). Two cases where the store-first premise doesn't hold, one of them a regression vs the first commit.

chrislusf:
Superseded by #10403, which redoes this on a single ordering invariant instead of per-site fences.

PR #10403, opened 2026-07-23
mount: version open file handles by filer log position

### LINKED ISSUE #10394: [mount] Metadata invalidation can leave an open FileHandle permanently stale

chrislusf:
Fixes #10394

A metadata event refreshing an open handle re-looked the path up; a transient failure left the handle permanently pinned to its old entry, since the subscription cursor had already advanced. The first attempt to fix this (#10395) applied the event's entry directly, and review then found a sequence of rollback races: each fix manufactured an ordering signal out of band from the state it protected — a delivered-event cursor, a pre-RPC ping, a stable-read retry — and each approximation had a window the next reproduction lived in. This redoes the fix on a single invariant instead: the filer log position is the version of any filer state, and the version travels with the value.

Filer side, lookup and remote-cache responses now carry a log timestamp stamped before the serving read — events are logged after their store write on that clock, so everything at or below the stamp is provably reflected in the returned entry. Mutation acks already carry it in their returned event, and the mount's local store pairs each read with a version cursor advanced under the same lock as the store write.

Mount side, every handle records the version its entry reflects, and one rule replaces the per-site reasoning: state at or below the handle's version must not be installed. Invalidations apply the event's own entry with the event's version — no lookup, so no transient-failure window — except under a cached parent, where the store entry is the ordered merge of the event and anything applied since. Opens version fresh handles from the lookup response; flushes, truncates, and remote-cache downloads version from their acks. Directory builds version the completed directory at the listing snapshot and re-invalidate buffered events at that version.

The test suite replays every race found in the previous attempt as a deterministic reproduction against a fake filer: rollback of a newer local flush in queued, cached, and read-through configurations; stale leftovers under uncached parents; the build window including abort; handles opened after an event was queued; events landing mid-lookup; and undelivered events at remote-cache time across a filer failover.

Known residuals, inherent to per-filer clocks: multi-filer clock skew can make a fence conservatively skip one event until the next arrives, and eventless acks from pre-upgrade filers leave the handle unversioned, degrading to today's behavior. The server-side copy path installs its result unversioned and converges when the copy's own committed event is delivered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented open file handles and cached entries from being overwritten or rolled back by older metadata events by fencing refreshes with filer log/version timestamps (including save/update acknowledgements, renames/deletes, queued notifications, and remote-entry downloads).
  * Improved metadata cache version tracking so invalidations and buffered directory-build events are applied consistently.
  * Added filer response log-position timestamps to better support consistent acknowledgement handling.
* **Tests**
  * Expanded coverage for versioned invalidation, rollback fencing, buffering/directory build behavior, concurrent open races, and remote-cache failover scenarios.


chrislusf (commented):
Read the whole thing against the previous attempt. The framing is right and worth stating plainly: the earlier fixes each invented an ordering signal beside the state it guarded, so every approximation left a window. Tying the version to the filer log position and shipping it with the value removes that class, and it also subsumes the two findings I raised on #10395 — the store-vs-event choice is now gated on `storeVersionTsNs >= eventTsNs` (so the mid-build stale-store read can't win), and read-through paths are no longer silently trusted.

I verified the fence argument holds within one filer: mutations hold the exclusive path lock across write and notify, so a stamp taken under the shared lock cannot land between a write and its event. Three comments below — one on where that argument stops holding, one on cost, one minor.

weed/server/filer_grpc_server.go
```diff
@@ -23,24 +23,39 @@ import (
...
 	"google.golang.org/grpc/status"
 )
 
+// fencedFindEntry reads an entry with an exact log-position fence: the stamp
+// and the read share the path lock mutations hold across write and notify, so
+// every event at or below the fence is in the entry and none above it are.
+// Any read whose result a client versions must go through here.
+func (fs *FilerServer) fencedFindEntry(ctx context.Context, path util.FullPath) (entry *filer.Entry, logTsNs int64, err error) {
+	pathLock := fs.entryLockTable.AcquireLock("fencedFindEntry", path, util.SharedLock)
+	logTsNs = time.Now().UnixNano()
```
chrislusf, line 36: The fence's exactness is a single-filer property: `entryLockTable` is in-process and `time.Now()` is this filer's clock, so with several filers the ordering between a fence from filer A and an event from filer B is whatever their clocks say. The PR body is honest about this, but I'd argue the residual is sharper than "conservatively skip one event": a skip leaves the handle holding state the event was supposed to correct, and since the subscription cursor has moved on, nothing retries — which is the same unhealable-staleness shape as #10394 itself, just with a window bounded by skew instead of by RPC latency. The opposite direction (fence behind, event re-applied) is harmless, absorbed by the base-equality check.

Since the two directions are not equally costly, it seems worth making the comparison refuse to guess. If the fence carried the identity of the filer that stamped it (alongside `log_ts_ns`), the mount could compare positions only within one clock domain and, when an event's originating signature differs from the filer that issued the fence, fall through to applying it — trading a benign re-apply for the stale-forever case. Events already carry `Signatures`, so the originating filer should be recoverable on the mount side; is there a reason not to pin the domain explicitly rather than let skew decide?

weed/mount/meta_cache/meta_cache.go
```diff
@@ -359,14 +386,171 @@ func (mc *MetaCache) FindEntry(ctx context.Context, fp util.FullPath) (entry *fi
...
+		if err := mc.localStore.KvDelete(ctx, key); err != nil {
+			glog.V(1).Infof("delete version record %s: %v", string(key), err)
+		}
+	}
+}
+
+// stampChildrenVersionLocked versions every listed child at the snapshot the
+// listing read. A zero snapshot (pre-upgrade filer) clears instead, or a
+// rebuilt entry reactivates its predecessor's stale record.
+func (mc *MetaCache) stampChildrenVersionLocked(ctx context.Context, dirPath util.FullPath, snapshotTsNs int64) {
```
chrislusf: This writes one KV record per direct child, inside `mc.Lock()`, on every directory build completion — so a 100k-entry directory becomes 100k `KvPut`s plus a full `ListDirectoryEntries` on the apply loop's critical path, blocking every other cache operation for the duration. That is a visible stall on exactly the directories where builds already hurt.

It also looks avoidable: `dirAbsenceFloors[dirPath]` is set to the same `snapshotTsNs` two lines up, and the only thing the per-child records add over it is a per-entry override. `entryVersionBlocksLocked` could treat the directory floor as the presence floor too — for an entry that exists with no record of its own, fall back to `dirAbsenceFloors[dir]` — and then every child a later event touches gets its own record naturally via `setEntryVersionLocked`, while untouched children cost nothing. Same fencing, one map write instead of O(children) store writes.

Related, on lifetime: tombstones are pruned only at build completion or eviction, and a recreate at the same name overwrites the key, so churn on repeating names is self-limiting — but a cached directory that accumulates *distinct* deleted names (unique temp files, rotated logs) grows tombstones without bound until something rebuilds it. Worth a word on the intended ceiling.

weed/mount/meta_cache/meta_cache.go
```diff
@@ -359,14 +386,171 @@ func (mc *MetaCache) FindEntry(ctx context.Context, fp util.FullPath) (entry *fi
...
+// reflected at fp. A tombstone fences with no entry present. A plain record
+// only blocks while its entry exists — records linger after a bulk folder wipe
+// and must not fence a recreate. For an absent entry the absence floor always
+// speaks, whatever older record remains.
+func (mc *MetaCache) entryVersionBlocksLocked(ctx context.Context, fp util.FullPath, tsNs int64) bool {
+	recordTsNs, tombstone := mc.getEntryVersionRecordLocked(ctx, fp)
+	if tombstone && recordTsNs >= tsNs {
+		return true
+	}
+	if _, err := mc.localStore.FindEntry(ctx, fp); err == nil {
```
chrislusf: Minor: this presence probe calls `localStore.FindEntry` directly, so it does not apply the TTL check `FindEntryWithVersion` does a few lines up. A TTL-expired entry therefore still counts as present here, and the path is judged by its (possibly older) version record rather than the directory absence floor — the floor is the more accurate answer once the entry has logically expired. Narrow, but the two readers disagreeing about what "exists" means is the kind of thing that gets rediscovered later.

weed/server/filer_grpc_server.go
chrislusf: No reason not to — you are right that the two directions are not symmetric, and the asymmetry is the whole argument. Implemented in 9d679cdc8.

Fences now carry `log_signature` alongside `log_ts_ns`, and the handle records it with the position. An event is fenced out only when the filer that logged it is the one that stamped the fence; every other event is applied. As you expected, the originating filer is recoverable without a new field on the event — `NotifyUpdateEvent` appends the logging filer's own signature to `Signatures`, and the aggregator relays events intact, so its presence is the domain test.

One scoping decision worth flagging: positions taken from *events* (rather than an RPC fence) carry signature 0 and keep comparing as before. Disambiguating which element of `Signatures` is the filer versus the client is not reliable, and it is not needed — a single filer's events reach the mount in log order over the subscription, so ordering there is already real rather than clock-derived. The signature therefore only pins the case where it was actually being guessed: fence-versus-event.

`TestEventFromAnotherFilerIsNotFencedByForeignClock` covers both halves — a cross-filer event below the fence applies, the fencing filer's own older event is still skipped.

The invalidation callback became a struct in the same commit; it was about to take a fifth positional parameter.

weed/mount/meta_cache/meta_cache.go
chrislusf: Agreed on both counts — done in c78fc9300.

The per-child stamping is gone. `dirAbsenceFloors` became `dirVersionFloors` and now answers for present children too: a child without a record of its own is versioned at its directory's listing snapshot, and only earns a record when a later event touches it. One map write per build replaces the `ListDirectoryEntries` plus O(children) `KvPut`s, with the same fencing — `TestBuildFloorVersionsChildrenWithoutPerChildRecords` asserts a built child has no record, still reads back at the snapshot, and still fences an event the snapshot covered.

The unversioned-listing case is the one place that still writes per child: with no snapshot there is no floor to fall back on, so the children's stale records are cleared instead, or a re-inserted entry would inherit its predecessor's version.

Tombstone ceiling is now stated at `setEntryTombstoneLocked`: a recreate at the same name overwrites the key, so repeating names are self-limiting; distinct names each hold a record until the directory is rebuilt or evicted, which prunes everything at or below the new snapshot. So the bound is distinct names deleted in a cached directory since its last listing.

weed/mount/meta_cache/meta_cache.go
chrislusf: Fixed in c78fc9300. The probe goes through `entryExistsLocked`, which applies the same TTL expiry the read path does, so the two readers now agree on what "exists" means.

Worth noting the case it actually changes, since the floor rework landed alongside: when the stale record is *older* than the directory floor the two answers now coincide anyway, because a path's version is `max(record, floor)`. The behaviour only diverges when the record is newer than the floor — there, treating an expired entry as present let a record describing content that has logically vanished fence a legitimate event. `TestExpiredEntryIsJudgedByDirectoryFloor` pins that direction (record 5000, floor 3000, event 4000 applies).

weed/server/filer_grpc_server_remote.go
```diff
@@ -297,6 +304,30 @@ func (fs *FilerServer) doCacheRemoteObjectToLocalCluster(ctx context.Context, re
...
+	}
+	if !filer.EqualEntry(current, entry) {
+		// The entry changed during the download; its writer supersedes the
+		// cached content. Return the current state, fenced at this read.
+		fs.filer.DeleteUncommittedChunks(ctx, chunks)
+		resp.Entry = current.ToProtoEntry()
+		resp.LogTsNs = commitLogTsNs
+		return resp, nil
+	}
+
```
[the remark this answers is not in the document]
chrislusf: Valid — fixed in c78fc9300. The commit-time re-read now returns `filer_pb.ErrNotFound` unwrapped when the object was deleted during the download, matching the initial read, so callers still surface a 404. The uncommitted chunks are cleaned up on both paths.

The test-assertion comment is fixed in the same commit (`got[0].tsNs == 78` is now asserted alongside `got[1]`).

PR #10419, opened 2026-07-24
s3: stop listing prefixes whose objects are all delete-marked

### LINKED ISSUE #10416: S3: delimited ListObjectsV2 returns a CommonPrefix for a prefix that has no current-version object (versioned bucket)
Hi!

First of all, thank you for the excellent software! 

I'm in the process of migrating our tests from Ceph to SeaweedFS and I found an issue that causes a regression.

**Version:** 4.40, official `chrislusf/seaweedfs` image
**Component:** S3 API gateway

In a versioning-enabled bucket, deleting the only object under a prefix creates a delete marker, so there is no longer any current-version object under that prefix. A `ListObjectsV2` scoped inside the prefix correctly returns no contents. However, a **delimited** `ListObjectsV2` at the parent still returns the prefix in `CommonPrefixes`.

AWS S3 (and Ceph RGW) do not do this: in the current-version view, a `CommonPrefix` is only returned when at least one current, non-delete-marked object shares it. SeaweedFS appears to derive `CommonPrefixes` from the filer directory tree, which stays populated by the retained version history (old versions + delete marker), so the prefix never disappears.

I see from #9884 that there's a good chance this won't be considered an issue, but I figured it was best to try. In our case, the phantom prefixes surface backups that have already been deleted. For some operations this is just a nuisance (https://github.com/pgbackrest/pgbackrest/issues/2769) but in others it causes a hard error (https://github.com/pgbackrest/pgbackrest/issues/2792). These issues are both against MinIO but they apply equally to SeaweedFS.

Our only recourse (that I can see) is to scan a level down and build our own common prefixes, but it seems like it would be better (and faster) to do this in SeaweedFS rather than have tools that need it roll their own.

Thank you!

chrislusf:
Fixes #10416.

In a versioning-enabled bucket, deleting the last object under a prefix writes a delete marker and keeps the version history, so the filer directory survives with nothing a current-version listing would return. A delimited `ListObjectsV2` at the parent still reported the path in `CommonPrefixes`, because SeaweedFS derives common prefixes from the directory tree rather than from the keys the listing returns. A listing scoped inside the prefix correctly came back empty, so the two answers disagreed and deleted backups stayed visible.

Before:

```
$ aws s3api delete-object --bucket b --key backup/20260101/manifest
$ aws s3api list-objects-v2 --bucket b --prefix backup/
{ }                                            # correct
$ aws s3api list-objects-v2 --bucket b --prefix backup/ --delimiter /
{ "CommonPrefixes": [ { "Prefix": "backup/20260101/" } ] }    # AWS and Ceph RGW return nothing
$ aws s3api list-objects-v2 --bucket b --delimiter /
{ "CommonPrefixes": [ { "Prefix": "backup/" } ] }             # and it propagates up
```

After, both delimited listings are empty, and `list-object-versions` still shows the version and its delete marker. Removing the delete marker brings the prefix back.

## What changed

`doListFilerEntries` now probes a plain directory before handing it to the delimiter roll-up: a directory that holds entries but none that a current-version listing returns is neither a `CommonPrefix` nor a path the trailing-slash directory probe answers for. The probe stops at the first key it finds and recurses into subdirectories.

Deciding a `.versions` directory is free: every write that changes an object's current version already stamps `Seaweed-X-Amz-Latest-Version-Is-Delete-Marker` onto that directory entry, and the probe's own `ListEntries` carries it. No version file is read and no pointer rescan runs on the list path. A `.versions` entry with no stamp leaves the current version unknown, which keeps the prefix.

Boundaries kept as they are:

- Empty directories still list as they do today. Mount and `mkdir` create them and empty-folder cleanup owns their lifetime, and the `prefix=dir/&delimiter=/` probe added in #9615 for hadoop-aws still answers for them.
- Directory key objects (`PutObject` on a trailing-slash key) are reported regardless of what happened below them; the key exists in its own right.
- `ListObjectVersions` is untouched.

## Cost

One extra `ListEntries` per candidate prefix, answered by its first entry in the normal case. Gated on the bucket having versioning configured, since a bucket that never had versioning cannot reach this state. Measured on a local single-node cluster, bucket root with 500 prefixes, `delimiter=/`:

| prefixes | non-versioned | versioned before | versioned after |
| --- | --- | --- | --- |
| 500 | 0.9 ms | 0.9 ms | 18 ms |
| 1000 | 1.4 ms | 1.4 ms | 35 ms |

The probes are serial, so on a remote filer the term that matters is one round trip per prefix returned. A page is bounded by `max-keys`; a page that also steps over deleted prefixes costs a probe for each of those too, bounded by the entry budget below.

A subtree that is entirely delete-marked costs a walk of that subtree, capped by a per-request budget of 10000 probed entries; past the budget prefixes are reported, which is what happened before this check existed.

Only the delete-marker stamp hides a prefix. An unstamped `.versions` directory reports, so a pointer that has not reached the filer serving the list cannot make a live prefix disappear.

## Tests

- `weed/s3api/s3api_object_handlers_list_deleted_prefix_test.go`: deleted prefix hidden at both levels, live prefix kept, no directory marker for the trailing-slash probe, empty directory unchanged, directory key object unchanged, both stamp values, an unstamped `.versions` directory, gate off for non-versioned buckets, paging steps over hidden prefixes.
- `test/s3/versioning/s3_deleted_prefix_test.go`: end-to-end against a running gateway, including that the version history survives and that deleting the delete marker restores the prefix.

## Known gap, not addressed here

An object written *before* versioning was enabled and deleted afterwards keeps its prefix listed, because the bare pre-versioning file is still returned as a current key: `list-objects-v2` returns it, `list-object-versions` reports both the null version and the delete marker with `IsLatest: true`, and only `HEAD`/`GET` agree that it is gone. That reproduces identically without this change, so the prefix staying visible is this listing agreeing with itself rather than a hole in the probe. Worth its own fix.

## Summary by CodeRabbit

- **Bug Fixes**
  - Refined S3 delimiter listings for versioned buckets: prefixes backed only by deleted/version-history entries are no longer returned as `CommonPrefixes` or emitted as directory markers.
  - Live prefixes, explicit directory-key objects, and genuine empty directories remain visible.
  - Hidden/deleted prefixes no longer affect `maxKeys` pagination, and delimiter results now update correctly as delete markers are created/removed.

- **Tests**
  - Added and expanded coverage for deleted-prefix `CommonPrefixes` behavior, directory-marker suppression, version-stamp/unstamped handling, non-versioned skipping, and pagination/truncation edge cases.

### DIFF
--- a/test/s3/versioning/s3_deleted_prefix_test.go
+++ b/test/s3/versioning/s3_deleted_prefix_test.go
@@ -1,0 +1,106 @@
+package s3api
+
+import (
+	"context"
+	"testing"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDeletedPrefixLeavesNoCommonPrefix covers the versioned bucket case where the only
+// object under a prefix is deleted. The delete marker and the version history stay, so
+// the directory survives in the filer, but a current-version listing has no key under
+// that path and AWS reports no CommonPrefix for it.
+func TestDeletedPrefixLeavesNoCommonPrefix(t *testing.T) {
+	client := getS3Client(t)
+	bucketName := getNewBucketName()
+
+	createBucket(t, client, bucketName)
+	defer deleteBucket(t, client, bucketName)
+	enableVersioning(t, client, bucketName)
+
+	putObject(t, client, bucketName, "backup/20260101/manifest", "first backup")
+	putObject(t, client, bucketName, "backup/20260102/manifest", "second backup")
+
+	listPrefixes := func(prefix string) []string {
+		t.Helper()
+		input := &s3.ListObjectsV2Input{
+			Bucket:    aws.String(bucketName),
+			Delimiter: aws.String("/"),
+		}
+		if prefix != "" {
+			input.Prefix = aws.String(prefix)
+		}
+		resp, err := client.ListObjectsV2(context.TODO(), input)
+		require.NoError(t, err)
+		prefixes := make([]string, 0, len(resp.CommonPrefixes))
+		for _, p := range resp.CommonPrefixes {
+			prefixes = append(prefixes, *p.Prefix)
+		}
+		return prefixes
+	}
+	listKeys := func(prefix string) []string {
+		t.Helper()
+		input := &s3.ListObjectsV2Input{
+			Bucket:    aws.String(bucketName),
+			Delimiter: aws.String("/"),
+		}
+		if prefix != "" {
+			input.Prefix = aws.String(prefix)
+		}
+		resp, err := client.ListObjectsV2(context.TODO(), input)
+		require.NoError(t, err)
+		keys := make([]string, 0, len(resp.Contents))
+		for _, c := range resp.Contents {
+			keys = append(keys, *c.Key)
+		}
+		return keys
+	}
+	deleteKey := func(key string) {
+		t.Helper()
+		_, err := client.DeleteObject(context.TODO(), &s3.DeleteObjectInput{
+			Bucket: aws.String(bucketName),
+			Key:    aws.String(key),
+		})
+		require.NoError(t, err)
+	}
+
+	assert.ElementsMatch(t, []string{"backup/20260101/", "backup/20260102/"}, listPrefixes("backup/"))
+
+	deleteKey("backup/20260101/manifest")
+	assert.Equal(t, []string{"backup/20260102/"}, listPrefixes("backup/"),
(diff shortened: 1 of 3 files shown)


dwsteele:
Thank for looking at this so quickly!

> ListObjectVersions is untouched.

Exactly what I would expect.

> An object written before versioning was enabled and deleted afterwards keeps its prefix listed

This seems reasonable.

> A subtree that is entirely delete-marked costs a walk of that subtree, capped by a per-request budget of 10000 probed entries; past the budget prefixes are reported, which is what happened before this check existed.

I tested this and it works exactly as specified, but I think it would be good if it were possible to increase the budget via configuration or remove it entirely. I think it makes sense for most use cases but in our case it is better to be accurate than fast and the timings above are fast enough that I think it is workable.

I'm happy to open a PR for that if you think it makes sense.

PR #10434, opened 2026-07-25
filer: fix credential handling on the proxyChunkId path

chrislusf:
Rebased onto master now that #10436 has landed.

## The bug

`filerHandler` dispatches the `?proxyChunkId=` branch and returns before the JWT gate, so whatever credential the proxy attaches on the caller's behalf is reachable without authentication.

The proxy attached a token from `maybeGetVolumeReadJwtAuthorizationToken`, which fell back to the **write** signing key when `jwt.signing.read.key` was unset. That is the configuration `scaffold/security.toml` steers a filer into — it notes read JWTs are "only supported with master+volume setup. Filer does not support this mode."

So with `jwt.signing.key` set and no read key, an anonymous `DELETE /?proxyChunkId=<vid>,<needle><cookie>` reached the volume server holding a write-key token scoped to exactly that fid, and the volume server honored it. An identical `DELETE` on any other path returns 401.

Moving the dispatch below the JWT gate would have broken every `weed mount -filerProxy` upload: those carry a volume JWT signed with `jwt.signing`, not `jwt.filer_signing`. So the fix is to stop the minting instead.

## Commits

**`claim the base fid when minting a volume read token`** — `GenJwtForVolumeServer` stamped the fid verbatim, but the volume server strips a trailing `_N` delta suffix before comparing. A token for a batch-assigned `3,01637037d6_1` was checked against `3,01637037d6` and never matched, so reading such a chunk returned 401 wherever a read key was configured. Pre-existing. Reuses the digit-guarded `baseFileId` helper from #10436.

**`don't mint a volume write token for an anonymous proxy caller`** — read tokens are signed with the read key only, and minted only for reads. The write-key fallback bought nothing on a read anyway: a volume server enforces read JWTs solely when that same key is set, so whenever the fallback fired the read was unchecked regardless. On a read with nothing to mint, the caller's `Authorization` is now dropped rather than relayed — on that path it is a *filer* credential, and forwarding it would hand a volume server a token it never used to see.

**`keep proxied writes out of the read concurrency semaphore`** — the semaphore is named and documented for reads, but was applied to every method. A write queued behind sixteen in-flight reads can wait past the 10s default expiry of the AssignVolume token it carries; `shouldReassignUpload` treats the resulting 401 as final, so the uploader replays the expired token instead of re-assigning. Only reachable once the filer stopped re-minting after the wait.

## Known residual gap

**This does not fully close anonymous proxied writes.** When `jwt.signing.key` is empty but `jwt.filer_signing.key` is set, the write still lands: the branch dispatches ahead of the filer's gate and the volume server has no write auth of its own to fall back on.

I could not close that without breaking `weed mount -filerProxy`, whose uploads carry only a volume JWT and so have no credential the filer gate would accept in that configuration. Closing it properly means giving the proxy write path a real filer-level credential — a design change rather than a patch, and the same shape as the fail-closed gate that had to be reverted in #9472. #10436 narrows the reachable surface to valid fids but does not close this.

## Tests

`proxyTestVolume` records request arrival separately from the header, so the negative assertions can't pass vacuously when the proxy never reaches the volume server. POST leads the write table — it is the only method any in-tree proxied uploader sends (`GenUploadUrlProxy`, mount, mq broker, filer sink) — and HEAD is covered on the read side. Also asserted: the minted token replaces rather than appends to a caller-supplied one; the semaphore entry is dropped on cleanup so `-count=N` doesn't grow the map.

Each of the five behaviours is mutation-tested: reverting any one of them individually makes its guarding test fail.

`go build ./weed/...`, `go vet ./weed/server/`, and `./weed/server/ ./weed/security/ ./weed/storage/needle/` all pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved proxy authentication and `Authorization` forwarding for both read and write requests, enforcing read-only credentials only for `GET`/`HEAD`.
  * Proxied reads now use minted read tokens when configured; otherwise the caller `Authorization` is dropped.
  * Proxied writes preserve the caller’s credentials and bypass read throttling.
  * Strengthened proxy chunk identifier validation and ensured request forwarding copies caller query parameters except `proxyChunkId`.

* **Tests**
  * Added unit tests validating credential behavior, read-only enforcement, semaphore bypass for writes, token correctness, and chunk id validation.

### DIFF
--- a/weed/server/filer_server_handlers_proxy.go
+++ b/weed/server/filer_server_handlers_proxy.go
@@ -51,7 +51,7 @@ func releaseProxySemaphore(host string) {
 }
 
 // isProxyReadMethod reports whether a proxied request only reads. Everything
-// else is treated as a write for credential purposes.
+// else is treated as a write for both credential and concurrency purposes.
 func isProxyReadMethod(method string) bool {
 	return method == http.MethodGet || method == http.MethodHead
 }
@@ -139,14 +139,20 @@ func (fs *FilerServer) proxyToVolumeServerURL(w http.ResponseWriter, r *http.Req
 		return
 	}
 
-	// Limit concurrent requests per volume server to prevent overload
-	volumeHost := proxyReq.URL.Host
-	if err := acquireProxySemaphore(ctx, volumeHost); err != nil {
-		glog.V(0).InfofCtx(ctx, "proxy to %s cancelled while waiting: %v", volumeHost, err)
-		w.WriteHeader(http.StatusServiceUnavailable)
-		return
+	// Limit concurrent reads per volume server to prevent overload. Writes are
+	// deliberately exempt: a proxied write carries the caller's AssignVolume
+	// token, which expires 10s after the assign by default, so queueing one here
+	// can push it past expiry and turn it into a 401 the uploader does not
+	// re-assign on.
+	if isProxyReadMethod(r.Method) {
+		volumeHost := proxyReq.URL.Host
+		if err := acquireProxySemaphore(ctx, volumeHost); err != nil {
+			glog.V(0).InfofCtx(ctx, "proxy to %s cancelled while waiting: %v", volumeHost, err)
+			w.WriteHeader(http.StatusServiceUnavailable)
+			return
+		}
+		defer releaseProxySemaphore(volumeHost)
 	}
-	defer releaseProxySemaphore(volumeHost)
 
 	proxyReq.Header.Set("Host", r.Host)
 	proxyReq.Header.Set("X-Forwarded-For", r.RemoteAddr)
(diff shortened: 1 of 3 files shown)


chrislusf:
Rebased onto the updated #10436. Note for anyone reading the earlier bot reviews on this PR: they were run against `620ba176`, the pre-rework commit, and do not describe the current state. That version had two regressions of its own — it dropped the caller's filer JWT onto volume servers on the read path, and it let a proxied write queue behind the read semaphore until its 10s AssignVolume token expired into a non-retryable 401. Both are fixed here.

`baseFileId` now lives in #10436 with the digit-guarded suffix strip, so the first commit here carries only the minting change that depends on it.

Each of the six behaviours in this stack is mutation-tested: reverting any one of them individually makes its guarding test fail.

The residual gap in the PR description still stands — anonymous proxied writes remain possible when `jwt.signing.key` is empty and `jwt.filer_signing.key` is set, because the branch dispatches ahead of the filer gate and the volume server has no write auth to fall back on. Closing that needs a real filer-level credential on the proxy write path, which is a design change rather than a patch.

PR #10441, opened 2026-07-25
filer: make the redis connection settings configurable

### LINKED ISSUE #10438: Redis connection pool issue with redis2_sentinel backend
Hello,

I am using SeaweedFS filer v4.17 with redis2_sentinel as metadata store. Recently, we observed Redis connection pool related errors in filer logs during Redis Sentinel master changes.

Environment:
- SeaweedFS: v4.17
- Filer store: redis2_sentinel
- Redis: 7.2.4
- Redis Sentinel: 3 nodes cluster

We have verified the following:
1. Redis Sentinel status is normal.
2. Redis master information reported by Sentinel is correct.
3. Filer configuration is using redis2_sentinel.
4. Redis client related configurations have been checked, including Sentinel configuration, and no abnormal configuration was found.

The observed scenario:
1. Redis master was running normally.
2. Redis Sentinel changed the master node (through failover).
3. The filer process continued running without restart.

After the master switch, filer occasionally reported:
`redis: connection pool: failed to get a usable connection after X attempts`

We would like to understand the expected behavior of the redis2_sentinel backend in this situation.
Questions:
1. When Redis Sentinel switches the master, how does SeaweedFS handle existing Redis connections in the connection pool?
2. Are Redis connections created before the master switch automatically detected as invalid and removed from the pool?
3. If old Redis connections remain in the pool temporarily after the master switch, can they cause errors such as "failed to get a usable connection"?
4. After Redis Sentinel master switch, should filer recover automatically without restart?

We would appreciate any suggestions on whether this behavior is expected or if there are additional configurations we should check.

chrislusf:
Fixes #10438

The sentinel stores hardcoded `ReadTimeout: 30s`, `WriteTimeout: 5s`, `MinRetryBackoff: 100ms` and `MaxRetryBackoff: 1m`. go-redis does retire the pooled connections to the demoted master on `+switch-master`, but anything already holding or picking one of those connections blocks for the full read timeout first, and `PoolTimeout` defaults to read timeout + 1s, so under load every other caller queues behind them for ~31s. That is what a filer looks like when it is "not recovering" from a sentinel failover, and there was no way to shorten it.

The other redis stores took the go-redis defaults with no knobs at all.

So: drop the hardcoded values and read the dial, timeout and pool settings from the store section, for every redis store (redis, redis2, redis3 and their cluster/sentinel variants). Each key keeps the go-redis default when unset, so the only behavior change is on the two sentinel stores, which now get the same 3s read timeout as everyone else and can be tuned back up with `read_timeout_millisecond = 30000` if a deployment needs it.

New keys, accepted by every redis store section:

```
max_retries                        # 0 is three retries, -1 disables retrying
min_retry_backoff_millisecond
max_retry_backoff_millisecond
dial_timeout_millisecond
read_timeout_millisecond
write_timeout_millisecond
pool_size
pool_timeout_millisecond
min_idle_conns
max_idle_conns
conn_max_idle_time_seconds
conn_max_lifetime_seconds
```

## Summary by CodeRabbit

* **New Features**
  * Added centralized Redis connection tuning options (timeouts, retries, connection pooling, idle connections, and connection lifetime).
  * Redis standalone, cluster, and sentinel configurations now consistently apply these settings.
  * Unset/negative values preserve existing Redis client default behavior.
* **Documentation**
  * Documented how to configure the connection tuning block and its scope across Redis store sections.
* **Tests**
  * Added coverage for reading and applying Redis tuning settings, including default-preservation and negative-value handling.

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -230,6 +230,23 @@ client_cert_path = ""
 client_key_path = ""
 # This changes the data layout. Only add new directories. Removing/Updating will cause data loss.
 superLargeDirectories = []
+# Connection tuning, accepted by every redis filer store section. Each key keeps the go-redis
+# default when left at 0, and a timeout cannot be turned off - a filer that never times out a read
+# is what stalls a failover. Shorter read timeouts make the filer notice a dead redis sooner, which
+# matters most for the sentinel store: every pooled connection to the old master has to time out
+# before a failover is absorbed.
+# max_retries = 0                        # 0 is three retries, -1 disables retrying
+# min_retry_backoff_millisecond = 0      # default 8
+# max_retry_backoff_millisecond = 0      # default 512
+# dial_timeout_millisecond = 0           # default 5000
+# read_timeout_millisecond = 0           # default 3000
+# write_timeout_millisecond = 0          # default 3000, follows read_timeout_millisecond
+# pool_size = 0                          # default 10 per CPU
+# pool_timeout_millisecond = 0           # default read_timeout_millisecond + 1000
+# min_idle_conns = 0
+# max_idle_conns = 0
+# conn_max_idle_time_seconds = 0         # default 1800
+# conn_max_lifetime_seconds = 0          # default unlimited
 
 [redis2_sentinel]
 enabled = false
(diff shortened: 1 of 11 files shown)


chrislusf:
Both addressed in the amended commit.

- scaffold: the tuning block now says it applies to every redis filer store section, not the ones below it.
- disabling a timeout: intentionally not reachable. go-redis spells that as a negative nanosecond count, and a filer that never times out a read is exactly what leaves a failover hanging - so non-positive values keep the go-redis default. Now stated in both the helper and the scaffold.

PR #10442, opened 2026-07-26
volume: fix EC decode/reconstruct index locality under -dir.idx

chrislusf:
EC→replicated decode fails under `-dir.idx` (and on multi-disk layouts) with `volume not found on disk`.

## Root cause
The reconstruct (`VolumeEcShardsToVolume`) rebuilds the `.dat` on the data disk but wrote the rebuilt `.idx` to the shared `-dir.idx` directory. The on-demand `VolumeMount` scans only the *data* directory and matches on `.idx`/`.vif`. With no `.idx` there it matched the volume's leftover EC `.vif` and skipped the volume as EC metadata, so the decoded volume never mounted.

## Fix
- **Resolve the EC `.ecx` local-first** — prefer the copy co-located with the shards over the shared `-dir.idx` copy, with a non-empty preference so a 0-byte local stub still yields to a valid sibling (preserves the cross-disk fallback and `TestMountEcShards_ZeroByteEcxIsIgnored`).
- **Co-locate the rebuilt `.idx`** with the `.dat` at the end of the reconstruct so the mount finds it; sweep `.ecx`/`.ecj` from both the data and index directories on `Destroy` so a stale copy cannot re-mount as a phantom EC volume.
- **New `VolumeConsolidateIndex` RPC** — once the EC shards are deleted, unmount the volume, move `.idx`/`.sdx` from the data disk back to the `-dir.idx` directory (rename, with a copy+remove fallback since `-dir.idx` is typically a separate filesystem), and remount. A no-op without `-dir.idx`. This lets an operator's index tier stay on its own media across an EC↔replicated transition.

The `.idx` for a normal volume was already resolved local-first (`volume_loading.go`); this brings the EC `.ecx` and the reconstruct path in line.

## Tests
`go build ./weed/...` and `go test ./weed/storage/...` pass (including the EC index tests). Exercised end-to-end across single-disk, `-dir.idx`, and multi-disk layouts in the enterprise tiering E2E suite.

## Summary by CodeRabbit

* **New Features**
  * Added an administrative gRPC operation to consolidate a volume’s index into the configured index directory.
* **Bug Fixes**
  * Improved EC index selection to prefer non-empty local `.ecx`, with correct fallback behavior for empty stubs.
  * Prevented phantom EC remounts by cleaning `.ecx`/`.ecj` from both data and shared index locations.
  * Improved alignment of rebuilt EC `.idx` files so mounted volumes find the correct index.
* **Tests**
  * Added unit/integration coverage for index consolidation and EC mount/destroy cleanup behavior.


chrislusf:
Thanks — addressed in `68e45fca`:

- **Maintenance-mode guard (CodeRabbit):** `VolumeConsolidateIndex` now calls `CheckMaintenanceMode()` after auth, matching `VolumeConfigure` and the other mutating volume RPCs.
- **Partial-move orphan on cross-device copy (greptile):** `RenameOrCopyFile` now rolls the copy back (removes `dst`) if `os.Remove(src)` fails on the EXDEV path, so a failed move never leaves two divergent copies — only the original `src` survives, and the reload reopens against it.
- **Log the recovery failure (CodeRabbit):** the specific `s.MountVolume` recovery that comment referenced was replaced earlier in this PR by the in-place `RelocateIndexTo` (no more unmount/remount, so no read gap). Its equivalent recovery — `v.load(...)` reopening against the old dir — now logs via `glog.Errorf` instead of discarding the error.

PR #10456, opened 2026-07-27
azure: authenticate with Entra ID instead of a storage account key

chrislusf:
Azure Blob backup and remote storage both demanded an account name plus an account key. Fleets that run many edge clusters have to distribute and rotate that key everywhere, which is exactly the thing workload identity exists to avoid.

Leaving the account key empty now authenticates through the azidentity chain, so a workload identity or a managed identity carries the authorization and the storage account grants it a data-plane role. `client_id` (`-azure.client_id` for `remote.configure`) pins a user-assigned identity when the host offers more than one: with a webhook-projected federated token that is a workload identity, otherwise a managed identity. Both call sites share one `NewAzBlobClient` helper, and azidentity was already a dependency of the Key Vault provider.

Unit tests cover the credential selection without needing an Azure account, since every azidentity constructor is offline-safe. The service URL is still hardcoded to `blob.core.windows.net`, so sovereign clouds remain unsupported.

## Summary by CodeRabbit

* **New Features**
  * Added Azure identity-based authentication for remote storage and Azure replication, including Entra ID, managed identity, and workload identity support.
  * Introduced optional `azure.client_id` / `azure_client_id` to support user-assigned identity authentication.
  * Environment variables are now used to fill in Azure account settings, with shared-key authentication still supported.

* **Documentation**
  * Updated `remote.configure` help and added an Azure identity configuration example.

* **Tests**
  * Added coverage for Azure credential selection and Azure remote storage client initialization paths.

### DIFF
--- a/seaweed-volume/proto/remote.proto
+++ b/seaweed-volume/proto/remote.proto
@@ -26,6 +26,7 @@ message RemoteConf {
 
   string azure_account_name = 15;
   string azure_account_key = 16;
+  string azure_client_id = 17;
 
   string backblaze_key_id = 20;
   string backblaze_application_key = 21;
--- a/weed/command/scaffold/replication.toml
+++ b/weed/command/scaffold/replication.toml
@@ -59,7 +59,11 @@ is_incremental = false
 # experimental, let me know if it works
 enabled = false
 account_name = ""
+# leave account_key empty to authenticate with Entra ID instead of a shared key,
+# picking up a workload identity or managed identity from the environment. The
+# identity needs a data-plane role such as Storage Blob Data Contributor.
 account_key = ""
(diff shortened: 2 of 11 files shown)


chrislusf:
Pushed three fixes.

The account key no longer falls back to `AZURE_STORAGE_ACCESS_KEY` when a client id is configured, so an old mounted secret cannot quietly pull the request back to shared key auth. Account names are now held to Azure's own rule of 3 to 24 lowercase letters and digits, since anything else lands in the service URL and can move the authority to another host. The scaffold and the `-azure.client_id` help now say that workload identity also reads `AZURE_TENANT_ID` and `AZURE_FEDERATED_TOKEN_FILE`, which the webhook projects.

The `-azure.account_key` flag description was already corrected to say "account key" in this branch.

PR #10458, opened 2026-07-27
s3: list the buckets an attached IAM policy grants

chrislusf:
An identity authorized by an attached IAM policy got only its own buckets from ListBuckets. Grant a user `s3:ListBucket` on a bucket that someone else created and it can GetObject/ListObjectsV2 against it fine, but the bucket never appears in the listing S3 clients use to populate their bucket picker, so the user sees nothing.

The owner-index fast path only holds for an identity whose grants name every bucket it can reach, and the routing check treated any policy as unenumerable. Now the names come out of the policy: statements allowing `s3:ListBucket` on a concrete bucket ARN become candidates, and the existing per-bucket permission check still decides what is listed, so a Deny still hides a bucket. A policy that can reach a bucket it does not name — a wildcard resource, a policy variable, a NotResource, an STS session policy — falls back to the /buckets scan that evaluates the policy per bucket.

Verified against a mini cluster with the reported policy: the user now sees the granted bucket, an identity with `Resource: arn:aws:s3:::*` sees every bucket it can list, and one with an explicit Deny keeps that bucket hidden. Unit tests cover the routing, and the ListBuckets integration test gains two policy identities.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved S3 bucket listing/visibility when permissions come from IAM policy attachments, including group-based permissions.
  - Corrected interpretation of `Resource` entries (explicit, wildcard, `NotResource`, denies, and policy variables) to prevent incorrect bucket exposure.
  - Hardened bucket visibility evaluation for requests with session-token credentials and for incomplete policy data.
- **Tests**
  - Expanded list-buckets and policy evaluation coverage with new non-admin identities and additional scenarios (wildcards, `NotResource`, deny, variables, mixed-case actions, and missing policies).

### DIFF
--- a/test/s3/normal/s3_list_buckets_test.go
+++ b/test/s3/normal/s3_list_buckets_test.go
@@ -35,7 +35,15 @@ func TestListBucketsPaginationAndOwnerIndex(t *testing.T) {
 	    {"name": "carol", "credentials": [{"accessKey": "carol", "secretKey": "carol_secret"}],
 	     "actions": ["List:alice-b1"]},
 	    {"name": "bob", "credentials": [{"accessKey": "bob", "secretKey": "bob_secret"}],
-	     "actions": ["Read:alice-b1"]}
+	     "actions": ["Read:alice-b1"]},
+	    {"name": "dana", "credentials": [{"accessKey": "dana", "secretKey": "dana_secret"}],
+	     "policyNames": ["ReadAliceB1"]},
+	    {"name": "erin", "credentials": [{"accessKey": "erin", "secretKey": "erin_secret"}],
+	     "policyNames": ["ListEveryBucket"]}
+	  ],
+	  "policies": [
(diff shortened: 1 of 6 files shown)


chrislusf:
Consolidated, with one deviation: `authorizationRoute` runs on every S3 request and only needs to know whether any policy is attached, so it keeps its allocation-free short-circuit rather than building and measuring a name slice. The helper now backs the two callers that genuinely need the list, `authorizeWithIAM` and the ListBuckets enumeration, so group eligibility is decided in one place.

chrislusf:
Fixed the multipart lookup — `lowerMultipartActionSet` keys the set in lower case, so `S3:UPLOADPART` now inherits an `s3:PutObject` grant the way the IAM authorizer does. Covered by a new policy_engine test.

Skipped the paired suggestion to normalize case in `CompiledStatement.MatchesAction`: that one backs `EvaluateStatement`, so loosening it would change what bucket policies actually authorize, not just what this classifier reads. The asymmetry is deliberate — the classifier is allowed to over-match because every candidate is re-checked by the real permission check, while the evaluator is not.

PR #10482, opened 2026-07-29
s3: register the advertised ip with the master

chrislusf:
`weed mini -ip=localhost` still registered its S3 gateway with the master under a completely different address:

```
$ weed shell -master=localhost:9333 <<< cluster.ps
* s3 servers 1
  * 10.21.155.124:18333        <- not on any local interface
* filers 1
  * localhost:8888.18888       <- honored -ip
```

`S3ApiServerOption` had no advertise-ip field. The registration address came from `BindIp`, which mini sets to the wildcard, so it fell through to `util.DetectedHostAddress()` and picked whatever interface sorted first — here a VPN `utun` address that stopped routing once the tunnel dropped.

IAM changes are pushed to every registered S3 server over gRPC, so the consequence shows up on every admin mutation:

```
W propagating_store.go:95 failed to propagate change to s3 server 10.21.155.124:18333:
  rpc error: code = DeadlineExceeded desc = context deadline exceeded
```

Creating a user took 10.014s — the full propagation deadline — before failing the push, and `cluster.ps` and the admin UI both listed a node nothing could reach.

Worth noting what this is *not*: identities still arrive, because the `/etc/iam/identities` metadata subscription delivers them independently of the gRPC push. I confirmed a freshly created key works against a gateway whose push had just timed out. This is a latency and visibility bug, not credential loss.

## Change

Adds `S3ApiServerOption.Ip` and an `advertisedHost()` helper with an explicit precedence chain — advertised ip, then bind address, then auto-detect — and wires the parent `-ip` through `server.go`, `filer.go` and `mini.go`. Standalone `weed s3` gains an `-ip` flag defaulting to empty, so the existing fallback behavior is unchanged for anyone who does not set it.

## Verification

Before, with `-ip=localhost`: s3 registers `192.168.1.98:18633`. After: `localhost:18633`, matching the filer.

Pointing a gateway at a deliberately unreachable advertised address reproduces the 10s stall and the warning above; with the fix, user creation drops to 10ms, no propagate warnings, the new key works immediately on S3, and a wrong key is still rejected.

`TestAdvertisedHost` covers the precedence chain. `go build ./weed/...` clean; `weed/s3api` and `weed/command` suites pass.

## Summary by CodeRabbit

- **New Features**
  - Added an `-ip` option to control the S3 gateway’s advertised cluster address.

- **Bug Fixes**
  - Improved advertised-host selection and fallback behavior when `Ip`/bind addresses are wildcard or unspecified.
  - Embedded S3 gateways (server, filer, mini) now consistently use the cluster/server identity address for advertisement.

- **Tests**
  - Added unit coverage for advertised-host precedence and fallback logic.

### DIFF
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -257,6 +257,7 @@ func runFiler(cmd *Command, args []string) bool {
 	startDelay := time.Duration(2)
 	if *filerStartS3 {
 		filerS3Options.filer = &filerAddress
+		filerS3Options.ip = f.ip
 		if *filerS3Options.bindIp == "" {
 			filerS3Options.bindIp = f.bindIp
 		}
--- a/weed/command/mini.go
+++ b/weed/command/mini.go
@@ -1223,6 +1223,7 @@ func runMini(cmd *Command, args []string) bool {
 	miniFilerOptions.masters = pb.ServerAddresses(actualPeersForComponents).ToServiceDiscovery()
 	miniFilerOptions.ip = miniIp
 	miniFilerOptions.bindIp = miniBindIp
+	miniS3Options.ip = miniIp
 	miniS3Options.bindIp = miniBindIp
 	miniWebDavOptions.ipBind = miniBindIp
 	miniOptions.v.ip = miniIp
--- a/weed/command/s3.go
+++ b/weed/command/s3.go
@@ -42,6 +42,7 @@ var (
 // When adding a new field, update all four flag registration sites.
 type S3Options struct {
 	filer                     *string
+	ip                        *string
 	bindIp                    *string
 	port                      *int
 	portHttps                 *int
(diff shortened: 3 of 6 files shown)


chrislusf:
Both bots flagged the same thing: `advertisedHost()` guarded the `0.0.0.0` literal but not the IPv6 wildcard. Valid — and `util.JoinHostPortStr` already anticipates bracketed hosts, so `[::]` reaching here isn't hypothetical.

Fixed in f2d02f17f, but with `net.ParseIP` + `IsUnspecified()` rather than extending the string list. That covers `::`, `[::]` and the expanded `0:0:0:0:0:0:0:0` form in one check, and it matches the existing pattern in `util/outbound_dial.go` instead of adding a second, hand-maintained notion of "wildcard" that would drift.

Host names parse as nil and are deliberately not treated as wildcards — `-ip=localhost` has to keep working, since that is the case this PR exists to fix.

Also applied the same guard to the advertised ip itself: `-ip=0.0.0.0` was previously returned verbatim, which is the same unreachable-registration bug by a different route.

Test now covers both wildcard forms, the bracketed and expanded spellings, a real IPv6 bind address, and wildcard-advertised-ip falling through to the bind address.

PR #10492, opened 2026-07-30
webdav: answer PROPFIND child stats from the listing

chrislusf:
A WebDAV directory listing goes through `golang.org/x/net/webdav`'s `walkFS`, which throws away the `FileInfo` that `Readdir` handed it and stats every child again - five times per entry between `walkFS`, `allprop` and `props`. Every one of those was a filer `LookupDirectoryEntry`, issued serially, and `WebDavFile.Stat` dropped the request context on top of it.

So a PROPFIND on a directory with 3000 subdirectories made 15006 filer lookups. Windows Explorer mounts a share, opens a large folder, and gives up.

This keeps the entries a listing already fetched in the request context and answers those stats from them. Positive entries only, populated from `Readdir` alone, discarded with the request, capped at 100k so one PROPFIND cannot balloon.

Measured against a leveldb2 filer on loopback, `weed webdav` in front of `weed server`:

| directory | lookups before | after | PROPFIND before | after |
|---|---|---|---|---|
| 3000 subdirs | 15006 | 6 | 1.41s | 0.035s |
| 6000 subdirs | | | 3.43s | 0.048s |
| 24000 subdirs | | | 13.4s | 0.248s |
| 60000 subdirs | | | 23.7s | 0.538s |
| 200 files | | | 0.102s | 0.005s |

Same bytes on the wire: response sizes are byte-identical and the `href`, `displayname`, `getcontentlength`, `getetag`, `getlastmodified` sets match the old server exactly, for directories and for files. MKCOL, PUT, GET, MOVE, COPY -Depth infinity, DELETE and PROPFIND at depth 0/1/infinity all behave as before.

## Summary by CodeRabbit

* **Performance Improvements**
  * Improved WebDAV directory browsing by reusing directory listing information within each request.
  * Reduced redundant WebDAV metadata fetches by caching per-request directory entries.

* **Bug Fixes**
  * Improved lookup consistency for WebDAV paths with or without trailing slashes.
  * Ensured request-scoped caching is isolated between separate WebDAV requests.
  * Prevented unbounded growth of temporary directory listing data by enforcing a cache size limit.

### DIFF
--- a/weed/server/webdav_listed_entries.go
+++ b/weed/server/webdav_listed_entries.go
@@ -1,0 +1,72 @@
+package weed_server
+
+import (
+	"context"
+	"net/http"
+	"strings"
+	"sync"
+)
+
+// A PROPFIND on a directory lists it once, then golang.org/x/net/webdav's walkFS
+// throws the listing's FileInfo away and stats every child again - one filer
+// lookup per entry, serially. listedEntries hands those stats back the entries the
+// listing already fetched, for the lifetime of a single request.
+type listedEntries struct {
+	mu     sync.Mutex
+	byPath map[string]*FileInfo
+}
+
+// maxListedEntries bounds the per-request memory. Past it, children fall back to
+// individual lookups.
+const maxListedEntries = 100000
+
+type listedEntriesKey struct{}
+
+func withListedEntries(ctx context.Context) context.Context {
+	return context.WithValue(ctx, listedEntriesKey{}, &listedEntries{byPath: make(map[string]*FileInfo)})
+}
+
+func listedEntriesFrom(ctx context.Context) *listedEntries {
+	if ctx == nil {
+		return nil
+	}
+	listed, _ := ctx.Value(listedEntriesKey{}).(*listedEntries)
+	return listed
+}
+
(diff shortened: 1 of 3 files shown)


chrislusf:
Both bots landed on the same thing from three angles, and all three are right: `stat`, `Readdir` and `Seek` still reached the filer on `context.Background()`. Fixed in f1831b3b1 - they now share a `requestContext()` helper, so a client that walks away stops the listing stream and the lookups behind it, and `Seek`'s `io.SeekEnd` stat reaches the held entries too.

`Write`, `Read`, `Close` and `removeAll` keep their own context on purpose: a cancelled request must not abandon a flush half done.

Re-verified after the change - the `href`, `displayname`, `getlastmodified`, `getcontentlength`, `getetag` and `getcontenttype` sets still match the pre-change server exactly for all five fixtures, and MKCOL/PUT/GET/Range/COPY/MOVE/DELETE plus PROPFIND at depth 0/1/infinity are unchanged. A 60000-entry PROPFIND is now 0.34s.

Unrelated to this PR, but noticed while testing: DELETE on a non-empty collection returns 405 on both the old and new server. Separate fix.

PR #10499, opened 2026-07-30
test: pin that a .vif replication outranks the superblock

chrislusf:
A volume's replication lives in two places that can disagree, and nothing tested which one wins.

`Store.ConfigureVolume` rewrites the **`.vif`** and never the replica-placement byte in the `.dat` superblock, so that byte keeps whatever the volume was created with for good. `readSuperBlock` reads it and then overrides it from the `.vif`:

```go
v.SuperBlock, err = super_block.ReadSuperBlock(v.DataBackend)

if v.volumeInfo != nil && v.volumeInfo.Replication != "" {
	...
	v.SuperBlock.ReplicaPlacement = replication
}
```

That override is the only reason a replication change takes effect and survives a remount. Invert it and **every replication change silently reverts on the next mount**, while the `.vif` on disk still records what the operator asked for — a durability setting going back to its old value with nothing to indicate it happened.

## Why a test rather than reading the code

The field immediately beside it resolves the **other way**: `version` takes the superblock over the `.vif`. Two fields, one function, opposite precedence, each one line to invert wrongly — and one of the two directions loses data durability silently.

Confirmed the test catches it. Disabling the override gives:

```
replication after reload = "000", want 010 from the .vif;
a reconfigured volume must not revert to the byte in its .dat
```

The second test covers the empty case: a `.vif` declaring no replication has to leave the superblock standing, or a volume whose replication was never configured would be forced to whatever the zero value parses as. It passes either way, which is the point — it constrains the fix rather than the bug.

`go test ./weed/storage/` green.

## Context

Found while tracing what a tiering replication change actually writes to disk. The Rust volume server in the enterprise tree implements the same override with a `// Match Go:` comment, against nothing that would notice if the two stopped matching; that side is being pinned separately.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for replication settings when volumes are reloaded.
  * Verified that explicitly configured replication values take precedence over existing volume metadata.
  * Confirmed that empty replication settings preserve the previously stored value, helping ensure consistent volume behavior across reloads.
  * Improved validation of volume information loading before replication settings are updated.

### DIFF
--- a/weed/storage/volume_super_block_replication_test.go
+++ b/weed/storage/volume_super_block_replication_test.go
@@ -47,9 +47,6 @@ func TestVifReplicationOutranksSuperBlock(t *testing.T) {
 	if err != nil {
 		t.Fatalf("load vif: %v", err)
 	}
-	if info == nil {
-		info = &volume_server_pb.VolumeInfo{}
-	}
 	info.Replication = "010"
 	if err := volume_info.SaveVolumeInfo(vifFile, info); err != nil {
 		t.Fatalf("save vif: %v", err)
--- a/weed/storage/volume_super_block_replication_test.go
+++ b/weed/storage/volume_super_block_replication_test.go
@@ -1,0 +1,104 @@
+package storage
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/volume_server_pb"
+	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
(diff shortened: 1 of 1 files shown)


chrislusf:
Good catch — verified and fixed in `31f4832`→latest.

`MaybeLoadVolumeInfo` assigns `volumeInfo = &volume_server_pb.VolumeInfo{}` before the existence check and every return is naked, with no path assigning nil, so the guard was unreachable and implied a contract the callee doesn't have. Dropped it.

`go vet` and `go test ./weed/storage/` still green.

PR #10503, opened 2026-07-31
filer: let a nested path rule turn worm off

### LINKED ISSUE #10500: merging logic of boolean in mergePathConf
Merging logic of boolean fields seems has different rule than string fields. A true value from parent/ancestor directory can not be overrided. It does make sense for "ReadOnly" since this is how bucket quota enforced. But for settings like "WORM", there is scenario that some child directory don't want WORM enabled while parent does.

Looks like a tri-state *bool type would be better for WORM? Or considering interface compatibility, an extra field ```OverrideWorm bool```?

chrislusf:
Fixes #10500.

`mergePathConf` merges string fields with `Nvl`, so the most specific rule wins, but merges every boolean with `||`, so an ancestor's `true` is unbeatable:

```
/buckets/b/      worm=true  ttl=7d
/buckets/b/tmp/  worm=false
→ MatchStorageRule("/buckets/b/tmp/x") = worm:true
```

`worm` becomes tri-state (`optional bool`, same field number and JSON name): unset inherits from the enclosing rule, set overrides it. `readOnly`, `fsync` and `disableChunkDeletion` keep the OR on purpose — a nested rule must not be able to lift the lock quota enforcement puts on a bucket.

### The migration

`filer.conf` is marshalled with `EmitUnpopulated`, so every rule ever written carries a literal `"worm": false`. Under field presence that would parse as an explicit opt-out and silently strip worm from nested paths on upgrade, which is the wrong direction for a compliance control.

So `FilerConf.version` — written as `0` and read by nobody until now — is stamped at `1` on every write, and a configuration older than that has an explicit `false` dropped back to unset at load time. A rolling upgrade is safe both ways: an old binary sees the flag omitted where it is unset and falls back to the OR, which is fail-closed.

### Notes

- `fs.configure` only writes `worm` when the flag is passed, so a rule created for some other setting no longer stamps a value; the `-worm` case in the post-merge fixup is gone since presence handles it now.
- `wormGracePeriodSeconds` / `wormRetentionTimeSeconds` still merge with `>0 wins`, so a nested rule can't reset them to 0. Left alone: 0 means "never expires" there, so only the strongest value is unreachable.

### Tests

`TestWormInheritance` covers inherit / override / re-enable deeper, `TestWormLegacyFalseIsNotAnOverride` pins the version-0 read, `TestWormSurvivesRoundTrip` pins the write side.

## Summary by CodeRabbit

* **New Features**
  * Path-level worm protection settings can now inherit from parent paths.
  * Explicitly enabling or disabling worm protection on nested paths is supported.
  * Configuration commands now document inheritance and include examples for overrides.

* **Bug Fixes**
  * Preserved worm protection settings when saving and reloading configurations.
  * Corrected handling of legacy configurations so unset values remain inheritable.

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -761,7 +761,8 @@ message FilerConf {
         string data_node = 11;
         uint32 max_file_name_length = 12;
         bool disable_chunk_deletion = 13;
-        bool worm = 14;
+        // unset inherits from the enclosing path rule, set overrides it
+        optional bool worm = 14;
         uint64 worm_grace_period_seconds = 15;
         uint64 worm_retention_time_seconds = 16;
     }
--- a/weed/filer/filer_conf.go
+++ b/weed/filer/filer_conf.go
@@ -15,6 +15,7 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/util"
 	"github.com/viant/ptrie"
 	jsonpb "google.golang.org/protobuf/encoding/protojson"
+	"google.golang.org/protobuf/proto"
 )
 
 const (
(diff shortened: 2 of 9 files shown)


chrislusf:
Both addressed in 1e34cc3be: `mergePathConf` now copies the value with `proto.Bool(*b.Worm)` instead of aliasing the trie entry's pointer, and the post-merge fixup in `fs.configure` says why worm is not in it.

PR #10521, opened 2026-08-01
s3: cover three untested STS paths

chrislusf:
Auditing the STS surface turned up three entry points with no meaningful test. All three commits are test-only.

**GetCallerIdentity had no handler test at all** — only XML marshalling. The case worth having is that a caller presenting session credentials is reported as `arn:aws:sts::…:assumed-role/<role>/<session>` and not as the user who minted the session; the test mints a session through `handleAssumeRole` and re-signs a `GetCallerIdentity` with the returned `ASIA` credentials. Static-key, bad-signature and unsigned cases round it out.

**AssumeRoleWithWebIdentity was never driven over HTTP with a real token.** Coverage reached the OIDC path either at the `IAMManager` service layer or through the `Authorization: Bearer` shortcut, so the public entry point an SDK actually calls was untested. This signs an RS256 token against an httptest JWKS server and drives `HandleSTSRequest` with the parameters in the query string and in the form body, plus an inline session policy reaching `PackedPolicySize`, a role whose trust policy names a different provider, and a token signed by an unknown key.

**Nothing checked that STS requests are audited.** STS responses go out through `WriteXMLResponse`, which never calls `PostLog`, so `track()` is the only thing that logs them — a route registered outside it would mint credentials with no audit trail. STS has three routing layers (explicit query-param routes, the authenticated POST dispatcher, the anonymous fallback), so a new action is easy to attach to the wrong one. The test registers the real router and covers all five actions plus the POST-body dispatcher. I confirmed the assertion is not vacuous: invoking the handler directly, bypassing `track()`, leaves the flag false.

One note that shaped an assertion: `AssumeRoleWithWebIdentity` does not embed the role's attached policies in the session token, while `AssumeRole` does. Authorization resolves them live by looking the role up at request time, so the session works but is not self-sufficient the way an `AssumeRole` session is. The test asserts effective permission rather than the claim contents so it stays true either way, but the asymmetry may be worth a separate look.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for STS request routing across query parameters, authenticated POST bodies, and anonymous requests.
  * Expanded validation of caller identity responses, credential types, signatures, and authorization outcomes.
  * Added end-to-end coverage for web identity role assumption, including trusted and untrusted roles, temporary credentials, policies, and invalid signing keys.

### DIFF
--- a/weed/s3api/s3api_sts_audit_routing_test.go
+++ b/weed/s3api/s3api_sts_audit_routing_test.go
@@ -1,6 +1,7 @@
 package s3api
 
 import (
+	"context"
 	"net/http"
 	"net/http/httptest"
 	"net/url"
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed in 7155aaa8b. Thanks — the routing comment was a real hole, and chasing it turned up a second one.

**Routing assertions were too weak (CodeRabbit).** Correct, and worse than it looked. `setupRoutingTestServer` builds `&STSHandlers{}` with a nil STS service, so every case was answering 503 without executing anything; a non-404 status was the only evidence the request had reached STS, and the POST-body case could have been served by `UnifiedPostHandler`'s IAM branch and still passed. The test now runs against a real STS service and asserts the STS XML namespace, which IAM and S3 responses do not carry. Three of the six cases now assert a 200 as well.

Rebuilding it surfaced a second problem in my own setup: `GetFederationToken` was returning 500 "failed to resolve caller policies", because I never wired a user store. Production always does (`s3api_server.go:374` passes the credential manager), so this was my harness diverging from reality rather than a product bug — but with the old weak assertions it would have gone unnoticed either way. Now mirrored.

**Session policy assertion (CodeRabbit).** Agreed — checking `SessionPolicy` was non-empty proved the policy travelled, not that it restricted anything. Now asserts `s3:GetObject` is allowed on the bucket the session policy names and refused on another bucket, which the role alone would have permitted.

**Missing `iat` (Greptile).** Fixed; the forged token now carries the same claim set a valid one does, so it cannot fail for want of a claim.

**Unknown key id (CodeRabbit).** Split rather than swapped. A wrong signing key under a published `kid` and a `kid` absent from the JWKS exercise different code — one fails signature verification, the other never resolves a key — and the first is the property I actually wanted. Both are covered now, and the subtest names say which is which.

**JWKS `Content-Type` (Greptile).** Set, and the handler now 404s on other paths instead of returning an empty 200.

17 subtests, all passing; `./s3api/...` and `./iam/...` green.

Issue #10468, opened 2026-07-28
After expanding my SeaweedFS cluster, I have issues with user permissions.


Thegsgs:
Hi,

I recently expanded my SeaweedFS cluster from 3 servers to 7 servers. Since the expansion, I have encountered two issues related to S3 authentication and authorization.

Issue 1: Newly created users cannot authenticate
When I create a new user and use the access key and secret generated by the admin panel, uploads fail with:
`upload failed: ./testfile.txt to s3://test-bucket/test/testfile.txt An error occurred (InvalidAccessKeyId) when calling the PutObject operation: The access key ID you provided does not exist in our records.`
The credentials are copied directly from the admin panel's export command.

Issue 2: Newly created buckets reject uploads
When I create a new bucket and assign an existing user as the bucket owner, uploads fail with:
`upload failed: ./testfile.txt to s3://test-bucket/testfile.txt An error occurred (AccessDenied) when calling the PutObject operation: Access Denied.`
The user is configured as the bucket owner and is also listed as an admin with actions:["Admin"].

While monitoring the server logs with `journalctl -u seaweedfs-server.service -f`, I can see:
`Jul 28 04:17:17 storage-leader-prod-01 weed[22081]: W0728 04:17:17.933851 auth_credentials.go:886 identity ci_cd_admin is associated with a non exist account ID, the association is invalid`

The ci_cd_admin user can still upload successfully to buckets that were created before the cluster expansion. The issue only affects newly created users and/or newly created buckets.
Running `weed shell` and s3.user.show shows the user ci_cd_admin with the expected credentials and actions:["Admin"].
Has anyone seen this before, or is there any additional information I can provide to help diagnose the issue?

SeaweedFS cluster version 4.40. The cluster was expanded by added 4 additional `weed server` instances.

tb3088:
you added 'weed volume' servers (ie more disk) or you added object servers (weed server) to the pool?

Thegsgs:
@tb3088 I have 7 `weed server`s linked in a cluster, Iv'e attached below my systemd configuration.

<img width="1070" height="596" alt="Image" src="https://github.com/user-attachments/assets/6b826c4a-9cb2-4eed-a50f-506c57d86d6d" />

tb3088:
you didn't actually answer my question. But I guess from the service script all of your nodes are the same - they each run a master, volume, filer and S3 protocol adapter. 

[...]

PR #10548, opened 2026-08-03
s3: register an identity's inline account instead of collapsing it into admin

chrislusf:
Every user created through the IAM API or the admin UI with an email address gets an account block attached inline on the identity (`weed/admin/dash/user_management.go`). No credential store ever populates the top-level `S3ApiConfiguration.Accounts` list, so that id was never in the S3 server's account map, and both load paths took the fallback:

```
identity ci_cd_admin is associated with a non exist account ID, the association is invalid
```

The fallback set the identity's account to `AccountAdmin`. Consequences:

- Distinct users all present `AccountAdminId` as `x-amz-account-id`, which is what `checkAccessByOwnership` compares against the bucket owner. Each one therefore passes ownership for the others' buckets.
- Objects and buckets they create are recorded with the admin id as owner, so ownership does not survive a later fix on its own.
- `GetAccountNameById` / `GetAccountIdByEmail` never resolve the real id, so ACL grantee validation and owner display report it as nonexistent.

#9962 fixed the same collapse for identities with no account block at all; the case where the identity carries an account the config never declared was left on the fallback.

An id missing from the account map means undeclared, not invalid, so register it. An email that another account already claimed is left alone, so an explicitly configured account still wins email lookups. Both `ReplaceS3ApiConfiguration` and `MergeS3ApiConfiguration` had the same three-way switch; they now share one helper, which also covers the push path — a propagated `PutIdentity` carries a single identity and no accounts list.

Surfaced while looking into #10468, whose reporter has this warning in their logs. It is not the cause of the stale-credential half of that report.

Tests cover the full load and the upsert/merge path: the inline account id survives, two users do not share an owner id, and the account resolves by id and by email.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed inline identity accounts being incorrectly associated with the administrator account.
  * Preserved distinct account ownership for separate identities.
  * Improved account lookup by ID, display name, and email.
  * Ensured identity updates correctly register and resolve referenced accounts.
  * Updated account metadata and email lookups when identity details change.
  * Preserved explicitly configured account settings over inline identity data.
  * Prevented conflicting email associations when accounts are replaced or updated.

### DIFF
--- a/weed/s3api/auth_account_collapse_test.go
+++ b/weed/s3api/auth_account_collapse_test.go
@@ -210,6 +210,60 @@ func TestInlineIdentityAccountIsRegisteredOnUpsert(t *testing.T) {
 	assert.Equal(t, "Alice Smith", iam.GetAccountNameById("100000000001"), "the display name must follow the update")
 }
 
+// Two inline accounts can carry the same email. The first to register keeps the
+// lookup, and the loser can claim it once the holder moves away — otherwise an
+// email freed by an update would resolve to nobody.
+func TestInlineAccountEmailClaimAndHandoff(t *testing.T) {
+	resetMemoryStore()
+
+	config := `{
+  "identities": [
+    {"name": "admin", "credentials": [{"accessKey": "admin_ak", "secretKey": "admin_sk"}], "actions": ["Admin"]}
+  ]
+}`
+	tmp, err := os.CreateTemp("", "s3-config-*.json")
+	require.NoError(t, err)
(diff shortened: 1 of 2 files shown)


tb3088:
what is the ramification of the previous behavior regarding multi-tenant and bucket access enforcement?
email address seems like a sketchy item on which to place any importance. Are there safeguards to make sure it's localized to a single (organization) account?

PR #10549, opened 2026-08-03
iceberg: let clients select their table bucket as the catalog warehouse

### LINKED ISSUE #10547: PyIceberg Sample code does not show namespaces
When using the sample python code after creating a namespace it gets rendered as:

```
from pyiceberg.catalog import load_catalog

catalog = load_catalog(
    name="seaweedfs",
    **{
        "type": "rest",
        "uri": "http://[server]:8181",
       "credential": f"creds",
        "scope": "catalog",
    }
)

namespaces = catalog.list_namespaces()
```

This then fails with:
```
pyiceberg.exceptions.ServerError: InternalServerError: all filers failed, last error: table bucket warehouse not found
```

I tried setting a warehouse to the name of the table bucket, in this case seaweed-iceberg as well as the arn of the table bucket but all still fail with the above message. 
I also tried setting the url to /v1/seaweed-iceberg and get an error that the auth tokens cannot be found

Is there a setting I am missing or how can I specify my table-bucket instead of the default one.

chrislusf:
Fixes #10547.

Every table bucket is a separate Iceberg catalog, but nothing in the sample code said so, and the catalog only understood one of the three ways a client can name one.

### The sample code connected without a warehouse

The PyIceberg snippet in the admin console passed only `type` and `uri`. With no warehouse the catalog falls back to a bucket literally named `warehouse`, which almost never exists, so the first call failed:

```
pyiceberg.exceptions.ServerError: InternalServerError: all filers failed, last error: table bucket warehouse not found
```

The snippet now sets `"warehouse": "s3://my-table-bucket/"`, alongside the `SCOPE 's3://my-table-bucket/'` the DuckDB example directly above it already showed.

### Only `s3://<bucket>/` was accepted as a warehouse

Setting the warehouse by hand did not help either. `s3://bucket/` was the sole recognized spelling; a bare table bucket name and the s3tables bucket ARN — the latter being what AWS S3 Tables itself takes as the warehouse — were both silently dropped and fell through to the same default bucket, producing the identical error. All three now resolve, in `/v1/config` and in the per-request `?warehouse=` fallback. A resolved name is validated as a table bucket name, so a warehouse that names no usable bucket falls back rather than reaching path construction.

### A missing table bucket was a 500

Naming a catalog that does not exist is a client-side mistake, but it came back as `InternalServerError` citing a bucket the client never mentioned. It is now 404 with the message spelling out how to select one:

```
table bucket warehouse not found: each table bucket is a separate catalog,
select one with warehouse=s3://<table-bucket>/ or /v1/<table-bucket>/
```

### Verification

Against `weed mini` with a table bucket named `seaweed-iceberg`, PyIceberg 0.10.0:

| warehouse | before | after |
| --- | --- | --- |
| unset (old sample code) | 500 `table bucket warehouse not found` | 404, message says how to select a bucket |
| `seaweed-iceberg` | 500 `table bucket warehouse not found` | `list_namespaces()` OK |
| `arn:aws:s3tables:us-east-1:admin:bucket/seaweed-iceberg` | 500 `table bucket warehouse not found` | `list_namespaces()` OK |
| `s3://seaweed-iceberg/` (new sample code) | OK | OK |

Full PyIceberg round-trip on the new snippet — create namespace, create table, append 3 rows, scan back 3 rows — passes. Unit tests cover each warehouse spelling through `/v1/config` and the error mapping.

## Summary by CodeRabbit

* **New Features**
  * PyIceberg examples now include a default warehouse location.
  * Warehouse settings accept S3 URLs, bucket names, and S3 Tables ARNs.
  * Warehouse configuration can determine the target bucket when no URL prefix is provided.

* **Bug Fixes**
  * Missing S3 Table buckets now return a clear 404 namespace error.
  * Unsupported or invalid warehouse values safely fall back to the default bucket.

### DIFF
--- a/test/s3tables/catalog/duckdb_oauth_test.go
+++ b/test/s3tables/catalog/duckdb_oauth_test.go
@@ -216,8 +216,12 @@ func TestOAuthTokenEndpoint(t *testing.T) {
 	t.Run("bearer token auth on catalog endpoint", func(t *testing.T) {
 		token := requestOAuthToken(t, env, env.accessKey, env.secretKey)
 
-		// Use the token to call the catalog
-		req, err := http.NewRequest(http.MethodGet, env.icebergURL()+"/v1/namespaces", nil)
+		// Point the call at a table bucket that exists, so anything but 200 is an
+		// auth or routing fault rather than the catalog reporting a missing bucket.
+		bucketName := "oauth-bearer-" + randomSuffix()
+		createTableBucketViaShell(t, env, bucketName)
+
+		req, err := http.NewRequest(http.MethodGet, env.icebergURL()+"/v1/"+bucketName+"/namespaces", nil)
 		if err != nil {
 			t.Fatalf("create request: %v", err)
 		}
@@ -225,17 +229,12 @@ func TestOAuthTokenEndpoint(t *testing.T) {
 
 		resp, err := http.DefaultClient.Do(req)
 		if err != nil {
-			t.Fatalf("GET /v1/namespaces with Bearer: %v", err)
+			t.Fatalf("GET /v1/%s/namespaces with Bearer: %v", bucketName, err)
 		}
 		defer resp.Body.Close()
 
-		// Auth should pass. We accept 200 (success) or 500 (missing warehouse bucket
-		// is an internal error, not an auth error). Reject 401/403/404/405.
 		body, _ := io.ReadAll(resp.Body)
-		switch resp.StatusCode {
-		case http.StatusOK, http.StatusInternalServerError:
-			t.Logf("Bearer token auth succeeded, status=%d", resp.StatusCode)
-		default:
+		if resp.StatusCode != http.StatusOK {
 			t.Fatalf("Bearer auth failed unexpectedly: status=%d body=%s", resp.StatusCode, body)
 		}
 	})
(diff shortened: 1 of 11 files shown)


chrislusf:
Addressed in 7547d8d6c, cbde470ab, 2de338543.

**CI — `TestOAuthTokenEndpoint/bearer_token_auth_on_catalog_endpoint`.** A real failure caused by this PR, now fixed. The subtest called `/v1/namespaces` with no warehouse and accepted `200 || 500` as proof that the Bearer token authenticated, since the default bucket never exists. Promoting the missing bucket to 404 landed it in the set the test read as an auth failure. Rather than widen the accepted statuses, the subtest now creates a table bucket and calls `/v1/<bucket>/namespaces`, so only 200 passes and the assertion actually distinguishes an auth fault from a routing one.

**CodeRabbit — assert the guidance message.** Fair, the message is the whole point of the mapping. `TestWriteManagerError` now checks the response body carries the original `table bucket ... not found` text plus both `warehouse=s3://<table-bucket>/` and `/v1/<table-bucket>/`, alongside the existing status and error-type checks.

**Greptile #1 — unencoded ARN in the query string.** Correct, the raw colons and slashes only worked because Go's parser tolerates them. The test now builds the query with `url.Values{}.Encode()`, which also covers the other cases.

**Greptile #2 — `defaults.warehouse` normalization.** Intentional, keeping as is. It has to be normalized: with a bare name or an ARN, echoing the client's value verbatim hands back something that is not a location at all, and clients derive default table locations from it. `s3://<bucket>` without the trailing slash matches how the catalog builds locations everywhere else (`fmt.Sprintf("s3://%s/%s", bucket, ...)` in `defaultNamespaceLocation`), so a client concatenating onto it lands on the same path the server would compute.

PR #10557, opened 2026-08-04
fix(volume): persist the leveldb needle map watermark at batch boundaries

SuperMohit:

The LevelDB needle map checkpoints a watermark every `watermarkBatchSize`
(10000) index entries so that a rebuild can skip the prefix of `.idx` already
applied to the `.ldb`.

`levelDbWrite` persists that watermark when its `updateWatermark` argument is
true. `Put` and `Delete` pass `watermark == 0` — but the local `watermark`
variable is set to `0` precisely on the writes that carry no checkpoint, and to
a real value on the batch boundary that does. The two cases are inverted:

```
recordCount % watermarkBatchSize != 0  ->  watermark 0, flag true
    -> re-persists a zero on 9999 of every 10000 writes
recordCount % watermarkBatchSize == 0  ->  watermark N, flag false
    -> drops the only value worth saving
```

So the stored watermark never leaves 0.

This is **not** a correctness bug: replaying `.idx` from offset 0 is a superset
of replaying from N, and replay is idempotent, so nothing fails loudly. That is
why it has gone unnoticed. The costs are:

1. `generateLevelDbFile` walks the entire index on every rebuild, so the
   optimization has never actually taken effect.
2. Every needle write pays a second leveldb `Put` to rewrite the same zero.

Pass `watermark != 0`, so the boundary write checkpoints and the ordinary
writes in between leave the key alone. Two characters, at both call sites.

# How is the PR tested?

New test `TestLevelDbNeedleMapWatermarkAdvances` drives a full batch of `Put`s
and a full batch of `Delete`s (tombstones are `.idx` entries too, so they
advance the watermark the same way) to cover both changed call sites.

It fails on master:

```
after 10000 puts: watermark = 0, want 10000
after 10000 deletes: watermark = 0, want 20000
```

and passes with the fix.

Measured on a 25000-needle volume: the stored watermark now reads 20000 instead
of 0, and a rebuild replays 5000 entries instead of 25000.

Existing tests pass, including `TestGenerateLevelDbFileStaleWatermarkRebuilds`
— relevant here because a working (non-zero) watermark makes that staleness
guard load-bearing for the first time.

Locally verified: `go vet`, `GOOS=linux GOARCH=386 go vet`, the tagged build
(`elastic gocdk sqlite ydb tarantool tikv rclone`), and `go test ./weed/storage/...`
including `-race`. The 32-bit test suite was compile-checked only, since 386
binaries do not execute natively on arm64.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved watermark updates so they occur at the correct batch boundaries.
  * Ensured watermark progression remains accurate when records are deleted.

* **Tests**
  * Added coverage verifying watermark persistence and advancement during updates and deletions.

### DIFF
--- a/weed/storage/needle_map_leveldb.go
+++ b/weed/storage/needle_map_leveldb.go
@@ -186,7 +186,7 @@ func (m *LevelDbNeedleMap) Put(key NeedleId, offset Offset, size Size) error {
 		watermark = (m.recordCount / watermarkBatchSize) * watermarkBatchSize
 		glog.V(1).Infof("put cnt:%d for %s,watermark: %d", m.recordCount, m.dbFileName, watermark)
 	}
-	return levelDbWrite(m.db, key, offset, size, watermark == 0, watermark)
+	return levelDbWrite(m.db, key, offset, size, watermark != 0, watermark)
 }
 
 func getWatermark(db *leveldb.DB) uint64 {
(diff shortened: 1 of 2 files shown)


chrislusf:
Thanks for this finding!!!

PR #10567, opened 2026-08-04
s3: honor a bucket owner recorded as an identity

### LINKED ISSUE #10562: [S3] Created objects use admin as owner instead of the bucket owner
**Environment**

* SeaweedFS version: 4.40
* Deployment: Kubernetes via SeaweedFS Operator
* S3 API enabled

**Description**

A bucket is created with a non-admin user as the owner, but objects uploaded to this bucket are assigned to the `admin` account instead of the configured bucket owner.

**Steps to reproduce**

1. Deploy SeaweedFS 4.40 using SeaweedFS Operator
2. Create a user through the SeaweedFS Admin UI:

```
User: test
Permissions:
- Read
- Write
- List
- Tagging
Bucket Scope: All Buckets
```

3. Create a bucket through the Admin UI:

```
Bucket: test
Owner: test
```

4. Upload an object to the bucket using mc:

```
mc cp web.yaml s3/test/
```
**Expected behavior**

The uploaded object should be owned by the bucket owner.

**Actual behavior**

The uploaded object is owned by admin.

The S3 container logs show:

```
I0804 16:22:49.693868 s3api_object_handlers_put.go:1206 setObjectOwnerFromRequest: using bucket owner admin (BucketOwnerEnforced)
I0804 16:22:49.694750 s3api_object_handlers_put.go:904 putToFiler: Metadata saved SUCCESS - path=/buckets/test/web.yaml, etag(hex)=7dd48201220342f60769acaa77e49efe, size=1105, partNumber=1
```

The object’s extended attributes contain:

```
Seaweed-X-Amz-Owner: admin
```

chrislusf:
Fixes #10562.

A bucket whose owner was set outside the S3 API — the admin UI, or `weed shell s3.bucket.create -owner` / `s3.bucket.owner` — records the owner as an identity name in `s3-identity-id` and never writes the account id (`Seaweed-X-Amz-Owner`) that the S3 API writes next to it at bucket creation. The S3 API only ever read the account id, so those buckets looked unowned and fell back to the default admin account. With the default `BucketOwnerEnforced` ownership, that account is handed every object written to the bucket:

```
setObjectOwnerFromRequest: using bucket owner admin (BucketOwnerEnforced)
Seaweed-X-Amz-Owner: admin
```

Reproduced on master with a bucket owned by a non-admin identity `test`:

```
$ echo "s3.bucket.create -name testb -owner test" | weed shell
bucket owner: test
$ AWS_ACCESS_KEY_ID=testkey aws --endpoint-url http://127.0.0.1:8333 s3 cp web.yaml s3://testb/
$ echo "fs.meta.cat /buckets/testb/web.yaml" | weed shell | grep Owner
    "Seaweed-X-Amz-Owner": "YWRtaW4=",     # admin
```

Two commits:

- Resolve the identity to its account when no account id is recorded, in the one place the bucket metadata and the bucket config both derive the owner from. Buckets already created this way are fixed with no migration, and an identity scoped to an explicit account resolves to that account rather than to its name.
- Reassigning a bucket's owner now drops the account id recorded at creation. It outranks the identity when the owner is resolved, so changing the owner of a bucket created through the S3 API had no effect on object ownership or the bucket ACL.

After the fix, the same run stamps `test`, `get-bucket-acl` reports `test` as the bucket owner, and reassigning the owner of an S3-API-created bucket takes effect on the next upload.

### DIFF
--- a/weed/admin/dash/bucket_management.go
+++ b/weed/admin/dash/bucket_management.go
@@ -310,7 +310,10 @@ func (s *AdminServer) SetBucketOwner(bucketName string, owner string) error {
 			bucketEntry.Extended = make(map[string][]byte)
 		}
 
-		// Set or remove the owner
+		// Set or remove the owner. The account id recorded by the S3 API goes with
+		// it: the S3 API derives it from the owning identity when it is absent, so
+		// leaving the old one behind would keep the previous owner of the objects.
+		delete(bucketEntry.Extended, s3_constants.ExtAmzOwnerKey)
 		if owner == "" {
 			delete(bucketEntry.Extended, s3_constants.AmzIdentityId)
 		} else {
--- a/weed/s3api/auth_credentials.go
+++ b/weed/s3api/auth_credentials.go
@@ -1443,6 +1443,16 @@ func (iam *IdentityAccessManagement) GetAccountNameById(canonicalId string) stri
 	return ""
 }
 
+// GetAccountIdByIdentityName resolves an identity name to the account its
+// resources are owned under. Bucket owners recorded outside the S3 API (the
+// admin UI, weed shell) name an identity, not an account.
+func (iam *IdentityAccessManagement) GetAccountIdByIdentityName(name string) string {
+	if identity := iam.lookupByIdentityName(name); identity != nil && identity.Account != nil {
+		return identity.Account.Id
+	}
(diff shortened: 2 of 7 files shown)


chrislusf:
@coderabbitai review

PR #10577, opened 2026-08-05
s3_lifecycle: bound shared metadata subscriptions

### LINKED ISSUE #10574: s3_lifecycle: no deadline/keepalive on dailyrun's shared-subscription gRPC stream — a half-dead stream wedges the job at stage "starting" forever
## Summary

An `s3_lifecycle` job can enter `stage: "starting"`, `progress: 0` and stay
there **indefinitely** — we captured one at **20.8 hours** (goroutine ages:
1250 minutes) against a known-good pass duration of ~2m8s, and have observed
6+ identical wedges over one week. While wedged, the job is externally
indistinguishable from a healthy long-running job: the worker heartbeats
normally, no log line is emitted by admin or worker, and the executor slot
stays occupied so no subsequent lifecycle pass can run. Net effect: S3
lifecycle expiry silently stops cluster-wide.

A SIGQUIT goroutine dump of the wedged worker pinpoints the cause.

## Goroutine-dump evidence (SeaweedFS `4.40`)

The job's `Execute` goroutine, blocked for exactly the age of the wedge:

```
goroutine 4108 [sync.WaitGroup.Wait, 1250 minutes]:
sync.(*WaitGroup).Wait                                          sync/waitgroup.go:206
weed/s3api/s3lifecycle/dailyrun.Run                              dailyrun/run.go:164
weed/worker/tasks/s3_lifecycle.(*Handler).executeDailyReplay     s3_lifecycle/handler.go:330
weed/worker/tasks/s3_lifecycle.(*Handler).Execute                s3_lifecycle/handler.go:259
weed/plugin/worker.(*Worker).handleExecuteRequest.func1          plugin/worker/worker.go:590
```

It is waiting on two classes of children, all blocked the same 1250 minutes:

**2 shared-subscription stream readers**, parked inside the gRPC transport's
blocking receive on an **open** stream that never delivers another frame:

```
goroutine [select, 1250 minutes]:
google.golang.org/grpc/internal/transport.(*recvBufferReader).readMessageHeaderClient
                                                                 internal/transport/transport.go:199
(grpc clientStream.RecvMsg / withRetry / GenericClientStream.Recv chain)
weed/s3api/s3lifecycle/reader.(*Reader).Run                      reader/reader.go:127
weed/s3api/s3lifecycle/dailyrun.startSharedSubscription.func2    dailyrun/run.go:316
created by dailyrun.startSharedSubscription                      dailyrun/run.go:315
```

**16 shard runners**, each starved of events those readers never forward:

```
goroutine [select, 1250 minutes]:
weed/s3api/s3lifecycle/dailyrun.drainShardEvents                 dailyrun/run.go:637
weed/s3api/s3lifecycle/dailyrun.runShard                         dailyrun/run.go:556
weed/s3api/s3lifecycle/dailyrun.Run.func2                        dailyrun/run.go:159
created by dailyrun.Run                                          dailyrun/run.go:153
```

## Mechanism

1. `Execute` → `executeDailyReplay` → `dailyrun.Run` opens a shared
   subscription (2 reader goroutines) plus 16 shard runners, then blocks on
   `WaitGroup.Wait` (run.go:164).
2. The subscription stream goes half-dead: it stays open at the transport
   level, but no further frames arrive. The readers' `Recv()` has **no context
   deadline and no gRPC keepalive** visible in the stack, so they block
   forever.
3. No events → all 16 `drainShardEvents` selects starve → the WaitGroup never
   reaches zero → `Execute` never returns.
4. Nothing above can recover it: `s3_lifecycle` is the only job type setting
   `ExecutionTimeoutSeconds: math.MaxInt32` / `JobTypeMaxRuntimeSeconds:
   math.MaxInt32` (deliberate, per the in-code comment and PR #9494), and the
   admin's stale-job sweep applies a 24h no-progress threshold while the
   worker still heartbeats — which it does, since only this one goroutine tree
   is stuck.

We did not determine *why* the stream stopped delivering (conntrack drop,
server-side subscription stall, …) — and arguably that should not matter: a
subscription client on an unbounded-lifetime job needs to defend itself with
deadlines/keepalives regardless of the trigger.

## What we ruled out first

- **Config drift:** deployment is stock chart + image `chrislusf/seaweedfs:4.40`;
  every relevant knob verified as an upstream default.
- **Filer health:** while a job was wedged, `GetFilerConfiguration` and
  `ListEntries` from the worker pod's own netns answered in ~1s, repeatedly.
- **S3 internal service health:** `SeaweedS3LifecycleInternal/LifecycleDelete`
  on both registered S3 endpoints answered sub-second, repeatedly, during a
  wedge.

Fresh probes all succeed because the defect lives in one *already-established*
stream — which is also why the failure produces zero log output: every
goroutine is "normally" blocked, nothing errors.

## Suggested fixes

1. **Keepalive and/or per-Recv deadline on the shared-subscription stream** in
   `dailyrun`/`reader` — a half-dead stream should surface as an error and
   fail (or resubscribe) the job, not block forever.
2. **Stall detection instead of a wall-clock cap:** the `math.MaxInt32` opt-out
   is understandable (a large lifecycle walk has no sane fixed bound), but
   "taking long" and "zero progress for N minutes" are distinguishable; the
   latter should at least log at warn, ideally fail the job.
3. **A log line per phase of `Execute()`** — the total silence between the
   `starting` stage update and first progress made this take days to localize.

## Secondary observation (can split into its own issue)

After we SIGQUIT'd the wedged worker to capture the dump, the container
restarted but the admin kept the job assigned to the old worker id at
`stage: starting` — worker-container restarts are not detected, so the fresh
worker process holds a job it has no in-memory state for until a heartbeat/
progress threshold eventually expires it.

## Environment

- SeaweedFS `4.40` (`chrislusf/seaweedfs:4.40`)
- Kubernetes; master + volume + filer + 2× S3 + admin + 1 worker
  (`-jobType=s3_lifecycle`)
- Reproduced 6+ times over one week; one wedge held >20h before we cleared it
- Full (unsanitized) goroutine dump retained locally; happy to share further
  extracts on request

chrislusf:
Fixes #10574.

## What changed

- Bound each daily replay subscription with `UntilNs = runNow`, so a pass ends once the filer has replayed through its frozen boundary instead of waiting for a later metadata event.
- Opt into the filer's idle stream heartbeats and add a 20-minute application-level receive timeout. The timeout is intentionally longer than the filer's 15-minute metadata-gap recovery budget.
- Cancel the shared fan-out whenever the reader exits on EOF, transport error, or timeout, ensuring every shard channel closes.
- Propagate unexpected shared-subscription errors from `dailyrun.Run` so the job fails visibly and can be retried.

## Root cause

`Reader.Run` called blocking `Recv()` without an application-level deadline. More importantly, when the reader returned, `startSharedSubscription` only wrote to `readerDone`; it did not cancel the fan-out. The fan-out therefore stayed blocked on its events channel, all shard drains stayed blocked on their channels, and the top-level wait group never completed.

## Reproduction

Against the unmodified base commit, a focused fake stream that returns EOF reproduces the downstream wedge deterministically: the reader exits, but the fan-out remains blocked and the test times out. A blocking fake stream covers the reported half-open `Recv()` case and verifies that the watchdog releases the fan-out and returns `reader.ErrReceiveTimeout`.

## Validation

- `go test -race ./weed/s3api/s3lifecycle/reader ./weed/s3api/s3lifecycle/dailyrun`
- `go test ./weed/s3api/s3lifecycle/... ./weed/worker/tasks/s3_lifecycle`
- `go vet ./weed/s3api/s3lifecycle/reader ./weed/s3api/s3lifecycle/dailyrun ./weed/worker/tasks/s3_lifecycle`

## Summary by CodeRabbit

* **New Features**
  * Added configurable receive timeouts for lifecycle metadata subscriptions, defaulting to 20 minutes.
  * Lifecycle processing now supports subscription end timestamps and idle heartbeats for more reliable stream handling.
* **Bug Fixes**
  * Prevented lifecycle runs from hanging when metadata streams stop responding.
  * Subscription channels now close cleanly after completion, transport errors, or timeout.
  * Subscription failures are surfaced as run errors.
  * Invalid negative timeout values are rejected.

### DIFF
--- a/weed/s3api/s3lifecycle/dailyrun/run.go
+++ b/weed/s3api/s3lifecycle/dailyrun/run.go
@@ -88,7 +88,15 @@ type Config struct {
 
 	// 0 -> unbounded.
 	EventBudget int
+
+	// SubscriptionReceiveTimeout bounds the gap between responses on the
+	// shared metadata stream. 0 uses the production default. The filer sends
+	// idle heartbeats once the reader catches up, while the generous default
+	// still permits the filer's bounded metadata-gap recovery to run.
+	SubscriptionReceiveTimeout time.Duration
 }
+
+const defaultSubscriptionReceiveTimeout = 20 * time.Minute
 
 // Run executes the daily replay for every shard in cfg.Shards
 // concurrently. Returns the first shard error; the rest log and run to
@@ -167,11 +175,12 @@ func Run(ctx context.Context, cfg Config) error {
 	// Tear down the shared subscription. cancelRead unblocks both the
 	// reader's gRPC stream and the fan-out's send loop; we wait on both
 	// so their goroutines don't outlive Run.
+	var subscriptionErr error
 	if cancelRead != nil {
 		cancelRead()
 		<-fanoutDone
 		if rerr := <-readerDone; rerr != nil && !errors.Is(rerr, context.Canceled) && !errors.Is(rerr, context.DeadlineExceeded) {
-			glog.V(2).Infof("daily_run: shared reader returned: %v", rerr)
+			subscriptionErr = fmt.Errorf("daily_run: shared subscription: %w", rerr)
 		}
 	}
 
@@ -183,6 +192,14 @@ func Run(ctx context.Context, cfg Config) error {
 			first = err
 		} else {
 			glog.V(1).Infof("daily_run: additional shard error: %v", err)
+		}
+	}
+	if subscriptionErr != nil {
+		errCount++
+		if first == nil {
+			first = subscriptionErr
+		} else {
+			glog.V(1).Infof("daily_run: additional subscription error: %v", subscriptionErr)
 		}
 	}
 	status := "ok"
@@ -277,9 +294,9 @@ func computeGlobalStartTsNs(ctx context.Context, cfg Config, runNow time.Time, m
 // Subscription floor is the caller-supplied globalStartTsNs (typically
 // min over per-shard cursors). Shards whose own startTsNs is fresher
 // filter out already-past events themselves inside drainShardEvents.
-// Events arriving with TsNs > runUpTo (the pass boundary) cause the
-// fan-out to cancel the reader and close all per-shard channels,
-// ending the pass.
+// The filer-side UntilNs bound normally ends the stream at runNow. Events
+// arriving with TsNs > runUpTo are retained as a defensive fallback: they
+// cause the fan-out to cancel the reader and close all per-shard channels.
 func startSharedSubscription(ctx context.Context, cfg Config, runNow time.Time, globalStartTsNs int64) (map[int]chan *reader.Event, chan error, chan struct{}, context.CancelFunc) {
 	shardSet := make(map[int]bool, len(cfg.Shards))
 	shardEvents := make(map[int]chan *reader.Event, len(cfg.Shards))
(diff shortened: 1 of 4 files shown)


chrislusf:
Closing in favour of #10578, which fixes the same root cause. Your receive-timeout watchdog is ported there (ebe8997b3) — including the 20-minute default and the reasoning for putting it above the filer's 15-minute `maxGapStall`, which was the right call and something I had missed. The wiring differs only because #10578 ends the pass by closing the reader's event channel rather than cancelling the fan-out.

Two things led me to build on the other branch rather than this one:

**Tail events get dropped at teardown.** `cancelReader()` leaves the fan-out's `select` with both `readerCtx.Done()` and buffered `events` ready, and Go picks uniformly at random — so events already received from the filer are discarded instead of routed. Measured on this branch with a 200-event stream that then EOFs:

```
PR #10577:  10 of 20 trials dropped part of the tail (196-199 of 200 routed)
PR #10578:   0 of 20
```

Not data loss — cursors don't advance past dropped events — but on an idle cluster the same tail is the tail again next pass, so expiry can be deferred repeatedly. Closing the channel drains it deterministically.

**A halted shard still wedges the pass.** A drain that stops mid-stream (`BLOCKED` / `RETRY_LATER` / an RPC error on dispatch) returns while the fan-out is still routing that shard's events; after 256 the buffer fills, the fan-out blocks on the send, and every other shard starves behind it. The watchdog doesn't reach this one either — the reader is blocked *pushing* to `Events`, not waiting in `Recv`. #10578 has a failing-then-passing regression for it.

One bug worth flagging since it was in both: `!errors.Is(rerr, context.Canceled)` doesn't match a canceled gRPC stream, which arrives as a status code. That condition fails in both directions — a `-runtime`-capped pass reports failure, a peer-cancelled stream reports success. #10578 stopped classifying by error entirely and reads intent instead (caller ctx ended, or the fan-out hit the boundary).

PR #10598, opened 2026-08-06
admin: dashboard counts chunks, not files

### LINKED ISSUE #10597: [UI] Total files count is inaccurate
Hi,

I got seaweedFS running with MongoDB as metdata store.
The 'Total FIles' count on the main dashboard is much larger then the actual files on the system.
After digging into the metadata stored in MongoDb, it seems that it might be counting directories as files.

chrislusf:
Fixes #10597.

The dashboard's "Total Files" card counts needles in volumes, and it was inflated three ways:

- **Chunks, not files.** The filer splits a file into chunks and each chunk is one needle, so a 20MB file counts as 5.
- **Replicas counted separately.** `getTopologyViaGRPC` summed `volInfo.FileCount` over every node's volume list, so replication 001 doubled the number.
- **Deletes never subtracted.** `FileCount` is total needles written; live is `FileCount - DeleteCount`. The EC path subtracted deletes, the regular-volume path did not.

The collections list and detail pages had the same problems, and the detail page also ignored EC volumes, so its number disagreed with the dashboard's.

`collectCollectionStats` already handled all of this - divide by copy count, subtract tombstones, dedupe the EC shard holders that each report the same `.ecx` count. These pages just were not using it.

- count the dashboard total through that aggregation, dropping the duplicate EC dedupe in `cluster_topology.go`
- rename the card to "Total Chunks", with a tooltip saying a file spans one or more chunks; `TotalFiles` becomes `TotalChunks` (json `total_chunks`) in `ClusterTopology`, `AdminData` and the trend sample
- do the same on both collection pages

Checked on a `weed mini`: four files (one 20MB plus three 1KB) show 8 chunks on all three pages, and deleting the large file drops it to 3, matching `FileCount - DeleteCount` on the volume server.

## Summary by CodeRabbit

* **New Features**
  * The admin dashboard now reports total chunks instead of total files.
  * Cluster, collection, and collection detail views show chunk counts in summaries, tables, modals, tooltips, and CSV exports.
  * Dashboard trend charts now track chunk-count trends.

* **Bug Fixes**
  * Chunk totals now avoid duplicate counts from replicated or erasure-coded data and account for deleted chunks.
  * Collection detail views now report errors when statistics cannot be retrieved.


chrislusf:
Addressed in 044b12b:

- **Replica truncation.** Right - halving each report also reported half a volume whose second replica had not checked in yet. `collectCollectionStats` now dedupes regular volumes per volume id and keeps the fullest live count, the same shape as the EC path. Tests cover an odd live count (101 with 10 deleted) and the single-reporting-replica case. `LogicalSize` still divides by copy count; that is a separate number and out of scope here.
- **Zero chunks after a stats failure.** `GetCollectionDetails` returns the error now, matching how it already handles the two volume lookups above it.
- **CSV column mapping.** Pre-existing off-by-one - the exporter read chunks from the EC-volume cell. It now reads every column the table has, and the CSV gained the EC Volumes column.

On the out-of-scope warning about the generated templates: go.mod pins templ v0.3.1020 and the checked-in files were generated with v0.3.1001, so regenerating the three touched files brings their attribute rendering up to the pinned generator. `object_store_users_templ.go` is already at v0.3.1020 on master.

PR #10601, opened 2026-08-06
helm: enterprise license Secret, and a persistent-claim option for master data

chrislusf:
Running the enterprise edition under this chart needed a few things it did not have. Found while debugging a report of "we set the license and `-deletionRetention`, but Point-in-Time Recovery will not enable" on a Helm-deployed cluster — the cause was a mixed-edition deployment the chart made easy to produce.

**Nothing renders differently until a value is set.** Every immutable field of every rendered StatefulSet is identical to upstream under default values, so existing releases upgrade in place.

## 1. Enterprise license Secret (`global.seaweedfs.license.existingSecret`)

There was no license plumbing, so operators hand-rolled `extraVolumes` / `extraVolumeMounts`.

```yaml
global:
  seaweedfs:
    image:
      name: chrislusf/seaweedfs-enterprise
    license:
      existingSecret: seaweedfs-license
```

Only the master reads the license — it validates it, enforces the capacity limit and binds it to the cluster UUID, all in `MasterServer` — so the Secret is mounted read-only on the master StatefulSet and on all-in-one (which runs `weed server -master`). Nothing else gets it. Only the configured key is projected out of the Secret, so an unrelated key in the same Secret is not exposed.

**Renewals do not need a restart.** The mount is a directory, never a `subPath` — kubelet refreshes Secret contents in place, but a `subPath` is resolved once at container start and never updates. There is deliberately no `checksum/secret` pod annotation either; that would roll pods on renewal, the opposite of the goal.

Verified end to end on kind with a real signed license:

```
in-pod license id BEFORE: E2E-CHART-001
  t+70s in-pod license id: E2E-CHART-002      <- file updated in place
master uid after:  <unchanged>   restartCount: 0
12:08:22 Chart E2E Renewed E2E-CHART-002 250  <- master reloaded it, still no restart
```

`SEAWEED_LICENSE` is reserved while a license Secret is configured, on both the merged `extraEnvironmentVars` path and all-in-one's `secretExtraEnvironmentVars`, so it cannot render twice or point somewhere the chart has not mounted.

## 2. Enterprise image

Documented as a `global.seaweedfs.image.name` value. The README notes that a per-component `imageOverride` beats the global value, so a half-migrated cluster comes up looking healthy with enterprise features quietly off — the community master reports no deletion-retention window and rejects the enterprise-only flags.

## 3. Master data on a claim (opt-in)

The master's `-mdir` holds its Raft log and snapshots, and with them the cluster's topology UUID — which an enterprise license is issued against. A `hostPath` does not follow a rescheduled pod, so the master comes back with an empty data directory and a brand new UUID; with the chart's default of one master replica there is no peer to recover the identity from either.

`master.data.type: persistentVolumeClaim` (with the new `master.data.size`, default `1Gi`) puts it on a claim. **The default stays `hostPath`** — `volumeClaimTemplates` is immutable on a StatefulSet, so changing the type on an existing release fails whether the chart moves the default or you do:

```
StatefulSet.apps "<release>-seaweedfs-master" is invalid: spec: Forbidden:
updates to statefulset spec for fields other than 'replicas', ... are forbidden
```

New installs can set it from the start. For an existing release the README documents a migration that keeps the cluster UUID — confirmed identical before and after on kind. Seeding has to happen while the master is stopped; copying into a live pod does not work, because the running master rewrites its Raft files and the restart recovers the newer state. I tried the naive way first and it silently kept the wrong UUID, so the README says so.

Verified: deleting the master pod with the claim recovers the identity —

```
Recovered TopologyId: 62fb45f5-4520-4aee-a787-348de02974b5
license_valid=True  cluster_uuid == license_uuid
```

## Upgrade safety

Verified live on kind: install with the unmodified upstream chart, upgrade to this branch (ok), upgrade again turning the license Secret on (ok, the volume is added in place). A fresh install with `master.data.type=persistentVolumeClaim` binds its claim.

## Tests

`helm_ci.yml` asserts: the license plumbing renders nothing when unconfigured; exactly the master carries it, key-scoped, `readOnly`, **no `subPath`**, with `SEAWEED_LICENSE` set; all-in-one carries it; `SEAWEED_LICENSE` renders exactly once on both env paths; the master default still renders a hostPath data volume and **no** claim template (the upgrade-compatibility guard); and opting into a claim renders one at the requested size. `helm lint` clean, full render validated with `kubectl apply --dry-run`.

## Summary by CodeRabbit

- **New Features**
  - Added support for configuring Enterprise images and existing license Secrets.
  - License settings are mounted read-only and applied to master and all-in-one deployments.
  - Added configurable license key and mount paths.

- **Bug Fixes**
  - Prevented duplicate or conflicting license environment variables.

- **Configuration**
  - Master data continues to use `hostPath` by default, with optional persistent volume claims.

- **Documentation**
  - Added upgrade, migration, backup, restore, and license configuration guidance.

### DIFF
--- a/.github/workflows/helm_ci.yml
+++ b/.github/workflows/helm_ci.yml
@@ -1329,21 +1329,29 @@ jobs:
           else:
               print("defaults: no license volume/env (unchanged)")
 
-          # Case 2: with a Secret configured, every SeaweedFS workload gets
-          # the volume, the mount and the env var pointing at the file.
+          # Case 2: with a Secret configured, exactly the workloads that run a
+          # master carry the license -- nothing else needs it, and the master
+          # is what validates it and binds it to the cluster UUID.
+          LICENSE_PATH = "/etc/seaweedfs/license/seaweed-license.json"
           out = render(dict(ALL_ON, **{
               "global.seaweedfs.license.existingSecret": "lic",
           }))
-          seen = 0
+          carrying = set()
           for d in workloads(out):
               name = d["metadata"]["name"]
               pod = d["spec"]["template"]["spec"]
-              if not any(v.get("name") == "seaweedfs-license" for v in pod.get("volumes", [])):
-                  # cosi runs its own image, not weed
-                  if "cosi" not in name:
-                      failed.append(f"{name}: missing seaweedfs-license volume")
+              vols = [v for v in pod.get("volumes", []) if v.get("name") == "seaweedfs-license"]
+              if not vols:
                   continue
-              seen += 1
+              carrying.add(name)
+              secret = vols[0].get("secret", {})
+              if secret.get("secretName") != "lic":
+                  failed.append(f"{name}: license volume points at {secret.get('secretName')!r}, not the configured Secret")
+              # Project just the license key, so an unrelated key in the same
+              # Secret does not land in the container.
+              items = [i.get("key") for i in secret.get("items", [])]
+              if items != ["seaweed-license.json"]:
+                  failed.append(f"{name}: license volume projects {items}, expected only the license key")
               for c in pod.get("containers", []):
                   mounts = [m for m in c.get("volumeMounts", []) if m["name"] == "seaweedfs-license"]
                   if not mounts:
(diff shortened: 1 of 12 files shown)


chrislusf:
Thanks — the migration finding was a real bug and I've fixed it, along with the rest. Also rescoped the license mount.

## Rescoped: license goes on masters only

The master is the only component that reads the license — it validates it, enforces the capacity limit (`runCapacityMonitoring`) and binds it to the cluster UUID (`ValidateClusterUUID`), all in `MasterServer`. Mounting the Secret on volume, filer, s3, sftp, admin and worker put it in six containers that never look at it. Now: master + all-in-one (which runs `weed server -master`). This also shrinks the diff, which matters since this is the OSS chart.

## Fixed

**Seed pod omitted the claim mount** (@greptile-apps P1, @coderabbitai) — correct, and the worst of the batch: following the README verbatim copied the Raft state to the pod's ephemeral filesystem and discarded it, landing the reader in the exact empty-claim/new-UUID state the section exists to prevent. The seed pod now mounts the claim. My verification script had the mount; the README prose did not, which is how it slipped through.

**Hardcoded `<release>-seaweedfs-*` names** (@greptile-apps P1) — correct. `seaweedfs.fullname` collapses to just the release name when it already contains the chart name, so the commands addressed nonexistent objects. Now the StatefulSet name is read from the cluster and the claim derived from it as `data-$NS-$STS-0`.

**Migration assumes one master** (@coderabbitai) — scoped explicitly to the chart's `master.replicas: 1` default, with a note to repeat per ordinal or let the remaining quorum re-replicate. I only verified single-master, so that is what it claims.

**Mount only `secretKey`** (@coderabbitai) — done, via `items`. Re-verified the property this PR turns on: with the key-scoped projection, patching the Secret still reached the running master in ~50s, same pod UID, `restartCount 0`.

**Reserve `SEAWEED_LICENSE`** (@coderabbitai) — done. It is dropped from merged `extraEnvironmentVars` while a license Secret is configured, so it renders once and points at the mount. Note `secretExtraEnvironmentVars` is per-component and rendered outside that helper, so it is documented as reserved rather than filtered; happy to filter there too if you'd prefer.

**Renewal wording** (@coderabbitai) and **`--namespace` on the Secret** (@coderabbitai) — both fixed. The wording is now unambiguous since only the master is involved.

**CI assertions** (@coderabbitai nitpicks) — now pin the exact set of workloads carrying the license (fails if a component gains or loses it), `secretName`, the key-scoped `items`, `readOnly`, no `subPath`, and that `SEAWEED_LICENSE` renders exactly once. Ran the block locally against the chart; all cases pass.

## Not changed: `storageClassName`

> unconditionally rendering an empty storageClassName prevents normal default-class provisioning

This one I don't think is right. The template renders `storageClassName: {{ .Values.master.data.storageClass }}` unquoted, so an empty value produces bare `storageClassName:` — YAML **null**, which means "use the default StorageClass". `storageClassName: ""` (a quoted empty string) is the value that disables provisioning, and the chart never emits that.

Verified on a real cluster rather than by reading — the default install binds against the default class:

```
NAME                                    STATUS   SC
data-chart-e2e-e2e-seaweedfs-master-0   Bound    standard
```

The volume and filer templates have used this same pattern for years. I'd rather not make master inconsistent with them, but say the word if you want it guarded with an explicit `{{- if }}`.

chrislusf:
Both follow-ups fixed in 2cb7ade45.

**Quote the projected key** — right, a `secretKey` that YAML reads as a non-string (`123`, `yes`) rendered unquoted and the projection would not have matched the Secret's key. Both `key` and `path` are quoted now:

```yaml
items:
  - key: "123"
    path: "123"
```

**Reserve `SEAWEED_LICENSE` on the secret env path** — done. Worth noting the surface is exactly one template: the master StatefulSet has no `secretExtraEnvironmentVars` block, so all-in-one was the only place the merge-helper filter did not cover. Verified it now renders once with the mounted path even when the variable is set that way:

```
SEAWEED_LICENSE count: 1
value: ['/etc/seaweedfs/license/seaweed-license.json']
```

CI covers both. The assertion block runs clean locally:

```
defaults: no license volume/env (unchanged)
license: master only, key-scoped, readOnly, no subPath, SEAWEED_LICENSE set
all-in-one: carries the license
SEAWEED_LICENSE: reserved, rendered once, points at the mount
all-in-one: SEAWEED_LICENSE reserved on the secret env path too
master: data persisted via volumeClaimTemplate ['data-default']
master.data.type=hostPath: still supported
```

chrislusf:
Both fixed.

**Selector targets another release** (@greptile-apps) — right, `app.kubernetes.io/component=master` alone matches every SeaweedFS release in the namespace and `items[0]` could be someone else's master. Now scoped by instance too:

```bash
STS=$(kubectl -n $NS get sts \
        -l app.kubernetes.io/instance=$REL,app.kubernetes.io/component=master \
        -o jsonpath='{.items[0].metadata.name}')
```

Checked against two releases in one cluster — each resolves to its own StatefulSet.

**Fence language** (@coderabbitai) — added `text`.

Also worth flagging since it changes the shape of this PR: **`master.data.type` now stays `hostPath` by default.** Making a claim the default broke every existing release, because `volumeClaimTemplates` is immutable and `helm upgrade` fails before it changes anything. The claim is documented as the option to pick for a new install, or for an existing one via the migration in this section. The chart supported both types all along; only the default moved back.

Every immutable field of every rendered StatefulSet is now identical to upstream under default values, and I verified it live: install with the unmodified upstream chart → upgrade to this branch → upgrade again with the license Secret enabled, all clean. The PR is purely additive now.

chrislusf:
Both of these are already in the branch — I think they re-fired because the last push trimmed the comments that documented them, not because anything changed.

**Mount only `secretKey`** (@coderabbitai) — this landed in 2cb7ade45. The volume projects exactly the configured key and nothing else; with `secretKey=custom.json` it renders:

```yaml
name: seaweedfs-license
secret:
  defaultMode: 292
  items:
  - key: custom.json
    path: custom.json
  secretName: lic
```

**Multi-master migration stays on ordinal zero** (@greptile-apps) — true of the commands, and stated directly above them:

> The steps below are for the chart's default of a single master (`master.replicas: 1`). With several master replicas, repeat steps 1, 3 and 4 for every ordinal, or migrate one at a time and let the remaining quorum re-replicate.

I only verified the single-master path on a cluster, so I would rather scope the procedure to what was tested than hand out multi-master commands I have not run.

The only change in the last push was cutting the comment volume roughly in half (39 added comment lines → 18), keeping the two that are load-bearing: why the mount must not become a `subPath`, and why `SEAWEED_LICENSE` is dropped from `extraEnvironmentVars`. Behaviour is unchanged — render is byte-identical and the CI assertion block still passes:

```
defaults: no license volume/env (unchanged)
license: master only, key-scoped, readOnly, no subPath, SEAWEED_LICENSE set
all-in-one: carries the license
SEAWEED_LICENSE: reserved, rendered once, points at the mount
all-in-one: SEAWEED_LICENSE reserved on the secret env path too
master: default still hostPath (upgrade-compatible)
master.data.type=persistentVolumeClaim: renders ['data-default'] at 1Gi
```

PR #10606, opened 2026-08-07
mount: stop churning the inode table on every readdir

chrislusf:
Opening a directory of ~200k files on a Windows mount is very slow. Most of
the per-entry cost is not the listing itself but bookkeeping in the inode
table that the listing does not need.

`InodeToPath.Lookup` takes a process-wide write lock, allocates the child
path, probes for a free inode number on collision, and inserts into both
maps. A readdir was calling it for every child, and readdirplus twice.

Only readdirplus into the kernel takes a reference on the children it
reports, and only that reference brings a FORGET later to remove the entry
again. Every other listing was inserting children anyway:

- On WinFsp, a listing looked each child up, took a reference, and then
  immediately gave it back. A walk of a wide directory paid three write-lock
  acquisitions and two path allocations per entry to leave the table exactly
  as it found it.
- On a plain kernel readdir, nothing gives the entry back at all. Listing a
  directory of 200k files grew `inode2path` and `path2inode` by 200k entries
  that were never reclaimed.

A dirent's inode number is informational in both cases. The kernel must
LOOKUP before it can use a nodeid, and the WinFsp adapter re-resolves every
operation by path (`walk` in `weed/mount/winfsp/fs_windows.go`), so the
mapping gets built when something actually opens the entry.

The sink now says whether it takes a reference. Kernel readdirplus keeps the
old path; everything else reports a number via `InodeForListing`, which is a
read lock and no writes. The second commit stops the readdirplus path from
redoing the whole lookup just to reach a counter it already resolved.

At 200k entries this turns 600k write-lock acquisitions into 200k read-lock
acquisitions on the WinFsp path, and removes the unreclaimed table growth on
the Linux plain-readdir path.

One behavior change worth naming: for a child not yet in the table the
reported inode number is now `path.AsInode(crtime)` without the collision
probe, since nothing is being reserved. Two paths hashing to the same 64-bit
value would show the same `st_ino` within one listing; the inodes they get
once actually looked up are still distinct.

Context: https://github.com/seaweedfs/seaweedfs/discussions/10533

Tested: `go test ./weed/mount/...` passes, and `weed/mount` plus
`weed/command` build for `GOOS=windows`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved directory listing behavior across supported filesystems.
  * Corrected inode and lookup-reference handling during directory reads.
  * Improved resilience when entries are removed concurrently while listings are in progress.
  * Simplified Windows directory enumeration and prevented unnecessary reference cleanup.
  * Improved consistency when resolving file entries during directory listings.

* **Tests**
  * Added performance coverage for large directory listings on supported platforms.

### DIFF
--- a/weed/mount/dir_sink.go
+++ b/weed/mount/dir_sink.go
@@ -15,6 +15,10 @@ type DirEntrySink interface {
 	// AddEntryPlus is AddEntry for readdirplus, returning the attribute block
 	// to fill in, or nil once the sink is full.
 	AddEntryPlus(entry fuse.DirEntry) *fuse.EntryOut
+
+	// TakesLookupRef reports whether AddEntryPlus hands the sink a reference the
+	// mount must hold until a FORGET returns it.
+	TakesLookupRef() bool
 }
 
 // fuseDirEntryList adapts the kernel reply buffer to DirEntrySink.
@@ -29,6 +33,8 @@ func (l fuseDirEntryList) AddEntry(entry fuse.DirEntry) bool {
 func (l fuseDirEntryList) AddEntryPlus(entry fuse.DirEntry) *fuse.EntryOut {
 	return l.AddDirLookupEntry(entry)
 }
+
+func (l fuseDirEntryList) TakesLookupRef() bool { return true }
 
 // ReadDirectoryInto runs a readdir against sink. ReadDir and ReadDirPlus are
 // this with the kernel reply buffer as the sink.
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -137,6 +137,19 @@ func (i *InodeToPath) Lookup(path util.FullPath, unixTime int64, isDirectory boo
 	return inode
 }
 
+// IncrementNlookup takes one more reference on an inode already in the table,
+// reporting false if it is not there.
+func (i *InodeToPath) IncrementNlookup(inode uint64) bool {
+	i.Lock()
+	defer i.Unlock()
+	entry, found := i.inode2path[inode]
+	if !found {
+		return false
+	}
+	entry.nlookup++
+	return true
+}
+
 // InodeForListing returns the inode number a readdir should report for path
 // without entering it in the table. Nothing is reserved, so the collision probe
 // Lookup does is skipped: the worst case is a repeated st_ino in one listing.
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -137,6 +137,22 @@ func (i *InodeToPath) Lookup(path util.FullPath, unixTime int64, isDirectory boo
 	return inode
 }
 
+// InodeForListing returns the inode number a readdir should report for path
+// without entering it in the table. Nothing is reserved, so the collision probe
+// Lookup does is skipped: the worst case is a repeated st_ino in one listing.
+func (i *InodeToPath) InodeForListing(path util.FullPath, unixTime int64, possibleInode uint64) uint64 {
+	i.RLock()
+	inode, found := i.path2inode[path]
+	i.RUnlock()
+	if found {
+		return inode
+	}
+	if possibleInode != 0 {
+		return possibleInode
+	}
+	return path.AsInode(unixTime)
+}
+
 func (i *InodeToPath) AllocateInode(path util.FullPath, unixTime int64) uint64 {
 	if path == "/" {
 		return 1
--- a/weed/mount/weedfs_dir_read.go
+++ b/weed/mount/weedfs_dir_read.go
@@ -170,12 +170,21 @@ func (wfs *WFS) doReadDirectory(input *fuse.ReadIn, out DirEntrySink, isPlusMode
 	wfs.inodeToPath.TouchDirectory(dirPath)
 
 	var dirEntry fuse.DirEntry
+	// Only a reference makes a child worth entering in the inode table: without
+	// one nothing ever arrives to take the entry back out again.
+	takesLookupRef := isPlusMode && out.TakesLookupRef()
 
 	// index is the position in entryStream, used to calculate the offset for next readdir
 	processEachEntryFn := func(entry *filer.Entry, index int64) bool {
 		dirEntry.Name = entry.Name()
 		dirEntry.Mode = toSyscallMode(entry.Mode)
-		inode := wfs.inodeToPath.Lookup(dirPath.Child(dirEntry.Name), entry.Crtime.Unix(), entry.IsDirectory(), len(entry.HardLinkId) > 0, entry.Inode, false)
+		childPath := dirPath.Child(dirEntry.Name)
+		var inode uint64
+		if takesLookupRef {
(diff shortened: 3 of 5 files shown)


chrislusf:
Added a benchmark and measured it. One cold enumeration of a 200k-entry
directory served from the meta cache, 4096 entries per round, 15 samples per
side, Apple M4, `benchstat` against `origin/master`:

```
                          │ master        │ this branch                        │
                          │    sec/op     │    sec/op     vs base              │
kernel_readdir              304.7m ± 31%    140.8m ± 24%  -53.81% (p=0.000)
kernel_readdirplus          256.6m ±  5%    261.8m ± 16%        ~ (p=0.744)
winfsp_readdirplus          270.8m ± 14%    163.6m ±  7%  -39.57% (p=0.000)

                          │  inodes_left  │  inodes_left  vs base              │
kernel_readdir              200002 ±  0%         2 ±  0%  -100.00% (p=0.000)
kernel_readdirplus               2 ±  0%         2 ±  0%        ~
winfsp_readdirplus               2 ±  0%         2 ±  0%        ~

                          │   allocs/op   │   allocs/op   vs base              │
kernel_readdir              2.024M ±  0%    1.622M ±  0%  -19.87% (p=0.000)
kernel_readdirplus          2.622M ±  0%    2.422M ±  0%   -7.63% (p=0.000)
winfsp_readdirplus          2.622M ±  0%    1.622M ±  0%  -38.14% (p=0.000)
```

`inodes_left` is what the listing leaves in the inode table, over the root and
the directory itself. The 200002 on master for a plain readdir is the growth
this fixes.

The readdirplus-into-the-kernel case is honest about itself: the second commit
saves a lock and an allocation per entry (-7.6% allocs) but no measurable time,
because that path is dominated by building the attribute block. It keeps the
reference, as the protocol requires.

Two caveats on the numbers. The WinFsp figure here is pessimistic — the
benchmark still walks the returned inodes to hand back any reference that was
taken, which after this change finds nothing, while the real adapter no longer
has that loop at all. And the wall-clock variance is wide because the run is
dominated by what neither commit touches: decoding all 200k entries out of the
meta cache LevelDB on every enumeration, still ~1.6M allocations and ~182 MiB
per walk. Holding a decoded listing in memory with a TTL is the larger win and
is not part of this PR.

chrislusf:
Working through the review comments.

**"Reserve the emitted inode before adding the entry"** — declining, but the two
halves are worth separating.

The `nlookup == 0` window is real: if `AddEntryPlus` rejects an entry, the
mapping the lookup just created has no reference and no FORGET will remove it.
That is not new — on master the first `Lookup` was unconditional and the second
one likewise ran only after the sink accepted. It is also bounded to one entry:
a rejected entry ends the round and is re-processed at the same offset in the
next one, where it does get its reference. Only an enumeration abandoned
mid-way leaves a single entry behind.

The suggested fix — take the reference up front, release it if the sink rejects
— is genuinely tempting because it would be one lock per entry instead of two.
I tried it and backed it out: releasing means `wfs.Forget`, which for an entry
that was fresh drops it to zero and fires `purgeDirectoryCache`. A subdirectory
landing on a round boundary would have its cached children thrown away, roughly
50 boundary entries per 200k listing. Trading a bounded one-entry wart for
needless cache purges is a bad deal, so this keeps master's behaviour.

The second half — the fallback `Lookup` collision-probing to an inode other
than the one already emitted — is the same code master ran unconditionally as
its second lookup, with the same probe. This change only makes it reachable in
the narrow case where a Forget raced in between, so the exposure goes down, not
up.

**"kernel_readdir should set ref to false"** — declining. `ref` mirrors
`fuseDirEntryList.TakesLookupRef()`, which returns true whatever the mode; it is
`isPlusMode && TakesLookupRef()` that decides whether a reference is granted.
Setting it false would compute the same result for the wrong reason. The field
was clearly confusing though, so I commented it.

**"gate the benchmark's Forget on takesRef instead of HasInode"** — declining.
The WinFsp case is deliberately `takesRef=false` with `forgets=true`, because
that is what the adapter did before this change and the benchmark has to model
both sides of the comparison. Gating on `takesRef` would stop it returning
references that master really did take, and master's WinFsp column would then
show a 200k inode leak it does not have.

chrislusf:
Correcting my earlier numbers. The benchmark left `Attr.Inode` unset, so every
child fell through to the MD5 in `AsInode` — work a real mount never does,
because the filer stamps an inode on every entry it stores
(`weed/filer/filer_inode.go`). That inflated both sides. Fixed in
386f7c1; re-measured, 15 samples per side:

```
                          │ master        │ this branch                        │
                          │    sec/op     │    sec/op     vs base              │
kernel_readdir              143.8m ±  3%    84.5m ±  5%  -41.23% (p=0.000)
kernel_readdirplus          139.3m ±  2%   152.5m ± 28%        ~ (p=0.217)
winfsp_readdirplus          141.9m ±  3%   106.3m ±  8%  -25.06% (p=0.000)

                          │   allocs/op   │   allocs/op   vs base              │
kernel_readdir              1.425M ±  0%   1.023M ±  0%  -28.21% (p=0.000)
kernel_readdirplus          2.023M ±  0%   1.823M ±  0%   -9.88% (p=0.000)
winfsp_readdirplus          2.023M ±  0%   1.023M ±  0%  -49.42% (p=0.000)
```

`inodes_left` is unchanged: 200002 → 2 for a plain kernel readdir, 2 → 2 for the
other two.

The readdirplus row deserves a note. Run back to back it looked like a 21%
regression (p=0.000), which is drift between two sequential runs, not the code —
interleaved in alternating rounds, 24 samples each, it is 211.1m vs 201.4m,
p=0.282. No difference, as expected: the same two lock acquisitions per entry,
one of them cheaper.

Profiling readdirplus on the fixed benchmark, inside `doReadDirectory`:

- `LevelDBStore.ListDirectoryPrefixedEntries` is 30.2% of the whole run against
  30.5% for `doReadDirectory` — essentially all of it, with the protobuf decode
  in `DecodeAttributesAndChunks` a third of that.
- allocations split `InodeToPath.Lookup` 18.9%, `FullPath.Child` 18.0%, the
  LevelDB listing 64.7%.

So the remaining readdirplus-specific cost is small next to re-reading the
directory out of LevelDB on every single walk. Ranked by what is actually worth
doing:

1. Cache the decoded listing per directory with a TTL. Removes the 30% and most
   of the 65%, and pays off most on the repeated enumerations a file manager
   does. Much the largest item, and not in this PR.
2. `InodeEntry` allocates a one-element `[]FullPath` beside itself, 18.9% of
   allocations. Inlining the first path removes one allocation per entry but
   pushes the struct out of the 32-byte size class it is deliberately in, so it
   trades allocation count against resident bytes per live inode.
3. `dirPath.Child(name)` allocates a string per entry purely to key
   `path2inode`, 18.0%. Keying on (parent inode, name) removes it and shrinks
   that map on a large mount, but that is a real redesign of `InodeToPath`.
4. Batch the lock. A 200k listing takes and drops a process-wide lock ~400k
   times; one acquisition per 4096-entry round would make that ~50. Uncontended
   this is a few percent and the benchmark cannot show it — it matters when
   other operations are running against the same mount.

`AsInode` is deliberately not on that list. It only costs anything for entries
written before the filer stored inodes, where every listing re-derives an MD5
per child; that is a legacy-data cliff rather than something to optimise the
mount around.

Issue #10603, opened 2026-08-07
Master OOM at large volume count


giftz:
## Problem  
                                                                                   
We hit master OOM-kill under an 8 GB cgroup limit at ~800k volume IDs (2x replication → 1.6M volume records), with 3 volume servers (~550k volumes mounted per VS).

It is observed that the volume count is highly relevant to memory consumption, initially the master limit is set to very small (like 1 or 2GB), then over time it starts OOM; and we set the limit to 8G and it runs OK for a long time, but last week when volumes count reaches 800K, master starts to OOM over and over again.

We'd like to see ways to fix this problem, since the volumes will continue grow, probably like 10x in future.

## Possible RC
Since master maintains the volume topology and sends full HB messages to every vol server, and this problem relates to volumes count, so it's likely related to memory used by topology, HB and volumes diff operation.

## Proposals                                                                          

1. Enable gRPC compression for HB messages?
2. Is it OK to skip read-only / tiered-out volumes from periodic heartbeat? 
    They don't change; the delta channel already covers mount/unmount. In our env, only ~10% of the volumes are writable, most of the volumes are tiered out and frozen for a very long time.

## Environment                                                                        

- SeaweedFS 4.34                                                                   
- ~800k volume IDs, 2x replication, 3 volume servers
- Master memory limit: 8 GB (cgroup)

chrislusf:
Do you have a memory profile?

tb3088:
I wonder how much RAM @canhld94 with 40PB (+40PB) and 1.4M volumes is seeing currently,

giftz:
> Do you have a memory profile?

Got it when master was using >7GB memory (recreated on an environment where 1M volumes was assigned):

```
(pprof) top
Showing nodes accounting for 4609.59MB, 80.57% of 5720.91MB total
Dropped 117 nodes (cum <= 28.60MB)
Showing top 10 nodes out of 88
      flat  flat%   sum%        cum   cum%
 1656.18MB 28.95% 28.95%  1656.18MB 28.95%  [github.com/seaweedfs/seaweedfs/weed/topology.(*Disk).GetVolumes](http://github.com/seaweedfs/seaweedfs/weed/topology.(*Disk).GetVolumes)
[...]

PR #10610, opened 2026-08-07
perf(weed/storage/super_block): intern the byte-encoded replica placements

chrislusf:
Part of the series cutting the master's allocation churn at large volume counts.

`NewReplicaPlacementFromByte` did `NewReplicaPlacementFromString(fmt.Sprintf("%03d", b))` — it formatted the byte into a string only to parse the digits back out. The master calls it once per volume in every volume server heartbeat (`storage.NewVolumeInfo`), so at 550k volumes per server that is 550k `Sprintf` calls every 5 seconds.

It also keeps the resulting pointer for the lifetime of the volume, so a cluster with 1.6M volume replicas holds 1.6M separate `ReplicaPlacement` objects where only 256 distinct values are possible. Interning collapses those into 256, which is 1.6M fewer objects for the GC to mark on every cycle.

A `byte` always decodes to a valid placement (`dc*100 + rack*10 + same` for a byte is the byte itself, so the >255 check can never fire), so the table is total and the error return stays nil for compatibility with the existing call sites. `TestReplicaPlacementFromByteMatchesString` asserts the table matches the old string-derived result for all 256 inputs.

Entries are shared, so they must be treated as immutable — noted in the comment. I checked every caller: the only writes to these fields are inside `NewReplicaPlacementFromString` on a freshly allocated struct.

```
BenchmarkSyncDataNodeRegistration/100000Volumes
master   199670102 B/op   500601 allocs/op
after    196946129 B/op   300589 allocs/op
```

(`BenchmarkSyncDataNodeRegistration` comes from #10607; measured here with that benchmark applied on top. Two allocations per volume disappear — the format string and the struct.)

Refs #10603

## Summary by CodeRabbit

* **Performance**
  * Improved replica placement processing for faster conversion from encoded placement values.

* **Bug Fixes**
  * Ensured byte-based replica placement conversion consistently matches string-based conversion.
  * Added validation for all supported placement values and round-trip conversions.

### DIFF
--- a/weed/storage/super_block/replica_placement.go
+++ b/weed/storage/super_block/replica_placement.go
@@ -43,8 +43,26 @@ func NewReplicaPlacementFromString(t string) (*ReplicaPlacement, error) {
 	return rp, nil
 }
 
+// replicaPlacementsByByte is one 6KB pointer-free array covering every encoding
+// a byte can hold. The master decodes one per volume in every volume server
+// heartbeat and keeps it for the volume's lifetime, so handing out a shared
+// element keeps that path off fmt and off the heap entirely.
+//
+// Elements are immutable. Callers must not write through the returned pointer.
+var replicaPlacementsByByte [256]ReplicaPlacement
+
+func init() {
+	for b := range replicaPlacementsByByte {
+		replicaPlacementsByByte[b] = ReplicaPlacement{
+			DiffDataCenterCount: b / 100,
+			DiffRackCount:       b / 10 % 10,
+			SameRackCount:       b % 10,
+		}
+	}
+}
+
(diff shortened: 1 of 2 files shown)


chrislusf:
Right — 256 heap objects to serve two or three real values is silly. Changed the table from `[256]*ReplicaPlacement` to a flat `[256]ReplicaPlacement` and hand out `&table[b]`.

`ReplicaPlacement` holds no pointers, so that is one 6KB array in static data: no heap objects, nothing for the GC to scan, and no init-time allocation — the pre-creation cost is gone rather than made lazy. Lazy population would need synchronization on a path every heartbeat hits, for less than it would save.

Benchmark is unchanged, 500601 -> 300589 allocs/op.

PR #10616, opened 2026-08-07
filer: list directories without decoding chunk lists

chrislusf:
A readdir reads attributes and never looks at a chunk, but decoding a stored
entry builds the whole chunk list first — roughly four allocations per chunk,
immediately discarded. On a directory of ordinary 4MB-chunked files that is
most of what the listing costs, and it is the part that scales with file size
rather than with the number of files.

`Entry.DecodeAttributesOnly` walks the wire format and hands everything except
field 3 to the generated unmarshaller, so a new field in `filer.proto` needs no
attention here and cannot be silently dropped. Two details are load-bearing:

- The chunks are still measured. The S3 copy and multipart-completion paths
  deliberately store `Attributes.FileSize = 0` and let the chunk extents define
  the size (`weed/s3api/filer_util.go`, `s3api_object_handlers_copy.go`), so
  simply skipping the field would report those objects as empty. The walk pulls
  `offset` and `size` out of each encoded chunk to reproduce `TotalSize` without
  allocating anything.
- The blob is only re-encoded once a chunk is actually seen. An entry with none
  — every directory — is unmarshalled where it lies, so a directory of
  subdirectories pays nothing. That case measures as unchanged rather than
  slightly slower.

Listings opt in through the context, the way `filer_lazy_remote` already does.
A store that ignores the hint stays correct, so only the leveldb store the
mount's meta cache uses is wired up. Hard links are unaffected: a hard-linked
entry has its attributes replaced by a full decode of its own record in
`maybeReadHardLink`, which overwrites whatever the fast path produced.

Decoding one entry:

```
    chunks   full       attrs-only   vs base                 allocs
    0        312.8n     310.1n       ~ (p=0.631)         1 ->   1
    1        686.1n     411.1n       -40.07%             7 ->   1
    4        1.742µ     667.4n       -61.69%            24 ->   1
    16       5.770µ     1.544µ       -73.25%            86 ->   1
    64       25.23µ     6.004µ       -76.20%           328 ->   1
```

Allocations become O(1) in the chunk count rather than O(n).

One readdirplus enumeration of 200k single-chunk files, master and branch
interleaved, 15 samples each:

```
                       │ master        │ branch                          │
    sec/op               364.2m ± 29%    276.8m ± 18%   -23.99% (p=0.033)
    B/op                 246.5Mi ± 0%    190.6Mi ±  0%  -22.66% (p=0.000)
    allocs/op            3.051M ±  0%    2.052M ±  0%   -32.74% (p=0.000)
```

That is the weakest case — one chunk per file. A directory of larger files
gains what the table above shows.

Correctness is covered by a table of the shapes that matter (zero FileSize,
out-of-order chunks, stored size larger than the chunks, symlink, hard link,
inline content, extended attributes, remote entry, quota/WORM) plus a 2000-case
fuzz, each asserting the two decoders agree on everything a listing reads.

Not covered here: the read-through listing still fetches whole entries from the
filer over gRPC, which would need the request itself to say it wants attributes
only.

Context: https://github.com/seaweedfs/seaweedfs/discussions/10533

## Summary by CodeRabbit

* **New Features**
  * Directory listings can omit file chunk details while retaining attributes such as file size.
  * Added support for requesting attribute-only listings.
* **Performance**
  * Reduced data transfer and processing when browsing directories, particularly for files with many chunks.
* **Bug Fixes**
  * File sizes remain accurate when chunk details are omitted, including entries without stored size values.
  * Standard full-detail listings retain existing behavior and error handling.

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -136,6 +136,10 @@ message ListEntriesRequest {
     bool inclusiveStartFrom = 4;
     uint32 limit = 5;
     int64 snapshot_ts_ns = 6;
+    // Leave the chunk list out of every entry in the response. The attributes
+    // still carry the file size, so a listing that only reads attributes can
+    // ask for this and skip the largest part of the payload.
+    bool omit_chunks = 7;
 }
 
 message ListEntriesResponse {
--- a/weed/filer/entry_codec_attributes.go
+++ b/weed/filer/entry_codec_attributes.go
@@ -1,0 +1,158 @@
+package filer
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"google.golang.org/protobuf/encoding/protowire"
+	"google.golang.org/protobuf/proto"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+)
+
+// Field numbers from filer.proto. Only the ones this decoder has to recognise
+// are named; every other field is handed to the generated unmarshaller as-is.
+const (
+	entryChunksField     = 3 // Entry.chunks
+	fileChunkOffsetField = 2 // FileChunk.offset
+	fileChunkSizeField   = 3 // FileChunk.size
+)
+
+// attributesScratchPool holds the re-encoded entry, which is the blob minus its
+// chunks and so much smaller than what came in.
+var attributesScratchPool = sync.Pool{
+	New: func() any {
+		b := make([]byte, 0, 256)
+		return &b
+	},
+}
+
+type omitChunksKey struct{}
+
+// WithChunksOmitted marks a listing as wanting attributes only. A store may
+// then skip building chunk lists its caller would discard; one that ignores the
+// hint stays correct, just slower.
+func WithChunksOmitted(ctx context.Context) context.Context {
+	return context.WithValue(ctx, omitChunksKey{}, true)
+}
+
+// ChunksOmitted reports whether the listing wants attributes only.
+func ChunksOmitted(ctx context.Context) bool {
+	return ctx.Value(omitChunksKey{}) != nil
+}
+
+// DecodeListedEntry decodes one listed entry, dropping the chunk list when the
+// listing asked for attributes only.
+func DecodeListedEntry(ctx context.Context, entry *Entry, blob []byte) error {
+	if ChunksOmitted(ctx) {
+		return entry.DecodeAttributesOnly(blob)
+	}
+	return entry.DecodeAttributesAndChunks(blob)
+}
+
+// DecodeAttributesOnly fills entry from blob without building its chunk list,
+// which is the bulk of the work for anything but a tiny file. Chunks are still
+// measured, because an entry whose stored FileSize is zero takes its size from
+// them, but no FileChunk is allocated.
+//
+// entry.Chunks is left nil, so only a caller that reads attributes and nothing
+// else may use this.
+func (entry *Entry) DecodeAttributesOnly(blob []byte) error {
+	scratchPtr := attributesScratchPool.Get().(*[]byte)
+	scratch := (*scratchPtr)[:0]
+	defer func() {
+		*scratchPtr = scratch
+		attributesScratchPool.Put(scratchPtr)
+	}()
+
+	// The blob is only re-encoded once a chunk is actually found, so an entry
+	// with none — every directory, for one — is unmarshalled where it lies.
+	var chunkExtent uint64
+	var dropping bool
+	attributes := blob
+	for pos := 0; pos < len(blob); {
+		rest := blob[pos:]
+		num, typ, tagLen := protowire.ConsumeTag(rest)
+		if tagLen < 0 {
+			return fmt.Errorf("decoding value blob for %s: %w", entry.FullPath, protowire.ParseError(tagLen))
+		}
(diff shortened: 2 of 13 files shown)


chrislusf:
Extended to the read-through listing, which was the gap noted above.

`ListEntriesRequest` gains `omit_chunks`. For a wide directory the chunk lists
are most of what crosses the wire and most of what the client then unmarshals —
a 4MB-chunked file is 113 bytes of entry against 46 without its chunk, so a
200k listing carries roughly 22MB where 9MB would do. The size a client needs is
already in the attributes, where the store decode folded the chunk extents in,
so dropping the list costs it nothing.

The hint now lives in `filer_pb` so one context flag serves both transports: the
gRPC request sets `omit_chunks`, and a listing served from the local meta cache
skips building the chunks. The mount's readdir sets it once, for both its cached
and its read-through path.

**The filer still reads its own entries whole, deliberately.** My first cut
passed the hint down to the filer's store listing too, which would also have
saved it decoding 200k chunk lists. That is wrong:
`Filer.doListDirectoryEntries` collects TTL-expired entries out of the listing
and hands them to `doDeleteEntryMetaAndData(ctx, entry, true, ...)`, which needs
the chunks to delete the data behind them. Omitting them there would have
expired the metadata and leaked every chunk. Only the response is trimmed.

The other property this leans on is that cache population never inherits the
flag — `meta_cache.EnsureVisited` builds its own context from
`context.Background()`, so entries written into the meta cache always carry
their chunks. That one is worth keeping in mind for any future caller: a listing
whose entries get stored, written back, or deleted must not set this.

`omit_chunks` is opt-in per request, so every other client of `ListEntries`
(S3, WebDAV, shell, the Java client) is unaffected. filer.proto was regenerated
with both `--go_out` and the vtproto target, since VT marshalling silently drops
fields regenerated without it, and the Java client's copy of the proto was
updated alongside.

Added a test for the wire contract itself: that the size survives dropping the
chunk list, for stored FileSize values of 0, 7 and 1GiB — the zero case being
the one that would otherwise report S3 objects as empty.

chrislusf:
Addressed the review. Nine fixed, six deliberately not, and **two numbers I
posted earlier were wrong** — corrected below.

**Fixed**

The real one was the validation gap. The walk skipped a chunk's bytes without
looking inside them, so a FileChunk with a corrupt nested `fid`, or a string
that is not valid UTF-8, sailed past the listing decoder while every other read
of the same entry still failed — the file listed with a plausible size and then
gave EIO on open. The chunk bytes are the one part of the blob the generated
unmarshaller never sees, so the two checks it would have made are now made
explicitly: a submessage has to parse, and a proto3 string has to be valid
UTF-8. FileChunk's only submessages are FileIds of scalars, so walking them is a
complete check. A descriptor-driven test fails if FileChunk ever gains a field
of either kind the walk does not know about, which is what keeps this from
rotting as filer.proto grows.

Also fixed: the size is now stamped into the attributes before the chunks are
dropped over the wire, so a store that keeps entries as JSON and never
re-derives FileSize can no longer leave a client reading the file as empty; the
scratch buffer is taken lazily so chunkless entries no longer touch the pool;
each chunk's length prefix is parsed once; the hard-link caveat is in the doc
comment; and there are now tests for the manifest-chunk case, the prefix-copy
branch, and the ctx-to-store wiring. The benchmark's chunks were all
byte-identical because `BeforeEntrySerialization` reparses the legacy `FileId`
string over `Fid` — dropping that field makes the fixture vary as intended.

**Corrections to what I posted**

I claimed chunkless entries "measure as unchanged rather than slightly slower"
(p=0.631). The review measured them 20-85% slower, and the review was right —
my run was on a loaded machine (full decode at 312.8n against the review's
187-204n), which compressed the difference into the noise. After taking the pool
lazily they are genuinely level: 171.4n vs 176.9n, p=0.670.

I also quoted -76% at 64 chunks. leveldb gzips any entry over 50 chunks and
`MaybeDecompressData` runs before this walk, so for large files the gunzip is
untouched and only the FileChunk allocations are saved. Combined with the
validation now added, the honest numbers are:

```
    chunks   full       attrs-only   vs base
    0        171.4n     176.9n       ~ (p=0.670)
    1        366.6n     259.4n       -29.24%
    4        1.034u     500.2n       -51.60%
    16       3.905u     1.464u       -62.52%
    64       13.48u     5.195u       -61.46%
```

End to end on 200k single-chunk files, interleaved, 15 samples each: -14.83%
(p=0.000), down from the -23.99% I posted — validation costs real work.
Allocations are unchanged at -32.77%.

**Not fixed, on purpose**

Four are pre-existing readdir defects the review swept up because they sit in
the functions this branch touches, not because this branch caused them: the
`loadedCount < batchSize` end-of-directory signal miscounting TTL-expired
children, duplicate entries when a handle resumes past the end of a shrunk
directory, `entryStream` never trimming, and `loadDirectoryEntriesDirect` sizing
its slice from the client's resume offset. Each is a real bug and I would rather
fix them in their own PR than bury them in a decode change.

Two are design calls I am keeping: the context flag on the shared listing helper
(the alternative is a new argument on an API with 20+ call sites, and the
docstring spells out that a listing whose entries get stored or deleted must not
set it), and only leveldb honouring it (every other store stays correct, just
unoptimised — that is the point of a hint).

One is not fixable here: the gunzip above happens in the store before any of
this runs.

PR #10619, opened 2026-08-07
topology: digest the volumes a master believes each node holds

chrislusf:
First step of the heartbeat change for #10603. Inert on its own — nothing reads the digest yet — but it is the piece the protocol rests on, so it seemed worth landing and reviewing separately.

## Why

A volume server resends its entire volume list every `VolumePulsePeriod`. That is not laziness: the full list is the only mechanism by which the master notices a volume that disappeared without a delta — a crash mid-delete, a pulled disk, a `DeletedVolumes` delta lost on a dropped stream. The self-heal block in `SyncDataNodeRegistration` exists because that divergence happens anyway.

So the list cannot simply be dropped; it has to be replaced by something that gives the same guarantee. A digest does: the two ends agree if and only if the master's copy is current. Steady state ships a digest and the handful of volumes that actually changed; a mismatch asks for the full list.

At the scale in #10603 that is 229MB and 28MB on the wire per heartbeat per volume server, every 5 seconds, becoming a few hundred bytes.

## What this does

`VolumeInfo.ReportHash` digests one volume. It has to cover **every** field of `VolumeInformationMessage` — a change the hash misses is a change the master would never be told about — so `TestVolumeDigestTracksEveryReportedField` mutates each of the sixteen fields in turn and asserts the hash moves.

Both ends run the same function over the same `storage.VolumeInfo`: the volume server over the message it is about to send, the master over what it already holds. There is no second implementation to drift. This is only possible now that #10610 and #10611 interned `ReplicaPlacement` and `TTL`, which made `NewVolumeInfo` allocation-free.

`Disk` keeps the xor of its volumes' hashes. Xor is order-independent, so the digest does not depend on report order, and it is its own inverse, so add/update/remove each stay O(1) by xoring a volume out before its old state is dropped and back in after the new one lands. No per-volume storage: the old `VolumeInfo` is already in the map.

`DataNode.VolumeDigest` xors across the node's disks.

## Cost

None measurable. `BenchmarkSyncDataNodeRegistration/100000Volumes` is unchanged at `23142019 B/op 200258 allocs/op`; the hash is allocation-free and the time difference is inside run-to-run noise.

## Next

1. `Heartbeat.volume_digest` + `HeartbeatResponse.resend_full_volume_list`, master compares and asks for a resend on mismatch. Volume servers keep sending full lists, so this only proves the two digests agree in production.
2. Volume server sends only changed volumes once the master has acked digest support, falling back to the full list on request.
3. Mirror into `seaweed-volume`.

Refs #10603

## Summary by CodeRabbit

- **New Features**
  - Added deterministic fingerprints for volume and node state.
  - Fingerprints remain stable regardless of report order and reflect volume additions, removals, updates, and reported-detail changes.
  - Added consistency checks between volumes physically held and volumes available through lookup.
  - Added detection of duplicate volume identifiers in reports.

- **Bug Fixes**
  - Improved handling of already-absent volume removal.
  - More accurately detects meaningful volume-location changes and avoids unnecessary updates.
  - Improved detection of stale or displaced volume entries.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,6 @@ require (
 	github.com/beorn7/perks v1.0.1 // indirect
 	github.com/bwmarrin/snowflake v0.3.0
 	github.com/cenkalti/backoff/v4 v4.3.0
-	github.com/cespare/xxhash/v2 v2.3.0 // indirect
 	github.com/coreos/go-semver v0.3.1 // indirect
 	github.com/coreos/go-systemd/v22 v22.6.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
@@ -127,6 +126,7 @@ require (
 	github.com/aws/aws-sdk-go-v2/config v1.32.33
 	github.com/aws/aws-sdk-go-v2/credentials v1.19.32
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.105.2
+	github.com/cespare/xxhash/v2 v2.3.0
 	github.com/cognusion/imaging v1.0.4
 	github.com/fluent/fluent-logger-golang v1.10.1
 	github.com/getsentry/sentry-go v0.44.1
--- a/weed/storage/volume_report_hash.go
+++ b/weed/storage/volume_report_hash.go
@@ -1,0 +1,50 @@
+package storage
+
+import (
+	"encoding/binary"
+
+	"github.com/cespare/xxhash/v2"
+)
+
+// ReportHash digests everything a volume server reports about a volume, so the
+// two ends of a heartbeat can agree on whether the master's copy is current
+// without shipping the volume list.
+//
+// It must cover every field of VolumeInformationMessage: a change the hash
+// misses is a change the master would never be told about. Volume servers hash
+// the message they are about to send, masters hash what they already hold, and
+// the two match only when the master is up to date.
+func (vi VolumeInfo) ReportHash() uint64 {
+	var buf [57]byte
+	binary.LittleEndian.PutUint32(buf[0:], uint32(vi.Id))
+	binary.LittleEndian.PutUint64(buf[4:], vi.Size)
+	binary.LittleEndian.PutUint64(buf[12:], uint64(vi.FileCount))
+	binary.LittleEndian.PutUint64(buf[20:], uint64(vi.DeleteCount))
+	binary.LittleEndian.PutUint64(buf[28:], vi.DeletedByteCount)
+	binary.LittleEndian.PutUint32(buf[36:], uint32(vi.ReplicaPlacement.Byte()))
+	binary.LittleEndian.PutUint32(buf[40:], uint32(vi.Version))
+	binary.LittleEndian.PutUint32(buf[44:], vi.Ttl.ToUint32())
+	binary.LittleEndian.PutUint32(buf[48:], vi.CompactRevision)
+	binary.LittleEndian.PutUint32(buf[52:], vi.DiskId)
+	if vi.ReadOnly {
+		buf[56] = 1
+	}
+	h := xxhash.Sum64(buf[:])
+
+	var modified [8]byte
+	binary.LittleEndian.PutUint64(modified[:], uint64(vi.ModifiedAtSecond))
+	h = foldReportHash(h, xxhash.Sum64(modified[:]))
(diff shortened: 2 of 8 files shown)


chrislusf:
Both confirmed by reproduction, both fixed. Thanks — P1 in particular would have been a silent data-availability regression, since the self-heal it disables is the only thing that repairs that state.

## P1: lookup-index divergence

Confirmed. Driving `SetVolumeUnavailable` into the state from `TestSyncDataNodeRegistrationReRegistersMissingVolume` left the digest bit-identical while `Lookup` returned nothing, so under delta heartbeats no resend would be requested and the self-heal in `SyncDataNodeRegistration` would never run again — permanently, because the affected volumes are exactly the read-only ones that never appear in `changed_volumes`.

The root problem is that the reported digest answers the wrong question for this. It answers "does the master hold what the server holds", and the server's report is identical either way — it cannot see this and never will. So the master has to notice on its own.

`8d5c26f` adds a second digest over volume ids on both sides of the split: `Disk.volumeIdDigest` for what the node holds, `DataNode.lookupDigest` for what the volume layouts will serve from it, maintained at the four places that change a node's lookup membership (`RegisterVolume`, `UnRegisterVolume`, `SetVolumeUnavailable`, `SetVolumeAvailable`). `HasConsistentVolumeIndex` compares them in O(1).

That is strictly better than the status quo: today the self-heal rescans every volume on every full heartbeat to find a divergence that is almost never there. Now the scan is gated on a digest mismatch. `TestVolumeIndexDigestSeesLookupDivergence` pins that the reported digest does *not* move while the index digests do.

## P2: duplicate volume ids

Confirmed too, and it is worse than a missed detection — it is unreconcilable, so a naive implementation would resend the full list forever.

I did not key the master by (volume id, disk id). Nesting `Disk.volumes` the way `ecShards` is nested would cost more memory per replica than this whole series has saved, which defeats the point of #10603.

`b1ee532` instead detects it and excludes the node: deduplicating the ids in `UpdateVolumes` already computes the count, so `len(actualVolumeIds) < len(actualVolumes)` is free. A marked node keeps sending full lists — today's behaviour, no resend loop — and clears the mark once the twin is unmounted. `TestVolumeDigestRefusesDuplicateVolumeIds` asserts both, and asserts the master genuinely cannot represent both copies, so it fails if that ever changes and the guard becomes unnecessary.

That the master mismodels duplicate volume ids at all is pre-existing and independent of this: `volumeCount` counts the pair once and the surviving `VolumeInfo` is whichever arrived last. Worth its own fix.

## Also

The protocol PR will carry a general version of the P2 lesson: if a full resend does not reconcile a node, stop asking and stay on full lists for that connection. Any divergence neither of us has thought of then degrades to today's behaviour instead of looping.

chrislusf:
Confirmed and fixed in `844af13`. You're right that fixing only `Set` is insufficient — both directions were broken, and the `Set` side was the dangerous one.

Reproduced first. `rack.go:96` deliberately refuses to merge a new id onto an address an older node still claims (`"don't incorrectly merge them"`), so two `DataNode` objects at one address is a supported state, not a corner case. Registering volume 1 on `n1` then on `n2` at the same address gave:

```
lookup resolves to 1 node(s); old servable=false fresh servable=true
  old node holds volume 1, cannot serve it, and reports a consistent index
  fresh node serves volume 1 but reports an inconsistent index
```

The displaced node reporting *consistent* is the bad one: that is precisely the silent unavailability the digest is meant to catch, reintroduced by the digest itself. The other direction is only a false alarm that would trigger a needless self-heal.

`Set` and `Remove` now return the node they displaced and removed, and `moveLookupOwnership(vid, from, to)` transfers the bit, with nil for an entry being created or dropped and a no-op when they are the same node. Both call sites of each are updated; `volumesBinaryState` uses the same list type for the readonly and oversized indexes but tracks no digest, so it is unaffected.

Two regression tests, one per direction: `TestVolumeIndexDigestFollowsDisplacedLookupEntry` (`Set` replaces) and `TestVolumeIndexDigestFollowsRemovedLookupEntry` (`Remove` matches a node that is not the one passed in). Both assert against `topo.Lookup`, so they check real servability rather than just digest equality.

One nearby hazard I did not touch: `VolumeLocationList.Refresh` rebuilds the list dropping stale nodes with no notification, which would silently strand digest bits the same way. It has no callers today. Say the word if you'd rather it were deleted than left as a trap.

PR #10628, opened 2026-08-07
topology: follow a volume that moved between a server's disks

chrislusf:
Pre-existing, found while checking whether the heartbeat digest in #10627 notices a volume moving between disks of one server. It notices a move within a disk type. It cannot notice a move across disk types, because the master never learns about it.

## What happens

`UpdateVolumes` asked only whether a volume id appeared *anywhere* in the report, so a volume that moved from an hdd disk to an ssd one was still "reported" and stayed in the hdd map, while the ssd map gained it as new. On `origin/master`:

```
copies=2 storedDiskType=""
```

The master then holds two copies of that volume forever. The volume count is overstated by one, and `GetVolumesById` returns whichever disk `dn.children` iterates first — so a lookup can hand back the disk the volume has already left.

Moves within one disk type were always handled: the volume stays in the same map and the disk id is updated in place.

## The fix

Track which disk types the heartbeat named each volume on, and treat a volume named on another disk as absent from this one. Disk types are interned to an index rather than stored per volume: a server reports a handful of them across hundreds of thousands of volumes, so a string header each would cost more than the map holding them. `BenchmarkSyncDataNodeRegistration/100000Volumes` moves from `23142575 B/op` to `23142640 B/op` — the interned type list and nothing else.

A volume named on two disks *at the same time* is a stale twin rather than a move — `store_duplicate_vid_test.go` covers the store handling that — and is still kept on both. Dropping one would tell the master a replica had vanished and could trigger re-replication of a volume that is fine.

That sharpens the duplicate marking added in #10619: a volume on two disk *types* is representable, since the two copies land in different maps, so it no longer disables digest comparison for the node. Only a volume named twice on one disk type is unrepresentable, and that is now what marks it.

## Why it matters beyond the count

Once heartbeats carry a digest, this is unreconcilable: the server digests one copy, the master digests two, and no amount of resending the full list closes the gap. The "already sent the list, do not ask again" guard in #10627 keeps that from looping, but the node would sit in full-list mode permanently — quietly losing exactly the saving the digest is for.

## Tests

`TestVolumeMovedBetweenDisks` covers a move within a type, across types, and back; `TestVolumeReportedOnTwoDiskTypesIsKept` and `TestVolumeReportedTwiceOnOneDiskTypeIsFlagged` separate the twin from the unrepresentable case; `TestVolumeDigestSurvivesADiskMove` checks the master ends up digesting what a server holding only the moved volume would report.

Refs #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume tracking when volumes move between disks or change disk types.
  * Prevented moved volumes from being incorrectly duplicated, removed, or reported as unavailable.
  * Improved handling of duplicate volume reports and volume indexing.
  * Preserved volume digest consistency during disk moves.
  * Correctly reports volumes removed from a node while ignoring disk-to-disk moves on the same node.

* **Tests**
  * Added coverage for volume moves, remounts, disk-type changes, duplicate reports, and removal notifications.

### DIFF
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -28,12 +28,9 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/topology"
 )
 
-// shouldBroadcastVolumeRemoval reports whether clients should be told the
-// volume has left this node. A volume moved between the node's own disks is
-// removed from one and added to the other, so it appears in both lists of the
-// same heartbeat, and clients apply additions before deletions -- passing the
-// removal on would drop a location that is still good. Mirrors the guard the ec
-// shard paths already apply.
+// A volume moved between the node's disks appears in both lists, and clients
+// apply additions before deletions, so passing the removal on would drop a
+// location that is still good.
 func shouldBroadcastVolumeRemoval(dn *topology.DataNode, vid needle.VolumeId) bool {
 	return !dn.HasVolumesById(vid)
 }
@@ -217,8 +214,7 @@ func (ms *MasterServer) SendHeartbeat(stream master_pb.Seaweed_SendHeartbeatServ
 			stats.MasterReceivedHeartbeatCounter.WithLabelValues("deletedVolumes").Inc()
 		}
 		if len(heartbeat.NewVolumes) > 0 || len(heartbeat.DeletedVolumes) > 0 {
-			// update master internal volume layouts first, so the removals below
-			// are judged against where the volumes ended up
+			// first, so the removals below see where the volumes ended up
 			ms.Topo.IncrementalSyncDataNodeRegistration(heartbeat.NewVolumes, heartbeat.DeletedVolumes, dn)
 
 			// process delta volume ids if exists for fast volume id updates
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -30,9 +30,12 @@ import (
 
 // A volume moved between the node's disks appears in both lists, and clients
 // apply additions before deletions, so passing the removal on would drop a
-// location that is still good.
+// location that is still good. Not HasVolumesById, which answers for ec shards
+// too: clients hold those separately and prefer the normal location, so a
+// replica that became ec shards has to be reported gone.
 func shouldBroadcastVolumeRemoval(dn *topology.DataNode, vid needle.VolumeId) bool {
-	return !dn.HasVolumesById(vid)
+	_, err := dn.GetVolumesById(vid)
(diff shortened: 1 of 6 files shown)


chrislusf:
Good catch, and confirmed — `5ba523e`.

`masterclient.go:366-373` applies `NewVids` and then `DeletedVids`, so a volume id in both lists ends up deleted on the client. Once the master follows a cross-disk move it produces exactly that pair, and clients would drop a location for a volume that never left the server.

The codebase already had the right answer for this, on the ec side: don't tell clients to delete a location the node still has.

```go
for _, s := range heartbeat.DeletedEcShards {
    if dn.HasEcShards(needle.VolumeId(s.Id)) {
        continue
    }
```

So the same guard now applies to volumes, via `shouldBroadcastVolumeRemoval`.

One thing that needed care beyond the guard itself: the delta branch built the message *before* calling `IncrementalSyncDataNodeRegistration`, so the guard would have been judged against pre-heartbeat state and would have swallowed a genuine unmount — a worse bug than the one being fixed. The topology update now runs first. `TestVolumeUnmountedViaDeltaIsBroadcastAsRemoved` exists specifically to fail if anyone moves it back.

Four tests, covering both directions on both paths: a cross-disk move on a full heartbeat and on a delta is not broadcast as a removal; a volume that really left, by either route, still is.

Worth noting this was reachable before this PR too, just by a different route: a volume server that unmounts and remounts a volume within one delta batch puts the id in both lists, and clients drop the location the same way. The guard covers that as well.

chrislusf:
Confirmed and fixed in `daef05c`. This was the more serious of the two problems on this branch.

`Disk.HasVolumesById` answers for ec shards as well as normal volumes (`disk_ec.go:98-102`), so once a volume was encoded the guard saw the shards, decided the node still had it, and swallowed the removal of the normal replica.

That is not a cosmetic staleness. Clients hold the two location classes separately and, on a tie in generation, `vid_map.go` prefers the normal one:

> A tie means one generation reported both, where the regular copies serve.

So the stale normal location would go on shadowing the shards, and reads would keep going to a server that no longer has the volume. `ec.encode` is routine, so this was readily reachable.

The guard now asks `dn.GetVolumesById`, which looks only at `d.volumes`. `TestVolumeReplacedByEcShardsIsBroadcastAsRemoved` fails against the previous version.

Worth noting the existing ec-side guard is cross-class on purpose and correct as it stands — it suppresses an ec removal while a normal volume of that id exists, which is the decode direction. The two guards look symmetrical and are not, so the reason is now written down at the call site.

PR #10629, opened 2026-08-07
mount: keep directory listings that have been walked through

chrislusf:
Every enumeration re-read the whole directory out of the local store and decoded
it again, so reopening a folder cost exactly what opening it did. That is most of
what a readdir spends. This is the item left open at the end of #10616.

A walk that covers a directory from its first child to its last now leaves the
decoded entries behind, and the next walk pages through them in memory.

```
                    time      bytes       allocations
    first walk    198.1ms   190.3 MiB   2.05M
    reopened       76.2ms    21.4 MiB   1.00M
                   -62%       -89%       -51%
```

200k-entry directory, medians of 8 samples in one run. What remains on the warm
path is the readdir's own per-entry work — building each child's path, resolving
its inode, filling the attribute block — not the listing.

**Only a complete walk is cached.** Nothing is published from a walk that stopped
early or started partway in, so `ls | head` on a huge directory does not pay for
the rest of it. The build is carried across pages and only published when the
store runs out without the caller having stopped it — the distinction that
matters, because a caller whose buffer filled has also produced a short page.

**Invalidation is not a timer.** The cache holds only what the meta cache holds,
so it is void whenever that is. Every write to the local store now goes through
one of four wrappers, each dropping the directory it touched under the same lock
that made the write; reaching past them for a write would leave a listing stale,
and there is a comment saying so. A rebuild drops the listing at begin and at
complete. A walk publishes only if its build is still the live one for that
directory, so a write landing halfway through cannot leave a stale listing
behind. A subscription gap drops everything, since the mount cannot know what it
missed.

**Expiry is the one thing that changes with no write at all**, so it is applied
when serving rather than invalidated for. A cached listing must not go stale
just because a child aged out while it sat there, and a TTL-driven invalidation
would be guesswork. Serving also has to step over expired children the way the
store does — counting them against the limit and advancing the cursor past them
— or pagination diverges between a hit and a miss. That is the same trap as the
truncation bug in #10624.

**Only chunkless listings are cached or served.** Those are what the readdir path
decodes since #10616, so an entry from the cache can never reach a caller that
wanted chunks.

`-dirListCacheEntries` sizes it, default 200k children (tens of megabytes),
0 disables. Least recently used directories are dropped to stay inside it.

Tests cover the hit, invalidation for each mutation kind (insert, update, delete,
delete-children, rename-in, purge, distrust-all), a partial walk, a mid-directory
listing, a write landing mid-walk, expiry applied on serve without any
invalidation, eviction, and the disabled case.

## Summary by CodeRabbit

- **New Features**
  - Added an in-memory directory listing cache to improve browsing performance for mounted filesystems.
  - Added the `-dirListCacheEntries` mount option to configure cache capacity. It defaults to 1,000,000 entries and can be set to zero to disable caching.
  - Cached listings support pagination, expiration filtering, and least-recently-used eviction.

- **Bug Fixes**
  - Directory listing caches are invalidated after filesystem changes to keep results accurate.

### DIFF
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -117,7 +117,7 @@ func init() {
 	mountOptions.cacheDirForWrite = cmdMount.Flag.String("cacheDirWrite", "", "buffer writes mostly for large files")
 	mountOptions.writeBufferSizeMB = cmdMount.Flag.Int64("writeBufferSizeMB", 0, "global cap on the per-mount write buffer (memory + swap) in MB, 0 means unlimited. Bounds /tmp growth when volume uploads stall")
 	mountOptions.cacheMetaTtlSec = cmdMount.Flag.Int("cacheMetaTtlSec", 60, "metadata cache validity seconds")
-	mountOptions.dirListCacheEntries = cmdMount.Flag.Int("dirListCacheEntries", meta_cache.DefaultListingCacheEntries, "directory children kept decoded in memory, so reopening a folder rereads nothing; 0 disables")
+	mountOptions.dirListCacheEntries = cmdMount.Flag.Int("dirListCacheEntries", meta_cache.DefaultListingCacheEntries, "directory children kept decoded in memory, so reopening a folder rereads nothing. A ceiling, not a reservation: only walked directories are held, at roughly half a kilobyte per child. 0 disables")
 	mountOptions.dataCenter = cmdMount.Flag.String("dataCenter", "", "prefer to write to the data center")
 	mountOptions.allowOthers = cmdMount.Flag.Bool("allowOthers", true, "allows other users to access the file system")
 	mountOptions.defaultPermissions = cmdMount.Flag.Bool("defaultPermissions", true, "enforce permissions by the operating system")
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -3,6 +3,8 @@ package command
 import (
 	"os"
 	"time"
+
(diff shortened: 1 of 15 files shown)


chrislusf:
Raised `-dirListCacheEntries` to 1,000,000.

Measured what one entry actually retains before picking the number, since the
flag is only useful if an operator can reason about it: a chunkless child holds
about **480 bytes**, so 200k was ~91 MiB and the new default bounds the cache
near **450 MiB**.

The reason a number that large is still reasonable is that it is a ceiling, not
a reservation — nothing is held for a directory that was never walked, so a
mount over ordinary directories never approaches it, and a mount that does walk
a million cached children is exactly the one that wants the speedup. Both the
constant's doc and the flag help now say what reaching it costs, so anyone on a
memory-tight host has the number to lower it against rather than discovering it.

Worth noting for anyone sizing this: a directory larger than the whole budget is
never cached at all, so raising the ceiling is also what lets a single very wide
directory be cached rather than skipped.

chrislusf:
Closing in favor of #10634: the kernel serves warm listings from the directory's page cache at a fraction of the code, with memory the kernel reclaims under pressure, and it also covers the read-through directories of #10631 which a mount-side cache never sees. The branch stays around for what the kernel cannot do — consumers with no kernel above the WFS layer, and serving lookup storms from a decoded listing — if either ever becomes pressing.

PR #10631, opened 2026-08-07
mount: read oversized directories through instead of caching them

chrislusf:
Fixes part of #10020.

Visiting a directory pulls every child from the filer into the mount's local
LevelDB before the first listing returns. For a directory of a few million
entries that is minutes of streaming before `ls` shows anything, gigabytes of
local store, and gigabytes of decoded entries in flight — paid in full by a
mount (a backup walker, say) that will only read the directory once.

A build that crosses `-cacheDirMaxEntries` (default one million, 0 caches
everything) now stops, cleans up the partial build, and marks the directory
read-through — the mode update-hot directories already use. Listings stream
from the filer with pagination and a consistent snapshot; lookups consult the
filer per entry, as for any uncached directory. Metadata events for its
children are already skipped for uncached directories, so it stays out of the
local store.

Three details carry the design:

- **The refusal is remembered.** The next visit fails fast instead of streaming
  a million entries to rediscover it. The memory is per-mount and in-memory, so
  a directory that shrinks is reconsidered at the next mount.
- **An oversized ancestor is stepped over, not fatal.** `EnsureVisited` pulls
  the whole ancestor chain, so without this a huge parent would wedge the
  caching of every directory beneath it forever.
- **Crossing the limit mid-walk carries on seamlessly.** The direct path keeps
  the same pagination state on the directory handle, so a walk that started
  from the cache resumes from the same position against the filer.

For the heap profile in the issue: most of the 10GB retained there is what
#10606/#10616/#10624 already remove — none of which are in v4.41, whose tag
predates them. This PR removes the remaining piece those do not touch: the cost
of pulling a huge directory into the local cache at all.

Tests cover the refusal (nothing cached, nothing left of the partial build,
second visit fails fast), the oversized-ancestor step-over, and that a
directory at the limit still caches normally.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable limit for directory entries cached locally, defaulting to 10,000.
  * Directories exceeding the limit now use direct access while preserving listing and pagination behavior.
  * Set the limit to 0 to disable the restriction.

* **Bug Fixes**
  * Improved browsing of very large directories and nested paths.
  * Prevented incomplete metadata from being retained after oversized directory detection.

* **Tests**
  * Added coverage for oversized directories, fallback behavior, cleanup, and normal caching within the limit.

### DIFF
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -20,6 +20,7 @@ type MountOptions struct {
 	concurrentWriters    *int
 	concurrentReaders    *int
 	cacheMetaTtlSec      *int
+	cacheDirMaxEntries   *int
 	cacheDirForRead      *string
 	cacheDirForWrite     *string
 	cacheSizeMBForRead   *int64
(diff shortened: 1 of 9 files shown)


chrislusf:
Default lowered to 10,000: only genuinely small directories get pulled whole into the local store. A read-through directory still gets local-speed repeat listings once the kernel is allowed to cache them, which is the companion change.

chrislusf:
The oversized-ancestor comment was right, and it was a real bug: on first
discovery, the ancestor's refusal cancelled the errgroup — aborting the
descendant's build — and the caller then marked the *listed* directory
read-through, leaving a perfectly cacheable directory streaming from the filer
until its inode was forgotten. My original test missed it by pre-marking the
ancestor, which only exercises the fast path.

Fixed in c87b608: a refusal for any directory other than the one being listed is
kept out of the group's result (it is already remembered for the next visit), so
the descendant's build completes and caches. Regression test covers the
first-discovery case with a per-directory fake filer, and fails without the fix.

PR #10635, opened 2026-08-08
wdclient: keep the location of a volume reported added and removed at once

chrislusf:
Hardening on the client side of the same problem as #10628, which fixes it on the master side.

A `VolumeLocation` message is about one server, so a volume id appearing in both `NewVids` and `DeletedVids` can only mean it moved between that server's own disks — it is still there. `updateVidMap` applied additions before removals, so the removal won and the client ended up with no location for a volume that never went anywhere.

Reachable today without #10628: a volume server that unmounts and remounts a volume within one delta batch produces exactly that pair.

## Not reordering

Applying removals first would give the right end state, but `addLocation` and `deleteLocation` take the lock separately, so it opens a window where the volume resolves nowhere — trading a wrong answer for an intermittent one, which is harder to diagnose.

Skipping the removal for ids that are also being added has no window and does not depend on the order the two lists happen to be applied in. Same for the ec lists, which have the same shape.

The index is only built when both lists are non-empty, so the full id list a client receives on connect — hundreds of thousands of ids with no removals — costs nothing.

## Why both sides

#10628 stops the master emitting the contradiction, which is the real fix and the only one that helps the clients already deployed. This one means a client is not at the mercy of the master getting that right, including against masters that predate it.

Four tests: a moved volume keeps its location, one that really left loses it, removals alongside a move still apply, and the ec path behaves the same.

Refs #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume tracking during disk moves so relocated volumes retain their server location.
  * Prevented volumes from being incorrectly removed when move and deletion notifications occur together.
  * Applied the same protection to erasure-coded volumes.
  * Ensured genuinely removed volumes are no longer listed.

* **Tests**
  * Added coverage for regular and erasure-coded volume moves, removals, and mixed move/removal updates.

### DIFF
--- a/weed/wdclient/masterclient.go
+++ b/weed/wdclient/masterclient.go
@@ -351,6 +351,22 @@ func (mc *MasterClient) tryConnectToMaster(ctx context.Context, master pb.Server
 	return nextHintedLeader
 }
 
+// addedVids indexes added ids that are also being removed. A volume moved
+// between a server's disks is reported both ways in one message, and the server
+// still has it -- acting on the removal would drop a good location, whichever
+// order the two lists happen to be applied in. Empty unless both lists are
+// non-empty, so the full id list a client gets on connect costs nothing.
+func addedVids(added, removed []uint32) map[uint32]struct{} {
+	if len(added) == 0 || len(removed) == 0 {
+		return nil
+	}
+	index := make(map[uint32]struct{}, len(added))
+	for _, vid := range added {
+		index[vid] = struct{}{}
+	}
+	return index
+}
+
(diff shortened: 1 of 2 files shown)


chrislusf:
The test passes as written — the second `updateVidMap` carried both lists, because it reused the same response object and only added `DeletedEcVids` to it, leaving `NewEcVids` set from the first call.

But that is exactly the reading the review made, and it is the natural one, so the objection stands even though the conclusion does not. A test whose whole point is an overlap should not leave the overlap implicit. Each update is now built explicitly via `moveEcResponse(new, deleted)`, and `TestEcVolumeRemovedLosesItsLocation` is added so the ec path covers both directions like the regular one.

PR #10639, opened 2026-08-08
rust: support the READS scrub mode

chrislusf:

#10621 introduces a `READS` scrub mode, but the Rust volume server hard-rejects unknown modes, so `volume.scrub`/`ec.scrub --mode=reads` would fail with `unsupported scrub mode 5` against Rust volume servers as soon as the shell offers the flag.

Mirror the mode into the Rust volume server. Both scrub handlers accept `READS`; regular volumes treat it as `FULL`, and the EC distributed needle walk gains the same read-recovery fallback as Go's `Store.ScrubEcVolume`: a failed interval read is reconstructed from the remaining shards — exercising parity data — before the shard is flagged broken. Scrub failure messages now carry the underlying read error, and the Rust proto's `VolumeScrubMode` comments match the Go proto's, both mirroring #10621.

Safe in either merge order with #10621: until the shell sends mode 5, this only teaches the server to accept it.

# How is the PR tested?

New unit test: a single-node EC volume scrubs clean under `FULL`; after unmounting one shard, `FULL` flags it broken while `READS` reconstructs the interval from the local survivors and reports clean. Full `cargo test` passes (427 tests).

## Summary by CodeRabbit

* **New Features**
  * Added a `READS` volume scrub mode to check whether degraded erasure-coded volumes remain readable.
  * Supports shard reconstruction when data is unavailable during read checks.
  * Expanded scrub mode documentation and behavior descriptions.

* **Bug Fixes**
  * Improved error reporting for failed shard reads and recovery attempts.
  * Preserved existing verification behavior for full scrub operations.

### DIFF
--- a/seaweed-volume/proto/volume_server.proto
+++ b/seaweed-volume/proto/volume_server.proto
@@ -699,10 +699,22 @@ message FetchAndWriteNeedleResponse {
 
 enum VolumeScrubMode {
     UNKNOWN = 0;
+    // Perform a quick scrub of volume index contents only.
     INDEX = 1;
+    // Performs a full scrub of the volume, reading and verifying every needle contained in it.
     FULL = 2;
+    // Performs a local scrub of volume contents hosted in this server. For regular volumes,
+    // this operation is equivalent to FULL; for EC volumes, only needles fully contained
+    // in local shards are verified.
     LOCAL = 3;
-    CHECKSUM = 4; // EC only: verify each local shard's raw bytes against the bitrot checksum sidecar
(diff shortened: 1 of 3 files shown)


chrislusf:
Superseded by #11006, which carried the Rust mirror in the same PR with refined semantics — READS keeps reporting the unreadable shards as broken and only errors on needles parity can no longer rebuild, and the handler forwards forceDeletedNeedlesCheck instead of hardcoding it off. The unit test from this branch lives on in #11027, adapted to those semantics.

PR #10640, opened 2026-08-08
heartbeat: send only the volumes that changed

chrislusf:
The step where the cost in #10603 actually goes away. Builds on #10619 and #10627.

A volume server has been resending its whole volume list every 5 seconds because that list was the only way the master could notice a volume that vanished without a delta. The digest now gives that guarantee, so the list can stop.

```
2000 volumes on the server
  full list heartbeat:     30004 bytes (2000 volumes)
  steady-state heartbeat:    131 bytes (0 changed)
```

At the 550k-volume scale in #10603 that is 28MB and ~229MB of master allocation per server every 5 seconds becoming a few hundred bytes, plus whatever genuinely moved. The read-only and tiered volumes that make up most of that cluster never appear at all.

## Wire

```proto
message Heartbeat {
  repeated VolumeInformationMessage changed_volumes = 28;
}
message HeartbeatResponse {
  bool volume_digest_supported = 9;
}
```

A partial list cannot travel in `volumes`: a master that did not understand it would read the absences as deletions and drop the lot. So changes get their own field, and are sent only once the master has said it compares digests — a server reaching an older master, or reconnecting to one, is back to full lists straight away with no deployment ordering to get right.

## Two things that would have been very bad

**`has_no_volumes` must come from a full list alone.** It was derived from the outgoing message; in delta mode a heartbeat with nothing to report carries nothing, which would have read as *this server holds no volumes* and dropped every volume on it. It is now false unless a full list is being sent, and `TestQuietHeartbeatDoesNotLookLikeAnEmptyServer` fails against the earlier version.

**The digest covers everything held, not what was sent.** Otherwise a master that correctly applied the changes could never confirm it was current.

## Rust

Mirrored, with a hazard the Go side does not have: mount and unmount deltas there are derived by diffing successive heartbeats against the last `volumes` list, so a partial list would have reported every volume it left out as unmounted. Collecting now returns the full set alongside the message, and all four sites that diff use that rather than what went on the wire.

## Reporting state

Per connection, so a reconnect or a different master starts from the full list. Keyed by (disk, volume) rather than volume id, since an id mounted on two disks would otherwise leave one copy's changes untold. A volume that goes away is forgotten, so one that comes back is reported again.

Its zero value means nothing has been told to anyone, which is the state that sends the whole list — a `Store` built as a literal, which tests do, would otherwise have panicked on the first heartbeat.

Go: `weed/storage`, `weed/topology`, `weed/server/...` pass. Rust: 400 tests pass.

Refs #10603

## Summary by CodeRabbit

* **New Features**
  * Heartbeats can report only volumes changed since the previous update when supported.
  * Masters can request a complete volume list when needed.
  * Incremental updates register previously unknown volumes and keep them available for service.

* **Bug Fixes**
  * Improved heartbeat handling across reconnects, remounted volumes, quiet updates, and concurrent reporting.
  * Preserved accurate volume digests and correct empty-volume reporting.
  * Ensured full volume lists are resent when requested during heartbeat processing.

### DIFF
--- a/seaweed-volume/proto/master.proto
+++ b/seaweed-volume/proto/master.proto
@@ -111,6 +111,10 @@ message Heartbeat {
   // from servers that do not compute it, and distinct from a digest of 0, which
   // is what a server holding no volumes reports.
   optional uint64 volume_digest = 27;
+  // Volumes whose reported state changed since the last heartbeat, sent in
+  // place of `volumes`. A master that does not understand this never sets
+  // volume_digest_supported, so it keeps being sent the whole list.
+  repeated VolumeInformationMessage changed_volumes = 28;
 }
 
 message HeartbeatResponse {
@@ -124,6 +128,9 @@ message HeartbeatResponse {
   // The master's view of this server's volumes disagrees with the reported
   // digest, so it needs the full volume list rather than changes alone.
   bool resend_full_volume_list = 8;
+  // The master compares volume digests, so a server that reports one may send
+  // changed_volumes in place of its whole list.
+  bool volume_digest_supported = 9;
 }
 
 message VolumeInformationMessage {
--- a/seaweed-volume/src/server/heartbeat.rs
+++ b/seaweed-volume/src/server/heartbeat.rs
@@ -529,7 +529,7 @@ async fn do_heartbeat(
                     info!("Heartbeat stopping");
                     return Ok(None);
                 }
-                let (current_hb, held_volumes) = collect_heartbeat_with_snapshot(config, state);
+                let (_current_hb, held_volumes) = collect_heartbeat_with_snapshot(config, state);
                 let current_volumes: HashMap<u32, _> = held_volumes.iter().map(|v| (v.id, v.clone())).collect();
                 let current_ec_shards = {
                     let store = state.store.read().unwrap();
@@ -756,7 +756,19 @@ fn collect_heartbeat_with_snapshot(
         &mut store,
         deleted_ec_shards,
         ec_shards.is_empty(),
+        true,
     )
+}
+
+/// Lists the volumes the server holds without touching reporting state or
+/// expiring anything, for callers that only need to diff against a previous
+/// snapshot and send a message of their own.
+fn collect_volume_snapshot(
+    config: &HeartbeatConfig,
+    state: &Arc<VolumeServerState>,
+) -> Vec<master_pb::VolumeInformationMessage> {
+    let mut store = state.store.write().unwrap();
+    build_heartbeat_with_ec_status(config, &mut store, Vec::new(), true, false).1
 }
 
 fn collect_heartbeat(
@@ -798,7 +810,7 @@ fn collect_location_metadata(
 #[cfg(test)]
 fn build_heartbeat(config: &HeartbeatConfig, store: &mut Store) -> master_pb::Heartbeat {
     let has_no_ec_shards = collect_live_ec_shards(store, false).is_empty();
-    build_heartbeat_with_ec_status(config, store, Vec::new(), has_no_ec_shards).0
+    build_heartbeat_with_ec_status(config, store, Vec::new(), has_no_ec_shards, true).0
 }
 
 /// Returns the heartbeat to send and, separately, every volume held. The
@@ -809,6 +821,7 @@ fn build_heartbeat_with_ec_status(
     store: &mut Store,
     deleted_ec_shards: Vec<master_pb::VolumeEcShardInformationMessage>,
     has_no_ec_shards: bool,
+    commit_report: bool,
 ) -> (master_pb::Heartbeat, Vec<master_pb::VolumeInformationMessage>) {
     const MAX_TTL_VOLUME_REMOVAL_DELAY: u32 = 10;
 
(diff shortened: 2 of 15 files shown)


chrislusf:
Both valid, both fixed. The second was worse than reported.

## Full-list request lost to the heartbeat it raced

`f549c01`. `begin` read the flag, the caller collected, and `commit` cleared it unconditionally — so a request arriving in that window was marked satisfied by a heartbeat that had already been built as a delta, and the master waited for another digest mismatch before asking again.

Now a counter, incremented by every request and by reset, is returned from `begin` and passed to `commit`, which clears the flag only if nothing has asked since. Same guard in both implementations, one test each.

## Committing a report the branch throws away

`c5c081a`. Correct, and the branch is worse than the comment describes: `current_hb` there is not merely unused for its volumes, it is **never sent at all**. That path builds its own `delta_hb` from the diff and sends that. So once collecting recorded what the master had been told, every mount or unmount marked the changed volumes as sent while nothing carrying them ever went out — the master learned of them only after a digest mismatch, and only then if one happened.

Snapshotting is now a separate thing that records nothing. It also no longer calls `delete_expired_ec_volumes`, which that path was already invoking and discarding the result of — since that call performs the deletion, the shards were being removed and the master never told. Pre-existing, but not worth leaving behind now that the path is being touched.

`test_snapshot_does_not_mark_volumes_reported` fails against the previous version.

The equivalent does not arise in Go: every `CollectHeartbeat` there is followed by a send.

Go `weed/storage`, `weed/topology`, `weed/server/...` pass; Rust 402.

chrislusf:
All three valid. The first two are the same mistake: I built the machinery to detect these in #10619 and then never asked it anything.

## 1. Delta heartbeats removed the lookup-index self-heal

`8fd1f60`. Correct, and it is the exact failure #10619 added `HasConsistentVolumeIndex` for — which I then left with no caller. The self-heal only runs on a full report; delta mode stops those; the volume server cannot see the divergence and its digest cannot show it. The affected volumes are read-only ones that never appear in `changed_volumes`, so nothing would ever have disturbed it.

`checkVolumeDigest` now asks whether the master's own two indexes agree, and requests the list when they do not. `TestDigestCheckAsksForTheListWhenTheLookupIndexDrifts` asserts the reported digest is *unchanged* across the divergence first, since that is the whole reason the index has to be checked separately.

## 2. Duplicate-id nodes went to deltas with an unverifiable digest

Same commit. My earlier reasoning — "asking would loop forever" — was right when the server always sent its whole list anyway, and wrong the moment it could stop. Such a node now stays on full lists, which is exactly the degradation intended: no digest of it can be verified, so nothing else would tell the master what it had stopped holding.

It does not loop: a heartbeat that already carried a full list is not asked for another, so the node settles into today's behaviour rather than ping-ponging. `TestDigestCheckSkipsNodesWithDuplicateVolumeIds` asserted the old contract and is replaced by `TestDigestCheckKeepsDuplicateNodesOnFullLists`, which checks both halves.

## 3. Every changed volume broadcast as a new location

`6f46f5e`. `ApplyVolumeChanges` now returns only the volumes the node did not already have, and only those reach `NewVids`. Growth is the common change and moves no location, so on a busy cluster the previous version told every connected filer about volumes it could already reach — bounded queues, so the cost is dropped real updates rather than just traffic.

Go `weed/storage`, `weed/topology`, `weed/server/...` pass under `-race`; Rust 402.

chrislusf:
Both valid, both fixed.

## 1. Repaired lookup entries were not announced

`fdc2a1a`. The repair restored servability on the master and stopped there. `newVolumes` came only from the disk map, which in the drift case already held the volume, so nothing reached `NewVids` — while the clients that heard the node drop out had already deleted the location.

Worse, the fix in `8fd1f60` made this the *only* outcome: the index check now sees the disagreement, requests a full list, the self-heal repairs the index, the check passes, and no further report is ever triggered. Master healthy, clients permanently stale. So the two changes had to land together.

Reaching the lookup index is what makes a volume servable, so recovering an entry there is an arrival as far as clients are concerned. Both `ApplyVolumeChanges` and `SyncDataNodeRegistration` now return it. No deduplication needed on the full path: volumes new to the disk map are registered into the layout earlier in the same call, so reaching the self-heal branch means only the index had lost it.

`TestRepairedLookupEntryIsAnnounced` runs both paths, and asserts the volume is unservable first so it is testing the repair rather than a plain registration.

## 2. Resend responses turned off preallocation

`a7b08fd`. Confirmed, and mine. `volume_size_limit` is guarded by `!= 0` on both implementations; `preallocate` is a bare bool applied unconditionally, so a response that omitted it read as the master switching it off until reconnect. A digest mismatch made that recurrent rather than a one-off.

Responses sent mid-stream now start from `heartbeatResponse()`, which carries the configured options, instead of being built field by field — the shape that caused this. No volume-server change, so nothing to mirror.

One instance of the same class predates this and is untouched: the not-the-leader response at `master_grpc_server.go:138` is also built bare. Its window is a single response before the server reconnects to the real leader, so it is a much smaller problem, but it is the same trap and worth a follow-up.

Go `weed/topology` and `weed/server` pass under `-race`.

PR #10644, opened 2026-08-08
volume: validate the file extension in CopyFile and ReceiveFile

chrislusf:
`CopyFile` and `ReceiveFile` are ungated cluster-internal RPCs (a peer volume server pulls/pushes whole files for replica sync and EC shard distribution). Both turn the client-supplied `Ext` and `Collection` into an on-disk path with no validation.

The EC-shard branches build the path via `util.Join` / `path.Join`, which path-clean a `..` away before the `os.Create`/`os.Open`:

- `Ext` like `/../../x` resolves outside the volume directory.
- `Collection` is folded in as `<collection>_<vid>`, so `../../x` escapes the same way.

A `ReceiveFile` with `IsEcVolume` set can then write outside the data directory, and `CopyFile` can read from outside it. The regular-volume branches concatenate raw and usually fail on the non-existent intermediate, but they build the path the same unchecked way.

Fix:
- `Ext` is constrained to a real suffix -- a leading dot followed by alphanumerics (`.dat`, `.idx`, `.vif`, `.ecx`, `.ecj`, `.ecsum`, `.ec00`..) -- checked by character class rather than an enumerated list so custom EC ratios keep working.
- `Collection` must be a single path element: separators and bare parent references are rejected, while ordinary names (including `.` and `-`) still pass.

Both are checked at the top of the two RPCs before any path is built.

Tests drive the real `ReceiveFile` EC path and `CopyFile` with a traversal `Ext` and `Collection` (rejected, nothing escapes), plus positive controls that a normal `.ec00` shard and a `my.bucket-1` collection still land inside the volume directory.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation for file extensions and collection names during file copy and receive operations.
  * Blocked path traversal attempts and malformed values before files are accessed or created.
  * Added structured error responses for invalid requests.

* **Tests**
  * Added coverage confirming traversal attempts are rejected and valid extensions continue to work safely.

### DIFF
--- a/test/volume_server/grpc/copy_receive_variants_test.go
+++ b/test/volume_server/grpc/copy_receive_variants_test.go
@@ -90,7 +90,7 @@ func TestCopyFileIgnoreNotFoundAndStopOffsetZeroPaths(t *testing.T) {
 
 	missingNoIgnore, err := grpcClient.CopyFile(ctx, &volume_server_pb.CopyFileRequest{
 		VolumeId:                 volumeID,
-		Ext:                      ".definitely-missing",
+		Ext:                      ".missing",
 		CompactionRevision:       math.MaxUint32,
 		StopOffset:               1,
 		IgnoreSourceFileNotFound: false,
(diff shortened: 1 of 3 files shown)


chrislusf:
Thanks for the reviews. Addressed both:

- Greptile: the `.definitely-missing` fixtures in `copy_receive_variants_test.go` tripped the new extension check on the hyphen; switched them to `.missing` (071be45).
- CodeRabbit: `Collection` was the same class of hole via `<collection>_<vid>`; added `checkVolumeCollection` at both RPC entry points and traversal coverage for `../../x` (29fb471).

PR #10648, opened 2026-08-08
filer: rebuild peer metadata subscriptions after a master reconnect

### LINKED ISSUE #10643: [filer] Peer metadata subscription is not restored after missing ClusterNodeUpdate ADD during Master reconnect
# Title

[filer] Peer metadata subscription is not restored after missing ClusterNodeUpdate ADD during Master reconnect

## Describe the bug

In a 3-Filer cluster using an embedded metadata database, a Filer metadata subscription can be permanently lost after the following sequence:

1. A peer Filer goes offline.
2. Other Filers receive the peer `REMOVE` event and stop/remove the corresponding metadata subscription.
3. The remaining Filers temporarily lose their Master `KeepConnected` stream.
4. The peer Filer comes back and the Master generates an `ADD` event while those Filers are disconnected.
5. The disconnected Filers reconnect to the Master, but the missing peer metadata subscription is not rebuilt.

After this happens, `cluster.ps` still shows all 3 Filers as online, but metadata replication from the recovered Filer stops permanently on the affected Filers.

Restarting the affected Filer immediately restores the subscription and metadata synchronization.

I originally observed this on SeaweedFS 4.33. I upgraded to SeaweedFS 4.41 and can still reproduce the same issue.

Possibly related issues:

* #6569 - `peer filer stops subscription and never restarted`
* #6618 - `3 filers, one seems to be not synced anymore`

---

## System Setup

SeaweedFS version:

```text
4.41
```

Docker image:

```text
docker.io/chrislusf/seaweedfs:4.41
```

Cluster topology:

```text
3 Masters
3 Filers

node1: 192.168.30.250
node2: 192.168.30.251
node3: 192.168.30.252
```

Master endpoints:

```text
192.168.30.250:27333
192.168.30.251:27333
192.168.30.252:27333
```

Filer HTTP endpoints:

```text
192.168.30.250:26888
192.168.30.251:26888
192.168.30.252:26888
```

Filer gRPC port:

```text
36888
```

Master configuration used by the Filers:

```text
-master=192.168.30.250:27333,192.168.30.251:27333,192.168.30.252:27333
```

Filer metadata storage:

```text
Embedded database
```

Each Filer uses its own embedded metadata database. No external PostgreSQL, MySQL, or other remote metadata database is used.

There is also a service VIP:

```text
192.168.30.240
```

but the reproduction below bypasses the VIP and accesses each Filer directly.

---

## Steps to reproduce

The following procedure reproduces the issue reliably by intentionally creating a window where a Filer `ADD` event can be missed.

### 1. Verify the healthy state

Run:

```bash
printf 'cluster.ps\n' | docker exec -i seaweed-filer /usr/bin/weed shell -master=192.168.30.250:27333,192.168.30.251:27333,192.168.30.252:27333
```

All three Filers should be visible.

Create a baseline file through node1:

```bash
echo "baseline-$(date +%s)" >/tmp/baseline.txt
curl -fsS -F file=@/tmp/baseline.txt http://192.168.30.250:26888/test-sync/baseline.txt
```

Verify that it is readable from all three Filers:

```bash
for ip in 192.168.30.250 192.168.30.251 192.168.30.252; do echo -n "$ip: "; curl -s -o /dev/null -w '%{http_code}\n' http://$ip:26888/test-sync/baseline.txt; done
```

Expected:

```text
192.168.30.250: 200
192.168.30.251: 200
192.168.30.252: 200
```

---

### 2. Stop node1 Filer

On node1:

```bash
docker stop seaweed-filer
```

On node2 and node3, wait until the node1 peer removal has been processed:

```bash
docker logs --since 2m seaweed-filer 2>&1 | grep -E '192\.168\.30\.250|stop subscribing'
```

The log contains messages similar to:

```text
noticed .filer 192.168.30.250
stop subscribing peer 192.168.30.250 meta change
```

At this point, the node1 metadata subscription has been stopped on node2/node3.

---

### 3. Pause node2 and node3 Filers

On node2:

```bash
docker pause seaweed-filer
```

On node3:

```bash
docker pause seaweed-filer
```

`pause` is intentional here.

Do not restart the Filers, because restarting a Filer re-discovers the existing peers and therefore repairs the missing subscription.

---

### 4. Break the existing Master KeepConnected stream

Check the current Master state and leader:

```bash
for ip in 192.168.30.250 192.168.30.251 192.168.30.252; do echo "===== $ip ====="; curl -s http://$ip:27333/cluster/status; echo; done
```

While node2/node3 Filers are still paused, restart the Master that currently owns the relevant Master connection / current leader.

For example:

```bash
docker restart seaweed-master
```

The purpose is to invalidate the existing Master `KeepConnected` stream while node2/node3 Filers are paused, so they cannot immediately reconnect.

Wait until the Master cluster has elected a leader again:

```bash
for ip in 192.168.30.250 192.168.30.251 192.168.30.252; do echo "===== $ip ====="; curl -s http://$ip:27333/cluster/status; echo; done
```

---

### 5. Start node1 Filer while node2/node3 are still paused

On node1:

```bash
docker start seaweed-filer
```

Wait until node1 Filer has registered with the Master again.

For example:

```bash
docker logs --since 2m seaweed-master 2>&1 | grep -E '\.filer|LockRing'
```

The important sequence is:

```text
node2/node3 receive node1 REMOVE
              ↓
node2/node3 stop node1 metadata subscription
              ↓
node2/node3 are paused
              ↓
their existing Master stream is invalidated
              ↓
node1 Filer comes back
              ↓
Master processes node1 ADD
              ↓
node2/node3 cannot receive this ADD event
```

---

### 6. Resume node2/node3 Filers without restarting them

On node2:

```bash
docker unpause seaweed-filer
```

On node3:

```bash
docker unpause seaweed-filer
```

Wait about 10-20 seconds for the Filers to reconnect to the Master.

Run:

```bash
printf 'cluster.ps\n' | docker exec -i seaweed-filer /usr/bin/weed shell -master=192.168.30.250:27333,192.168.30.251:27333,192.168.30.252:27333
```

All three Filers are visible again:

```text
filers 3

192.168.30.250
192.168.30.251
192.168.30.252
```

However, the metadata sync time from node1 on node2/node3 no longer advances.

On node2/node3, there is also no active metadata subscription connection to node1:

```bash
ss -nt | grep '192.168.30.250:36888'
```

and there is no new subscription log such as:

```text
loopSubscribeToOneFiler read 192.168.30.250
```

---

### 7. Verify that metadata replication is broken

Create a new file through node1:

```bash
echo "repro-$(date '+%F %T')" >/tmp/repro.txt
curl -fsS -F file=@/tmp/repro.txt http://192.168.30.250:26888/test-sync/repro.txt
```

Check the file directly through each Filer:

```bash
for ip in 192.168.30.250 192.168.30.251 192.168.30.252; do echo -n "$ip: "; curl -s -o /dev/null -w '%{http_code}\n' http://$ip:26888/test-sync/repro.txt; done
```

Actual result:

```text
192.168.30.250: 200
192.168.30.251: 404
192.168.30.252: 404
```

The cluster membership itself looks healthy, but metadata written through node1 is no longer replicated to node2/node3.

The problem persists indefinitely rather than recovering automatically.

---

### 8. Restarting the affected Filers repairs the problem

Restart the affected Filer on node2/node3:

```bash
docker restart seaweed-filer
```

After restart, the missing peer subscription is recreated.

Logs again contain something similar to:

```text
loopSubscribeToOneFiler read 192.168.30.250
subscribing remote 192.168.30.250 meta change
```

The metadata sync time starts advancing again, and the test file becomes available through node2/node3.

This indicates that the failure is related to missing peer-subscription state rather than volume/data availability.

---

## Actual behavior

After reconnecting to the Master:

```text
Master cluster membership:
    node1 = online
    node2 = online
    node3 = online

cluster.ps:
    filers 3
```

but on node2/node3:

```text
metadata subscription to node1 = missing
```

As a result:

```text
metadata sync time from node1 stops advancing
new metadata from node1 is not replicated
```

The condition does not recover automatically.

Restarting the affected Filer repairs it.

---

## Expected behavior

After a Filer reconnects to the Master, its active metadata peer subscriptions should eventually match the Master's current Filer membership.

For example, if the current Master membership is:

```text
192.168.30.250
192.168.30.251
192.168.30.252
```

but the local `MetaAggregator` does not have a subscription to:

```text
192.168.30.250
```

the missing subscription should be recreated automatically.

A transient Master connection failure or a missed cluster membership event should not permanently stop Filer metadata replication.

This is especially important when each Filer uses its own embedded metadata database, because peer metadata replication is required to keep the independent embedded metadata stores synchronized.

---

## Suspected root cause

From reading the current Filer metadata aggregation logic, peer subscription lifetime appears to depend on `ClusterNodeUpdate` events.

When a peer `REMOVE` event is processed, the corresponding subscription is stopped and removed:

```text
ClusterNodeUpdate REMOVE
        ↓
close peer stop channel
        ↓
delete peer subscription state
        ↓
loopSubscribeToOneFiler exits
```

Creating the subscription again depends on receiving another peer `ADD` event:

```text
ClusterNodeUpdate ADD
        ↓
create peer subscription state
        ↓
start loopSubscribeToOneFiler
```

`loopSubscribeToOneFiler` can retry an existing subscription after network errors, but once the subscription goroutine has been stopped by a peer `REMOVE`, it needs another `ADD` event to be created again.

The problematic case therefore appears to be:

```text
REMOVE received
    ↓
subscription removed
    ↓
Master stream disconnected
    ↓
peer comes back
    ↓
ADD happens while client is disconnected
    ↓
client reconnects to Master
    ↓
ADD is not replayed
    ↓
missing subscription is never reconciled
```

It appears that after reconnecting the Master `KeepConnected` stream, the Filer does not reconcile the current `MetaAggregator` peer subscriptions against the Master's current Filer membership.

This can leave the cluster membership and subscription state permanently inconsistent.

---

## Possible fix

One possible solution would be to reconcile Filer peer subscriptions after the Master connection is re-established.

For example:

```text
Master KeepConnected reconnects
              ↓
get current Filer membership
              ↓
compare it with current MetaAggregator subscriptions
              ↓
create missing subscriptions
              ↓
remove stale subscriptions
```

A periodic reconciliation could also provide an additional safety mechanism.

The important property would be:

```text
current peer subscriptions
        eventually equal
current Master Filer membership
```

instead of relying exclusively on every individual `ClusterNodeUpdate ADD/REMOVE` event being received.

---

## Additional context

This seems very similar to #6569, where a peer Filer subscription was reported as stopped and never restarted until the affected Filer was restarted.

#6618 also shows a case where all three Filers were present in `cluster.ps`, while metadata sync timestamps for one peer stopped progressing.

The issue is reproducible on both:

```text
SeaweedFS 4.33
SeaweedFS 4.41
```

The deployment uses an embedded metadata database on each Filer rather than a shared external database.

I can provide additional Filer/Master logs or run a patched build if needed.

chrislusf:
Fixes #10643

A filer only learns about its peers from `ClusterNodeUpdate` events on the master `KeepConnected` stream, and those are broadcast to whoever is connected at that moment. Nothing replays them, so a filer that is disconnected when a peer registers never subscribes to that peer, and its metadata replication from that peer stops until the filer is restarted.

The master now sends the current filer membership of the group right after a client's stream is established, next to the volume locations and the initial lock ring snapshot it already sends there. The peer add is also made idempotent: a repeated add for a peer that is already followed used to close the subscription and start a new one from `time.Now()`, dropping the events in between — with the membership replay that would happen on every master reconnect.

### Tests

`test/metadata_subscribe/peer_resubscribe_test.go` runs the reported sequence against real processes — master, two filers, no mocks:

1. an entry created on filer2 reaches filer1, so the peer subscription works to begin with
2. filer2 is killed, and the test waits for filer1 to log `stop subscribing peer` — the state the bug needs
3. filer1 is `SIGSTOP`ed and the master is restarted, breaking filer1's stream while it cannot reconnect
4. filer2 starts again and registers with the master while filer1 is deaf to it
5. filer1 is resumed, reconnects to the master, and a new entry created on filer2 has to reach it

Without the master commit this fails after 90s with `/peer-resubscribe/after-reconnect not on filer1: no entry is found in filer store`, which is the issue. With it, it passes in ~11s.

`test/multi_master/cluster_membership_test.go` covers the protocol end: a filer joining the master is told about the filer already registered, and the membership arrives *after* the volume locations — `masterClient` only reads `VolumeLocation` out of the first message it receives, so an update sent ahead of them would be dropped.

Note this only replays adds. A peer that leaves while a filer is disconnected still leaves a subscription retrying against a gone filer — noisy, but it recovers on its own if the peer comes back.

## Summary by CodeRabbit

* **New Features**
  * Newly connected clients automatically receive current filer cluster-node updates.
  * Cluster node updates can be listed by filer group and node type.
  * Existing filer registrations are shared with newly connected cluster members.

* **Bug Fixes**
  * Repeated peer-add notifications preserve active subscriptions instead of restarting them.
  * Filer metadata subscriptions reliably resume after master connection interruptions.
  * Connection streams terminate cleanly when initial updates cannot be sent.


chrislusf:
Added the regression test in `test/multi_master/cluster_membership_test.go`, on the existing 3-master harness rather than a mocked stream: a second filer joining the master must be told about the filer already registered. It fails against the pre-fix master with `DeadlineExceeded`, and the first-message assertion fails if the membership is sent ahead of the volume locations, which is the ordering a real filer would silently drop since `masterClient` only reads `VolumeLocation` out of the first message.

Skipping the `stream.Send` failure assertion: that path is the same `return sendErr` as every other send in `KeepConnected`, and reaching it in a test needs a mock stream behind a raft-leader `MasterServer`, which is more harness than the branch is worth.

PR #10653, opened 2026-08-08
topology: track volume size only where writes can land

chrislusf:
First of three trimming the master's resident volume layout, the part #10603 left untouched. Measured at 800k volume ids with 90% read-only, the shape reported there:

| index | before | per vid |
|---|---|---|
| `Disk.volumes` | 140.1 MB | 184 B |
| **`sizeTracking`** | **79.1 MB** | **104 B** |
| `vid2location` | 42.5 MB | 56 B |
| `readonlyVolumes` | 40.0 MB | 52 B |
| `crowded` | 9.3 MB | 12 B |

`sizeTracking` is the largest, and almost all of it is unreachable. It exists to decay pending-assign estimates so the master does not overfill a volume before heartbeats catch up — and nothing is ever assigned to a read-only volume, so an entry for one can never be consulted.

```
800k volumes, 90% read-only
  sizeTracking   79.1 MB -> 8.4 MB
  crowded         9.3 MB -> 1.2 MB
  layout total  311.3 MB -> 232.4 MB
```

The `crowded` drop is a side effect rather than a second change: a read-only volume now returns before reaching the threshold check that would have marked it. The other path that marks volumes crowded, `CollectDeadNodeAndFullVolumes`, still has no writability check and is a separate PR.

## Read-only, not unwritable

The gate is deliberately read-only rather than not-writable. A volume held out of the writable list *for capacity* keeps its entry, because `fullSince` on that entry is what enforces the recovery delay — dropping it would let a full volume bounce straight back into the writable list. `TestSizeTrackingSurvivesAFullVolume` covers that.

## Where the gate had to go

`initSizeTracking` alone is not enough: `UpdateVolumeSize` creates the entry too, and it runs for every volume on every full heartbeat, so it would simply put them all back.

`UpdateVolumeSize` takes the flag from its caller rather than consulting the layout's own read-only index, which can be stale — that index is updated by `RegisterVolume` and the explicit setters, but not by `EnsureCorrectWritables`, so a volume that flips to read-only between registrations is not reflected there. The heartbeat is ground truth.

Four tests: read-only volumes are not tracked, a volume that turns read-only gives its entry back, one that turns writable is tracked again from its reported size, and a full volume keeps its entry.

Refs #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume capacity tracking for read-only and unavailable replicas.
  * Prevented read-only or full-capacity volumes from being incorrectly added to tracking.
  * Restored tracking when volumes become writable again, using their latest reported size.
  * Removed outdated crowded-volume status when volumes become read-only.
  * Repaired missing volume location information during availability updates.

* **Tests**
  * Added coverage for replica report order, read-only transitions, full volumes, and capacity tracking behavior.

### DIFF
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -636,7 +636,7 @@ func (t *Topology) SyncDataNodeRegistration(volumes []*master_pb.VolumeInformati
 			// it is back.
 			newVolumes = append(newVolumes, v)
 		}
-		if vl.UpdateVolumeSize(v.Id, v.Size, v.CompactRevision, v.ReadOnly) {
+		if vl.UpdateVolumeSize(v.Id, v.Size, v.CompactRevision) {
 			vl.AdjustActiveVolumeCountAfterRecovery(v.Id)
 		}
 	}
@@ -712,7 +712,7 @@ func (t *Topology) ApplyVolumeChanges(changed []*master_pb.VolumeInformationMess
 			newVolumes = append(newVolumes, vi)
 		}
 		vl.EnsureCorrectWritables(&vi)
-		if vl.UpdateVolumeSize(vi.Id, vi.Size, vi.CompactRevision, vi.ReadOnly) {
+		if vl.UpdateVolumeSize(vi.Id, vi.Size, vi.CompactRevision) {
 			vl.AdjustActiveVolumeCountAfterRecovery(vi.Id)
 		}
 	}
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -636,7 +636,7 @@ func (t *Topology) SyncDataNodeRegistration(volumes []*master_pb.VolumeInformati
 			// it is back.
 			newVolumes = append(newVolumes, v)
 		}
-		if vl.UpdateVolumeSize(v.Id, v.Size, v.CompactRevision) {
+		if vl.UpdateVolumeSize(v.Id, v.Size, v.CompactRevision, v.ReadOnly) {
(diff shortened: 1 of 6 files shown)


chrislusf:
Both valid, both fixed in `64257aa`.

## Replica report order

Correct. `RegisterVolume` makes a volume unwritable if *any* replica is read-only, so asking the replica whose heartbeat happened to arrive was the wrong question — a writable replica reporting after a read-only one put the tracking straight back.

It now asks the volume, through `isAllWritable`, which reads each location's current state from the disk maps. That is ground truth for the heartbeat being processed, unlike the layout's own read-only index, which `EnsureCorrectWritables` does not update and so goes stale on exactly this transition.

That also removed the caller-supplied flag, so the signature is back to what it was and the eighteen test call sites this had churned are untouched.

`TestSizeTrackingIgnoresReplicaReportOrder` runs a writable and a read-only replica in both orders.

Cost is a map lookup per replica on a full-list heartbeat: about 19ms per 100k volumes, no extra allocations (`23142091 -> 23142653 B/op`). That path is rare since #10640, so I took the determinism over the microseconds.

## The crowded entry

Also correct, and it undercut the point of the change — the entry was simply moved rather than released. `removeFromCrowded` now goes with the delete, and `TestCrowdedEntryReleasedWithSizeTracking` covers a volume that is crowded while writable and then turns read-only.

That makes the crowded drop measured here real rather than incidental. The other feeder is still #10655.

PR #10654, opened 2026-08-08
topology: keep per-node volume state with the location it describes

chrislusf:
Second of three trimming the master's resident volume layout for #10603. Independent of #10653.

`readonlyVolumes` and `oversizedVolumes` are `map[VolumeId]*VolumeLocationList` — volume id to the list of nodes reporting that state. That is the same key space `vid2location` already holds, kept a second and third time, and at 800k volumes the read-only one alone is 40MB.

Nothing ever asks *which* nodes. Both are only ever read through `IsTrue`, which with `ExistCopies()` reduces to "is the list non-empty":

- `topology_vacuum.go:347` — skip read-only volumes
- `volume_layout.go` `Stats` — count a read-only volume at its size rather than the size limit
- `ensureCorrectWritables` and the capacity-recovery path — is it oversized

So the state moves onto the location list as a bit per entry, and both indexes go with the `volumesBinaryState` type.

```
800k volumes, 90% read-only
  readonly index   40.0 MB -> 0
  lookup index     42.5 MB -> 48.6 MB   (the bits)
  layout total    311.3 MB -> 277.4 MB
```

## Keeping the bits aligned

The bits index the location slice, so `Remove` shifts them down with the entries — otherwise a removed node's marking would slide onto whichever node took its place. A node replacing another at the same address inherits the slot, which matches what already happens to the location itself.

Bitmasks rather than a wider element type because `list` is read directly from fifteen places in the vacuum code; this leaves those untouched and costs the same 8 bytes per volume. Past 32 replicas of one volume a flag is not recorded rather than landing on another node's slot — `TestLocationFlagsIgnoreUntrackableReplicaCounts` pins that.

## The one place the key spaces could have differed

`SetVolumeReadOnly` guards on the volume id existing, not on the node being in its location list, so a flag for an absent node would have had nowhere to live. Its only caller iterates `Topo.Lookup` and passes a node from that list, so it cannot arise.

Four tests on the bookkeeping: flags follow their node when an earlier one is removed, go when their node goes, survive a replacement at the same address, and are dropped rather than misplaced beyond the tracked width.

Refs #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tracking of read-only and oversized volume locations.
  * Corrected volume availability, writability, recovery, and maintenance decisions when node locations change.
  * Preserved location status correctly when nodes are removed, replaced, or refreshed.
  * Prevented untracked replicas from affecting volume status.
  * Improved vacuum operations by using the current read-only status of volume locations.
  * Ensured volume status remains accurate after topology updates and stale location cleanup.

### DIFF
--- a/weed/topology/topology_vacuum.go
+++ b/weed/topology/topology_vacuum.go
@@ -344,7 +344,7 @@ func (t *Topology) vacuumOneVolumeLayout(grpcDialOption grpc.DialOption, volumeL
 // it so a benignly read-only (full/oversized) volume can be reclaimed.
 func (t *Topology) vacuumOneVolumeId(grpcDialOption grpc.DialOption, volumeLayout *VolumeLayout, c *Collection, garbageThreshold float64, locationList *VolumeLocationList, vid needle.VolumeId, preallocate int64, skipReadOnly bool) {
 	volumeLayout.accessLock.RLock()
-	isReadOnly := volumeLayout.readonlyVolumes.IsTrue(vid)
+	isReadOnly := volumeLayout.vid2location[vid].AnyReadOnly()
 	isEnoughCopies := volumeLayout.enoughCopies(vid)
 	volumeLayout.accessLock.RUnlock()
 
--- a/weed/topology/volume_layout.go
+++ b/weed/topology/volume_layout.go
@@ -35,73 +35,6 @@ const (
 	NoWritableVolumes             = "No writable volumes"
 )
 
-type stateIndicator func(copyState) bool
-
-func ExistCopies() stateIndicator {
-	return func(state copyState) bool { return state != noCopies }
-}
-
-type volumesBinaryState struct {
-	rp        *super_block.ReplicaPlacement
-	name      volumeState    // the name for volume state (eg. "Readonly", "Oversized")
-	indicator stateIndicator // indicate whether the volumes should be marked as `name`
-	copyMap   map[needle.VolumeId]*VolumeLocationList
-}
-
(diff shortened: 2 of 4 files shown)


chrislusf:
Valid, fixed in `6ebb421`. `Refresh` rebuilds the location list, so leaving the masks alone left bits describing whoever moved into the dropped entries' place — exactly the aliasing the shift in `Remove` exists to prevent, missed on the other path that reorders locations.

It now rebuilds both masks alongside the list.

Worth recording how nearly the regression test failed to earn its keep. My first check removed the `setFlag` calls from the rebuild, which left the masks at zero, so `AnyReadOnly` returned false and the test passed. That is a different bug from the one being guarded against. Restoring the actual pre-fix body — `dnll.list = l` and nothing else — fails as it should:

```
--- FAIL: TestLocationFlagsRebuiltByRefresh
    the flag did not move with its location, so it now describes another server
```

`Refresh` still has no callers, which is why this was latent. I would rather delete it than keep dead code carrying an invariant, but that is a separate call.

chrislusf:
Right, and this is the gap I found while verifying the fix and then did not close — the test caught a stranded flag but not a dropped one, because the only assertion was that clearing it works, which an already-empty mask satisfies.

`18d9b8f` asserts the flag survived the rebuild before checking it survived against the right location. Both failure modes now fail it:

```
flags left stranded (the original bug)
  --- FAIL: the flag did not move with its location, so it now describes another server

flags wiped instead of rebuilt
  --- FAIL: the flag was dropped rather than carried across the rebuild
```

The other two comments here and on #10653 are on the earlier diffs; `6ebb421`, `64257aa` and the sleep removal on #10655 already cover them.

PR #10655, opened 2026-08-08
topology: mark a volume crowded only if it can take writes

chrislusf:
Last of three trimming the master's resident volume layout for #10603. Independent of #10653 and #10654.

The writable-volume refresh loop feeds `chanCrowdedVolumes` for any volume past the growth threshold, with no check that writes can land there. In a tiered cluster the read-only volumes sit near the size limit — that is usually why they were tiered — so nearly every volume ends up in the crowded set.

Those entries change no decision. `GetWritableVolumeCount` already reports crowded as its intersection with the writable list, from #10522, precisely because a crowded set that outgrows writables demanded growth forever. So a read-only volume in there is discounted at every read and only takes space.

## Why here rather than at the map

The other feeder, `UpdateVolumeSize`, stops before the threshold check for read-only volumes once #10653 lands, which is what leaves this path as the one that still adds them. Gating `setVolumeCrowded` itself would mean testing membership of `writables`, which is a slice — an O(writables) scan per crowded volume per pass. The source has the volume in hand and the check is free.

Insertion is what is gated, not removal: #7793 deliberately stopped removing entries when a volume becomes unwritable, to keep transient flips from flapping the set, and that is untouched. A volume that was crowded while writable keeps its entry.

## Tests

`TestCrowdedVolumesAreOnesThatCanTakeWrites` puts a writable and a read-only volume both past the threshold and asserts only the first is reported; it fails against the previous version. `TestCrowdedCountStillMatchesWritables` checks the count growth actually decides on is unchanged.

The refresh loop's channels are unbuffered and normally drained by a goroutine `StartRefreshWritableVolumes` owns, so the test drains them itself.

Refs #10603

## Summary by CodeRabbit

* **Bug Fixes**
  * Crowded-volume reporting now includes only writable volumes that exceed the growth threshold.
  * Read-only volumes are excluded from crowded-volume results, preventing them from being considered available for additional writes.
  * Improved consistency between crowded-volume totals and writable-volume counts.

* **Tests**
  * Added coverage verifying writable and read-only volume handling.
  * Added checks confirming crowded-volume counts remain accurate across topology changes.

### DIFF
--- a/weed/topology/crowded_writable_test.go
+++ b/weed/topology/crowded_writable_test.go
@@ -28,21 +28,23 @@ func TestCrowdedVolumesAreOnesThatCanTakeWrites(t *testing.T) {
 	// The channels are unbuffered and normally drained by the writable-refresh
 	// loop, which is not running here.
 	crowded := map[needle.VolumeId]bool{}
-	collected := make(chan struct{})
+	stop, done := make(chan struct{}), make(chan struct{})
 	go func() {
+		defer close(done)
 		for {
 			select {
 			case v := <-topo.chanCrowdedVolumes:
 				crowded[v.Id] = true
 			case <-topo.chanFullVolumes:
-			case <-collected:
+			case <-stop:
 				return
 			}
 		}
 	}()
 	topo.CollectDeadNodeAndFullVolumes(time.Now().Unix()-1, topo.volumeSizeLimit, 0.9)
 	time.Sleep(50 * time.Millisecond)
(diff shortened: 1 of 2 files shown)


chrislusf:
Right, removed. The channels are unbuffered, so every send has been received by the time the sweep returns, and `<-done` already covers the recording that follows each receive. The sleep only suggested the result turned on timing.

Ran the test five times without it, plus under `-race`.

PR #10656, opened 2026-08-08
s3api: fix ListObjectsV2 dropping objects under a partial prefix

### LINKED ISSUE #10652: ListObjectsV2 with small MaxKeys returns incomplete results (missing objects)
**Describe the bug**

When calling ListObjectsV2 with a small MaxKeys value (e.g. MaxKeys = 1), SeaweedFS returns incomplete results. The pagination terminates early — NextContinuationToken becomes null (or IsTruncated = false) before all matching objects have been enumerated. This causes objects to be silently missing from the listing.

The issue is reproducible with a prefix that matches multiple keys in nested paths. Increasing MaxKeys to a larger value (e.g. 1000) masks the problem — all objects are returned correctly in a single response.

**Environment**

SeaweedFS version: 4.41
Deployment:
```
  seaweedfs:
    container_name: ${COMPOSE_PROJECT_NAME}_seaweedfs
    profiles: [ resources ]
    image: chrislusf/seaweedfs:latest
    entrypoint: /bin/sh -c
    command: |
      "echo '{
        \"identities\": [
          {
            \"name\": \"anonymous\",
            \"actions\": [
              \"Read\",
              \"List\"
            ]
          },
          {
            \"name\": \"seaweedfs_admin\",
            \"credentials\": [
              {
                \"accessKey\": \"seaweedfs\",
                \"secretKey\": \"seaweedfs123\"
              }
            ],
            \"actions\": [
              \"Admin\",
              \"Read\",
              \"List\",
              \"Tagging\",
              \"Write\"
            ]
          }
        ]
      }' > /etc/seaweedfs/config.json && \
      weed server -master.volumeSizeLimitMB=100 -volume.max=0 -volume.preStopSeconds=1 -dir=/data -s3 -s3.config /etc/seaweedfs/config.json"
    restart: unless-stopped
    ports:
      - 9333:9333   # Master
      - 8080:8080   # Volume
      - 8888:8888   # Filer
      - 8333:8333   # S3
    volumes:
      - seaweedfs-data:/data
    networks:
      front:
        ipv4_address: 10.0.0.14
```

**To Reproduce**

Create a bucket list-objects-issue
Put the following objects into the bucket:
- foo0.bin
- foo1/bar1.bin
- foo1/foo2/foo3.bin
- foo11/foo11.bin
- foo4/bar4/foo4.bin
- bar3/foo3/bar3.bin

Call ListObjectsV2 with Prefix = "foo1", MaxKeys = 1, no delimiter, following NextContinuationToken until it becomes null.

**Expected behavior**

All three matching keys should be returned across paginated responses:
- foo1/bar1.bin
- foo1/foo2/foo3.bin
- foo11/foo11.bin

**Actual behavior**

Not all keys are returned. The pagination terminates prematurely after 1–2 pages. The exact number of returned keys may vary between runs.

**Reproduction code (C# / AWS SDK for .NET)**

``` csharp
using System.Diagnostics;
using Amazon.Runtime;
using Amazon.S3;
using Amazon.S3.Model;

const string endpoint = "localhost:8333";
const string accessKey = "seaweedfs";
const string secretKey = "seaweedfs123";
const string region = "";
const bool withSsl = false;

var config = new AmazonS3Config
{
    ServiceURL = $"{(withSsl ? "https" : "http")}://{endpoint}",
    ForcePathStyle = true,
    UseHttp = !withSsl,
    AuthenticationRegion = region,
};

var client = new AmazonS3Client(new BasicAWSCredentials(accessKey, secretKey), config);

const string bucketName = "list-objects-issue";
const string prefix = "foo1";

HashSet<string> names =
[
    "foo0.bin",
    $"{prefix}/bar1.bin",
    $"{prefix}/foo2/foo3.bin",
    $"{prefix}1/foo11.bin",
    "foo4/bar4/foo4.bin",
    "bar3/foo3/bar3.bin",
];

byte[] data = new byte[1_000];
Random.Shared.NextBytes(data);

foreach (string name in names)
{
    var request = new PutObjectRequest
    {
        BucketName = bucketName,
        Key = name,
        InputStream = new MemoryStream(data),
        AutoCloseStream = false,
    };
    await client.PutObjectAsync(request);
}

string? continuationToken = null;

HashSet<string> expectedNames = names.Where(name => name.StartsWith(prefix)).ToHashSet();
HashSet<string> actualNames = new();

do
{
    ListObjectsV2Request request = new()
    {
        BucketName = bucketName,
        Prefix = prefix,
        Delimiter = null,
        MaxKeys = 1,
        ContinuationToken = continuationToken
    };
    ListObjectsV2Response response = await client.ListObjectsV2Async(request);

    if (response.S3Objects != null)
    {
        foreach (var obj in response.S3Objects)
        {
            actualNames.Add(obj.Key);
        }
    }

    continuationToken = response.NextContinuationToken;
}
while (continuationToken != null);

Debug.Assert(expectedNames.SetEquals(actualNames));
```

**Additional context**

The issue does not reproduce with MaxKeys = 1000 — all objects are returned in a single page.

hsdfat:

`ListObjectsV2` (and `ListObjects`) with a prefix that names only part of a directory name — e.g. `data/a`, which matches both `data/a/...` and `data/ab/...` — silently drops objects when paginating with a small `max-keys`. The response reports `IsTruncated=false` (empty `NextContinuationToken`) before all matching keys have been returned. A large `max-keys` (e.g. 1000) that fits everything in one page masks the problem.

Fixes #10652.

Minimal repro (`max-keys=1`): with keys `data/a/1`, `data/a/2`, `data/ab/3` and prefix `data/a`, pagination returns only `data/a/1` and `data/a/2` — `data/ab/3` is missing.

The bug is in `normalizePrefixMarker`. On resume, when the continuation marker had descended into `data/a/`, it took the listing directory from the *marker* (`alignedDir = "data/a"`) and discarded the `a` name prefix. The listing then only enumerated inside `data/a/`, finished, and reported not-truncated — never returning to the parent `data/` to continue with the sibling `data/ab/` that also matches the partial prefix.

The fix resolves the listing directory and name prefix from the *prefix* (via `toDirAndName`) instead of the marker, so `doListFilerEntries` continues to siblings sharing the name prefix after the marker's subtree is exhausted. The special-case branch is removed; the existing general path already handles this correctly, and no existing `normalizePrefixMarker` test exercised that branch.

# How is the PR tested?

- Added `Test_normalizePrefixMarker` cases for a partial name prefix whose marker resumes inside a matching subdirectory (`data/a` + `data/a/1` → dir `data`, prefix `a`, marker `a/1`). Written test-first: fails before the change, passes after.
- `go test ./weed/s3api/` passes for the full package with no regressions.
- Verified end-to-end against a locally built server (`weed server -s3`): the reported case now returns all objects across `max-keys=1/2/3` pagination, matching a single `max-keys=1000` listing.

Follow-up in this PR: `delimiter=/` with a partial prefix was also broken, and the marker change above turned its silent truncation into a non-terminating loop. `buildTruncatedNextMarker` rebuilt the CommonPrefixes continuation token as `requestDir + "/" + prefix + "/" + name + "/"`, which only holds when `prefix` names a directory; for `prefix=data/a` it produced `data/a/ab/` for the sibling prefix `data/ab/`. It now returns the emitted CommonPrefix, which is already the full key path with its trailing delimiter.

Still out of scope, separate root cause: directory-key objects (an explicit `d/sub/` marker alongside real `d/sub/a`, `d/sub/b` children).

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed object listing pagination so markers in similarly named directories no longer hide sibling entries.
  * Improved prefix handling for nested, top-level, empty, and trailing-slash paths.
  * Listing results now remain consistently scoped to the requested prefix.

* **Tests**
  * Added coverage for partial-prefix pagination and marker handling across matching and sibling directories.


chrislusf:
Addressed both review comments, pushed to this branch.

**CommonPrefixes continuation token** — confirmed, and worse than described. I checked it end to end against `weed mini` with keys `data/a/1`, `data/a/2`, `data/ab/3`, `data/ac/4`:

| | `prefix=data/a&delimiter=/&max-keys=1` |
|---|---|
| before this PR | page 1 returns `data/a/`, token `data/a/a/`, page 2 is empty and not truncated — `data/ab/` and `data/ac/` silently dropped |
| with this PR only | page 1 `data/a/` token `data/a/a/`, page 2 `data/ab/` token `data/a/ab/`, page 3 `data/ab/` token `data/a/ab/` — **never terminates** |

So the malformed token predates the PR, but the marker change turned a silent truncation into an infinite loop with duplicates. Fixed here rather than in a follow-up.

Root cause is `buildTruncatedNextMarker` rebuilding the token as `requestDir + "/" + prefix + "/" + name + "/"`, which is only right when `prefix` names a real directory. The emitted CommonPrefix is already the full key path with its trailing delimiter, so the fix tracks that string and returns it directly. The `prefix` parameter is gone.

**Sibling-marker test case** — added, along with a `buildTruncatedNextMarker` case for the partial-prefix token.

Verification: paginated listings now match the single-page listing across 48 combinations (prefixes `""`, `data/a`, `data/`, `data`, `logs/2026`, `logs/2026/`, `logs/`, `top` x delimiter on/off x `max-keys` 1/2/3), with pagination required to terminate and tokens to advance. Base binary fails 7 of those, this PR alone fails 2, this PR with the fix fails 0. `go test ./weed/s3api/...` passes.

PR #10658, opened 2026-08-09
filer: add initial chunk-aligned HLS TS VOD support (part 1)

Dimonyga:
## Summary

This is **Part 1** of an incremental VOD/HLS feature for the filer.

It adds an opt-in `/hls/` virtual namespace for storing a VOD asset as a **single logical filer entry** while exposing a conventional HLS media playlist with plain `.ts` segment URLs.

The main goal is to make HLS segment reads map naturally to SeaweedFS storage chunks instead of serving a large source object through many arbitrary HTTP Range requests. This is especially useful for VOD origins and older HLS clients/TVs that expect regular MPEG-TS segment URLs and do not use byte-range HLS.

Part 1 intentionally keeps the media model small and explicit. More automatic media handling is planned as Part 2; see the roadmap below.

## How Part 1 works

Ingest accepts two multipart parts:

1. an FFmpeg-style **single-file HLS VOD playlist** using `EXT-X-BYTERANGE`
2. the corresponding **single MPEG-TS media payload**

The playlist is used only as ingest metadata. It describes the segment boundaries inside the single TS file.

During ingest, SeaweedFS preserves those boundaries in storage:

- a storage chunk never spans two HLS segments
- segments larger than the normal filer chunk size are split into multiple chunks
- chunk splits are aligned to the 188-byte MPEG-TS packet size
- the normal filer `maxMB` limit is reused rather than introducing a separate media chunk size

The HLS client never sees byte ranges. SeaweedFS generates a normal playlist such as:

```m3u8
#EXTM3U
#EXT-X-VERSION:3
#EXT-X-TARGETDURATION:6
#EXT-X-MEDIA-SEQUENCE:0
#EXT-X-PLAYLIST-TYPE:VOD
#EXTINF:6.000,
0.ts
#EXTINF:6.000,
1.ts
#EXT-X-ENDLIST
```

and serves:

```text
GET /hls/videos/movie/index.m3u8
GET /hls/videos/movie/0.ts
GET /hls/videos/movie/1.ts
...
```

A segment read uses the existing filer streaming/prefetch path and reads only the storage chunks belonging to that segment.

## Configuration

The feature is disabled by default.

In `filer.toml`:

```toml
[filer.options]
hls_ts_enabled = true
```

HLS playback follows normal filer read-JWT behavior by default.

`security.toml` has a separate playback policy:

```toml
[jwt.filer_signing.hls]
enabled = true
```

Setting it to `false` allows public `GET`/`HEAD` playback under `/hls/` even when filer read JWT is configured.

HLS ingest never bypasses the normal filer write-JWT policy.

JWT prefix scopes are checked against the underlying filer path, not the virtual `/hls/` path.

## Preparing media for Part 1

Part 1 does **not** accept an MP4 directly. The source must first be prepared as single-file HLS MPEG-TS.

A complete example for a regular MP4 is:

```bash
mkdir -p hls-out

ffmpeg -i input.mp4 \
  -map 0:v:0 \
  -map 0:a? \
  -c:v libx264 \
  -preset medium \
  -crf 20 \
  -pix_fmt yuv420p \
  -c:a aac \
  -b:a 192k \
  -ar 48000 \
  -force_key_frames 'expr:gte(t,n_forced*6)' \
  -sc_threshold 0 \
  -f hls \
  -hls_time 6 \
  -hls_playlist_type vod \
  -hls_segment_type mpegts \
  -hls_flags single_file+independent_segments \
  -hls_segment_filename hls-out/video.ts \
  hls-out/index.m3u8
```

This produces:

```text
hls-out/index.m3u8
hls-out/video.ts
```

The generated `index.m3u8` is only an ingest description. SeaweedFS stores its own segment metadata and generates the playback playlist dynamically.

Upload example:

```bash
curl -X POST \
  -F 'playlist=@hls-out/index.m3u8;type=application/vnd.apple.mpegurl' \
  -F 'media=@hls-out/video.ts;type=video/mp2t' \
  http://127.0.0.1:8888/hls/videos/movie
```

Playback then becomes:

```bash
curl http://127.0.0.1:8888/hls/videos/movie/index.m3u8
curl -o segment.ts http://127.0.0.1:8888/hls/videos/movie/0.ts
```

If the input is already HLS-TS compatible and has suitable keyframe placement, FFmpeg can also remux instead of re-encoding, but Part 1 deliberately leaves source preparation outside SeaweedFS.

## Current supported scope

Part 1 currently supports:

- VOD playlists only (`EXT-X-ENDLIST` is required)
- single-file HLS media playlists using `EXT-X-BYTERANGE`
- one media URI shared by all segments
- contiguous byte ranges covering the media payload
- MPEG-TS media with valid 188-byte packet framing
- normal filer storage options and `maxMB` semantics
- `GET`/`HEAD` virtual playlist and `.ts` segment serving

It deliberately does not try to preserve playlist state that the generated playlist cannot yet represent safely. In particular, ingest rejects unsupported state such as:

- `EXT-X-KEY`
- discontinuities / discontinuity sequence
- `EXT-X-MAP`
- `EXT-X-GAP`
- I-frame-only playlists

Part 1 also does **not** provide:

- direct MP4 ingest
- automatic GOP/keyframe discovery
- live HLS
- master/multi-variant playlists
- transcoding
- automatic rendition generation
- rich audio/video track metadata
- MP4/MPEG-TS media introspection beyond the validation needed for safe storage and playback

## Storage and failure behavior

The implementation also handles several storage-specific cases:

- oversized HLS segments are split using the regular filer chunk limit
- chunk boundaries remain MPEG-TS packet aligned
- failed uploads clean up uncommitted chunks
- malformed media/playlist input returns client errors, while storage/assign/upload failures are surfaced as server errors
- truncated or trailing media payloads are rejected
- internal segment metadata is stored under the existing `x-seaweedfs-*` namespace so it is not accidentally emitted as a potentially large HTTP response header by normal filer GETs
- metadata and sequence arithmetic are checked for integer overflow

## Part 2 roadmap

This PR is intentionally the initial storage/serving foundation.

The next part is planned to remove most of the manual preparation required above.

The intended Part 2 workflow is to upload a **prepared fast-start MP4 with suitable GOP/keyframe placement directly**, without generating an external HLS playlist first. SeaweedFS would then automatically:

- inspect MP4 container metadata available from the fast-start layout
- discover sync samples/keyframes and GOP boundaries
- derive HLS segment timing and byte/sample boundaries
- build and persist the internal segment index automatically
- extract media and track metadata, including audio streams/languages where available
- build the HLS representation without requiring the caller to supply `index.m3u8`
- support container repackaging for HLS-TS-compatible codecs without requiring an unnecessary video re-encode

Further follow-ups can build on that metadata layer for richer rendition handling, master playlists and additional media formats without changing the basic filer storage model introduced here.

Keeping these pieces out of Part 1 is intentional: this PR establishes the storage layout, routing, authorization and segment-serving behavior first, while leaving the media parser/remux surface small enough to evolve separately.

## Tests

The change includes focused tests for:

- playlist parsing and validation
- implicit/explicit byte ranges and sequence handling
- overflow and malformed-range cases
- unsupported stateful HLS tags
- MPEG-TS packet validation
- segment-to-storage-chunk splitting
- oversized segment handling and TS packet alignment
- truncated/trailing media detection
- streaming reader EOF behavior
- virtual route parsing
- inherited/overridden filer chunk size semantics
- HLS JWT playback policy
- internal metadata isolation
- segment sequence lookup and response header behavior

## Summary by CodeRabbit

* **New Features**
  * Added opt-in HLS MPEG-TS support for uploading and streaming single-file VOD content.
  * Added playlist parsing, validation, rendering, and packet-aligned segment delivery.
  * Added virtual HLS routes for playlists, segments, and ingestion.
  * Added configurable HLS playback authentication, enabled by default.
  * Added support for `HEAD` requests, cache headers, and selected response metadata.

* **Bug Fixes**
  * Improved validation and error handling for malformed playlists, incomplete media, invalid routes, and storage failures.


Dimonyga:
@coderabbitai review

Dimonyga:
@coderabbitai resume

Dimonyga:
@chrislusf take a look at this if you can. I understand this is a huge change, but I can't yet imagine how to break it down into smaller pieces. This is just the bare minimum.

PR #10664, opened 2026-08-09
s3: take bucket sizes from the master's summary

chrislusf:
First of Phase 3 for #10603. The remaining memory is footprint rather than churn, and this is the acute part of it: a recurring spike that does not depend on the cluster growing.

`s3api`'s bucket size metrics loop pulls the entire volume list once a minute and adds it up itself, to arrive at six numbers per collection.

```
VolumeList over 550k volumes    184.6 MB allocated, 17.8 MB on the wire
CollectionStatistics               176 bytes allocated in 9 objects, 47 bytes on the wire
```

That 184.6MB is also held live on the master while the response marshals, so it lands on top of whatever else is in flight.

## Where the aggregation went

To the master, which can do it in one pass allocating per collection rather than per volume. Regular volumes count once each for logical totals and once per replica for physical, taken from the lookup index — already keyed by volume, so no set of seen ids is needed. Ec shards are node-local so their sizes sum across holders, while the file and delete counts describe the volume and resolve once every holder has been seen.

## Replicas that disagree

These numbers enforce bucket quotas, so I held the summary to producing what the listing produced — and the equivalence test immediately found a case where it could not.

Replicas of one volume disagree while a write is landing or a heartbeat is late. Walking a full listing took whichever replica the map iteration reached first, so **the existing answer moves between runs**. The summary takes the largest instead: stable, and it never reports usage below what some replica already holds, which is the safer direction for a quota.

`TestCollectionStatisticsMatchesAFullListing` pins equivalence for everything else against the old walk, kept as a reference implementation in the test.

## Tests

The three tests that covered the removed walker asserted absolute expected values, which is stronger than equivalence, so their cases moved with the logic: `TestCollectionStatisticsTakesTheLargestEcFileCount` (a holder still loading .ecx reports zero and must not pin the count down) and `TestCollectionStatisticsCountsRegularAndEcTogether`.

## One thing measuring caught

The first version allocated 83.9MB — `largest = &replica` heap-escaped a `VolumeInfo` per volume. Keeping the largest replica on the stack is what gets this to 176 bytes. Worth stating because a change whose whole purpose is removing a 184MB allocation would have shipped an 84MB one.

Refs #10603

## Summary by CodeRabbit

- **New Features**
  - Added collection-level statistics for file counts, deletion metrics, logical and physical sizes, and volume counts.
  - Added an endpoint for retrieving statistics across collections.
  - Improved bucket size reporting using summarized collection data.

- **Bug Fixes**
  - Improved accuracy for replicated and erasure-coded storage by preventing double-counting and separating logical from physical sizes.

- **Tests**
  - Added coverage for mixed storage types, replica selection, deduplication, and statistics accuracy.


chrislusf:
## Replica selection understating live usage — valid, fixed in `dc06155`

Right, and it contradicted the reason I gave for picking the largest. Quotas are enforced on `Size - DeletedByteCount`, and the replica with the biggest raw size can be the one that has deleted the most:

```
replica A  size 1000, deleted 900  -> 100 live
replica B  size  900, deleted   0  -> 900 live
```

Picking A by raw size reports 100 bytes live for a volume holding 900, which leaves an over-quota bucket writable — the opposite of what "never below what some replica already holds" was supposed to guarantee. It now selects on live bytes. `TestCollectionStatisticsPicksTheReplicaHoldingTheMostLiveData` reports `100` against the previous version.

## Stale fixture comment — fixed

Correct, it described the fixture before I changed the sizes to agree. Reworded, including your point about why disagreement has to be covered separately: the listing the equivalence test compares against is itself order dependent there.

## Ec delete counts — not changing this

The premise does not hold. `file_count` and `delete_count` do not both describe the volume:

```proto
uint64 file_count = 8;   // total needles in the .ecx index (live + tombstoned)
uint64 delete_count = 9; // node-local tombstones in the .ecj deletion journal
```

and `EcVolume.FileAndDeleteCount` is explicit about why they aggregate differently:

> deleteCount = len(deletedNeedles) — the number of unique runtime deletes tracked in memory... Because a needle delete is applied on exactly one shard holder, **the admin aggregation sums deleteCount across nodes** to get the volume's true delete total.

`.ecx` is the sealed index every holder carries, so max is right for `file_count`. Deletes land on one holder, so sum is right for `delete_count`. Taking the max would under-count deletes and overstate live usage — which for a quota is the dangerous direction.

Worth recording that this points at a real bug elsewhere: `Topology.CollectionEcVolumeStats`, which backs the `Statistics` RPC and the shell, takes the max for *both*. That under-reports deletes for any ec volume whose shards span nodes. Pre-existing and a separate fix, but the two implementations disagreeing is what made this look like an inconsistency here.

chrislusf:
## Cross-volume deletions cancelling live usage — valid, fixed in `ff39254`

Correct, and it is the same mistake as the replica one a layer up: the selection floored that volume's live size at zero, then the totals went on to add its raw counters as if it had not.

```
volume 1   size  100, deleted 500
volume 2   size 1000, deleted   0
collection size 1100, deleted 500  ->  600 live, for a bucket holding 1000
```

A volume's contribution to the collection's deletions is now capped at what it holds, so it can only ever cancel its own bytes. `TestCollectionStatisticsDeletionsNeverExceedTheVolume` reports 600 against the previous version.

## Lock order — checked, no inversion

`vl.accessLock` then `dn.RLock` is the order the package already uses: `RegisterVolume` and `SetVolumeAvailable` both hold `vl.accessLock` across a `dn.GetVolumesById` (`volume_layout.go:135`, `:854`). This follows it rather than adding a second order.

The opposite order does not exist. No `DataNode` method reaches a `VolumeLayout` while holding its own lock — the places that look like it, `UnRegisterDataNode` and `CollectDeadNodeAndFullVolumes`, call `dn.GetVolumes()` and take the node lock only for the copy, then touch the layout after it is released. Calling `GetVolumesById` *before* taking `vl.accessLock` is sequential, not a nesting, so there is nothing to invert against.

I would rather not move these reads behind a node-owned query on that basis: it would not remove a lock order, only relocate the one that already exists.

## Ec delete counts

Unchanged, for the reason in my previous comment: `delete_count` is documented as node-local and `EcVolume.FileAndDeleteCount` states the aggregation sums it across holders, while `file_count` comes from the volume-wide `.ecx` and takes the max. Taking the max for both would under-count deletes and overstate live usage, which for a quota is the direction that lets a bucket through.

PR #10665, opened 2026-08-09
storage: share the volume strings a cluster repeats

chrislusf:
Second of Phase 3 for #10603, on the master's resident footprint. Independent of #10664.

A cluster has a handful of collection names, disk types and remote backends, but decoding a heartbeat allocates a fresh string for each of them on every volume. A master holding a million volumes holds a million copies of `"somecollection"`.

`unique.Make` shares one copy. It clears entries by weak reference, so a collection that goes away is not pinned.

```
800k volumes, registered from a heartbeat that has been over the wire
  plain     227 -> 211 B/volume
  tiered    238 -> 214 B/volume
```

Tiered volumes gain more because the backend name shares too. At the 1.6M replicas in #10603 that is 26MB, or 38MB tiered.

`RemoteStorageKey` is deliberately left alone: it is unique per volume, so interning it would fill the table rather than share anything.

## The harness was hiding this

My first attempt measured no change at all, because building the messages in Go reuses one string literal across all 800k of them — there were never a million copies to remove. Only a heartbeat that has actually been marshalled and unmarshalled produces the distinct allocations production has. The numbers above come from that; the earlier per-volume figures I quoted on this issue were measured the other way and did not reflect this cost.

`TestRepeatedVolumeStringsAreShared` builds its strings from bytes for the same reason, and compares the backing pointers rather than the values, so it fails if the sharing stops. It also asserts `RemoteStorageKey` is *not* shared, so interning it later is a deliberate act rather than an accident.

## Cost

`BenchmarkSyncDataNodeRegistration/100000Volumes` moves 44.7ms to 46.2ms, about 3%, with allocations unchanged — decoding still allocates the string, this only stops it being retained.

Refs #10603

## Summary by CodeRabbit

* **Performance**
  * Reduced memory usage for repeated volume metadata by reusing identical strings.
  * Preserved empty values and ensured unique remote storage keys remain distinct.
  * Maintained shared metadata safely during concurrent operations and after garbage collection.

* **Tests**
  * Added coverage for shared volume strings, short volume information, empty values, non-remote storage scenarios, concurrency, and garbage collection.

### DIFF
--- a/weed/storage/volume_info.go
+++ b/weed/storage/volume_info.go
@@ -3,7 +3,7 @@ package storage
 import (
 	"fmt"
 	"sort"
-	"unique"
+	"sync"
 
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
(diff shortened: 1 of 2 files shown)


chrislusf:
Both valid, and the first one matters more than it looks. Fixed in `e1cb203`.

I probed the lifetime rather than reasoning about it:

```
shared across GC while a value is retained: false
shared across GC with nothing retained:     false
```

So `unique.Make` does not return the same canonical value after a collection, even while a caller still holds the string it handed back. Sharing was therefore per interning round, not per value.

That is worse here than it first appears. Volumes are interned when they are first reported, and since #10640 a server sends only what changed — so most volumes are interned once at connect and never again. Every collection cycle after that hands the next new volume its own copy, which is exactly the case this change exists to remove.

Swapped for a table that holds what it canonicalises. It only grows, so it stays restricted to values drawn from a small set — collection, disk type, remote backend; a cluster with ten thousand collections keeps a few hundred kilobytes. Never the remote storage key, which is unique per volume, and there is a test asserting it is *not* shared so interning it later has to be deliberate.

The saving is unchanged: `227 -> 211 B/volume`, `238 -> 214` tiered. `BenchmarkSyncDataNodeRegistration/100000Volumes` moves 44.7ms to 49.0ms for the read lock per volume, on the full-list path only.

Added the GC-boundary test you asked for, `TestVolumeStringsStaySharedAcrossCollection`, which fails against the `unique.Make` version, plus one for concurrent reports since the table is now shared mutable state.

PR #10674, opened 2026-08-09
master: let VolumeList ask for the volumes it wants

chrislusf:
Continuing Phase 3 for #10603.

`VolumeListRequest` carried nothing, so every caller was answered with the whole cluster. A dashboard opening one volume's page was served all 800k of them and threw away the rest — and the master built every one of those messages first, on the request path.

```
                          allocated   volumes listed
whole cluster               152.6 MB           800000
one collection of 20          8.9 MB            40000
one volume                     0.0 MB               1
```

## What the filter selects

Only what is listed under a disk — volumes and ec shards. The topology, its disks and their counters are always reported in full, because a caller reading free space or replica placement needs the cluster whichever volumes it asked about. That is what lets the s3 capacity probe below narrow its request without changing its answer.

```proto
string collection = 1;
uint32 volume_id = 2;
bool default_collection_only = 3;
```

Empty and zero take everything, matching how `volume.list` already reads its own `-collectionPattern` and `-volumeId`, and how `ExportClusterVolumeList` and `GetClusterEcShards` read the collection they are handed. A caller that passes its own "no filter" string straight through is therefore answered too much rather than answered wrongly — the failure mode is a slow page, not a page missing most of its rows.

That leaves the default collection unnameable, since the empty string is how "every collection" is spelled, so it gets a field of its own. A named collection wins over it, being the more specific of the two. Both rules are covered in `TestNewVolumeFilterReadsTheRequest`, and I checked each fails against the opposite implementation rather than assuming it.

An older client sends none of it and is answered exactly as before. Callers still filter on their own side after receiving, because an older *master* ignores the fields and sends everything.

## Callers switched

- **Volume details page** — was scanning the cluster for one volume and its replicas.
- **EC volume details page** — same, for one volume's shards.
- **SOSAPI capacity probe** — summed one collection out of a listing of every volume. Cluster capacity is unchanged: it comes from the disk counters.

Left alone deliberately: the maintenance scanner and the collection/volume list pages genuinely want everything, and the JSON exporter's duplicate-id scan is documented to cover the whole cluster even when a collection is selected — filtering it server-side would blind it to a duplicate whose two copies disagree about their collection, which is exactly what it looks for.

Refs #10603

## Summary by CodeRabbit

* **New Features**
  * Added filtering of volume topology by collection and volume ID.
  * Added support for selecting only the default collection.
  * Volume and EC volume details now retrieve targeted topology data.
  * Bucket capacity and usage calculations now use collection-specific topology.
  * Filtered results preserve relevant disk and topology metadata.

* **Bug Fixes**
  * Improved compatibility with older masters that do not apply request filters.
  * Preserved disk identity when filters match no volumes.

### DIFF
--- a/weed/admin/dash/ec_shard_management.go
+++ b/weed/admin/dash/ec_shard_management.go
@@ -559,7 +559,7 @@ func (s *AdminServer) GetEcVolumeDetails(volumeID uint32, sortBy string, sortOrd
 
 	// Get detailed EC shard information for the specific volume via gRPC
 	err := s.WithMasterClient(func(client master_pb.SeaweedClient) error {
-		resp, err := client.VolumeList(context.Background(), &master_pb.VolumeListRequest{})
+		resp, err := client.VolumeList(context.Background(), &master_pb.VolumeListRequest{VolumeId: volumeID})
 		if err != nil {
 			return err
 		}
@@ -571,6 +571,7 @@ func (s *AdminServer) GetEcVolumeDetails(volumeID uint32, sortBy string, sortOrd
 						for _, diskInfo := range node.DiskInfos {
 							// Process EC shard information for this specific volume
 							for _, ecShardInfo := range diskInfo.EcShardInfos {
+								// An older master ignores the filter.
 								if ecShardInfo.Id == volumeID {
 									collection = ecShardInfo.Collection
 									dataCenters[dc.Id] = true
--- a/weed/admin/dash/volume_management.go
+++ b/weed/admin/dash/volume_management.go
@@ -309,7 +309,7 @@ func (s *AdminServer) sortVolumes(volumes []VolumeWithTopology, sortBy string, s
 }
 
 // GetVolumeDetails retrieves detailed information about a specific volume
-func (s *AdminServer) GetVolumeDetails(volumeID int, server string) (*VolumeDetailsData, error) {
+func (s *AdminServer) GetVolumeDetails(volumeID uint32, server string) (*VolumeDetailsData, error) {
 	var primaryVolume VolumeWithTopology
 	var replicas []VolumeWithTopology
 	var volumeSizeLimit uint64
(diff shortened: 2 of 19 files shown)


chrislusf:
Both reviewers were right about `ToDiskInfo`, and chasing it turned up a second race underneath.

**The reported one.** I had moved the capacity calculation above `d.RLock()`, so `len(d.volumes)` read the map unsynchronised where the original read it inside the lock. Moved back in, folded into the commit that introduced it so no commit in this PR ships the race.

**The one that found.** Writing a test to prove the fix, `-race` still failed — on `DiskUsages.ToDiskInfo`, which is a different method I had not touched. It reads counters that heartbeats write with `atomic.AddInt64`, and iterates `d.usages` without the `RLock` its neighbour `FreeSpace()` takes. `Disk.ToDiskInfo` and `DiskUsageCounts.FreeSpace` read them the same way. Pre-existing, but on exactly the path this PR exercises, and it made my own fix untestable under `-race`. Fixed in its own commit via a `snapshot()` that loads each counter atomically — the codebase already reads them that way in `data_node.go`.

`TestVolumeFilterListsWhileVolumesChange` lists while a heartbeat writes; I confirmed it fails on both bugs by reintroducing each.

**CodeQL** flagged `uint32(volumeID)`. The bound check above it is sound, but CodeQL does not track it — `VacuumVolume` a few lines down has the same conversion with an `lgtm[go/incorrect-integer-conversion]` suppression, so this now matches.

chrislusf:
The CodeQL fix I pushed first did not work, and was the wrong shape anyway.

I had copied the `lgtm[go/incorrect-integer-conversion]` comment from `VacuumVolume` a few lines below. That is legacy LGTM.com syntax which GitHub's CodeQL does not honour, so it suppressed nothing — the alert came back on the same line. Worth knowing that comment is not doing anything where it already sits either.

The alert was also genuinely mine rather than a line shift: it exists only on this PR's merge ref, because `GetVolumeDetails` had no integer conversion in it before.

Fixed properly by not converting. The id is now parsed at the width it is used at — `strconv.ParseUint(volumeIDStr, 10, 32)` in the handler — and `GetVolumeDetails` takes a `uint32`, which drops the range check, the conversion and the suppression together, and matches `GetEcVolumeDetails(volumeID uint32, ...)` next to it.

PR #10676, opened 2026-08-09
master: stream volume listings

chrislusf:
Continuing Phase 3 for #10603. Stacked on #10674 — review that first; this branch targets it, and I will retarget to master once it lands.

At 800k volumes a listing is **36MB on the wire but 305MB as messages**, and the master built all of it before sending any, then held it while gRPC encoded it. The maintenance scanner asks for one every 30 minutes; the dashboard asks on page loads.

```
topology alone                  434.9 MB
VolumeList,       live           341.3 MB   (305 MB of messages + 36 MB encoding it)
VolumeListStream, live             4.4 MB   (162 batches, 1.6M volumes)
```

Streaming does not allocate meaningfully less. What changes is how much has to be **live at once**, which is what sets the heap ceiling — and with Go sizing the heap against the live set, a 341MB spike on a master already holding 435MB is what an 8GB cgroup runs out of when two arrive together.

## Shape

The topology goes out first, listing nothing; every message after carries volumes for one of its disks.

Batches are built under their disk's lock and sent outside it, so a slow reader stalls the stream rather than the topology. They therefore do not share one instant — but neither did a single listing, which takes each disk's lock in turn, so a volume moving during either can be seen twice or not at all. Only 4 bytes per volume are carried between the pass that takes the ids and the passes that build the messages.

## Old masters

`pb.ReceiveVolumeList` hides which kind of master answered: one that returns `Unimplemented` is asked the old way and its single reply cut into the same batches. Both paths hand over a topology that lists **no** volumes, so a caller cannot accidentally come to depend on finding them there and then break against the other kind of master.

That symmetry is the whole risk of the fallback, so it is tested against both: `TestReceiveVolumeListIsTheSameEitherWay` and `TestCollectVolumeListRebuildsTheListing` each run over a real gRPC connection against a master that streams and one that does not. The first version of the fallback aliased each disk's slice into the batch it handed over, so reassembly appended a slice onto itself and doubled every volume — both tests catch that, and only on the non-streaming path, which I verified by reintroducing it.

## Caller

The maintenance scanner. It reassembles client-side via `pb.CollectVolumeList`, which keeps the scan byte-identical: `ActiveTopology` splits disks by the disk ids carried on the volumes, so it needs them in the topology. The master-side win lands regardless — that is where #10603 OOMs.

The dashboard's list pages can move the same way in a follow-up; they are user-triggered rather than periodic, and the detail pages already stopped pulling the cluster in #10674.

## What the snapshot does and does not promise

The volumes a listing carries belong to disks it named — `StreamVolumes` takes the listing that went out and streams only those, so the two walks agree by construction and there is never a batch the client cannot place.

It is not an instant. A disk registering mid-listing is in neither half and is reported by the next one. A node that disconnects and reconnects under the same identifiers mid-listing has its new volumes attached to the counters the header carried, since disks are matched by identifier rather than object. Excluding it instead would drop a whole node's volumes, and the maintenance scanner reads replica counts off this listing — a missing replica is acted on, a slightly old `VolumeCount` is not.

None of that is new: `ToDiskInfo` reads a disk's counters and then takes its lock to read its volumes, so an unstreamed listing mixes the two across the same reconnect. The window widens from microseconds to seconds; the property was never atomic, and making it so would mean holding the topology for the length of a listing, which is what this PR exists to stop doing.

Refs #10603

## Summary by CodeRabbit

* **New Features**
  * Added streaming volume-list retrieval for faster, more efficient topology and volume discovery.
  * Volume data is delivered in batches, including erasure-code shard details and disk metadata.
  * Added compatibility fallback for environments that do not support streaming.
  * Maintenance scanning now uses the improved volume-list retrieval process.

* **Bug Fixes**
  * Prevented duplicate volume entries and preserved topology metadata during reconstruction.
  * Ensured streamed results remain consistent with the announced topology.
  * Improved handling of topology changes while volume information is being collected.

### DIFF
--- a/weed/admin/maintenance/maintenance_scanner.go
+++ b/weed/admin/maintenance/maintenance_scanner.go
@@ -6,6 +6,7 @@ import (
 	"time"
 
 	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/worker/types"
 )
@@ -101,7 +102,8 @@ func (ms *MaintenanceScanner) getVolumeHealthMetrics() ([]*types.VolumeHealthMet
 	glog.V(1).Infof("Collecting volume health metrics from master")
 	err := ms.adminClient.WithMasterClient(func(client master_pb.SeaweedClient) error {
 
-		resp, err := client.VolumeList(context.Background(), &master_pb.VolumeListRequest{})
+		// Streamed, so the master never builds the whole listing to send it.
+		resp, err := pb.CollectVolumeList(context.Background(), client, &master_pb.VolumeListRequest{})
 		if err != nil {
 			return err
 		}
--- a/weed/pb/master.proto
+++ b/weed/pb/master.proto
@@ -27,6 +27,8 @@ service Seaweed {
   }
   rpc VolumeList (VolumeListRequest) returns (VolumeListResponse) {
   }
+  rpc VolumeListStream (VolumeListRequest) returns (stream VolumeListStreamResponse) {
+  }
   rpc LookupEcVolume (LookupEcVolumeRequest) returns (LookupEcVolumeResponse) {
   }
   rpc VacuumVolume (VacuumVolumeRequest) returns (VacuumVolumeResponse) {
@@ -417,6 +419,24 @@ message VolumeListRequest {
 message VolumeListResponse {
   TopologyInfo topology_info = 1;
   uint64 volume_size_limit_mb = 2;
+}
+
+// VolumeListStream answers the same request as VolumeList without either end
+// holding every volume in the cluster at once. At 800k volumes the reply is
+// 36MB on the wire but 305MB as messages, which the master built in full
+// before sending any of it.
+message VolumeListStreamResponse {
+  // Sent once, first, listing no volumes: the topology, its disks and their
+  // counters. Every message after carries volumes for one of those disks.
+  VolumeListResponse header = 1;
+  // Which disk this batch is from. A disk arrives over as many batches as it
+  // takes, so append rather than assign.
+  string data_center = 2;
+  string rack = 3;
+  string data_node = 4;
+  string disk_type = 5;
+  repeated VolumeInformationMessage volume_infos = 6;
+  repeated VolumeEcShardInformationMessage ec_shard_infos = 7;
 }
 
 message LookupEcVolumeRequest {
--- a/weed/pb/master_pb/master.pb.go
+++ b/weed/pb/master_pb/master.pb.go
@@ -2895,6 +2895,106 @@ func (x *VolumeListResponse) GetVolumeSizeLimitMb() uint64 {
 	return 0
 }
 
+// VolumeListStream answers the same request as VolumeList without either end
+// holding every volume in the cluster at once. At 800k volumes the reply is
+// 36MB on the wire but 305MB as messages, which the master built in full
+// before sending any of it.
+type VolumeListStreamResponse struct {
+	state protoimpl.MessageState `protogen:"open.v1"`
+	// Sent once, first, listing no volumes: the topology, its disks and their
+	// counters. Every message after carries volumes for one of those disks.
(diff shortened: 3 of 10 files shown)


chrislusf:
Greptile's finding was real, and the failure mode was worse than a wrong answer: a heartbeat registering a disk between the header and the batches made `CollectVolumeList` return an error, so one new disk lost the entire 30-minute maintenance scan.

Those volumes belong to a disk the listing's topology never named, so there is nowhere to put them. They are now skipped and counted rather than fatal, which also restores what an unstreamed listing did — it reads each node's disks once, so a disk registering mid-traversal did not appear there either.

`TestCollectVolumeListSurvivesADiskAddedMidStream` covers it: the listing succeeds, keeps every volume it was told about, and does not grow a disk its topology never named.

Rebased on the updated #10674, which now also fixes a pre-existing counter race on this same path.

chrislusf:
The second of these was right, and the fix belonged further up than where I had put it.

The first comment describes behaviour I had already changed — an unknown-disk batch stopped failing the listing in an earlier push. But the second one is the sharper version and it stood: skipping those batches means the maintenance scanner computes **replica counts** from a listing that quietly lost volumes, and a volume short a replica is something it acts on.

Handling that at the client was the wrong place. The master sent the topology, then walked it a second time to stream the volumes, and nothing tied the two walks together — so a disk registering in between produced a batch the client had nowhere to put, and then had to decide what to do about it.

`StreamVolumes` now takes the listing that went out and streams only the disks it named. The two walks agree by construction, there is no unplaceable batch to drop, and the snapshot is exactly the topology the client was given. A disk arriving mid-listing is in neither half and is reported by the next one — which is also what an unstreamed listing did, since it reads each node's disks once.

`TestStreamVolumesStaysInsideTheAnnouncedTopology` registers a node between the listing and the walk and asserts nothing is streamed for it. I checked it fails against the unbounded walk — and had to fix the first attempt at that check, which didn't compile, so "FAIL" was proving nothing.

The client-side skip stays as a backstop for a master that does not bound its walk, but it is now unreachable against this one.

chrislusf:
Second round, both fixed.

**The fallback could restart a listing that had already begun.** Right, and the second of the two paths is the worse one. `ReceiveVolumeList` fell back to the unary RPC on any `Unimplemented`, including one raised by the *caller's own callback* — so a caller that happened to return that code had the entire listing silently replayed into it. The other path is a master that sends its topology and then gives up with that code, which re-ran `onTopology` and delivered every batch twice.

It now falls back only when the stream said nothing at all. Past its first message the master plainly does have the method, so there is nothing to fall back to.

Two tests, and both fail against the unguarded version — the callback one reports being called twice, which is the doubling itself.

**A cancellable context for the scanner.** Fair, but not this PR's to make. `ScanForMaintenanceTasks()` takes no context, its caller in `maintenance_manager.go` has none to give, and `getDefaultReplicaPlacement` right beside it already uses `context.Background()`. The unary `VolumeList` this replaced had no deadline either, so nothing here got less cancellable — plumbing one through the scanner is a change worth making on its own, where the timeout can be chosen against how long a full-cluster listing actually takes rather than guessed at here.

chrislusf:
The scenario is real — I checked rather than assumed. `UnRegisterDataNode` ends with `dn.Parent().UnlinkChildNode(dn.Id())`, so a disconnect really does remove the node, and `GetOrCreateDataNode` then builds a fresh `DataNode` and fresh `Disk` on reconnect. Matching by the four identifiers does accept that as the original.

I am leaving it, because the alternative is worse for the caller this exists for.

Matching by object identity instead would exclude the replacement disk, and its volumes would not be listed at all. That is precisely the failure established two comments ago: the maintenance scanner reads replica counts off this listing, and a volume short a replica is something it acts on. Trading "one disk's counters are from a few seconds earlier" for "an entire node's volumes are missing" moves the harm from cosmetic to actionable.

It is also not a new class of inconsistency. `ToDiskInfo` reads the disk's counters and then takes the lock to read its volumes, so an unstreamed listing mixes the two across a reconnect in exactly the same way — the window is microseconds rather than seconds, but the property was never atomic. Nothing here can be made a true snapshot without holding the topology for the length of a listing, which is the thing this PR exists to stop doing.

What the streamed listing promises is that the volumes it carries belong to disks it named, and that holds either way. I have said so in the PR body rather than leaving it in a thread.

PR #10678, opened 2026-08-09
filer: format adapters align storage chunks to file structure (HLS TS, parquet)

chrislusf:
This generalizes the idea in #10658. Instead of an HLS-specific virtual namespace, the filer gets a small format adapter system: an adapter reduces one container format to extent sizes, an alignment quantum, and an opaque payload. The core never learns what a TS packet or a row group is, and adapters never see chunks, file ids, or authorization. HLS TS and parquet ship as the first two adapters.

## Why align chunks to file structure

The chunk is the unit of lookup, caching, CRC, EC recovery, and tiering. When chunk boundaries coincide with the file's internal structure, a segment or row-group read maps to exactly its covering chunks: one lookup and one volume round trip instead of two, whole-needle reads that verify CRC, caches that hold exactly the hot bytes, and half the fan-out on EC and cloud-tiered volumes. Arbitrary fixed-size cuts give none of that when readers fetch sub-ranges.

## How it works

The upload loop now accepts a boundary source instead of a fixed chunk size; the fixed size stays as the default implementation. A `Layout` (extent sizes + align + adapter payload) is derived by the adapter and drives the cuts: chunks never span extents, extents larger than `maxMB` split on the align quantum. The layout persists in one compact `x-seaweedfs-` extended attribute, keyed by extent sizes rather than chunk ids so it survives chunk manifest folding — about 3 bytes per extent.

Everything runs on the entry's real path, so JWT prefix scopes, WORM, and read-only rules apply unchanged. No new namespace, no auth bypass knob, no config.

- `POST /path?format.ingest=hls-ts` ingests a multipart `index` sidecar (the FFmpeg-generated EXT-X-BYTERANGE playlist) plus the `media` bytes, cutting chunks on segment boundaries at 188-byte packet alignment.
- `GET /path?format.view=hls-ts` serves a rewritten playlist with plain numbered segment URLs for clients that do not speak byte-range HLS; `&seq=N` streams one segment through the normal prefetch path with entry ETag, preconditions, and HEAD.
- `POST /path?format.repack=parquet` derives the layout from the stored bytes (parquet keeps its index in the footer, so ingest-time cutting is impossible) and rewrites the chunks cut at row-group starts, swapping the entry and queueing the old chunks for deletion. Parquet needs no view; readers already fetch row groups by offset, alignment alone delivers the benefit.

```
ffmpeg -i input.mp4 ... -f hls -hls_playlist_type vod -hls_flags single_file \
  -hls_segment_filename video.ts index.m3u8

curl -F 'index=@index.m3u8' -F 'media=@video.ts;type=video/mp2t' \
  'http://filer:8888/videos/movie.ts?format.ingest=hls-ts'
curl 'http://filer:8888/videos/movie.ts?format.view=hls-ts'          # playlist
curl 'http://filer:8888/videos/movie.ts?format.view=hls-ts&seq=0'    # segment

curl -X POST 'http://filer:8888/data/part-0001.parquet?format.repack=parquet'
```

## Boundaries that keep it maintainable

Adapters implement identity plus whatever capabilities fit (`Sniffer`, `Indexer`, `SidecarIndexer`, `Viewer`) and are discovered by type assertion; they register from `init()` like filer stores. Viewers return a plan (a rendered document or an extent number) and the server executes it, so streaming stays on the optimized path and adapters stay pure functions. A `formattest` kit holds every adapter to no-panic parsing of truncated input, since indexers parse untrusted bytes inside a storage daemon.

The layout is advisory: a stale or corrupt one 404s its views while plain reads stay untouched. It is also bound to a digest of the chunk list it described, so a write through any other path (offset writes, appends, mounts) invalidates the views until re-ingest or repack, even when the file size is unchanged.

## Limitations

- Repack refuses hard-linked, remote, and server-side-encrypted entries. Versioned entries are fine: every object version owns its chunk list. Within a filer, writes serialize on the per-path entry lock - gRPC writers, renames, plain-HTTP overwrites, and repack all take it, and WORM is revalidated under it. The lock is filer-local, so repack also re-reads the entry from the store at commit time and aborts with 409 if anything that fed the repack changed - chunk fingerprint, file size, TTL and both expiry anchors, the S3-expiry flag, hard-link or remote state - or WORM was enabled. The cross-filer window shrinks to the swap itself, inside which repack races like any ordinary writer. Closing it entirely needs owner routing like the S3 gateway's, which is the follow-up; until then route writes for a path to one filer. TTL'd entries repack with their remaining lifetime, anchored the way expiry actually works (Mtime for S3-expiring entries, Crtime otherwise) and rounded up to a volume-representable TTL so chunks never expire before their entry.
- Ingest is explicit; nothing sniffs uploads on the write path.
- S3 surface (per-bucket repack policy, align hints on PUT) and a background repack worker are follow-ups; this PR is the storage and serving foundation they build on.

Tests cover the layout codec and cutter, playlist parsing/rendering/rejections, parquet extent derivation against footer metadata, truncation fuzzing for both adapters, and the full existing weed/server suite passes.

### DIFF
--- a/weed/filer/filechunk_manifest.go
+++ b/weed/filer/filechunk_manifest.go
@@ -236,7 +236,10 @@ func doMaybeManifestize(saveFunc SaveDataAsChunkFunctionType, inputChunks []*fil
 	for i := 0; i+mergeFactor <= len(dataChunks); i += mergeFactor {
 		chunk, err := mergefn(saveFunc, dataChunks[i:i+mergeFactor])
 		if err != nil {
-			return dataChunks, err
+			// Return the manifests already written plus the chunks not yet
+			// wrapped: a complete, deletable representation of every byte, so
+			// callers can clean up or keep a usable chunk list.
+			return append(chunks, dataChunks[i:]...), err
 		}
 		chunks = append(chunks, chunk)
 		remaining -= mergeFactor
--- a/weed/filer/filechunk_manifest_test.go
+++ b/weed/filer/filechunk_manifest_test.go
@@ -77,6 +77,38 @@ func TestDoMaybeManifestize(t *testing.T) {
 		actual, _ := doMaybeManifestize(nil, mtest.inputs, 2, mockMerge)
 		assertEqualChunks(t, mtest.expected, actual)
 	}
+}
+
+// A mid-run merge failure must still return every chunk that exists: the
+// manifests already written plus the chunks not yet wrapped, so callers can
+// delete or keep a complete set.
+func TestDoMaybeManifestizePartialFailure(t *testing.T) {
+	inputs := []*filer_pb.FileChunk{
+		{FileId: "0", IsChunkManifest: true},
+		{FileId: "1", IsChunkManifest: false},
+		{FileId: "2", IsChunkManifest: false},
+		{FileId: "3", IsChunkManifest: false},
+		{FileId: "4", IsChunkManifest: false},
+	}
+	calls := 0
+	failingMerge := func(saveFunc SaveDataAsChunkFunctionType, dataChunks []*filer_pb.FileChunk) (*filer_pb.FileChunk, error) {
+		calls++
+		if calls > 1 {
+			return nil, fmt.Errorf("merge failed")
+		}
+		return mockMerge(saveFunc, dataChunks)
+	}
+	actual, err := doMaybeManifestize(nil, inputs, 2, failingMerge)
+	if err == nil {
+		t.Fatalf("doMaybeManifestize() expected an error")
+	}
+	expected := []*filer_pb.FileChunk{
+		{FileId: "0", IsChunkManifest: true},
+		{FileId: "12", IsChunkManifest: true},
+		{FileId: "3", IsChunkManifest: false},
+		{FileId: "4", IsChunkManifest: false},
+	}
+	assertEqualChunks(t, expected, actual)
 
 }
 
--- a/weed/format/format.go
+++ b/weed/format/format.go
@@ -1,0 +1,83 @@
+// Package format maps the internal structure of container file formats onto
+// storage chunk boundaries.
+//
+// An adapter translates one format into three things the core understands: a
+// list of extent sizes, an alignment quantum, and an opaque payload. Adapters
+// never see chunks, file ids, or authorization; the core never learns what an
+// MPEG-TS packet or a parquet row group is.
+package format
+
+import (
+	"context"
+	"errors"
+	"io"
+	"net/url"
+)
+
+// LayoutKey is the filer Extended key holding the encoded Layout. The
+// x-seaweedfs- prefix keeps it out of HTTP response headers.
+const LayoutKey = "x-seaweedfs-format-layout"
+
+// ErrNoSuchView reports that a view request addresses nothing servable; the
+// server answers 404.
+var ErrNoSuchView = errors.New("no such view")
+
+// Layout describes how a file's structure maps to byte extents.
+type Layout struct {
+	Format      string  // adapter name
+	ExtentSizes []int64 // extent lengths in file order; they sum to the file size
+	Align       int64   // quantum for cutting inside an oversized extent; 1 cuts anywhere
+	Payload     []byte  // adapter-owned metadata, opaque to the core
+}
+
+// Hint carries the cheap identification signals available to Sniff.
+type Hint struct {
+	Name        string
+	ContentType string
+	Size        int64
+	Head        []byte
+	Tail        []byte
+}
+
+// Format is the mandatory adapter identity. Capabilities beyond it are
+// discovered by type assertion.
+type Format interface {
+	Name() string
+	Sniff(h Hint) bool
+}
+
+// Indexer derives a Layout from the complete stored bytes.
+type Indexer interface {
+	Index(ctx context.Context, r io.ReaderAt, size int64) (*Layout, error)
+}
+
+// SidecarIndexer derives a Layout from an external index document supplied at
+// ingest, before the media bytes arrive.
+type SidecarIndexer interface {
+	IndexSidecar(sidecar []byte) (*Layout, error)
+}
+
+// Object is everything a Viewer may know about the file it serves.
+type Object struct {
+	Name   string
+	Size   int64
+	Layout *Layout
+}
+
+// ViewRequest carries the request parameters of a ?view= request.
+type ViewRequest struct {
+	Query url.Values
+}
+
+// ViewPlan tells the server what to serve. The server executes it on the
+// normal streaming path; adapters stay pure functions of request and layout.
+type ViewPlan struct {
+	ContentType string
+	Body        []byte // rendered document; when nil, stream Extent instead
+	Extent      int
+}
+
+// Viewer answers ?view= requests.
+type Viewer interface {
+	View(req ViewRequest, obj Object) (*ViewPlan, error)
+}
--- a/weed/format/format.go
+++ b/weed/format/format.go
@@ -47,6 +47,12 @@ type Hint struct {
 // discovered by type assertion.
 type Format interface {
 	Name() string
+}
+
+// Sniffer cheaply recognizes the format from identification signals. Repack
+// gates on it before parsing, and policy-driven detection will rely on it;
+// ingest-only adapters, whose files carry their layout from birth, skip it.
+type Sniffer interface {
 	Sniff(h Hint) bool
 }
 
--- a/weed/format/format.go
+++ b/weed/format/format.go
@@ -18,6 +18,10 @@ import (
 // x-seaweedfs- prefix keeps it out of HTTP response headers.
 const LayoutKey = "x-seaweedfs-format-layout"
 
+// ViewParam is the query parameter selecting a format view on GET requests.
+// Adapters rendering self-referential URLs must use it.
+const ViewParam = "format.view"
+
 // ErrNoSuchView reports that a view request addresses nothing servable; the
 // server answers 404.
 var ErrNoSuchView = errors.New("no such view")
--- a/weed/format/formattest/formattest.go
+++ b/weed/format/formattest/formattest.go
@@ -1,0 +1,64 @@
+// Package formattest is the conformance kit format adapters must pass:
+// indexers parse attacker-controlled bytes inside a storage daemon, so they
+// must never panic and every layout they accept must validate.
+package formattest
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/seaweedfs/seaweedfs/weed/format"
+)
+
+// IndexTruncations feeds progressively truncated copies of a valid file to the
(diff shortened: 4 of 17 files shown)


chrislusf:
Addressed the review findings:

- The chunk-leak reports (greptile, coderabbit) had one root cause: doMaybeManifestize returned only the flat data chunks on a mid-run merge failure, dropping manifests already written. It now returns the written manifests plus the not-yet-wrapped remainder - a complete, deletable set - so the ingest/repack cleanup deletes everything, and AppendToEntry (which keeps the returned list after logging the error) no longer loses wrapped chunks. Covered by a new test. The separate suggestion to delete both slices in the handlers is unnecessary with this fix: DeleteUncommittedChunks resolves manifest chunks to their data chunks.
- formatChunkSizeLimit now floors at 4MB when neither the query nor -maxMB provides a positive limit; extent chunks are buffered in memory, so absent must not mean unlimited. The int32 narrowing warnings are noise: ParseInt with bitSize 32 already bounds the value.
- formatRepack now closes its ChunkReadAt, releasing the private reader cache.
- Added the missing chunk-count assertion in the align-clamp cutter test.
- Dropped the arithmetic capacity hint flagged by CodeQL.

Not changed: preserving Crtime on ingest overwrite is already handled - filer.CreateEntry finds the existing entry and routes through UpdateEntry, which carries the old Crtime forward.

chrislusf:
All three fixed:

- Repack now forces the entry's TtlSec onto the storage option before uploading, so chunk lifetime always matches the metadata; a ?ttl= query cannot desynchronize them.
- Repack emits NotifyUpdateEvent after the store update, mirroring the gRPC UpdateEntry path, so sync/backup/replication subscribers learn the new chunk ids.
- The cutter no longer materializes interior cuts: it keeps only the extent start offsets (bounded by MaxExtentCount) and derives each cut arithmetically, so a sidecar declaring a petabyte extent costs O(extents) memory, not O(cuts). Covered by a new test probing a 1<<50-byte extent.

chrislusf:
Took all five nitpicks, two of them differently than suggested:

- The query parameters are now format.ingest, format.repack and format.view, following the mv.from/cp.from dotted convention, so pass-through clients sending a bare ?format= or ?view= can no longer collide with these endpoints. Requests naming both ingest and repack get a 400. The PR description and the playlist URLs the hls-ts view renders are updated to match.
- Instead of naming the two positional bools, I removed one: the small-content permission is now derived from the boundary source itself (fixed-size chunking allows it, meaningful boundaries never do), so no call site can silently re-enable inlining.
- View responses now state Accept-Ranges: none; they always answer with whole documents or whole extents. Full range support on extents can come later if a real client needs it.
- Validate now bounds the format name to the same 256 bytes DecodeLayout enforces, so nothing encodes that cannot decode.
- hls-ts IndexSidecar validates its layout before returning it, making the formattest invariant enforced rather than emergent.

chrislusf:
All five taken:

- Repack rejects S3-versioned entries (entry.IsS3Versioning) - object versions can share one chunk list, and repack's old-chunk deletion would corrupt siblings. Stays rejected until chunk ownership is tracked.
- Plain HTTP overwrites now commit under the per-path entry lock, same as gRPC writers, renames, and repack (which holds it exclusively for its whole read-index-rewrite-swap span). The lost-update interleaving is gone; the full weed/server suite passes with the added locking.
- TTL'd entries repack with their remaining lifetime (Crtime + TtlSec - now), and an already-expired entry is rejected, so needle expiry stays aligned with entry expiry instead of restarting.
- View responses derive their validator from the media ETag plus the encoded layout plus the request's view parameters, and preconditions check against that, so re-ingesting identical bytes with a different sidecar invalidates cached playlists and segments instead of feeding 304s.
- The media-sequence decode bound now mirrors the ingest bound (MaxInt64-(count-1)); a boundary playlist round-trips through ingest, playlist render, and last-segment lookup in a new test.

chrislusf:
Dropped the S3-versioned guard: the one path that ever shared chunks across versions was the metadata-only self-copy fixed in #10594, so every version owns its chunk list and repack on a versioned entry only touches that version.

chrislusf:
Addressed all five:

- WORM is revalidated under the commit lock. Repack moved its check inside its lock; ingest and saveMetaData recheck at commit time, so a WORM enable landing during an in-flight upload now wins.
- Remaining TTL rounds up to the smallest volume-representable value (roundUpToVolumeTTL, unit-tested incl. round-tripping through SecondsToTTL/ReadTTL): 3599s becomes a full hour rather than 59m, and sub-minute remainders become 1m rather than permanent chunks.
- The expiry anchor now mirrors FindEntry exactly: S3-expiring entries age from Mtime via GetS3ExpireTime, everything else from Crtime. A recently overwritten S3 object gets its full remaining lifetime.
- Layouts are bound to an md5 of the chunk list they described (x-seaweedfs-format-layout-chunks). Offset writes, appends, or mount writes change the chunks while preserving Extended; views now detect the mismatch and 404 until re-ingest or repack, covering every mutation path rather than just the HTTP ones.
- On HA: agreed the entry lock is filer-local. I scoped the claim honestly in the code comments and PR description - single-filer serialization is what this PR provides, and cross-filer safety needs owner routing like the S3 gateway's route-by-key, which is a follow-up rather than something to half-build here. Repack and ingest are explicit operations, so routing them to one filer is an operational rule until then.

chrislusf:
All three:

- Repack now verifies against the store at commit time: it re-reads the entry under its local lock right before the swap, compares the chunk fingerprint, revalidates WORM, and answers 409 Conflict on any change. The new entry is built from that fresh read, so a concurrent metadata-only update on another filer is carried forward rather than clobbered. The unguarded cross-filer window shrinks from the whole repack (seconds to minutes on large files) to the swap itself; closing it entirely still needs owner routing, which stays the named follow-up.
- The fingerprint now digests every read-relevant chunk field: offset, file id, size, modification timestamp, cipher key, compression, manifest status, and SSE type. The truncate-then-sparse-extend sequence is caught by the size and timestamp fields; tests cover size, timestamp, and manifest-flag mutations.
- An unrepresentable remaining TTL (the narrow band above 68 years) now stores chunks with no volume TTL instead of a shortened one: they outlive the entry rather than predecease it, and the test asserts the round-trip for every representable value with no skipped case.

chrislusf:
Widened the conflict check to everything the repack consumed: the commit now digests the chunk fingerprint plus file size, TtlSec, Crtime, Mtime, the S3-expiry flag, and hard-link/remote state (repackSourceIdentity, table-tested per field), and answers 409 when any of it moved. A TTL cleared on another filer while chunks stayed put now conflicts instead of producing a permanent entry over expiring chunks - the swap can never pair fresh metadata with chunks built from stale inputs.

On the final check-to-swap atomicity: I looked at both suggested closures. A store-backed CAS does not exist across the pluggable filer stores, and the existing gRPC UpdateEntry write-condition is itself enforced under the same filer-local lock, so it adds nothing cross-filer. That leaves owner routing as the only real closure, and I would rather build it properly (the S3 gateway's route-by-key shape, one PR, covering ingest and repack and plain HTTP writes together) than half-embed it here. With the full-state conflict in place, the residual microsecond window no longer lets repack revert changes it observed - inside it, repack races exactly like any two concurrent HTTP writers do on a multi-filer deployment today. If you want owner routing in this PR rather than as the follow-up, say so and I will add it.

chrislusf:
Took the reject option: format entries are never written with inline content (ingest disables the small-content path, repack requires chunks), so nonempty Content is now disqualifying everywhere - views answer 404 stale, repack rejects it up front (its reader only sees chunks and would have dropped the inline bytes), the extent path's Content-serving branch is removed, and repackSourceIdentity digests Content so it conflicts if set concurrently during a repack. Tested via the identity table.

chrislusf:
Guarded both conversion sites: repack answers 400 and views answer 404 stale when entry.FileSize exceeds MaxInt64, so the wrap can no longer size a negative sniff buffer or slip past Validate's skip-when-negative contract. The viewer Object construction sits behind the same view guard.

Dimonyga:
I like this generalized format-adapter approach much more than my original HLS-specific implementation.

One thought: could we take the `repack` idea a bit further and make it a generic **post-upload rebuild** operation, instead of having special ingest paths for formats that need a sidecar?

For example, the workflow for HLS-TS could be:

1. Upload `video.ts` normally, exactly like any other filer object.
2. Later call something like:

```text
POST /videos/movie.ts?format.rebuild=hls-ts
```

and provide the HLS playlist as the sidecar/index.
3. The adapter derives the desired layout from the playlist.
4. SeaweedFS reads the already stored object, rewrites its chunks according to that layout, and atomically replaces the chunk list.

This is basically what `format.repack=parquet` already does, except that HLS needs an external index to describe the desired extents.

So conceptually the generic operation could be:

```text
stored object + format adapter + optional sidecar
    -> Layout
    -> rebuild chunks
    -> atomically replace entry
```

Parquet would derive the layout from the object itself, while HLS-TS would derive it from the supplied playlist.

I think this has a few nice properties:

* normal uploads remain completely format-agnostic;
* there is no special upload protocol for HLS or future formats;
* existing files can be rebuilt after the fact, not only at ingest time;
* rebuild can be retried independently if parsing/repacking fails;
* adapters only need to describe the desired layout, while the filer owns the generic rewrite/swap logic;
* future formats can use either an internal index, an external sidecar, or potentially both.

It also means we don't need to know the format or have the playlist at the moment the object is uploaded. Upload first, optimize its physical layout later.

Maybe `repack` already has almost all of the required machinery and could simply be generalized into `rebuild`/`repack` with an optional sidecar, rather than keeping `format.ingest` as a separate path.

What do you think?

Dimonyga:
Another interesting consequence of making rebuild/repack generic is that it opens the door to **automatic layout optimization based on actual access patterns**.

Once the filer has a generic mechanism that can take an existing object, derive a new layout, and rebuild its chunks, the decision to run it no longer has to come from the user.

For example, SeaweedFS could analyze request logs / range access patterns and detect objects where the current physical chunk layout is inefficient:

* many repeated reads of the same byte ranges;
* ranges consistently crossing chunk boundaries;
* high read amplification;
* excessive chunk/volume fan-out for a logical read;
* hot regions inside very large objects;
* known formats where access patterns strongly suggest a better structural layout.

A background optimizer could then identify the format, generate or derive an appropriate layout, and schedule a rebuild automatically.

Conceptually:

```text
request/access log
    -> detect inefficient access pattern
    -> identify format / derive candidate layout
    -> estimate expected benefit
    -> rebuild/repack object
    -> observe resulting read amplification
```

That makes the format adapters more than just an ingest feature. They become a generic storage optimization framework.

It could also be a very interesting commercial feature: automatic storage-layout optimization could be offered as part of a paid SeaweedFS subscription/support tier. The core format/rebuild mechanism remains useful and generic, while the more advanced policy engine, workload analysis, recommendations, and automatic optimization can live on top of it.

In other words, instead of requiring users to know that a particular file should be repacked, SeaweedFS could eventually learn from the workload and optimize the physical representation itself.

PR #10683, opened 2026-08-10
volume: decode IPv6 transition addresses in the remote-endpoint guard

chrislusf:
## What

`FetchAndWriteNeedle` validates a caller-supplied S3 endpoint with `validateRemoteEndpoint` / `checkBlockedIP` so a volume server won't dial loopback, link-local, RFC 1918, CGNAT, or metadata addresses. That guard normalizes only `::ffff:`-mapped IPv4. IPv6 transition addresses that embed an internal IPv4 in another form slipped through:

| form | prefix | example | decodes to |
|---|---|---|---|
| NAT64 | `64:ff9b::/96` | `64:ff9b::a9fe:a9fe` | 169.254.169.254 |
| 6to4 | `2002::/16` | `2002:a00:1::` | 10.0.0.1 |
| Teredo | `2001:0000::/32` | `2001:0:4136:e378:8000:63bf:80ff:fffe` | 127.0.0.1 |
| IPv4-compatible | `::/96` | `::7f00:1` | 127.0.0.1 |

The plain IPv4 forms are all refused, but these encodings passed both the up-front check and the dial-time guard. On a network with the matching relay (NAT64 is common in IPv6-only cloud), the address routes to the embedded internal host.

## Change

`checkBlockedIP` now extracts the embedded IPv4 from those four transition forms and re-checks it against the same deny list, after the existing checks so `::` / `::1` still match unspecified / loopback first. A transition address that embeds a public IPv4 still passes, so legitimate destinations are unaffected. Mirrored in the Rust volume server (`seaweed-volume`).

Tests cover each form decoding to an internal target (rejected) and one decoding to a public target (allowed), in both Go and Rust.

## Summary by CodeRabbit

* **Security Improvements**
  * Enhanced remote endpoint validation to detect and block IPv6 transition addresses that resolve to private or loopback IPv4 destinations.
  * Added coverage for NAT64, 6to4, Teredo, and IPv4-compatible address formats.
  * Public transition addresses continue to be accepted when safe.

### DIFF
--- a/seaweed-volume/src/remote_storage/endpoint_guard.rs
+++ b/seaweed-volume/src/remote_storage/endpoint_guard.rs
@@ -17,7 +17,7 @@
 //! narrow TOCTOU window (a hostname that resolves to a public IP here and then
 //! flips to a blocked one when the SDK dials) remains as a follow-up.
 
-use std::net::{IpAddr, Ipv4Addr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr};
 
 /// AWS/Azure/GCP IPv4 instance-metadata-service (IMDS) address. It is
 /// link-local and thus already covered by [`is_link_local`], but is named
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed the CodeRabbit review in 6f05a316d:

- NAT64 detection now requires the full `64:ff9b::/96` well-known prefix (bytes 4-11 zero) before decoding the low 32 bits, in both Go and Rust. Other `64:ff9b:` prefixes embed the IPv4 elsewhere, so they're left untouched rather than mis-decoded. Added a case asserting `64:ff9b:1::a9fe:a9fe` is not decoded.
- Added public-target (8.8.8.8) allow cases for 6to4, Teredo, and IPv4-compatible alongside the existing NAT64 one, so each decoder is exercised on both a blocked and an allowed destination.

6to4 (`2002::/16`), Teredo (`2001:0000::/32`), and IPv4-compatible (`::/96`) already match their complete prefixes, so no change was needed there.

PR #10689, opened 2026-08-10
feat(shell): support parallel volume moves in volume.balance

kmlebedev:
## Description

This is a follow-up to [PR #10689](https://github.com/seaweedfs/seaweedfs/pull/10689).
It keeps the opt-in `volume.balance -maxParallelization` feature, but removes the
separate source/target selection implementation from the sequential and parallel
paths.

## What changed

- Add the small, policy-neutral `weed/topology/balancer` node-load selector. It
  chooses the most/least loaded eligible node deterministically, with node ID as
  the tie-breaker.
- Reuse that selector in `ecbalancer` for global EC balancing, rack target
  selection, and per-volume node selection. EC-specific shard, rack, disk, and
  replica-placement rules remain in `ecbalancer`.
- Add one `volumeBalancePlan`/`planBalanceMove` path for `volume.balance`.
  Sequential mode applies the plan after a successful move; parallel mode
  reserves the same plan under the existing mutex before starting the RPC.
- Keep replica placement checks, physical-disk high-water protection,
  `volumesPerExec`, reservation rollback, and target blacklisting unchanged.
- Keep the existing verbose error output before blacklisting a target on
  `ErrVolumeNoSpaceLeft`. The review finding from the earlier patch is already
  fixed in PR #10689, so no duplicate logging was added.

The full EC planner is intentionally not reused for normal volumes: EC shards
have data/parity, rack, disk, and shard-bit semantics that do not apply to a
regular volume. The shared selector is the common balancing primitive while
those domain-specific policies stay isolated.

## Usage

```text
volume.balance -apply -maxParallelization=8
```

The default remains `-maxParallelization=1`, preserving sequential behavior.

## Validation

```text
go test ./weed/topology/balancer ./weed/storage/erasure_coding/ecbalancer ./weed/shell -count=1
go test -race ./weed/topology/balancer ./weed/storage/erasure_coding/ecbalancer ./weed/shell -run 'Test(PickNodeByLoad|BalanceParallel|BalanceDoesNotDrainOntoOneNode|BalanceSkipsPhysicallyFullTarget|Plan)' -count=1
```

Both pass. `go vet` reports the pre-existing protobuf lock-copy warning in
`weed/shell/command_volume_delete_empty_test.go`; it is unrelated to this
change.

## Related work

- [Current volume balance implementation](https://github.com/seaweedfs/seaweedfs/blob/master/weed/shell/command_volume_balance.go)
- [EC balance command](https://github.com/seaweedfs/seaweedfs/blob/master/weed/shell/command_ec_balance.go)
- [Previous parallel balance attempt #6222](https://github.com/seaweedfs/seaweedfs/pull/6222)
- [Parallel replication fix #6789](https://github.com/seaweedfs/seaweedfs/pull/6789)
- [Current implementation in PR #10689](https://github.com/seaweedfs/seaweedfs/pull/10689)

## Summary by CodeRabbit

* **New Features**
  * Added configurable parallel volume balancing, while preserving sequential operation by default.
  * Improved balancing decisions for more even distribution while respecting capacity, placement, and replica-safety requirements.
* **Bug Fixes**
  * Improved handling of failed moves, unavailable targets, and insufficient storage.
  * Prevented duplicate volume placement during balancing.
  * Improved deterministic selection when nodes have equal load scores.


chrislusf:
This is in the right direction. My concern is that the volume balancing and ec volume balancing admin workers have more logic on candidate selection, move accounting, and inflight reservation. Needs more refactoring work to reuse the logic. Adding this would complicating the effort.

kmlebedev:
> This is in the right direction. My concern is that the volume balancing and ec volume balancing admin workers have more logic on candidate selection, move accounting, and inflight reservation. Needs more refactoring work to reuse the logic. Adding this would complicating the effort.

@chrislusf done and tested in production

chrislusf:
The refactoring mentioned above is now up as #10727: the volume-move and EC-shard-move sequences are extracted into a shared, unit-testable `weed/operation/volume_move` package used by both the shell commands and the maintenance workers, with pre-delete verification and cleaned-up failure/rollback semantics. Once that merges, this PR can rebase into a much thinner change: plan the moves under the existing mutex and execute `Mover.LiveMoveVolume` through `ErrorWaitGroup`, the same way `ec.balance -maxParallelization` already fans out.

PR #10719, opened 2026-08-12
ec: confirm a surviving copy before deleting a duplicate EC shard

chrislusf:
The dedup phase of EC balancing deletes a shard it believes exists elsewhere. It copies nothing first, so the shard surviving on another node is the *only* thing that makes the delete safe — and it took the plan's word for that:

```go
func (t *ECBalanceTask) executeDedupDelete(...) error {
    t.unmountShard(...)   // no copy
    t.deleteShard(...)    // no check that any other copy exists
}
```

The plan comes from the master's topology, and the topology can name a location that holds nothing — such a server answers `CopyFile not found ec volume id N` when something later tries to read the shard there. A shard listed on a phantom location *and* a real one looks duplicated, so dedup deletes one. When it picks the real one, the last copy is gone and **the job returns success**. The loss only surfaces much later, as a rebuild that cannot assemble enough shards:

```
ec.rebuild -collection <c> -apply
missing shard 42561.3
missing shard 42561.6
failed to copy 42561.11 ...: CopyFile not found ec volume id 42561
failed to copy 42561.12 ...: CopyFile not found ec volume id 42561
error: rebuildEcFiles: ec rebuild: input shard 1 size 3146776576 != 3125805056
```

The move phase already refuses to work on trust — it verifies the shard registered on the destination before removing the source, and there is an explicit guard rejecting same-node cross-disk moves because the delete "would erase the shard after a skipped copy". Dedup is the more dangerous of the two, since it deletes without copying at all, and had none of that.

## The change

- `detectDuplicateShards` records the node it chose to keep (`Move.KeepNode`).
- It is carried to the worker in `EcBalanceTaskParams.dedup_keep_node` (new field 5, additive).
- Both executors verify before deleting: the worker task (`verifyShardsOnKeepNode`) and the shell's `ec.balance` (`verifyEcShardOnKeepNode`), reusing the existing `VerifyShardsAcrossServers` primitive that the move path already uses.
- A keep node that cannot be queried is treated as **unknown, not confirmed**, and blocks the delete — otherwise a network blip is enough to lose a shard.
- An empty keep node also blocks, so a task from an older planner cannot delete unverified.

Only `worker.pb.go` is regenerated; `worker_grpc.pb.go` was left alone since the service definitions are unchanged and regenerating it only produced tool-version churn.

## Tests

`fake_ec_volume_server_test.go` adds an in-process volume server implementing the EC subset, which tracks **what is actually on disk** separately from what the plan claims — the distinction the bug turns on. `dedup_safety_test.go` drives the real destructive path against it:

| test | without the guard |
|---|---|
| topology claims two locations, only one has the shard, dedup targets the real one | **FAIL** — deleted the only copy, `err=<nil>` |
| shard genuinely on two servers | PASS — duplicate removed, survivor kept |
| keep node unreachable | **FAIL** — deleted anyway |

Verified by reverting just the guard: the first and third fail exactly as above, and pass with it restored. The middle one guards against "fixing" this by disabling dedup.

Found while investigating a cluster where 227 EC volumes had fallen below the reconstruction threshold, with the master still advertising shards that were on no disk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented erasure-coded deduplication from deleting the only confirmed copy of a shard.
  * Deduplication now verifies that the retained copy is reachable, belongs to the correct collection, and contains the required shard before removing duplicates.
  * Safely aborts operations when the retained node is missing, unreachable, unverifiable, or lacks the shard.
  * Preserved existing dry-run behavior while improving safety checks for applied operations.


chrislusf:
Both addressed in 36ec48fa7.

**Collection not preserved during verification** (@coderabbitai) — real, and fixed. I checked how far it goes: `DiskLocation.ecVolumes` is `map[needle.VolumeId]*EcVolume`, so a *single* node cannot hold volume N for two collections — the inventory it returns for N is unambiguous locally. The exposure is across nodes: the keep node could hold a different collection's volume N, and the check would approve on that basis.

Rather than widen the RPC and update every caller, I used what the response already carries — `EcShardInfo.Collection`, populated by the server via `ToEcShardInfo`. The check now lives in a new `erasure_coding.VerifyShardsOnServer(ctx, collection, volumeID, server, shardIDs, dialOption)`, shared by both executors so they cannot drift. `VerifyShardsAcrossServers` and its other callers (including the pre-existing `verifyShardsOnDestination`) are untouched; tightening those is a separate change with a wider blast radius.

Regression test added as suggested: `TestDedupRefusesWhenTheKeepNodeHoldsAnotherCollection` stands up two servers holding volume 42561 shard 0 in *different* collections. Verified it is not vacuous — with the collection comparison removed it fails with `dedup deleted shard 42561.0 after matching another collection's volume (err=<nil>)`.

**Unbounded keep-node inventory request** (@greptile-apps) — correct, and only the shell path. The worker task already inherits a deadline from `EcBalanceTaskParams.TimeoutSeconds`; `verifyEcShardOnKeepNode` used `context.Background()`, so a node that accepts the connection without answering would stall the whole `ec.balance` run rather than report an unconfirmable survivor. Now bounded by `ecShardVerifyTimeout` (30s), and a timeout blocks the delete — same treatment as an unreachable node, since neither is confirmation.

All four dedup tests pass; `erasure_coding`, `ecbalancer`, `ec_balance` and `shell` suites are green.

Issue #9212, opened 2026-04-24
EC volumes with shards distributed across multiple disks on same node are not fully registered with master on startup


omorfley666:
**SeaweedFS version:** 4.20 50f25bb5c

**Environment:**
- 4 volume server pods, each with 4 physical disks mounted at `/data1`, `/data2`, `/data3`, `/data4`
- EC encoding: 10+4 (14 shards total)
- Kubernetes deployment

**Description**

When an EC volume has its shards distributed across multiple disks on the same volume server node, the master only registers a subset of those shards upon startup. The remaining shards are physically present on disk but invisible to the master.

At startup, the volume server logs show the following message for affected volumes:

```
EC volume 15: distributed EC (.dat removed) with 2 shards
EC volume 15: distributed EC (.dat removed) with 1 shards
EC volume 15: distributed EC (.dat removed) with 3 shards
```

The message appears multiple times per volume - once per disk that contains shards of that volume. The shard count increments with each disk but the master ultimately only registers a partial count.

**Example - volume 15:**

Physical state on disks:
- `volume-0 /data2`: ec00
- `volume-1 /data1`: ec13
- `volume-1 /data3`: ec03
- `volume-2 /data1`: ec02, ec06, ec10
- `volume-2 /data2`: ec01, ec04, ec09
- `volume-2 /data4`: ec08, ec12
- `volume-3 /data2`: ec05
- `volume-3 /data4`: ec07, ec11

All 14 shards (ec00–ec13) are physically present across the 4 nodes. However, the master reported only 6 shards visible:

```
volume-0: shards [0]
volume-1: shards [3]
volume-2: shards [8, 12]
volume-3: shards [7, 11]
```

This caused the volume to appear as `Critical` in the master topology despite all data being intact on disk.

`ec.rebuild` dry-run confirmed the volume as unrepairable:
```
ec volume 15 is unrepairable with 6 shards (need 10), skipping
```

**Workaround we applied**

We manually moved all shards of each affected volume onto a single disk per node (consolidating to `/data1`), then restarted the volume server pod. After restart, the volume server correctly registered all shards from that single disk location and reported them to the master.

Before consolidation - volume 15 visible to master: **6 shards**
After consolidation and restart - volume 15 visible to master: **11 shards**

The consolidation required also copying the `.vif` file alongside the `.ec??` shards, as without `.vif` the shards were not registered at all.

[...]

PR #10721, opened 2026-08-12
test: walk an EC volume through a multi-disk node's whole life

chrislusf:
A multi-disk volume server keeps one `.ecx` / `.ecj` / `.vif` set per volume on a single disk while `ec.balance` scatters the shards across the others. Every EC operation on such a node crosses that split, and each transition is handled by a **different** mechanism:

- startup registration on the sidecar disk — per-disk scan (`loadAllEcShards`)
- startup registration on the other disks — cross-disk reconcile (`reconcileEcShardsAcrossDisks`, #9212)
- a shard arriving on a disk with no local sidecars — mount-time `.ecx` lookup (`findEcxIdxDirForVolume`)
- deleting the sidecar disk's shards — the node-wide pass-2 sidecar guard in `VolumeEcShardsDelete`

Each mechanism has its own test, but nothing exercised them as the **sequence a production node actually lives through**, where the output state of one transition is the input of the next. A regression in any hop has the same signature: shards that exist on disk while the master's view says otherwise — and every topology-driven repair (`ec.rebuild`, `ec.balance`, dedup) then works against the wrong shard set. That signature is exactly what a recent support case looked like from the outside, which is where this layout, volume id and collection come from.

`TestMultiDiskLifecycle_SpreadLayout` walks one volume through: cold start over a spread layout (6/5/3 across three disks, sidecars only on the first) → balance the sidecar disk's shards away (sidecars must stay: eight shards still read through them) → **reboot in that state** (sidecar-only disk + shard-only disks, the steady state of a balanced node) → a fresh shard landing on a shard-only disk and mounting cross-disk. After every hop it asserts both directions: what is registered matches what should be, and what is on disk was not touched.

`TestMultiDiskLifecycle_SidecarDiskLost` pins the failure floor when the sidecar disk itself dies: the surviving disks' shards may drop out of the registered view — nothing can read them without the `.ecx` — but their files must survive untouched, so restoring the sidecars (or the disk) restores the volume. Losing one disk must never cost the other disks' shards their data. Today the registered view in that state is empty node-wide; the test documents that and will catch either a fix (shards become visible) or a regression (files start disappearing).

Both tests pass on master — this is regression coverage for a class of state transitions that currently only holds together because three separate mechanisms agree with each other.

## Summary by CodeRabbit

* **Tests**
  * Added coverage for multi-disk erasure-coded volume lifecycles, including cold starts, balancing, restarts, and mounting newly available shards.
  * Verified shard files remain intact when sidecar disks are unavailable.
  * Added validation for degraded visibility when a sidecar disk is lost without deleting surviving shard data.

### DIFF
--- a/weed/storage/store_ec_multidisk_lifecycle_test.go
+++ b/weed/storage/store_ec_multidisk_lifecycle_test.go
@@ -1,0 +1,273 @@
+package storage
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/volume_server_pb"
+	"github.com/seaweedfs/seaweedfs/weed/stats"
+	"github.com/seaweedfs/seaweedfs/weed/storage/erasure_coding"
+	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
+	"github.com/seaweedfs/seaweedfs/weed/storage/types"
+	"github.com/seaweedfs/seaweedfs/weed/storage/volume_info"
+	"github.com/seaweedfs/seaweedfs/weed/util"
+)
+
+// These tests walk one EC volume through the life a multi-disk production node
+// gives it: shards spread across disks with the .ecx / .ecj / .vif sidecars on
+// only one of them, then a balance moving the sidecar disk's shards away, then
+// a restart, then a fresh shard arriving. Each transition is exercised against
+// the real Store so the invariant under test is the one production depends on:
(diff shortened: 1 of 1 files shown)


chrislusf:
These landed after merge; addressed in follow-up #10723 (assert exact per-disk shard identity at every phase instead of counts, assert `countRegistered == 0` on sidecar-disk loss, and defer the first store's closer so an early Fatalf no longer leaks its goroutine).

PR #10722, opened 2026-08-12
test: drive ec.encode/balance/rebuild E2E with a byte-identical payload check

chrislusf:
The existing multi-disk EC integration test (`TestMultiDiskECBalanceNoShardLoss`) asserts on **shard counts**. Counting cannot distinguish a healthy volume from one a repair reassembled out of the wrong inputs — both have fourteen shards. The whole risk with mixed-generation or phantom shards is a rebuild that *succeeds* and produces fourteen plausible-but-wrong shards; a count check waves that through.

This drives the real shell commands against a live 3-node × 4-disk cluster and **reads the stored bytes back after every step**, so a rebuild from the wrong inputs fails here where a count check would pass.

An 8 KB random payload (no structure, so wrong bytes can't accidentally match) is stored, then:

- **`ec.encode`** → 14 shards, payload verified
- **`ec.balance`** → still 14, payload verified
- **damage** → two shard files removed straight off disk, then the volume servers restarted so the master relearns the reduced set from disk; payload still readable from the surviving 12
- **`ec.rebuild -apply`** → back to 14, payload verified

The rebuild output matches the shape of the support case that motivated this:

```
rebuildOneEcVolume test 3
missing shard 3.0
127.0.0.1:8090 copied 3.1 from 127.0.0.1:8091
missing shard 3.2
127.0.0.1:8090 copied 3.3 from 127.0.0.1:8091
...
```

Two ordering facts the test pins, both of which surfaced while writing it and both of which cost real debugging time:

1. **`ec.rebuild` is driven by the master's topology, not disk truth.** Removing shard files without a rescan leaves the master listing all 14, and `ec.rebuild` then prints `rebuildEcVolumes for "test"` and does nothing — no `rebuildOneEcVolume`, no error. The volume has to be relearned (here, via a volume-server restart) before a repair can target the right set. An operator who removes files and immediately runs rebuild sees a clean no-op and assumes success.
2. **The shell lock is dropped when the restart disconnects the master**, so it must be retaken before the rebuild.

Runs in ~45s, gated on `testing.Short()` like the sibling test, and stable across repeated runs. The final assertion is that the master's EC shard view and disk truth agree on all 14 after the rebuild — the rebuilt shards are real, not phantom registrations.

## Summary by CodeRabbit

* **Tests**
  * Added end-to-end validation for multi-disk erasure-coding workflows.
  * Verifies encoding, balancing, shard recovery after simulated disk damage, and data integrity across volume-server restarts.
  * Confirms shard metadata remains consistent with on-disk data throughout the lifecycle.

### DIFF
--- a/test/erasure_coding/multidisk_shell_lifecycle_test.go
+++ b/test/erasure_coding/multidisk_shell_lifecycle_test.go
@@ -1,0 +1,410 @@
+package erasure_coding
+
+import (
+	"bytes"
+	"context"
+	"crypto/rand"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+	"google.golang.org/grpc"
+
+	"github.com/seaweedfs/seaweedfs/weed/operation"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
+	"github.com/seaweedfs/seaweedfs/weed/shell"
+	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
+)
+
+// TestMultiDiskECShellLifecycle drives the real shell commands — ec.encode,
+// ec.balance, ec.rebuild — against a live multi-disk cluster, through the
+// sequence a production repository goes through, with one invariant checked
+// after every step: the bytes a client stored come back identical. Shard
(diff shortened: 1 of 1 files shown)


chrislusf:
Addressed in 96ef97a02.

**Unbounded payload verification requests** (@greptile-apps) — correct. `readFid` used the default `http.Client`, so a master or restarting volume server that accepted the connection and then stalled would block past the 30s `Eventually` deadline and the 360s test context. Both requests (the master lookup and the location GET) now go through a shared `http.Client{Timeout: 10s}`, so a slow location fails and is retried rather than hanging the run. Re-ran the full test: still passes (~42s).

PR #10724, opened 2026-08-12
[shell] volume.balance io byte per second

kmlebedev:

To avoid 100% disk utilization,

We limit the data transfer limit when moving

# How is the PR tested?

reuse option from volume.move

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Volume balancing now supports configurable parallel operations and I/O bandwidth limits.
  * Balancing more evenly distributes volumes while preserving replica placement and preventing unsafe disk usage.
  * Improved handling of failed moves and rollback during balancing.

* **Bug Fixes**
  * Enhanced volume selection and load balancing for more predictable results.
  * Improved erasure-coded data balancing across nodes and racks.

* **Tests**
  * Added coverage verifying parallel balancing preserves all volumes and achieves an even distribution.

### DIFF
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -512,21 +512,22 @@ func selectVolumesByActive(volumeSize uint64, volumeByActive *bool, volumeSizeLi
 	}
 }
 
-func (c *commandVolumeBalance) balanceSelectedVolume(diskType types.DiskType, volumeReplicas map[uint32][]*VolumeReplica, nodes []*Node, sortCandidatesFn func(volumes []*master_pb.VolumeInformationMessage)) (err error) {
-	if c.maxParallelization > 1 {
-		return c.balanceSelectedVolumeParallel(diskType, volumeReplicas, nodes, sortCandidatesFn)
-	}
-
-	selectedVolumeCount, volumeCapacities := uint64(0), float64(0)
-	var nodesWithCapacity []*Node
-	volumeSizeLimitMb := c.volumeSizeLimitMb
+// planBalance computes the state shared by both the sequential and parallel
+// balancing paths: the nodes that have spare capacity, the density function
+// used to rank them, and the ideal volume ratio this pass is aiming for.
+// ok is false when there is no capacity to balance against (volumeCapacities
+// == 0); callers should return nil without doing any work in that case.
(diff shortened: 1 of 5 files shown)


chrislusf (commented):
Rebased onto master (clean, no conflicts — pushed to the branch; the merge commit is flattened into the 6-commit series). Build and `weed/shell` + `ecbalancer` + `topology/balancer` tests pass on the rebase. Review:

**Verified sound:**
- `rollbackParallelMove` is a faithful inverse of `adjustAfterMove` (selectedVolumes, replica location, per-disk VolumeInfos/count/free-bytes all symmetric).
- The reserve-under-mutex design is right: each reservation updates the planner state before the next plan is computed, so parallel reservations can't double-pick a volume or overcommit a target; `failedTargets` reads/writes are all under `balanceMu`.
- The `NodeLoad`/`PickMostLoaded`/`PickLeastLoaded` refactor in `ecbalancer` is behavior-preserving: the old strict-inequality scans took the first node in sorted order on ties, and the id tie-break selects the same node.

**One bug to fix — sequential no-space handling regressed:** in the restructured `balanceSelectedVolume`, a `util.ErrVolumeNoSpaceLeft` failure hits `continue` in the `for hasMoved` loop while `hasMoved` is still false, so the loop condition ends the whole pass for that disk type. The old code's `continue` targeted the inner target loop and moved on to the next target. Simplest fix: give the sequential path the same `failedTargets` map the parallel path uses (pass it instead of `nil`, add the target on no-space, and keep looping) so one physically-full target doesn't stop balancing.

**Sequencing with #10727:** `moveVolume` still passes `v.ReadOnly` as `skipTailError` — fine against current master, but #10727 extracts the move sequence into a shared `volume_move.Mover` whose `LiveMoveVolume` drops that parameter (the already-readonly rule is internalized) and adds pre-delete verification and abort cleanup. I'd land #10727 first: the rebase here is then mechanical (drop the last argument), and the parallel executor inherits the verification/rollback safety for free. One nuance after that rebase: a `volume_move.ErrSourceKeptReadonly` failure means the cluster state is deliberately parked (source readonly) — the in-memory reservation rollback stays correct, but the run should stop rather than keep planning around that volume.

**Also:** this PR supersedes #10689 (same parallel machinery plus the IO limit and the planner unification) — worth closing #10689 to keep review in one place.

**Minor:**
- The per-move plan line (`"%s %.2f %.2f:%.2f\t"`) used to print unconditionally and is now verbose-only — intended?
- `planBalanceMove` pre-filters `eligibleVolumes` with `isGoodMove` for every candidate target on every reservation — O(targets × source volumes) per move. Fine at typical scale; may deserve a cap or memoization for 10k-volume sources.

PR #10737, opened 2026-08-13
feat(shell): parallelize volume balance moves

kmlebedev:

Follow-up to [PR #10689](https://github.com/seaweedfs/seaweedfs/pull/10689),
rebased on [PR #10727](https://github.com/seaweedfs/seaweedfs/pull/10727).

PR #10727 extracts the volume-move sequence into the shared
`weed/operation/volume_move` package. This change adds parallel
planning/reservation to `volume.balance` and reuses the actual `ec.balance`
fan-out model through a common task-group executor built on `ErrorWaitGroup`.

## What changed

- Add the opt-in `volume.balance -maxParallelization` flag; the default remains
  `1`, preserving sequential behavior.
- Use one balance loop for sequential and parallel operation. It reserves a
  batch of one move at parallelism `1`, or up to `maxParallelization` moves
  otherwise, then executes the batch with the same task-group executor.
- Plan and reserve each batch while holding the existing `balanceMu`.
  Reservations update the in-memory selected volumes, replica locations, and
  disk accounting before any move RPC starts, so concurrent workers cannot
  select the same volume or target state.
- Extract `executeParallelTaskGroups` from `ec.balance`: it runs tasks within
  each group serially, then fans independent groups out through
  `ErrorWaitGroup`.
- Make `ec.balance` pass one task group per EC volume, preserving its current
  protection against concurrent updates to a volume's `.ecx`, `.ecj`, and
  `.vif` sidecars.
- Make `volume.balance` pass one task group per already-reserved volume, and
  execute that task with `volume_move.Mover.LiveMoveVolume`. Thus both commands
  use exactly the same concurrency, throttling, and aggregated-error path.
- Remove the now-redundant split between sequential and parallel balance loops,
  the intermediate plan type, and the unused `attemptToMoveOneVolume` helper.
- Reuse the lock and remote-storage safety validation with volume-server
  evacuation; a failed reservation always restores the in-memory topology.
- Keep only the shell-level rollback needed for an unsuccessful reservation;
  copy/tail/delete/verification and remote cleanup remain in the shared
  `volume_move` package from #10727.
- Preserve `volumesPerExec`, replica-placement checks, physical-disk
  high-water protection, lock-loss handling, and target blacklisting on
  `ErrVolumeNoSpaceLeft`. The verbose error is printed before blacklisting.

No changes to `weed/operation/volume_move` are included in this PR; it should be
rebased after #10727 merges. The remaining changes are limited to the shell
executor, the two shell balance commands, and their tests.

## Usage

```text
volume.balance -apply -maxParallelization=8
```

## Validation

```text
go test ./weed/operation/volume_move ./weed/shell -count=1
go test -race ./weed/shell -run 'TestExecuteParallelTaskGroups|TestBalanceParallel|TestBalanceDoesNotMoveRemoteStorageVolumes|TestBalanceDoesNotDrainOntoOneNode|TestBalanceSkipsPhysicallyFullTarget|TestBalanceVolumesPerExec' -count=1
```

The focused tettps://github.com/seaweedfs/seaweedfs/pull/10689)sts pass. The full non-race package test also passes. A full
`go test -race ./weed/operation/volume_move ./weed/shell -count=1` is blocked by
the existing `TestFsMvMovesIntoExistingDestinationDirectory` attempt to bind a
Unix socket under `/var/folders`; the new balance tests pass under race.

## Related work

- [Current volume balance implementation](https://github.com/seaweedfs/seaweedfs/blob/master/weed/shell/command_volume_balance.go)
- [EC balance command](https://github.com/seaweedfs/seaweedfs/blob/master/weed/shell/command_ec_balance.go)
- [Shared move refactor #10727](https://github.com/seaweedfs/seaweedfs/pull/10727)
- [Previous parallel balance attempt #6222](https://github.com/seaweedfs/seaweedfs/pull/6222)
- [Parallel replication fix #6789](https://github.com/seaweedfs/seaweedfs/pull/6789)
- [Original parallel balance PR #10689](h

## Summary by CodeRabbit

* **New Features**
  * Volume balancing now supports configurable parallel moves for faster rebalancing.
  * Independent volumes can be balanced concurrently while preserving move order within each volume.
  * Balancing supports per-move I/O limits and idle timeouts.

* **Bug Fixes**
  * Remote-storage volumes are preserved on their source.
  * Failed moves are rolled back safely, with unavailable targets excluded and unrelated operations continuing.

### DIFF
--- a/weed/shell/command_ec_common.go
+++ b/weed/shell/command_ec_common.go
@@ -1026,19 +1026,19 @@ func (ecb *ecBalancer) executePhase(byID map[string]*EcNode, moves []ecbalancer.
 		}
 		byVol[m.VolumeID] = append(byVol[m.VolumeID], m)
 	}
-	ewg := NewErrorWaitGroup(ecb.maxParallelization)
+	taskGroups := make([][]ErrorWaitGroupTask, 0, len(order))
 	for _, vid := range order {
-		group := byVol[vid]
-		ewg.Add(func() error {
-			for _, m := range group {
-				if err := ecb.executeMove(byID, m); err != nil {
-					return err
-				}
-			}
-			return nil
-		})
+		movesForVolume := byVol[vid]
+		taskGroup := make([]ErrorWaitGroupTask, 0, len(movesForVolume))
+		for _, move := range movesForVolume {
+			move := move
+			taskGroup = append(taskGroup, func() error {
+				return ecb.executeMove(byID, move)
+			})
+		}
+		taskGroups = append(taskGroups, taskGroup)
 	}
-	return ewg.Wait()
+	return executeParallelTaskGroups(ecb.maxParallelization, taskGroups)
 }
 
 // verifyEcShardOnKeepNode confirms the node a dedup move chose to keep actually
--- a/weed/shell/command_volume_balance.go
+++ b/weed/shell/command_volume_balance.go
@@ -39,6 +39,7 @@ type commandVolumeBalance struct {
 	volumeByActive     *bool
 	applyBalancing     bool
 	volumesPerExec     int
+	ioBytePerSecond    int64
 	maxParallelization int
 	movedCount         int
 	byDiskUsage        bool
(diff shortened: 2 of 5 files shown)


chrislusf (commented):
Checked this against the other in-flight work — no duplicated effort: this correctly supersedes #10689/#10724 (both closed) and builds on the shared mover the way the #10724 review suggested; the reservation/rollback design (`failBalanceMove`/`rollbackBalanceMove`) matches what was verified sound there, and `executeParallelTaskGroups` unifying the ec.balance fan-out is a nice touch. It is also complementary to #10749 (throughput for EC shard copies, replicate, and worker-driven moves): together they cover the whole matrix, and the file overlap in `command_ec_common.go`/`common.go` is in different hunks, so whichever lands second rebases trivially.

Two things to fix:

1. **A kept-readonly failure can be misclassified as a benign no-space skip.** `failBalanceMove` classifies by `strings.Contains(err, ErrVolumeNoSpaceLeft)` — but the mover's abort path wraps the *original* error together with `volume_move.ErrSourceKeptReadonly` when it deliberately parks the source (e.g. the incomplete-target cleanup failed after a no-space copy). Such an error contains both signals; the current check blacklists the target and continues while the source volume sits readonly on purpose. Check `errors.Is(err, volume_move.ErrSourceKeptReadonly)` first and abort the run — that state needs the operator, not more planning.

2. **Help-text nit**: `-ioBytePerSecond` says "default 0 is unlimited", but 0 actually falls back to the volume server's own `-maintenanceBytePerSecond` (that's how `VolumeCopy` treats an unset limit; #10749 documents the same convention for the EC and worker paths). Worth matching the wording so operators don't assume unthrottled.

One note, not a request: the `NodeLoad`/`PickMostLoaded`/`PickLeastLoaded` shared-picker refactor from #10724 didn't carry over — it was behavior-preserving and fine; if you still want it, it'd make a good standalone PR.

PR #10738, opened 2026-08-13
EC worker: clear stale/interrupted shards at task start and on failure

chrislusf:
## Problem

The plugin-worker EC encode task cleared stale EC shards from a prior interrupted encode only at 55% progress (after mark-readonly, copy, and generate), using a **generation-fenced** teardown. Two cases left orphan shards behind:

1. A **retried** encode's prior attempt carries the *same* admin-issued `encodeTsNs`, and the server's teardown fence preserves same-or-newer generations — so the prior attempt's shards were never cleared.
2. Shards left by an **interrupted distribute** often have an *unreadable* `.vif` generation (the sidecar never landed), which the fence also preserves.

Both survive the next volume-server restart as orphans and make detection refuse the volume ("Manual intervention required").

## Fix

- **Step 0 preflight**: move the stale-shard cleanup to the very start (before mark-readonly), and switch it to the server's **blanket** (generation-independent) teardown — the same wipe the shell `ec.encode` pre-cleanup already uses. The admin dedupe key prevents a concurrent newer encode of the volume, and the blanket path aborts rather than clobber a live newer mount. The preflight also rejects a plan with no targets/source before touching the source.
- **`rollbackDistribute`**: a failure *after* distribute begins but *before* verify commits the EC copy now tears down the shards it wrote and restores the sources to writable — so a terminally-failed encode (a single-attempt job, or the last of a retry series, which has no successor preflight) leaves nothing behind. Not invoked past verify, where we are committed to the EC copy.

## Testing

New `ec_task_preflight_test.go` (preflight clears stale shards; rollback clears distributed shards and keeps the source; preconditions reject empty plans). Full package incl. the e2e encode tests pass. Also verified end-to-end on a live 5-node cluster: an interrupted-then-failed encode that previously stranded ~9 orphan shards for the target volume now leaves **0**, source intact.

## Summary by CodeRabbit

* **Bug Fixes**
  * Added preflight validation to prevent erasure-coding operations when targets or source replicas are missing.
  * Stale mounted shards are removed before encoding begins.
  * Failed distribution, mounting, or verification now rolls back partial changes while preserving source data.
  * Verified erasure-coded shards are retained if cleanup afterward fails.

* **Tests**
  * Added coverage for validation, stale-shard cleanup, rollback behavior, and source-volume preservation.

### DIFF
--- a/weed/worker/tasks/erasure_coding/ec_task.go
+++ b/weed/worker/tasks/erasure_coding/ec_task.go
@@ -961,6 +961,16 @@ func (t *ErasureCodingTask) ensureCleanEcStart(ctx context.Context) error {
 	if len(t.targets) == 0 {
 		return fmt.Errorf("no EC shard targets for volume %d; refusing to mark source readonly", t.volumeID)
 	}
+	// A non-empty slice is not enough: a target with an empty Node (or no
+	// assigned shards) is silently skipped by cleanupStaleEcShards and by
+	// distributeEcShards, so a plan of only such entries would pass the length
+	// check and mark the source readonly before failing. Reject any malformed
+	// target here, before the first destructive step.
+	for i, target := range t.targets {
+		if target == nil || target.Node == "" || len(target.ShardIds) == 0 {
+			return fmt.Errorf("malformed EC shard target %d for volume %d; refusing to mark source readonly", i, t.volumeID)
+		}
+	}
 	if t.server == "" && len(t.getReplicas()) == 0 {
 		return fmt.Errorf("no source replica for volume %d", t.volumeID)
 	}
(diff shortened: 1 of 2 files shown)


chrislusf:
Addressed the review feedback in 3253769:

- **coderabbit (malformed targets):** `ensureCleanEcStart` now validates each target (nil / empty `Node` / empty `ShardIds`) before the first destructive step, not just the slice length — with regression cases (`TestEnsureCleanEcStartRejectsMalformedTarget`).
- **greptile (writable source on failed teardown):** when `cleanupStaleEcShards` returns an error, `rollbackDistribute` now leaves the source **readonly** rather than exposing it for writes beside stale/possibly-mounted shards (which could diverge, and orphan cleanup won't remove a writable source). The next preflight or an operator reconciles once the shards are reachable.

PR #10740, opened 2026-08-13
erasure_coding: share the EC shard teardown primitive

chrislusf:
Stacked on #10738.

The unmount + full-teardown of EC shards was duplicated: the plugin-worker EC task had `unmountAndDeleteEcShards` and the shell had `unmountAndDeleteEcShardsQuiet`, byte-identical apart from a fence parameter and a sentinel error. That duplication is exactly how the teardown fence semantics drifted between the two paths (see #10738).

Distribute, mount, and verify already live in `weed/storage/erasure_coding` and are shared by both callers; this moves the teardown there too, as `UnmountAndDeleteEcShards` plus the shared `ErrFullTeardownNotAcked` sentinel. Both paths now call the one function, so the fence semantics can't diverge again. The shell keeps a thin type-converting wrapper and aliases the sentinel — **behavior is unchanged**.

Follows the same shell/worker de-duplication pattern as #10727.

Worker + shell EC tests pass through the shared helper.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved cleanup of stale erasure-coded storage shards.
  * Standardized shard unmounting and deletion across cleanup workflows.
  * Added clearer handling when servers complete cleanup but do not confirm full teardown.
  * Improved error reporting for shard unmount and deletion failures.

### DIFF
--- a/weed/shell/command_ec_common.go
+++ b/weed/shell/command_ec_common.go
@@ -2,7 +2,6 @@ package shell
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"regexp"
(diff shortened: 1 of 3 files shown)


chrislusf:
Rebased onto the updated #10738 (no code changes needed from review — the bots flagged no inline issues here). Still a pure de-duplication with unchanged behavior.

PR #10741, opened 2026-08-13
shell: roll back a failed ec.encode instead of leaving readonly volumes and orphan shards

chrislusf:
`ec.encode` marks the source volumes readonly and generates EC shards before it verifies the shards and deletes the originals. If a step in between failed, the command just returned the error: the volumes were left **readonly** and the partially-produced EC shards survived as **orphans**, cleaned up only if the operator re-ran `ec.encode` (via `clearPreexistingEcShards`).

This adds a deferred rollback that runs when the batch fails before the originals are deleted: it tears down the EC shards produced this run and restores the sources to writable, reusing the existing `clearPreexistingEcShards` and `markVolumeReplicaWritable` helpers. Once the shards are verified recoverable the batch is committed to the EC copy and does not roll back. Both rollback steps are idempotent, so a failure before the volumes were marked readonly is safe.

This is the shell counterpart to the plugin-worker `rollbackDistribute` in #10738. Independent of that PR.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when erasure-coding operations fail before source data is removed.
  - Automatically cleans up incomplete encoded shards and restores original data sources to a writable state.
  - Preserves the original error while recording any cleanup issues for diagnosis.
  - Once encoded data has been successfully verified, the verified copy is retained even if source deletion encounters an error.

### DIFF
--- a/weed/shell/command_ec_encode.go
+++ b/weed/shell/command_ec_encode.go
@@ -217,7 +217,7 @@ func chunkVolumeIds(volumeIds []needle.VolumeId, batchSize int) [][]needle.Volum
 	return batches
 }
 
-func processEcEncodeBatch(commandEnv *CommandEnv, writer io.Writer, volumeIds []needle.VolumeId, rp *super_block.ReplicaPlacement, diskType types.DiskType, maxParallelization int, applyBalancing bool, collectionForMessage string) error {
+func processEcEncodeBatch(commandEnv *CommandEnv, writer io.Writer, volumeIds []needle.VolumeId, rp *super_block.ReplicaPlacement, diskType types.DiskType, maxParallelization int, applyBalancing bool, collectionForMessage string) (err error) {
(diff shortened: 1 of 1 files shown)


chrislusf:
Addressed the review feedback in d1d53562f:

- **coderabbit (replica-placement race):** the rollback now re-reads volume locations before restoring writable (falling back to the pre-encode snapshot only if the re-read fails), so a replica `doEcEncode` marked readonly from its own later snapshot is also restored.
- **greptile (rollback-boundary test coverage):** the shell `ec.encode` flow has no cluster test harness (this package's tests are pure-function unit tests), so the rollback is composed from already-tested helpers (`clearPreexistingEcShards`, `markVolumeReplicaWritable`) rather than a new integration test. The equivalent plugin-worker rollback in #10738 does carry integration coverage.

PR #10749, opened 2026-08-13
feat: throughput limits for replicate, EC shard, and worker-driven moves

chrislusf:
Follow-up to #10727: the shared mover carried `IoBytePerSecond` only where the protocol had it — the volume-copy phase. This closes the remaining gaps so every data-moving path can be rate-limited, with 0 always meaning "the receiving volume server's own `-maintenanceBytePerSecond` governs".

- **proto**: `VolumeEcShardsCopyRequest` gains `io_byte_per_second = 10`; the balance/ec_balance task params and configs gain the same.
- **volume server**: `VolumeEcShardsCopy` throttles with **one `WriteThrottler` per request**, shared across the shard, `.ecx`, `.ecj`, `.vif`, and `.ecsum` copies so the limit caps the transfer as a whole — the same shape as `VolumeCopy`.
- **volume_move**: `ReplicateVolume` accepts the limit; `EcMoveOptions.IoBytePerSecond` flows through `MoveEcShards`/`CopyAndMountEcShards` into the copy request; fake-client tests assert the limit reaches both request types.
- **shell**: `ec.balance -ioBytePerSecond` (alongside `-maxParallelization`); `volume.tier.move`'s replication top-up now honors the command's existing `-ioBytePerSecond` instead of running unthrottled; `volume.fix.replication`/`volume.tier.upload`/`ec.encode` keep the server-rate default.
- **worker**: balance and ec_balance configs gain `io_byte_per_second` (surfaced in the admin config schema), carried through detection and plugin job parameters into task params and handed to the shared mover; batch balance jobs inherit the limit from their detection results.

Note: the limit is per copy stream, so `maxParallelization × ioBytePerSecond` is the aggregate ceiling — a per-server budget is possible follow-up work.

Testing: `go build ./...`; `go test -race ./weed/operation/volume_move/`; shell, balance, ec_balance, and all `test/plugin_workers` suites green. The same feature ships in the enterprise tree with the Rust volume server side included (the Rust gRPC EC-copy drain path was previously entirely unthrottled).


chrislusf:
Both findings confirmed and fixed in ab12a60ed: `deriveBalanceWorkerConfig` and `deriveECBalanceWorkerConfig` now read `io_byte_per_second` from the plugin configuration values (clamped at zero) into the task Config, and both plugin descriptors expose the field with defaults so the configuration form carries it — a configured limit now reaches plugin-driven detection instead of silently reverting to the server maintenance rate. Same fix mirrored to the enterprise tree.

PR #10755, opened 2026-08-14
volume: validate replica targets and restrict gcs credentials in FetchAndWriteNeedle

chrislusf:
Follow-up to the `FetchAndWriteNeedle` remote-endpoint hardening (#10720, and the azure extension in #10754). Two more caller-supplied inputs in the same handler were passed through without validation.

### Replica upload targets (`replicas[].url`)
The replica leg formats the target straight into an `http://<target>/<fileId>?type=replicate` upload URL. A target that carries a scheme, path or query, or that points at loopback / link-local / unspecified, would redirect the forwarded write. Each target is now required to be a bare `host:port` whose host is on a permitted network, reusing the existing address deny-list. Cluster peers legitimately sit on private networks, so RFC 1918 / CGNAT stay allowed, and `-volume.allowUntrustedRemoteEndpoints` still opts out.

### gcs credentials (`gcs_google_application_credentials`)
This value could name a local filesystem path, which the SDK reads from disk while handling the request. Only inline JSON is accepted here now; the server-side `GOOGLE_APPLICATION_CREDENTIALS` env var still supplies a path.

### Notes
- Each change is its own commit, with tests.
- The replica-target guard is mirrored in the Rust volume server (`seaweed-volume`). The Rust server only implements the S3-compatible backends, so it has no gcs path to change.
- The azure endpoint (the third caller-supplied input in this handler) is handled separately by #10754.

## Summary by CodeRabbit

* **New Features**
  * Added replica-target validation requiring bare host-and-port values and checking resolved addresses before uploads.
  * Private and CGNAT addresses are supported for replica communication.
  * GCS credential inputs now distinguish between inline credentials and filesystem paths.

* **Bug Fixes**
  * Blocked loopback, link-local, metadata-service, unspecified, and other unsafe addresses.
  * Invalid replica targets are rejected before replica writes begin.
  * Added protection against DNS rebinding during replica connections.

### DIFF
--- a/seaweed-volume/src/remote_storage/endpoint_guard.rs
+++ b/seaweed-volume/src/remote_storage/endpoint_guard.rs
@@ -112,6 +112,14 @@ fn embedded_transition_ipv4(v6: Ipv6Addr) -> Option<Ipv4Addr> {
 /// Returns an error if `ip` is not safe to dial from a server that can reach
 /// cluster-internal hosts. Mirrors Go's `checkBlockedIP`.
 pub fn check_blocked_ip(endpoint: &str, ip: IpAddr) -> Result<(), String> {
+    check_blocked_ip_policy(endpoint, ip, false)
+}
+
+/// Like [`check_blocked_ip`], but `allow_private` keeps RFC 1918 / CGNAT
+/// reachable for callers whose target legitimately sits on an internal network
+/// (peer volume servers), while still blocking loopback, link-local (IMDS) and
+/// unspecified. Mirrors Go's `checkBlockedIPPolicy`.
+pub fn check_blocked_ip_policy(endpoint: &str, ip: IpAddr, allow_private: bool) -> Result<(), String> {
     // Normalize IPv4-mapped IPv6 (`::ffff:a.b.c.d`) to its IPv4 form so the
     // IPv4 deny rules apply. The OS routes these to the embedded IPv4 address,
     // so without this `::ffff:127.0.0.1` / `::ffff:169.254.169.254` would slip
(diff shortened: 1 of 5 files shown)


chrislusf:
Thanks — both of you flagged the same two real gaps, addressed in the latest push:

**1. Replica upload DNS rebinding.** `validateReplicaTarget` only checked the up-front DNS answer, while `UploadData` re-resolved the hostname at connect time through the default client — the same TOCTOU the S3 path already closes. The replica upload now goes through a guarded client (`newGuardedHTTPClientPolicy(target, allowPrivate=true)`) whose dialer re-applies the address policy at connect time, so a replica hostname that rebinds to loopback/link-local/unspecified after validation is refused at dial. Private peers stay allowed. Added `TestGuardedReplicaDialerRebind`. (The Rust S3 path likewise validates DNS up front without a connect-time re-check; the replica path matches that posture, with the same re-check noted as a shared follow-up.)

**2. Rust validation left the local write running.** The `?` on a failed replica returned after `spawn_blocking` had already been kicked off, so the client got an error while the needle could still be written. Both Go and Rust now validate every replica target *before* the local write, so a bad target fails the request cleanly with no partial write — and the two implementations mirror each other.

chrislusf:
Addressed the two comments on the latest push:

**1. `validateReplicaTarget` — require an explicit port.** Good catch: the doc says "bare host:port" but the code fell back to accepting a portless host. It now rejects when `net.SplitHostPort` fails, so a bare hostname/IP literal without a port can't reach URL construction. Added `missing port literal` / `missing port hostname` cases; the Rust `validate_replica_target` mirrors this (rejects when there's no `host:port`, still accepts `[ipv6]:port`).

**2. `gcsCredentialsArePath` — trimming leading whitespace: intentionally not done.** This predicate deliberately matches the gcs client's own inline detection, which is `strings.HasPrefix(util.ResolvePath(creds), "{")` with no trim (`ResolvePath` only expands `~`, it doesn't trim). If I trimmed here, `"  {json}"` would pass this guard as "inline" but the client would then treat it as a path and `os.ReadFile` it — reintroducing the exact readable/absent/permission-denied distinction this guard removes. Keeping it un-trimmed means anything the client would read from disk is rejected first. Skipping this one to preserve that parity.

PR #10766, opened 2026-08-15
admin: show capacity per storage tier and stop counting remote-tiered bytes as local disk usage

### LINKED ISSUE #10765: Volume server capacity incorrectly reported in admin UI with Cloud Tiering
I have noticed that the capacity of Volume servers appears to be incorrectly reported when using Cloud Tiering to S3. It appears that the size of the data file is counted into the used size, even after uploading it to S3. 

Usage reported in dashboard:
Image

Actual usage
```
Filesystem                Size      Used Available Use% Mounted on
/dev/sdg                 63.9G     44.1G     19.9G  69% /data
Filesystem                Size      Used Available Use% Mounted on
/dev/sde                 63.9G     34.9G     29.0G  55% /data
Filesystem                Size      Used Available Use% Mounted on
/dev/sdg                 63.9G     31.0G     33.0G  48% /data
```

However, the Volume server UI seems to report accurate values (t1 rack below):

Image

chrislusf:
Fixes #10765.

## What

Volumes and EC shards can live on different tiers — local disk types (`hdd`, `ssd`, custom tags) and, for volumes, a remote/cloud tier. The admin UI flattened all of that into single used/capacity numbers, and computed "used" by summing `VolumeInformationMessage.Size` over every volume. For a remote-tiered volume that field reports the **cloud object's size** (the local `.dat` is unlinked after `volume.tier.move`, and `Size` comes from the remote `DataBackend.GetStat()`), so the dashboard counted cloud bytes against local disk capacity — the inflation reported in #10765. The volume server's own UI is right because it uses statfs and lists remote volumes separately.

## How

Everything needed already rides the `VolumeList` response (per-disk-type `DiskInfo` with statfs bytes, `RemoteStorageName`/`Size` per volume, `DiskType`/`ShardSizes` per EC shard), so this is an admin-side aggregation and display change only — no proto, heartbeat, or volume server changes.

- **Accounting split**: `VolumeServer.DiskUsage` now counts only bytes on local disks; the cloud bytes move to a new `RemoteSize` field. `TotalSize` keeps its logical meaning (all data managed, wherever it lives).
- **Dashboard "Storage Tiers" table** (new `CollectTierStats` in `dash/tier_stats.go`): one row per local disk type plus one per remote storage name, with volume/EC-shard counts, logical data size, statfs disk used, capacity (statfs, with the slot-based fallback for older servers), and a usage bar.
- **Servers tables** (dashboard + `/cluster/volume-servers`): the capacity column shows `local / capacity` with a cloud badge for remote bytes; the details modal gains a "Remote Data" row.
- **Volumes page**: remote-tiered volumes get a cloud badge with their remote storage name in the disk type column (shown whenever any volume is remote).
- **EC shards page**: per-shard `Size` is now filled from `ShardSizes` instead of the hard-coded 0 (the "not in the API response" comment was stale).

Edge cases: `-keepLocalDatFile` volumes are attributed to the remote tier for data-size purposes (their local copy still shows in statfs disk-used); servers that don't report `disk_total_bytes` fall back to slot-based capacity, and the usage bar falls back to logical size. EC shards can't be remote-tiered today; when that lands, remote EC rows fold into the same table.

## Testing

- New table-driven test `dash/tier_stats_test.go` covering mixed disk types, two remote storages, EC shards, empty-disk-type normalization, and the missing-statfs fallback.
- `go build ./weed/...`, `go vet` on the touched packages, and the full `weed/admin` test suite pass. Templates regenerated with the pinned templ v0.3.1020.

## Summary by CodeRabbit

* **New Features**
  * Added Storage Tiers dashboard statistics for volumes, erasure-coded shards, capacity, usage, and data size.
  * Added remote-storage indicators and badges to volume and volume-server views.
  * Volume-server details now show remote data separately from local disk usage.
* **Bug Fixes**
  * Corrected erasure-coded shard size reporting.
  * Improved local and remote storage usage calculations, including capacity fallback handling.
  * Added clearer disk-type and remote-storage labeling.
* **Tests**
  * Added coverage for local, remote, and erasure-coded storage tier statistics.


chrislusf:
Addressed the review in 2fe934fad:

- **Mixed statfs/legacy tiers**: a disk that predates `disk_total_bytes` now folds its logical bytes into the tier's `DiskUsed`, so a tier mixing old and new volume servers no longer underreports usage. This also makes the Disk Used cell and the Usage bar always agree, so the `DataSize` fallback inside `UsagePercent` is gone. Added the suggested test case (an old server sharing the hdd tier with a statfs-reporting one).
- **Overflow**: `UsagePercent` now uses float math instead of `used * 100`.
- **VolumeSizeLimitMb**: `getTopologyViaGRPC` defaults a zero value to 30000 MB, matching `GetClusterVolumeServers`, and passes the normalized value to `CollectTierStats` and the slot-based capacity estimate.
- **Header**: the dashboard volume-servers column now reads "Usage / Capacity".
- **`hdd` default**: extracted `displayDiskType` and reused it in `countUniqueDiskTypes`.

Not done here: regenerating all `*_templ.go` at v0.3.1020. The v0.3.1001/v0.3.1020 mix predates this PR (the three files this PR touches were generated with v0.3.1020, same as the neighboring `admin_templ.go` already on master); a repo-wide regeneration is a separate mechanical chore and would bury this diff in unrelated churn.

PR #10769, opened 2026-08-15
ec.encode: require the shards to agree on size before deleting the volume

chrislusf:
## The gap

An encode deletes the volume it encoded. `verifyEcShardsBeforeDelete` is what stands in the way, and it asks two questions:

- `RequireRecoverableShardSet` — do enough shards exist?
- `ecShardsClumpedOnOneNode` — are they spread across nodes?

Both are about **presence**. Nothing asks whether those shards are *whole*. A shard truncated by a partial write, a half-finished copy, or a disk that filled up is counted the same as a good one, and the source volume — the only thing that could rebuild it — is deleted on that count.

## The change

Every shard of a volume takes one piece of each block row, so they are all written to the same length. A shard that disagrees with its siblings is incomplete, and that is visible without reading any data: `shard_sizes` already travels in the volume-server heartbeat (`VolumeEcShardInformationMessage` field 7) and is reachable through `ShardsInfo.Size`.

So compare them, and hold the deletion back when they disagree, naming the odd shard and its holder:

```
volume 7 ec shards disagree on size, so at least one is incomplete
  (1048576 bytes: server1:8080.0 server1:8080.1; 4096 bytes: server2:8080.2)
```

The check sits beside the existing recoverability check inside the same retry loop, so a size that has not propagated yet is retried rather than treated as failure.

### Not blocking encodes that cannot be checked

The failure mode to avoid is a wrong **reject**: refusing to delete leaves the volume beside its shards, which is the mixed state this whole path exists to get out of. A volume server that predates shard-size reporting, or one that has not heartbeated sizes yet, reports zero — those are skipped rather than read as a disagreement. Same for nil inventories and empty topologies. The tests pin each of those cases.

## Test

`TestRequireUniformShardSizes` covers a uniform set on one holder and spread across holders, a truncated shard (named with its holder), two copies of one shard that disagree, and the four "cannot check" shapes that must still let the encode through.

Unit tests cannot answer the question that decides whether this is safe — whether real encodes actually produce uniform sizes — so it was run against the live-cluster suite: the whole `test/erasure_coding` package, 384s, 0 failures, and **no encode was rejected**.

## Scope

This is the encode half of the same gap [#10768](https://github.com/seaweedfs/seaweedfs/pull/10768) closes for the decode half, where the volume was deleted-adjacent check was `.dat`/`.idx` non-empty. The two are independent.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved erasure-coded volume verification by validating shard sizes from the newest generation.
  * Detects inconsistent nonzero shard sizes before the original volume is deleted.
  * Ignores unavailable zero-size reports and preserves per-node shard information when combining results.
  * Added clearer, deterministic diagnostics identifying affected servers and shards.

* **Tests**
  * Added coverage for generation filtering, matching and conflicting sizes, missing inventories, truncated shards, duplicate shard reports, and legacy data.

### DIFF
--- a/weed/ec/ec_encode.go
+++ b/weed/ec/ec_encode.go
@@ -9,6 +9,7 @@ import (
 	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


chrislusf:
Both reviews caught the same real bug — thank you, this would have bitten in production, not just in tests.

The size check collected every shard the master reports, while `RequireRecoverableShardSet` beside it counts only the newest `EncodeTsNs`. Since a re-encode can change the ratio, an orphaned older generation has shards of a different length *by nature*, so merging them made a healthy current set look inconsistent — and because the orphan keeps being reported, every retry would fail and the encode would sit on both the volume and its shards indefinitely. That is precisely the mixed state this check exists to get out of, so the check would have caused the condition it was meant to prevent.

Fixed in `dec9b9c5` by collecting sizes the way `CollectEcShardBitsByNode` collects bits — fenced to the newest `EncodeTsNs`, with unstamped entries forming the single legacy generation, so the two checks now judge the same set.

Three regression tests go through the topology collection path rather than the helper directly, as suggested:

- `TestShardSizeCheckIgnoresOlderEncodeGenerations` — a 512 KiB orphan alongside a uniform 1 MiB current generation must not veto the delete, and the orphan's holder must not be collected at all
- `TestShardSizeCheckStillCatchesShortShardInNewestGeneration` — fencing must not blind the check: a truncated shard *within* the newest generation still holds the deletion back
- `TestShardSizeCheckHandlesUnstampedGenerations` — volumes encoded before generation stamping report zero, which is one legacy generation rather than an orphan to drop

Re-running the live-cluster suite now to confirm the fenced collector behaves on a real encode; will report.

PR #10772, opened 2026-08-16
iceberg: let table properties override the worker config

chrislusf:
The Iceberg maintenance worker took its target file size and snapshot retention entirely from the global plugin config. Table properties were never read, so a table asking for 512 MiB files got whatever the plugin config said — and a writer honouring `write.target-file-size-bytes` and a compactor rewriting to a different size will rewrite each other's output forever. Every other Iceberg implementation lets the table's own properties win here; StarRocks resolves property, then session variable, then engine default, and Spark and Trino do the same.

Four properties now override the worker config:

| Property | Config field |
| --- | --- |
| `write.target-file-size-bytes` | `TargetFileSizeBytes` |
| `write.delete.target-file-size-bytes` | `DeleteTargetFileSizeBytes` |
| `history.expire.max-snapshot-age-ms` | `SnapshotRetentionMs` |
| `history.expire.min-snapshots-to-keep` | `MaxSnapshotsToKeep` |

The last two already meant the right thing. `expireSnapshots` keeps the current snapshot, keeps the newest `MaxSnapshotsToKeep`, and expires past that only beyond the retention window — Iceberg's own predicate, just fed from the wrong place. (`MaxSnapshotsToKeep` is a floor rather than a cap despite the name; renaming the published plugin config key is a separate question.)

Detection pays nothing for this: `scanTablesForMaintenance` already parses the metadata envelope, so `Properties()` is in hand. Execution costs one metadata read, done once for the whole job rather than per operation, so compaction committing new metadata halfway through cannot change the settings underneath it.

The first commit is a prerequisite refactor with no behaviour change. Retention was stored in hours, so `history.expire.max-snapshot-age-ms` could only be truncated to whole hours — any sub-hour value rounds to zero and then gets clamped back up to the 168 hour default, which is the loosest possible retention rather than the tightest. Config now carries milliseconds internally; the `snapshot_retention_hours` plugin key is unchanged.

A malformed, zero or negative property is logged at V(1) and ignored, so a misconfigured table falls back to the plugin config instead of stopping its own maintenance.

## Summary by CodeRabbit

* **New Features**
  * Iceberg maintenance now supports per-table settings for target file sizes and snapshot retention.
  * Snapshot retention supports sub-hour precision for more accurate expiration scheduling.
  * Invalid, missing, or non-positive table settings safely fall back to configured defaults.
  * Retention and orphan cleanup values are safely bounded to prevent overflow issues.

* **Bug Fixes**
  * Maintenance now evaluates the latest table metadata before applying operations.
  * Metadata loading failures now stop the job with an error instead of using outdated settings.

### DIFF
--- a/test/s3tables/maintenance/maintenance_integration_test.go
+++ b/test/s3tables/maintenance/maintenance_integration_test.go
@@ -600,9 +600,9 @@ func testExpireSnapshots(t *testing.T) {
 
 	handler := icebergHandler.NewHandler(nil)
 	config := icebergHandler.Config{
-		SnapshotRetentionHours: 0, // instant expiry — everything eligible
-		MaxSnapshotsToKeep:     1, // keep only the current snapshot
-		MaxCommitRetries:       3,
+		SnapshotRetentionMs: 0, // instant expiry — everything eligible
+		MaxSnapshotsToKeep:  1, // keep only the current snapshot
+		MaxCommitRetries:    3,
 	}
 
 	result, _, err := handler.ExpireSnapshots(context.Background(), client, bucket, path.Join(ns, tbl), config)
@@ -932,9 +932,9 @@ func testFullMaintenanceCycle(t *testing.T) {
 
 	// Step 1: Expire snapshots
 	expireConfig := icebergHandler.Config{
-		SnapshotRetentionHours: 0, // instant expiry
-		MaxSnapshotsToKeep:     1,
-		MaxCommitRetries:       3,
+		SnapshotRetentionMs: 0, // instant expiry
+		MaxSnapshotsToKeep:  1,
+		MaxCommitRetries:    3,
 	}
 	result, _, err := handler.ExpireSnapshots(ctx, client, bucket, tablePath, expireConfig)
 	require.NoError(t, err)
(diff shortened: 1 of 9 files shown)


chrislusf:
CI and review feedback addressed.

**CI (Go Vet 32-bit).** `test/s3tables/maintenance/maintenance_integration_test.go` also constructed `Config` with `SnapshotRetentionHours` — an incomplete rename on my side, now folded into the commit that did the rename. It sat outside `weed/`, so `go build ./...` skipped it (test file) and my package-scoped `go test` never reached it. Re-verified with `GOOS=linux GOARCH=386 go vet` over `weed/worker/tasks/iceberg`, `weed/s3api/s3tables` and `test/s3tables/...`.

**Metadata fallback drops table overrides (Greptile, P1).** Valid, and the failure mode is exactly what this PR exists to prevent — a transient read failure would silently fall back to the plugin config and rewrite the table to a size it never asked for, which is the thrash loop. It now fails the job instead. I originally chose log-and-continue to keep the diff additive; that was the wrong trade for a path that decides file sizes.

**Third commit is new.** While checking the overflow report on the stacked PR I found the same class of bug reachable from today's `orphan_older_than_hours`: `collectOrphanCandidates` converts the cutoff to a `time.Duration`, and past ~2.5M hours the multiplication wraps negative, putting the cutoff in the future so every file walked looks like an orphan. That deletes data a concurrent writer has not yet committed. Clamped in `applyThresholdDefaults`, which covers both the plugin config and the API path on the stacked PR.

PR #10773, opened 2026-08-16
s3tables: add the maintenance configuration APIs

chrislusf:
Rebased onto master now that #10772 has merged; this PR is the remaining three commits.

Adds the five maintenance APIs the S3 Tables control plane was missing: `Get`/`PutTableBucketMaintenanceConfiguration`, `Get`/`PutTableMaintenanceConfiguration`, and `GetTableMaintenanceJobStatus`. Until now the worker took a single global plugin config with wildcard bucket/namespace/table filters, so there was no per-table override and no way to read back what a run did.

### One idea

Iceberg table properties, the AWS maintenance settings, and our plugin config are three vocabularies for the same handful of numbers, so the whole thing is one resolver plus the storage to feed it.

| Knob | AWS setting | Worker field |
| --- | --- | --- |
| Compaction target | `icebergCompaction.targetFileSizeMB` | `TargetFileSizeBytes` |
| Snapshot age | `icebergSnapshotManagement.maxSnapshotAgeHours` | `SnapshotRetentionMs` |
| Snapshot floor | `icebergSnapshotManagement.minSnapshotsToKeep` | `MaxSnapshotsToKeep` |
| Orphan age | `icebergUnreferencedFileRemoval.unreferencedDays` | `OrphanOlderThanHours` |

Resolution order is table property, then per-table configuration, then per-bucket configuration, then plugin config. Detection pays nothing for the two new attributes — `scanTablesForMaintenance` already has the entries and their `Extended` maps in hand.

### The one call worth arguing about

Properties beating the maintenance configuration means `PutTableMaintenanceConfiguration` is advisory for knobs the table also declares: you Put 256, Get returns 256, and the worker still uses the table's 512. That diverges from AWS, and the reason is mechanical — a writer honouring `write.target-file-size-bytes` and a compactor rewriting to a different size rewrite each other's output forever. `table_properties_override` (default on) flips the two layers for anyone who wants the control plane authoritative.

`status` is deliberately outside that contest. Iceberg has no property for "don't compact this table", so a disabled type drops its operations and nothing can re-enable them. The kill switch always holds.

### Storage

Two new extended attributes. `s3tables.maintenance` holds the configuration verbatim in its wire shape, so `Get` is a passthrough and no translation layer can drift; Put merges a single type so configuring compaction does not drop snapshot management. `s3tables.maintenanceStatus` is written only by the worker, kept separate so operator and worker writes never contend on one attribute.

Both writes assert the attribute's prior value. Worth flagging: `validateUpdateEntryPreconditions` does per-key CAS but `UpdateEntry` rewrites the whole entry, so writers asserting *different* keys can still clobber each other. `persistPlanningIndex` already lives with this; separate keys plus bounded retries narrow it, and the real fix is partial-xattr update semantics in the filer, which belongs in its own issue.

### Divergences from AWS

Dispatch is by `X-Amz-Target` rather than REST paths, inherited from the existing thirty actions rather than introduced here. `Get` on an unconfigured resource returns an empty `configuration` map instead of synthesised defaults, since our effective defaults live in a layer this API cannot express. Job status is last-run only — there is no retained history behind it. Manifest and delete-file rewrites have no AWS equivalent and ride with `icebergCompaction`, so disabling compaction stops all rewriting.

The encryption trio is still missing and nothing here depends on it.

## Summary by CodeRabbit

- **New Features**
  - Added S3 Tables maintenance configuration for compaction, snapshot management, and unreferenced-file removal at bucket and table levels.
  - Added REST operations to update and retrieve settings and view maintenance job status.
  - Added automatic compaction strategy selection and configurable table-property precedence.
  - Added support for AWS ARN partitions beyond the standard commercial partition.

- **Bug Fixes**
  - Improved concurrent updates with conflict detection and automatic retries.
  - Preserved maintenance settings during table renames and removed them during deletion.
  - Added safer handling for invalid settings, disabled operations, and numeric overflow conditions.

### DIFF
--- a/weed/s3api/s3api_tables.go
+++ b/weed/s3api/s3api_tables.go
@@ -76,8 +76,9 @@ func (s3a *S3ApiServer) registerS3TablesRoutes(router *mux.Router) {
 		s3TablesApi.SetDefaultAllow(true)
 	}
 
-	// Regex for S3 Tables Bucket ARN
-	const tableBucketARNRegex = "arn:aws:s3tables:[^/:]*:[^/:]*:bucket/[^/]+"
+	// Regex for S3 Tables Bucket ARN. The partition is not fixed to "aws":
+	// aws-cn and aws-us-gov ARNs are valid and have to reach the handlers.
+	const tableBucketARNRegex = s3tables.ARNPrefixPatternStr + ":[^/:]*:[^/:]*:bucket/[^/]+"
 
 	// REST-style S3 Tables API routes (used by AWS CLI)
 	targetMatcher := func(r *http.Request, rm *mux.RouteMatch) bool {
@@ -144,11 +145,11 @@ func (s3a *S3ApiServer) registerS3TablesRoutes(router *mux.Router) {
 	router.Methods(http.MethodGet).Path("/tables/{tableBucketARN:" + tableBucketARNRegex + "}/{namespace}/{name}/maintenance-job-status").MatcherFunc(serviceMatcher).
 		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("GetTableMaintenanceJobStatus", buildGetTableMaintenanceJobStatusRequest)), "S3Tables-GetTableMaintenanceJobStatus"))
 
-	router.Methods(http.MethodPost).Path("/tag/{resourceArn:arn:aws:s3tables:.*}").MatcherFunc(serviceMatcher).
+	router.Methods(http.MethodPost).Path("/tag/{resourceArn:" + s3tables.ARNPrefixPatternStr + ":.*}").MatcherFunc(serviceMatcher).
 		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("TagResource", buildTagResourceRequest)), "S3Tables-TagResource"))
-	router.Methods(http.MethodGet).Path("/tag/{resourceArn:arn:aws:s3tables:.*}").MatcherFunc(serviceMatcher).
+	router.Methods(http.MethodGet).Path("/tag/{resourceArn:" + s3tables.ARNPrefixPatternStr + ":.*}").MatcherFunc(serviceMatcher).
 		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("ListTagsForResource", buildListTagsForResourceRequest)), "S3Tables-ListTagsForResource"))
-	router.Methods(http.MethodDelete).Path("/tag/{resourceArn:arn:aws:s3tables:.*}").MatcherFunc(serviceMatcher).
+	router.Methods(http.MethodDelete).Path("/tag/{resourceArn:" + s3tables.ARNPrefixPatternStr + ":.*}").MatcherFunc(serviceMatcher).
 		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("UntagResource", buildUntagResourceRequest)), "S3Tables-UntagResource"))
 
 	router.Methods(http.MethodGet).Path("/get-table").MatcherFunc(serviceMatcher).
--- a/weed/s3api/s3api_tables.go
+++ b/weed/s3api/s3api_tables.go
@@ -133,6 +133,17 @@ func (s3a *S3ApiServer) registerS3TablesRoutes(router *mux.Router) {
 	router.Methods(http.MethodDelete).Path("/tables/{tableBucketARN:" + tableBucketARNRegex + "}/{namespace}/{name}/policy").MatcherFunc(serviceMatcher).
 		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("DeleteTablePolicy", buildDeleteTablePolicyRequest)), "S3Tables-DeleteTablePolicy"))
 
+	router.Methods(http.MethodPut).Path("/buckets/{tableBucketARN:" + tableBucketARNRegex + "}/maintenance/{type}").MatcherFunc(serviceMatcher).
+		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("PutTableBucketMaintenanceConfiguration", buildPutTableBucketMaintenanceConfigurationRequest)), "S3Tables-PutTableBucketMaintenanceConfiguration"))
+	router.Methods(http.MethodGet).Path("/buckets/{tableBucketARN:" + tableBucketARNRegex + "}/maintenance").MatcherFunc(serviceMatcher).
+		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("GetTableBucketMaintenanceConfiguration", buildGetTableBucketMaintenanceConfigurationRequest)), "S3Tables-GetTableBucketMaintenanceConfiguration"))
+	router.Methods(http.MethodPut).Path("/tables/{tableBucketARN:" + tableBucketARNRegex + "}/{namespace}/{name}/maintenance/{type}").MatcherFunc(serviceMatcher).
+		HandlerFunc(track(s3a.authenticateS3Tables(s3TablesApi.handleRestOperation("PutTableMaintenanceConfiguration", buildPutTableMaintenanceConfigurationRequest)), "S3Tables-PutTableMaintenanceConfiguration"))
+	router.Methods(http.MethodGet).Path("/tables/{tableBucketARN:" + tableBucketARNRegex + "}/{namespace}/{name}/maintenance").MatcherFunc(serviceMatcher).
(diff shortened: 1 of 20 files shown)


chrislusf:
Rebased onto the updated #10772 and addressed both review findings.

**Maintenance kill switch fails open (Greptile, P1).** Correct and the most serious of the three — a filer lookup failure or an unreadable attribute collapsed to an empty configuration, which `filterDisabledOperations` reads as "nothing disabled", so orphan removal could delete files the operator had explicitly switched off. A kill switch that fails open is worse than no kill switch.

`parseMaintenanceConfiguration`, `lookupMaintenanceConfiguration` and `loadMaintenanceConfiguration` now return errors. Execution fails the job; detection skips the bucket or table, which proposes no work and so also fails closed.

**Orphan retention duration overflows (Greptile, P1).** Also correct. `unreferencedDays * 24` overflows on the way to the `time.Duration` conversion, wrapping the cutoff into the future and making every walked file an orphan candidate. Fixed in two places: `daysToHours` saturates here, and the clamp added in #10772 bounds the result to what a `time.Duration` can hold — that clamp also fixes the same bug reachable today through `orphan_older_than_hours`, independent of this API.

**Still open, unchanged.** `UpdateEntry`'s `ExpectedExtended` is a genuine CAS — the precondition check and the write are serialized under an exclusive per-path lock (`filer_grpc_server.go:636-653`), so I was wrong earlier to suggest the filer needed new semantics for the single-filer case. What remains is that these writers assert only their own key while writing the whole entry, so a concurrent write to a *different* attribute can still be lost. Asserting the full `Extended` snapshot closes that with no filer change, and I would rather do it as a follow-up across all the s3tables writers — `persistPlanningIndex` and the policy and tag handlers have the same shape — than widen this PR. The cross-filer case is separate again: the lock is in-process, and these control-plane paths do not participate in route-by-key.

chrislusf:
All six addressed, one commit each.

**1. Status persistence undoing a kill switch (P1).** Correct, and worse than a lost write — an advisory status write silently re-enabling maintenance an operator disabled. Both writers now assert the entry's whole attribute set rather than their own key, including the target key when absent so a concurrent create also fails the precondition. No filer change needed: `ExpectedExtended` is already a real CAS under the per-path lock, the assertion was just too narrow.

**2. nonCurrentDays ignored (P1).** Correct. Confirmed against the SDK doc comments — `unreferencedDays` is time-to-mark-non-current, `nonCurrentDays` is the further wait before deletion. `remove_orphans` deletes in one step rather than marking, so the cutoff is now the sum; a 3/10 configuration waits thirteen days. Saturating add, and the existing clamp still bounds the result.

**3. Unreachable through AWS CLI/SDK (P1).** Correct, and I had this backwards earlier — I called `X-Amz-Target` "the shape of this surface", but `registerS3TablesRoutes` is explicitly the path the CLI uses and I had added nothing to it. Five REST routes added on the published bindings, with builders that take the type from the path and the value from the body. Tests assert the paths route as AWS spells them.

**4. Bucket-level disabling missing from job status (P2).** Correct. The handler read only the table attribute, so a bucket-level disable of unreferenced file removal reported `Not_Yet_Run` or a stale success. It now merges bucket and table configuration, and the merge helper moved into this package so the worker shares one implementation instead of duplicating it.

**5. ARN field casing (P2).** Correct — `tableBucketArn` where every other response in the package spells it `tableBucketARN`. One line.

**6. Compaction strategy discarded (P2).** Correct, and it broke this PR's own premise that storage is the verbatim wire shape. `strategy` is modelled now; the worker maps `binpack` and `sort` onto its rewrite strategy and lets `auto` defer to the worker configuration. `z-order` is rejected at the API rather than accepted and quietly binpacked, since the worker cannot produce it.

Also removed unrelated formatting churn: an earlier `gofmt -w weed/` had swept ten pre-existing unformatted files into this branch. The diff is back to the thirteen files this work actually touches.

One thing I did not change: `TargetFileSizeMB` and the day counts are `*int32` in the AWS model and `int64` here. The clamps make the overflow unreachable either way, so I left it rather than churn the types.

chrislusf:
All three addressed.

**1. First-time disables still erasable (P1).** Right, and this is the second time I have half-closed this. My previous fix asserted the attributes that existed at read time, which does nothing for one created *after* that read — exactly the first-time `PutTableMaintenanceConfiguration` case, where the new configuration lands, passes the per-key checks, and is then deleted by the stale whole-entry write.

The precondition now covers every attribute this package defines, absent ones asserted absent, so a concurrent create fails the write and the caller re-reads. You were also right that the metadata commit and planning index writers share the exposure: both rewrite the same entries and both previously asserted only their own keys, so a successful maintenance commit could erase a configuration. Both use the shared snapshot now.

That is a bounded fix, not a general one — it protects the keys this package knows about. An exact whole-map CAS would need a proto change to `UpdateEntryRequest`, which I would rather raise separately than smuggle in here.

**2. auto silently ignored (P2).** Correct. It was accepted, stored and read back while leaving the worker on its default, so a sorted table configured `auto` compacted with binpack. Implemented per the AWS definition — sort where the table declares a usable sort order, binpack otherwise. The choice needs the table metadata, so it is made where the rewrite plan resolves; an unsorted table falls back rather than failing the way an explicit `sort` does.

z-order is still rejected. The worker cannot produce it, and accepting it would put us back in exactly the failure you flagged last round: stored one way, executed another. Happy to revisit if you would rather take the AWS-compatibility side of that trade.

**3. Numeric settings unvalidated (P2).** Correct. Confirmed the range from the three API pages — all five are 1..2147483647 — and it is enforced now. The fields became pointers so an explicit `0` is distinguishable from an omitted one and can be rejected rather than silently ignored; that also fixes a round-trip bug where a stored zero vanished on re-marshal. The worker's clamps stay as defence for values that never went through the API.

Also stripped the unrelated formatting churn again — a `gofmt -w weed/` had re-swept those ten pre-existing files after I removed them. I have stopped formatting the whole tree; only touched files now.

chrislusf:
All three addressed.

**1. Regular xattr writers still undoing the kill switch (P1).** Right, and this is the third round on this bug — I keep fixing the writer in front of me instead of the class. `updateExtendedAttribute` had the CAS, but `setExtendedAttribute`, `setExtendedAttributes`, `removeExtendedAttributes` and `deleteExtendedAttribute` were still unconditional whole-entry writes, and those are what the metadata, policy and tag handlers actually call. Any of them landing on a stale snapshot deletes a maintenance configuration written in between.

They now share a single read-modify-write loop, so the snapshot precondition and the bounded retry apply wherever this package rewrites an entry. One consequence worth knowing: these helpers can now return `ErrConcurrentUpdate` after five attempts, where before they silently clobbered. The maintenance handlers map that to 409; other handlers surface it as a 500, which is worth tidying separately.

**2. Rename dropping the configuration (P1).** Correct, and the leak direction is the nastier half — the old name keeps the settings and hands them to whatever is created there next. Both attributes now move with the table and are cleared from the source, and the decoupled-delete cleanup drops them too. The test sets a disabled configuration, renames, and asserts it arrived and did not linger.

**3. Partition-restricted ARNs (P2).** Correct. Fixed properly rather than only for the new routes, since the regex is shared: the router now takes the partition-tolerant prefix from the parser so the two cannot drift, and `buildARN` derives the partition from the region so an emitted ARN parses back. That should close the open CodeRabbit thread as well.

This one was pre-existing across all thirty-odd actions rather than something these five introduced, so tell me if you would rather I pulled it out into its own PR — it does change behaviour for every existing action, in that they now accept ARNs they previously rejected.

chrislusf:
All three addressed.

**1. Rename swallowing a concurrent maintenance write (P1).** Correct, and my rename fix opened it: copying attributes from an early read and clearing the source at the end leaves a window where a Put succeeds, misses the copy to the destination, and is then deleted by the cleanup. It returns success and vanishes.

The cleanup now clears the source only while it still holds exactly what was copied, and returns 409 otherwise — that keeps the existing invariant that a failed rename never loses the table, since the source is left intact. Put checks the catalog identity inside the same conditional mutation, so it also cannot write onto a name a rename or delete has already soft-deleted. The assertion covers every attribute the rename carries, not just the maintenance ones, since policy and tags have the same window.

**2. Generators still emitting arn:aws (P2).** Correct, and the more useful half of this is the second sentence: my round-trip test was worthless for this because parsing accepts any partition, so a commercial ARN returned in a GovCloud region passes it. The three generators now go through the partition-aware builder, along with the plain S3 ARN used for IAM resource matching, and the test asserts the prefix the region implies rather than only that the name survives.

**3. Incomplete partition resolver (P2).** Correct — aws-iso-e, aws-iso-f and aws-eusc were missing. Added, with the region table in the test extended to cover each one.

Worth saying plainly: this is the fourth round on the concurrency of these entry writes, and each round I have fixed the specific caller in front of me rather than the shape of the problem. The invariant I should have stated at the outset is that every whole-entry write must assert what it read, and every read-then-write pair separated in time must assert across the gap.

PR #10782, opened 2026-08-17
filer: do not sweep children when deleting a folder non-recursively

### LINKED ISSUE #10779: s3: empty-folder cleaner can delete an object committed after its emptiness check (acknowledged write lost)
### Version

4.41 (also current master: `weed/filer/empty_folder_cleanup/empty_folder_cleaner.go` is byte-identical between the two, md5 `f6099931caee2dc989cba8631cbded47`).

### What happens

The S3 empty-folder cleaner can delete a row that a concurrent PUT committed after the emptiness check, so an object that S3 acknowledged with 200 is gone. A HEAD issued a few milliseconds later returns 404.

This is the same class of bug as the one fixed for the multipart staging tree in #10273, on a different path.

### The window

`executeCleanup` counts the folder and, if empty, calls `deleteFolder`, which uses `isRecursive=false`:

```go
// empty_folder_cleanup/empty_folder_cleaner.go:359
func (efc *EmptyFolderCleaner) deleteFolder(ctx context.Context, folder string) error {
	return efc.filer.DeleteEntryMetaAndData(ctx, util.FullPath(folder), false, false, false, false, nil, 0)
}
```

`isRecursive=false` does re-list the folder and refuse a non-empty one, so the `countItems` call is not the window:

```go
// filer_delete_entry.go:83
entries, _, err := f.ListDirectoryEntries(ctx, entry.FullPath, lastFileName, includeLastFile, PaginationSize, "", "", "")
...
// filer_delete_entry.go:89
if lastFileName == "" && !isRecursive && len(entries) > 0 {
	return fmt.Errorf("%s: %s", MsgFailDelNonEmptyFolder, entry.FullPath)
}
```

That refusal fires only when the listing comes back non-empty. When it comes back empty the loop exits and control reaches an unconditional bulk delete of the folder's children:

```go
// filer_delete_entry.go:134
if storeDeletionErr := f.Store.DeleteFolderChildren(ctx, entry.FullPath); storeDeletionErr != nil {
```

So a row committed between `ListDirectoryEntries` returning empty at `:83` and `DeleteFolderChildren` executing at `:134` is removed by the bulk delete. The client has already been told the write succeeded.

### Observed, from a single job's own artifacts

ClickHouse CI runs stateless tests against SeaweedFS and verifies each upload with a HEAD immediately after it. That verification is what surfaced this. Millisecond ordering for one loss, filer store leveldb, single host:

```
16:18:17.675651  EmptyFolderCleaner: deleting empty folder /buckets/test/test/jys   (seaweedfs.log)
16:18:17.676290  PUT of the object returns success, 282 bytes                       (client-side upload log)
16:18:17.678805  verifying HEAD of that object -> 404                               (client log)
16:18:17.681060  client logs "Nothing to abort", i.e. it issued no DELETE           (client log)
```

The object was never deleted by the client, and it was not eventually-consistent: it stayed missing.

Why this job hits it: object keys are `<prefix>/<3 chars>/<29 chars>`, so the bucket holds a few thousand shallow folders that empty and refill continuously. In that one job 34,819 of 45,578 cleaner log lines were deletions, peaking at 3,238 deletions per minute, and the specific folder above was deleted four times.

### Reproduction status, stated as a negative

I could not land a loss synthetically and I would rather say so than present a weak reproducer. Against real 4.41 I drove 35,464 cleaner folder deletions against roughly 60,000 acknowledged PUTs without losing one. Two collisions I did land were both correctly refused at `filer_delete_entry.go:89` with `fail to delete non-empty folder`, which is direct evidence that the re-list guard works and that the remaining window is narrow.

My reading is that the window is a couple of store round-trips wide, so it needs real store latency to open. My harness was single-host leveldb; every prior report on this cleaner (#8963, #8035, #6515) involved MariaDB, Cassandra or a network filer. I am happy to keep working a reproducer against a MySQL or Postgres filer store if that would be more useful than the ordering above.

Three things that make a naive harness read as "no bug", in case they save someone time:

1. The default 2m `empty_folder_cleanup_delay` means a short write window never pops the queue at all, so deletions stay 0.
2. Writers protect the folders they are trying to race. Any PUT fires `OnCreateEvent`, which cancels the queued cleanup for that folder, so hammering PUTs measures nothing by construction.
3. `cleanupProcessor` ticks on a fixed 30s `DefaultProcessorSleep` and the pop loop drained 5,000 folders in about 0.6s in my runs, so a load generator started after tick detection misses the window entirely.

### Note on the ClickHouse side

Nothing is blocked on this. ClickHouse CI has mitigated it by pushing `filer.options.s3.empty_folder_cleanup_delay` past a job's lifetime, so the cleaner never becomes eligible while the suite runs (ClickHouse/ClickHouse#114660). I am reporting it because the underlying race can drop an acknowledged write for any S3 client whose key layout produces folders that empty and refill.

chrislusf:
Fixes #10779

`doBatchDeleteFolderMetaAndData` lists a folder's children and returns `fail to delete non-empty folder` if it finds any, then calls `Store.DeleteFolderChildren` unconditionally. On the non-recursive path that bulk sweep has nothing legitimate to remove — it is only reached once the listing came back empty, so the only rows it can possibly delete are ones inserted after the check.

The S3 empty-folder cleaner deletes through exactly this path. An object written between the listing and the sweep loses its entry after the write was already acknowledged, and the failure is silent on both sides: the client has its 200, and the cleaner logs an ordinary `deleting empty folder`. The chunks leak too, because the cleaner passes `shouldDeleteChunks=false` and the empty listing collected nothing, so the needles stay on the volume unreferenced until someone runs `volume.fsck`.

The window is reachable on any workload that shards object keys over a small fixed set of shallow prefixes. A layout like `<prefix>/<3 chars>/<random>` spreads objects uniformly across a few thousand folders, so once the live object count drops below the folder count, nearly every delete empties a folder and nearly every write revives one that has been idle long enough to be queued for cleanup. The set of folders pending deletion and the set about to be written are then the same set, and a single run can log tens of thousands of folder deletions, each carrying the window.

ClickHouse hit exactly this in their stateless CI. Their S3 disk generates keys as `<prefix>/[a-z]{3}/[a-z]{29}`, and one job logged 34,819 folder deletions at a peak of 3,238/min. The symptom is `Object <key> in bucket <bucket> suddenly disappeared`, raised by `s3_check_objects_after_upload` doing a HEAD immediately after upload. They worked around it in ClickHouse/ClickHouse#114660 by pushing `empty_folder_cleanup_delay` out to 24h.

That workaround does not fix the bug. The delay changes how often the deletion runs, not what it does when it runs, so a long delay defers the problem rather than removing it.

The `.uploads` guard added in #10273 covered one subtree where this first surfaced. The window is not specific to multipart staging — every object folder has it.

## Fix

Sweep only when the delete is recursive, or when the whole-bucket shortcut skipped the listing and relies on the sweep to drop the table.

## Reproduction

A stress harness driving the cleaner's exact call against a concurrent writer, on a real `Filer` over leveldb2, destroyed 33 entries out of 105,521 creates across 7,554 folder deletions. End to end over S3 against a stock `weed server` in that CI's configuration, 9 objects were destroyed out of 456,762 PUTs, each loss matching a `deleting empty folder` line for that same folder in the same second. With this change the same harness reports 0 losses over 291,628 creates and 13,243 folder deletions.

The test added here is the deterministic form of that: it hooks the store listing to create an entry inside the window, then asserts the entry survives the folder delete. It fails on master and passes with the fix. It also pins what this change does *not* fix — the folder entry is still removed, so the surviving entry is reachable by path but absent from listings.

## Follow-up

That remaining exposure is handled in #10783, stacked on this branch: the cleaner records the folders it deleted each pass and restores any that turn out to hold entries. Closing it inside this function instead would need the emptiness check and the entry delete to be atomic, which is per-store — one statement on SQL, a Lua script on redis, but not expressible on leveldb2, where children hash by `md5(dir)` and the directory entry by `md5(parent)` into different partitions.

https://claude.ai/code/session_01HdLXMUopwgofPb1ZEmiE6r

### DIFF
--- a/weed/filer/filer_delete_entry.go
+++ b/weed/filer/filer_delete_entry.go
@@ -78,7 +78,8 @@ func (f *Filer) doBatchDeleteFolderMetaAndData(ctx context.Context, entry *Entry
 	var chunksToDelete []*filer_pb.FileChunk
 	lastFileName := ""
 	includeLastFile := false
-	if !isDeletingBucket || !f.Store.CanDropWholeBucket() {
+	listedChildren := !isDeletingBucket || !f.Store.CanDropWholeBucket()
+	if listedChildren {
 		for {
 			entries, _, err := f.ListDirectoryEntries(ctx, entry.FullPath, lastFileName, includeLastFile, PaginationSize, "", "", "")
 			if err != nil {
@@ -131,8 +132,12 @@ func (f *Filer) doBatchDeleteFolderMetaAndData(ctx context.Context, entry *Entry
 
 	glog.V(3).InfofCtx(ctx, "deleting directory %v delete chunks: %v", entry.FullPath, shouldDeleteChunks)
 
-	if storeDeletionErr := f.Store.DeleteFolderChildren(ctx, entry.FullPath); storeDeletionErr != nil {
-		return fmt.Errorf("filer store delete: %w", storeDeletionErr)
+	// a non-recursive delete already proved the folder empty above, so sweeping the
+	// children now can only remove entries that raced in after that listing
+	if isRecursive || !listedChildren {
+		if storeDeletionErr := f.Store.DeleteFolderChildren(ctx, entry.FullPath); storeDeletionErr != nil {
+			return fmt.Errorf("filer store delete: %w", storeDeletionErr)
+		}
 	}
 
 	f.NotifyUpdateEvent(ctx, entry, nil, shouldDeleteChunks, isFromOtherCluster, signatures)
--- a/weed/filer/leveldb2/empty_folder_race_test.go
+++ b/weed/filer/leveldb2/empty_folder_race_test.go
@@ -2,11 +2,13 @@ package leveldb
 
 import (
 	"context"
+	"errors"
 	"os"
 	"testing"
 
 	"github.com/seaweedfs/seaweedfs/weed/filer"
 	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
 	"github.com/seaweedfs/seaweedfs/weed/util"
 )
 
@@ -64,5 +66,13 @@ func TestNonRecursiveFolderDeleteKeepsRacingChild(t *testing.T) {
 
 	if _, err := testFiler.FindEntry(ctx, child); err != nil {
 		t.Errorf("entry created during the folder delete was removed: %v", err)
+	}
+
+	// The folder entry itself still goes, so the surviving entry is reachable by
+	// path but absent from listings until the folder comes back. Pinning that here
+	// keeps the remaining exposure visible; tighten it if the two steps ever become
+	// atomic.
+	if _, err := testFiler.FindEntry(ctx, dir); !errors.Is(err, filer_pb.ErrNotFound) {
+		t.Errorf("folder entry should still be removed, got %v", err)
 	}
 }
--- a/weed/filer/leveldb2/empty_folder_race_test.go
+++ b/weed/filer/leveldb2/empty_folder_race_test.go
@@ -1,0 +1,68 @@
+package leveldb
+
+import (
+	"context"
+	"os"
+	"testing"
+
+	"github.com/seaweedfs/seaweedfs/weed/filer"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/util"
+)
+
+// listHookStore runs a hook once, right after a directory listing returns, so a
+// test can act inside the window between a delete's emptiness check and the
+// removal of the folder.
+type listHookStore struct {
+	filer.FilerStore
+	hook  func()
+	fired bool
+}
+
+func (s *listHookStore) ListDirectoryPrefixedEntries(ctx context.Context, dirPath util.FullPath, startFileName string, includeStartFile bool, limit int64, prefix string, eachEntryFunc filer.ListEachEntryFunc) (string, error) {
+	lastFileName, err := s.FilerStore.ListDirectoryPrefixedEntries(ctx, dirPath, startFileName, includeStartFile, limit, prefix, eachEntryFunc)
+	if s.hook != nil && !s.fired {
+		s.fired = true
+		s.hook()
+	}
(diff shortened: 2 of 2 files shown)


chrislusf:
Thanks both — two findings here, one addressed and one declined.

**CodeRabbit, on keeping the directory entry (`filer_delete_entry.go:135-140`).** Correct, and the PR description says so: this change stops the child being destroyed but the directory entry still goes, leaving the child reachable by path and missing from listings.

I've extended `TestNonRecursiveFolderDeleteKeepsRacingChild` to assert exactly that, so the remaining exposure is pinned by a test instead of only described in prose. It asserts the folder entry *is* removed, since that is the honest current behaviour — tighten it to the opposite when the two steps become atomic.

On making them atomic here: not achievable in this function. The check and the delete are separate store operations and the stores differ in what they can promise. SQL can do it in one statement with a `NOT EXISTS` guard, redis with a Lua script, but leveldb2 hashes children by `md5(dir)` and the directory entry by `md5(parent)` — different partitions, so there is no single atomic batch to put them in. Doing this properly means a per-store capability alongside `CanDropWholeBucket` with a fallback path, which is a much larger change than the data-loss fix it would be riding along with.

The practical answer is #10783, stacked on this branch: it records the folders deleted in each pass and restores any that turn out to hold entries, deferred a pass so it lands after the writer's parent-lookup-to-insert gap.

**Greptile, on failed expiration cleanup (`filer_delete_entry.go:141`).** Real, and I'm accepting it rather than fixing it.

The scenario needs an expired TTL entry whose best-effort deletion in `doListDirectoryEntries` failed. It is then filtered from the listing, the folder reads as empty, and with this change its metadata survives the parent's removal instead of being swept.

Two reasons not to keep the sweep for it. Cleaning up a failed TTL deletion is the TTL path's job, not a side effect of an unrelated folder delete — and the chunks already leak in that case regardless, since expired entries never reach `chunksToDelete`. And it isn't permanent: the entry stays a candidate for deletion on any later listing of that directory, which resumes as soon as a write recreates the folder.

Weighed against the alternative — a bulk delete acting on a stale listing, which is what silently destroys acknowledged writes — leaving orphaned metadata from a failed TTL delete is the better trade.

PR #10783, opened 2026-08-17
filer: restore a folder that received an entry while it was deleted

chrislusf:
Follows #10782, now merged, and rebased onto it.

#10782 stopped the empty-folder cleaner from destroying an object that arrives between the emptiness check and the delete. What it left behind is the milder half of the same race: the entry survives, but the directory holding it does not.

The cleaner checks that a folder is empty and then deletes it, and those two steps are not atomic. An entry created in between ends up reachable by its own path — `FindEntry` keys on the full path and never consults the parent — but absent from `ListObjectsV2` and every other traversal, because the directory that would have carried it is gone. It stays that way until some later write into the same folder happens to recreate the parent.

That is why this is a follow-up rather than part of the fix: a HEAD still succeeds, so read-after-write verification never sees it. It shows up as an object missing from a listing, and for disk layouts that derive their state from listing the bucket rather than from local metadata, as data that has effectively vanished.

## The metadata stream already knows

A folder is recorded before it is deleted. For an entry to be orphaned it has to be created after the emptiness check inside the delete, which is after that record — so its create event always arrives while the folder is still being watched, and that event names the exact directory.

So the cleaner does not go looking. `OnCreateEvent` matches the directory against the recently deleted folders, and a hit means that folder is known to need putting back. No polling, no listing per folder per pass, and no guessing when the racing write will land.

The folder is then restored with the mode, uid, gid and user/group names read from it *before* it was deleted, so a restore cannot hand back a directory that grants access the original denied. Missing ancestors are restored too, shallowest first, since the eager parent cascade may have taken them in the same pass. One listing is done before restoring, to skip the work when the entry has since gone away again.

The observation window is no longer a guess at the race. It bounds how far behind the event stream may run before a folder stops being watched — relevant because a write on a peer filer reaches this one through the aggregator.

Remaining limits, stated rather than implied: the watched set is in memory, so a filer restart between the delete and the restore leaves the orphan until a later write recreates the parent, and the set is bounded, with anything past the bound logged rather than dropped silently. In both cases the entry itself is never lost, only its place in listings.

## The redis stores need the opposite fix

On leveldb and the SQL stores a child records its own directory, so a listing is derived and the orphan is recoverable. On redis it is not: `ListDirectoryEntries` is nothing but a read of `genDirectoryListKey(dir)`, and `DeleteEntry` opened by deleting that key outright.

So the racing entry lost its membership immediately, and no repair could reach it. The listing is now dropped in `DeleteFolderChildren`, alongside the children it describes, and left alone in `DeleteEntry`. The plain redis store was leaking the key altogether. redis3 needed two changes: `removeChildren` clears the skip list nodes but not the list itself, and `removeChild` was writing an empty list back rather than deleting it, so an emptied directory left a header that nothing would collect once the sweep was skipped.

## Tests

`TestEmptyFolderCleaner_restoreFoldersWrittenDuringDelete` covers the event path: a create event for the deleted folder puts it back, one for a different folder is ignored, an entry that went away again is not restored, observation expires, a failed restore is retried, and a delete that reports failure is still observed.

`TestEnsureDirectoryEntryRestoresRacingParent` drives the real thing over leveldb2 — it creates an entry inside the delete window, asserts the folder has dropped out of its parent's listing, then asserts the restore brings it back. `TestEnsureDirectoryEntryRestoresExactAttributes` asserts a `0700` directory returns as `0700` with its owner intact.

`TestDeleteEntryKeepsChildListing` (redis2, redis3), `TestDeleteFolderChildrenRemovesTheListing` (redis) and `TestRemovingTheLastChildDropsTheListing` (redis3) cover the store changes. They run against a live redis under `RUN_REDIS_TESTS=1`; verified against redis 7 both ways.

https://claude.ai/code/session_01HdLXMUopwgofPb1ZEmiE6r

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved folder deletion handling when new entries are created during cleanup.
  - Automatically restores affected folders while preserving their original permissions and ownership.
  - Added retries and safeguards for incomplete or interrupted cleanup operations.
  - Corrected directory listing cleanup so child listings remain available until their contents are removed, then are deleted completely.
  - Limited retained deletion tracking to prevent excessive resource usage.

- **Tests**
  - Added coverage for concurrent folder changes, restoration, metadata preservation, and directory-list cleanup.

### DIFF
--- a/weed/filer/empty_folder_cleanup/empty_folder_cleaner.go
+++ b/weed/filer/empty_folder_cleanup/empty_folder_cleaner.go
@@ -301,12 +301,29 @@ func (efc *EmptyFolderCleaner) restoreFoldersWrittenDuringDelete() {
 	efc.deletedDropped = 0
 	var restore []*deletedFolder
 	for path, folder := range efc.deleted {
-		switch {
-		case folder.writtenTo:
+		// The window applies whatever the folder's state is. Checking writtenTo first
+		// would keep a folder whose restore keeps failing forever, re-counting it on
+		// every pass.
+		if time.Since(folder.deletedAt) >= DefaultObservationWindow {
+			delete(efc.deleted, path)
+			continue
+		}
+		if folder.writtenTo {
 			restore = append(restore, folder)
 			delete(efc.deleted, path)
-		case time.Since(folder.deletedAt) >= DefaultObservationWindow:
-			delete(efc.deleted, path)
+		}
+	}
+	// A cascade takes ancestors along with the folder. Rebuild those from what they
+	// were too: leaving them to the descendant's restore would mint them from the
+	// descendant's own attributes, handing back access the ancestor did not grant.
+	for i := 0; i < len(restore); i++ {
+		ancestor, _ := util.FullPath(restore[i].path).DirAndName()
+		for ancestor != "" && ancestor != "/" {
+			if folder, found := efc.deleted[ancestor]; found {
+				restore = append(restore, folder)
+				delete(efc.deleted, ancestor)
+			}
+			ancestor, _ = util.FullPath(ancestor).DirAndName()
 		}
 	}
 	efc.mu.Unlock()
@@ -331,18 +348,22 @@ func (efc *EmptyFolderCleaner) restoreFoldersWrittenDuringDelete() {
 			retry = append(retry, restore[i:]...)
 			break
 		}
-		// The entry may have been removed again between the event and now, in which
-		// case there is nothing to hold and the folder can stay gone.
-		count, err := efc.countItems(ctx, folder.path)
-		if err != nil {
-			glog.V(2).Infof("EmptyFolderCleaner: cannot count %s before restoring it: %v", folder.path, err)
-			retry = append(retry, folder)
-			continue
+		// An event named this folder, but the entry may have been removed again since,
+		// in which case there is nothing to hold and it can stay gone. Ancestors pulled
+		// in above carry no event and are rebuilt regardless, since the folder below
+		// them needs them.
+		if folder.writtenTo {
+			count, err := efc.countItems(ctx, folder.path)
+			if err != nil {
+				glog.V(2).Infof("EmptyFolderCleaner: cannot count %s before restoring it: %v", folder.path, err)
+				retry = append(retry, folder)
+				continue
+			}
+			if count == 0 {
+				continue
+			}
 		}
-		if count == 0 {
-			continue
-		}
-		glog.V(1).Infof("EmptyFolderCleaner: restoring %s, %d entries arrived while it was being deleted", folder.path, count)
+		glog.V(1).Infof("EmptyFolderCleaner: restoring %s, written to while it was being deleted", folder.path)
 		if err := efc.filer.EnsureDirectoryEntry(ctx, util.FullPath(folder.path), folder.attrs); err != nil {
 			glog.V(2).Infof("EmptyFolderCleaner: failed to restore %s: %v", folder.path, err)
 			retry = append(retry, folder)
@@ -354,13 +375,35 @@ func (efc *EmptyFolderCleaner) restoreFoldersWrittenDuringDelete() {
 	}
 	efc.mu.Lock()
 	for _, folder := range retry {
-		if _, found := efc.deleted[folder.path]; !found && len(efc.deleted) >= DefaultMaxDeletedKept {
-			efc.deletedDropped++
-			continue
+		if _, found := efc.deleted[folder.path]; !found {
+			efc.makeRoomForDeletedLocked()
 		}
 		efc.deleted[folder.path] = folder
 	}
 	efc.mu.Unlock()
+}
+
+// makeRoomForDeletedLocked drops the oldest of a small sample when the set is full.
+// The newest folders are the ones whose race is still live, so they must not be the
+// ones given up; sampling keeps this cheap under heavy deletion rates.
+func (efc *EmptyFolderCleaner) makeRoomForDeletedLocked() {
+	if len(efc.deleted) < DefaultMaxDeletedKept {
+		return
+	}
+	const sampleSize = 32
+	oldestPath, seen := "", 0
+	for path, folder := range efc.deleted {
+		if oldestPath == "" || folder.deletedAt.Before(efc.deleted[oldestPath].deletedAt) {
+			oldestPath = path
+		}
+		if seen++; seen >= sampleSize {
+			break
+		}
+	}
+	if oldestPath != "" {
+		delete(efc.deleted, oldestPath)
+		efc.deletedDropped++
+	}
 }
 
 // executeCleanup performs the actual cleanup of an empty folder
@@ -460,11 +503,8 @@ func (efc *EmptyFolderCleaner) executeCleanup(folder string, triggeredBy string)
 	if efc.deleted == nil {
 		efc.deleted = make(map[string]*deletedFolder)
 	}
-	if len(efc.deleted) < DefaultMaxDeletedKept {
-		efc.deleted[folder] = &deletedFolder{path: folder, attrs: attrs, deletedAt: time.Now()}
-	} else {
-		efc.deletedDropped++
-	}
+	efc.makeRoomForDeletedLocked()
+	efc.deleted[folder] = &deletedFolder{path: folder, attrs: attrs, deletedAt: time.Now()}
 	efc.mu.Unlock()
 
 	glog.Infof("EmptyFolderCleaner: deleting empty folder %s (triggered by %s)", folder, triggeredBy)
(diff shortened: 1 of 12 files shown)


chrislusf:
Thanks — both findings were valid and are fixed in ef7cc2b.

**Failed repairs are forgotten (`empty_folder_cleaner.go:275-282`).** Right: the list was drained before the work, so a transient store error on either `countItems` or `EnsureDirectoryEntry` lost the folder permanently, leaving its entries out of listings until some later write recreated the parent — the exact outcome this pass exists to prevent.

Failures now go back for the next pass, subject to the same cap, with the overflow accounted for in `deletedDropped` rather than dropped silently. Covered by a new subtest where the first check fails and the second succeeds, asserting nothing is restored on the failing pass and the folder is restored on the next one.

**Repair widens directory access (`filer.go:380-381`).** Also right, and the more serious of the two. The original attributes went with the deleted entry, and minting a fixed `0755` with a zero owner meant a directory that had been private could come back world-readable and root-owned.

`EnsureDirectoryEntry` now walks up to the nearest ancestor still present and takes its mode, uid, gid and user/group names, falling back to `0755` only when nothing above it survives — which requires the whole chain including the bucket to be gone. Walking up rather than reading only the immediate parent matters because the eager parent cascade can have taken several levels in the same pass. `TestEnsureDirectoryEntryInheritsAncestorOwnership` covers it with a `0700` ancestor owned by uid 4242.

One caveat worth recording, unrelated to these comments: this restore is store-dependent. On leveldb and SQL, and on redis v1, the racing entry keeps its listing membership so the check sees it. On redis2 and redis3, `DeleteEntry` opens by deleting the directory's whole index (`Del(genDirectoryListKey(fullpath))`), so the entry's membership is gone before the check runs — the repair cannot see it and would not restore visibility if it did. Fixing that means moving the index cleanup into `DeleteFolderChildren` so `DeleteEntry` stops wiping it unconditionally, which is the same argument as #10782 one level down. Happy to add it here or keep it separate.

chrislusf:
All six addressed in a10192a. Four were fixes, one changed the design, one I can only narrow rather than close.

**1 and 2 — restoring from ancestor metadata, and swallowed lookup errors.** Both correct, and they had a common root: the restore was reconstructing attributes it should never have had to guess. It now reads the folder's own mode, uid, gid and user/group names *before* deleting it, and puts exactly those back. The `0111` OR is gone with it, so a `0700` directory returns as `0700`, not `0711` — the new `TestEnsureDirectoryEntryRestoresExactAttributes` pins that. The ancestor walk that swallowed non-`ErrNotFound` errors is deleted outright rather than patched.

**3 — the deferred check is not a delay.** This is the one that changed the design, and you're right that a ticker send coalesces: `processCleanupQueue` running past its interval means the next pass starts immediately, so "next pass" bounds nothing.

A folder is now kept under observation for a bounded wall-clock window (`DefaultRestoreCheckWindow`, 2m) and re-checked on every pass until it expires, instead of being checked once and discarded. That gives a real lower bound independent of tick spacing.

I want to be straight that this narrows the exposure rather than closing it, and I've said so in the code comment rather than implying otherwise. A writer stalled past the window still slips through. Closing it needs the emptiness check and the delete to be atomic, or a generation protocol — per-store work that belongs on its own, not smuggled into this.

**4 — partial delete failures bypass restoration.** Correct, and the redis ordering makes it concrete. The folder is now recorded *before* the delete, so a failure return no longer decides whether it is observed. Covered by "a delete that reports failure is still observed".

**5 — a restart loses pending repairs.** Correct, and not fixed. Candidates are in memory, so a restart between delete and restore leaves the orphan until a later write recreates the parent. Durable tracking is a real design change and I would rather not bolt it on here. Worth noting the underlying entry is never lost — it stays readable by path, and this only affects when it reappears in listings.

**6 — redis3 leaks an index key per cleaned folder.** Correct, and my own regression: with the sweep skipped and `DeleteEntry` no longer touching the listing, nothing collected the emptied header.

Fixed at the root rather than at the delete: `removeChild` was writing an empty serialized list back, so the header survived removal of the last child. It now deletes the key when the list empties, which also matches what the ZSet and SET stores get from redis for free. `TestRemovingTheLastChildDropsTheListing` covers it, verified failing without the change.

Full filer, s3api and server suites pass, with the redis tests run against redis 7 rather than skipped.

chrislusf:
Rebased onto master now that #10782 has merged, retargeted to `master`, and reworked to detect the race from the metadata stream instead of polling for it.

**The design change.** Re-checking every deleted folder on a timer was the wrong instrument: it cost a listing per folder per pass, and no timer can be more than a guess about when the racing write lands — which is what the earlier review of the "next pass" delay was really pointing at.

The event stream already carries the answer. A folder is recorded before it is deleted, so an entry that can be orphaned is necessarily created after that record, and its create event names that exact directory. `OnCreateEvent` now matches against the recently deleted folders, so the cleaner is *told* which folder needs putting back rather than inferring it. The window stops being a guess at the race and becomes what it should have been: how far behind the event stream may run before a folder stops being watched.

**On the standing inline comments**, all from the previous round and all addressed in 7b6356b before this change:

- *Failed repairs are forgotten* — a failed check or restore now keeps the folder instead of dropping it.
- *Repair widens directory access* / *Repair still widens access* / *Propagate ancestor lookup errors* — these had a common root, so they are fixed together. The restore no longer reconstructs attributes from an ancestor at all. It reads the folder's own mode, uid, gid and user/group names before deleting and puts exactly those back, and the `0111` OR is gone, so `0700` returns as `0700`. The ancestor walk that swallowed non-`ErrNotFound` errors is deleted rather than patched.
- *Restart discards failed repairs* — correct, and still not fixed. The watched set is in memory. It is now stated as a limit in the PR description rather than left implicit. The entry itself is never lost; only its place in listings is delayed until a later write recreates the parent.

Full filer, s3api and server suites pass, with the redis tests run against redis 7 rather than skipped.

chrislusf:
All five addressed in 79b8310 and 8042c1b. Each was written against the code before the event-driven rewrite, but every one still held against the current version, so thanks — the rewrite did not retire any of them.

**Retained folders never expire** (`empty_folder_cleaner.go`). Still true after the rewrite, by a different route: the written-to check ran before the age check, so a folder whose `EnsureDirectoryEntry` kept failing was picked up, retried, put back with its flag intact, and counted again on every pass forever. The window now applies first, whatever state the folder is in, so a folder that cannot be restored ages out instead of accumulating work. Covered by "a folder whose restore keeps failing stops being observed".

**The cap drops the newest deletions.** Also still true — the slice trim became a map insert that turned away the folder being recorded. That is backwards for exactly the reason given: the newest folder is the one whose race is still live, while the older entries are already close to ageing out. It now gives up an older one, chosen as the oldest of a 32-entry sample, which keeps the cost flat under the deletion rates that make the cap reachable at all.

**Nil entry from `FindEntry`** (`filer.go`). Correct — `DirectoryAttributes` dereferenced it directly, and the cleaner calls this right after the count check, so the folder really can be gone by then. Returns `filer_pb.ErrNotFound` instead, which the cleaner already treats as a reason not to delete.

**`Mode.Perm()` drops setgid, setuid and sticky.** Correct, and a nastier version of the widening problem than the earlier round: a setgid directory would come back without it and silently stop propagating group ownership. The mode is now taken whole apart from the type bits.

**Ancestor permissions remain unrestored** (`empty_folder_cleaner.go`). Correct. Leaving an ancestor to the descendant's restore meant `ensureParentDirectoryEntry` minted it from the descendant's attributes plus execute bits. Restoring a folder now pulls in any ancestor the same cascade deleted and rebuilds it from its own recorded attributes, shallowest first, so the descendant never has to invent one. Ancestors carry no create event, so they skip the entry count and are rebuilt because the folder below them needs them. Covered by "an ancestor taken by the cascade is restored with its own attributes", which asserts both the ordering and the attributes.

**Empty-list deletion is not retry-safe** (`redis3/kv_directory_children.go`, separate commit). Correct: if the delete fails after the name is gone, the retry finds nothing to remove, reports no changes and returns before reaching the delete, so the header stays for good. That path now takes the header when the list is empty, which makes it self-healing on the next removal in that directory.

Full filer, s3api and server suites pass, with redis tests run against redis 7.

azat:
@chrislusf thanks for fixing this promptly! But I have a question - is it still possible to see empty response for `ListObjectV2` even if there are some children? (If cleaner removes the directory, but after/in-the-middle, new child has been created)

chrislusf:
Yes — with what shipped in 4.42 the restore waits for the create event and then the cleaner's next pass, so a listing can miss the object for up to 30s, and a filer restart in between loses the observation entirely. #10799 closes it: the cleaner re-lists a folder after deleting it, and writers now create the parent directories after the entry rather than before, so whichever of the two acts second sees what the other did.

Issue #10784, opened 2026-08-17
Request: Add comparison with RustFS to the README


ylw510:
I noticed that the README currently includes feature comparisons with MinIO, Ceph, and GlusterFS. However, I didn't see a similar comparison with [RustFS](https://github.com/rustfs/rustfs), which is another emerging open‑source object storage solution written in Rust. Could we add a comparison with RustFS?

chrislusf:
Added in #10788.

I read through the RustFS tree at 1.0.0-rc.2 first. It turns out to be a MinIO reimplementation in Rust rather than a separate design: byte-compatible on-disk format, same `xl.meta` container and magic, same `format.json` erasure layout, same default parity table, and its own docs describe one-way MinIO to RustFS drive migration.

Because the storage model is the same one, a standalone section would have restated the MinIO section almost line for line. So RustFS is folded into it instead, and that section is now "Compared to MinIO, RustFS", the way "Compared to GlusterFS, Ceph" already works. RustFS also gets its own row in the comparison table.

Two of the existing points picked up detail while merging, both about the shared architecture. The meta file is written on every drive of the erasure set, and a tags-only or retention-only change rewrites it on all of them, so the write amplification does not shrink with object size. And the storage layout requirement is now concrete: erasure sets are 2 to 16 drives and must divide the drive list symmetrically, and capacity moves a whole pool at a time in either direction.

ylw510:
> Added in [#10788](https://github.com/seaweedfs/seaweedfs/pull/10788).
> 
> I read through the RustFS tree at 1.0.0-rc.2 first. It turns out to be a MinIO reimplementation in Rust rather than a separate design: byte-compatible on-disk format, same `xl.meta` container and magic, same `format.json` erasure layout, same default parity table, and its own docs describe one-way MinIO to RustFS drive migration.
> 
> Because the storage model is the same one, a standalone section would have restated the MinIO section almost line for line. So RustFS is folded into it instead, and that section is now "Compared to MinIO, RustFS", the way "Compared to GlusterFS, Ceph" already works. RustFS also gets its own row in the comparison table.
> 
> Two of the existing points picked up detail while merging, both about the shared architecture. The meta file is written on every drive of the erasure set, and a tags-only or retention-only change rewrites it on all of them, so the write amplification does not shrink with object size. And the storage layout requirement is now concrete: erasure sets are 2 to 16 drives and must divide the drive list symmetrically, and capacity moves a whole pool at a time in either direction.

[...]

PR #10788, opened 2026-08-17
readme: fold RustFS into the MinIO comparison

chrislusf:
Adds RustFS to the file system comparison, as requested in #10784.

RustFS is a MinIO reimplementation in Rust and keeps MinIO's storage model down to a byte-compatible on-disk format, so rather than a standalone section this folds it into the MinIO one, the way "Compared to GlusterFS, Ceph" already does. The shared paragraphs move to present tense since RustFS is still developed, and the note about MinIO being unmaintained stays.

Two points gained detail while merging, both about the shared architecture:

- The meta file is written on every drive of the erasure set, and a tags-only or retention-only change rewrites it on all of them, so the write amplification does not shrink with object size.
- The storage layout requirement is concrete: erasure sets are 2 to 16 drives and must divide the drive list symmetrically, and capacity moves a whole pool at a time in either direction.

Claims come from the RustFS tree at 1.0.0-rc.2, not from its marketing: the on-disk format, erasure layout, set-size and symmetry rules, and the metadata write fan-out are all in its own source and architecture docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,8 +68,7 @@ Table of Contents
     * [Compared to GlusterFS, Ceph](#compared-to-glusterfs-ceph)
     * [Compared to GlusterFS](#compared-to-glusterfs)
     * [Compared to Ceph](#compared-to-ceph)
-    * [Compared to Minio](#compared-to-minio)
-    * [Compared to RustFS](#compared-to-rustfs)
+    * [Compared to MinIO, RustFS](#compared-to-minio-rustfs)
 * [Dev Plan](#dev-plan)
 * [Installation Guide](#installation-guide)
 * [Disk Related Topics](#disk-related-topics)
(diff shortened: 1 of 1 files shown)


chrislusf:
Went through the review comments against the RustFS source at 1.0.0-rc.2. Took four, skipped two.

Fixed in 2c058b3:

- Single drive really does run without parity. `FormatV3::new` picks `FormatBackend::ErasureSingle` when the set length is 1, and `default_parity_count(1)` is 0, so "erasure codes everything on write" was wrong for that case. Now scoped to multi-drive deployments.
- The protocol list overstated support, and by more than the review caught. `rustfs/Cargo.toml` has `default = ["ftps", "webdav"]`, so Swift needs a non-default build too, not just SFTP. Reworded for both.
- Pinned the version to 1.0.0-rc.2 so the claim dates itself.
- Table cell now says "separate meta file per drive for each file", which is the part that was actually understated.

Skipped:

- Changing `### Compared to RustFS ###` to drop the trailing hashes. Every heading in this README is closed ATX, so following the linter here would make the one new heading the odd one out. The lint rule is worth a separate pass over the whole file if we want it.
- Adding upstream citations to this section alone. None of the other comparison sections carry them, and the version pin covers the staleness concern that motivated the request.

PR #10797, opened 2026-08-17
go 1.26

chrislusf:
The etcd client v3 3.7.1 bump in #10789 moves the go.mod directive from 1.25.8
to 1.26. Every builder image we pin is still `golang:1.25`/`1.24`, and the
official Go images set `GOTOOLCHAIN=local`, so they cannot pick up a newer
toolchain on their own:

    go: go.mod requires go >= 1.26 (running go 1.25.13; GOTOOLCHAIN=local)

That is what fails the PostgreSQL Gateway Tests on #10789
(https://github.com/seaweedfs/seaweedfs/actions/runs/32072598094/job/95518885757),
and it would fail every other containerized job the same way once the directive
lands.

Same sweep as the earlier `go 1.25` bump (3d9f7f6f8), plus the spots it missed:
the `container: image: golang:1.24-alpine` jobs in kafka-tests.yml, the
test/postgres and telemetry/server Dockerfiles, tls-rotation and
vacuum-integration, and test/fuse_integration's GO_VERSION.

Language-level fallout from the same directive flip was handled separately in #10794.

Verified against #10789's go.mod: `go mod download` fails on `golang:1.25-alpine`
and succeeds on `golang:1.26-alpine` (go1.26.6).

## Summary by CodeRabbit

* **Chores**
  * Upgraded build, test, and development environments to Go 1.26.
  * Updated automated integration, compatibility, and end-to-end test workflows to run with Go 1.26.
  * Updated containerized testing and service build environments to use Go 1.26 base images.
  * Refreshed testing documentation and configuration to reference Go 1.26.

### DIFF
--- a/.github/workflows/ec-integration-tests.yml
+++ b/.github/workflows/ec-integration-tests.yml
@@ -30,7 +30,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v7
       with:
-        go-version: ^1.25
+        go-version: ^1.26
       id: go
 
     - name: Check out code into the Go module directory
--- a/.github/workflows/kafka-quicktest.yml
+++ b/.github/workflows/kafka-quicktest.yml
@@ -42,7 +42,7 @@ jobs:
     - name: Set up Go 1.x
       uses: actions/setup-go@v7
       with:
-        go-version: ^1.25
+        go-version: ^1.26
         cache: true
         cache-dependency-path: |
           **/go.sum
--- a/.github/workflows/kafka-tests.yml
+++ b/.github/workflows/kafka-tests.yml
@@ -43,7 +43,7 @@ jobs:
       matrix:
         container-id: [unit-tests-1]
     container:
-      image: golang:1.24-alpine
+      image: golang:1.26-alpine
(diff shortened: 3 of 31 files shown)


chrislusf:
Addressed:

- `test/foundationdb/Dockerfile.build`: stale `# Using Go 1.24 to match project requirements` comment now says 1.26.
- `test/s3/iam/README.md`: the sample workflow uses `go-version-file: 'go.mod'` instead of a hardcoded version, and the prerequisite line moves off Go 1.19.
- Swept the two remaining stale prerequisites next to Dockerfiles this PR touches: `seaweedfs-rdma-sidecar/README.md` (was 1.23+) and `test/kafka/kafka-client-loadtest/README.md` (was 1.21+).
- The `test/fuse_integration/README.md` Actions/Docker mismatch was already fixed before the review landed; both examples say 1.26.

Not changing `^1.26` to `'1.26'`. `^` is what this repo has always used for these jobs and it is deliberate: setup-go reads it as `>=1.26.0 <2.0.0`, so these workflows track the newest stable Go. That is also why nothing here broke earlier - `^1.25` was already installing 1.26. Pinning to the minor would freeze them instead.

`docker/Dockerfile.go_build:64` keeps `(Go 1.24+)` - that is the version FIPS 140-3 mode arrived in, not a toolchain pin.

PR #10799, opened 2026-08-17
filer: close the empty-folder race by checking after each mutation

chrislusf:
Follows #10782 and #10783.

#10782 stopped the empty-folder cleaner from destroying an entry that arrives between the emptiness check and the delete. #10783 put back the directory that such an entry loses, by watching the metadata stream for its create event. That narrowed the window rather than closing it, and the restore only ran on the cleaner's next pass, so the entry stayed out of listings for up to a ticker interval.

The question on #10783 was whether `ListObjectsV2` can still return nothing for a prefix that has children. It can, and it is worth being precise about why, because the answer decided the shape of this change.

## Every listing starts at the parent

`normalizePrefixMarker` splits a prefix at its last slash, after trimming a trailing one when there is no marker: the part before becomes the directory to list, the part after a name filter. So `data/abc/` lists `data` looking for `abc`, and the traversal has to see `abc` there before it can descend into it. A directory whose own entry is gone is therefore invisible to every prefix shape, including one that points straight at it, and everything it holds goes with it.

That matters most for object storage disks that derive their filesystem from listings rather than from local metadata. There a missing directory is not one absent key, it is a directory that loads with no files at all. Those layouts also tend to empty and refill the same directory name repeatedly, which is exactly the churn the cleaner acts on.

## Both sides act, then look

The cleaner now re-lists a folder once it has removed it, and puts it back if something is in there.

The writer now creates the directories holding an entry *after* the entry rather than before it. A parent checked first can be taken before the entry lands, and nothing looks again; a parent created afterwards cannot be undone by a delete that was authorised before the insert.

Either alone is another narrowing. Together they close it. Call the writer's insert `Tc` and its parent creation `Tp`, the cleaner's delete `Td` and its re-list `Tr`. The writer creates the parent if `Tp > Td`, the cleaner restores it if `Tr > Tc`. For an entry to be orphaned both have to miss, which needs `Tp < Td` and `Tr < Tc`; with `Tc < Tp` and `Td < Tr` that gives `Tr < Tc < Tp < Td < Tr`. Whichever of the two acts second sees what the other did.

This does not depend on the event stream, on the observation window holding, or on a watched set surviving a restart, which were the three ways the previous fix could still be beaten. The deferred path stays as the retry route for a re-check or a restore that fails. Nothing is added to the write path: it is the same single parent lookup as before, on the other side of the insert.

Going second has a cost. The entry is already stored when the parent fails, and it is left there: deleting it by path would destroy a concurrent create that has already succeeded through the update branch, and nothing can tell the two apart, since the update keeps the inode and crtime and mtime only survives the store to the second. A failed create leaving an entry behind is within what a write is allowed to do; deleting an acknowledged one is not.

That entry is not announced. `MetaAggregator` replicates create events straight into a peer's store when the stores are not shared, so emitting one for a write that failed would put the same parentless row on every filer in the cluster rather than leaving it on the one whose next write into that folder repairs it. The row stays local and unlogged, which also means a later delete of it reaches peers with no matching create - a no-op there, since they never had the row.

The order of the events that are emitted is unchanged: the directory events still precede the entry's, because the entry's is emitted at the end of `CreateEntry` in both arrangements. What does change is that the entry is in the store before its parent's event goes out, so anything that lists a directory on seeing it created can now find it already occupied.

A filer that dies between the two steps also leaves the entry without its parent, where before it would have left an empty directory. The entry stays readable by path, and the next write into the same folder recreates the parent.

## A restore should not widen a directory

A writer creating a missing parent has only the entry it is inserting to go on, so a `0700` directory comes back `0751` - `os.ModeDir | 0640 | 0111`. That is not new, it is what implicit parent creation has always done, but the change above means it now happens on the racing path too.

The cleaner reads the real attributes before deleting, so `EnsureDirectoryEntry` now intersects the restorable bits with the ones it saved rather than replacing them - replacing would grant a bit the current directory denied whenever the two modes are not nested, `0705` against a saved `0750` coming back as `0750`. A directory tightened deliberately since is therefore left alone. It cannot tell a deliberate widening from an inferred one and narrows either way, which is the safe direction to be wrong in. The mode bits a restore carries over are named once now and shared with `DirectoryAttributes`, since `Perm()` alone would drop setgid, setuid and sticky.

## Tests

`TestEmptyFolderCleaner_restoreIfWrittenTo` covers the cleaner half: an entry that landed during the delete is restored without waiting for its event, a restored folder does not cascade into its parent, a folder that really is empty still does, and a re-check that fails leaves the folder under observation.

`TestCreateEntryCreatesParentAfterTheEntry` drives the writer half over leveldb2, deleting the folder from the store while the entry is landing and asserting the entry is reachable by listing afterwards. `TestCreateEntryRollsBackWhenTheParentCannotBeCreated` covers the error path. `listHookStore` grew an `afterInsert` hook for this and became `hookStore` rather than gaining a sibling type.

`TestEnsureDirectoryEntryNarrowsAWiderRestore` and `TestEnsureDirectoryEntryKeepsATightenedDirectory` cover the mode. Each test above fails with its own production change disabled.

### DIFF
--- a/weed/filer/empty_folder_cleanup/empty_folder_cleaner.go
+++ b/weed/filer/empty_folder_cleanup/empty_folder_cleaner.go
@@ -383,6 +383,39 @@ func (efc *EmptyFolderCleaner) restoreFoldersWrittenDuringDelete() {
 	efc.mu.Unlock()
 }
 
+// restoreIfWrittenTo puts folder back when an entry landed in it while it was being
+// deleted. True means the parent is no longer empty, so the caller must not cascade.
+func (efc *EmptyFolderCleaner) restoreIfWrittenTo(ctx context.Context, folder string, attrs DirectoryAttributes) bool {
+	count, err := efc.countItems(ctx, folder)
+	if err != nil || count > 0 {
+		// Ask for the retry here rather than leaving it to a create event that may
+		// already have gone by: this pass can be the only sight of the entry.
+		efc.mu.Lock()
+		if observed, found := efc.deleted[folder]; found {
+			observed.writtenTo = true
+		}
+		efc.mu.Unlock()
+	}
+	if err != nil {
+		glog.V(2).Infof("EmptyFolderCleaner: cannot re-check %s after deleting it: %v", folder, err)
+		return false
+	}
+	if count == 0 {
+		return false
+	}
+
+	glog.V(1).Infof("EmptyFolderCleaner: restoring %s, written to while it was being deleted", folder)
+	if err := efc.filer.EnsureDirectoryEntry(ctx, util.FullPath(folder), attrs); err != nil {
+		glog.V(2).Infof("EmptyFolderCleaner: failed to restore %s: %v", folder, err)
+		return false
+	}
+
+	efc.mu.Lock()
+	delete(efc.deleted, folder)
+	efc.mu.Unlock()
+	return true
+}
+
 // makeRoomForDeletedLocked drops the oldest of a small sample when the set is full.
 // The newest folders are the ones whose race is still live, so they must not be the
 // ones given up; sampling keeps this cheap under heavy deletion rates.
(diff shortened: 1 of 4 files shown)


azat:
Thanks a lot!

But it seems that the race is still there, in other words it is still possible that `ListObjectsv2` may not return an object after it has been created? Is it correct?

PR #10803, opened 2026-08-18
fix(filer): bound aggregated metadata reads by peer watermarks

kisow:

The aggregated metadata subscription (`SubscribeMetadata`) merges per-filer sources with a single scalar cursor: once it passes a timestamp T, anything a source materializes below T afterwards is silently skipped — no error on either side, the offset keeps advancing. 

Reproduced on a 3-filer cluster feeding `filer.backup`: with two filers' flush landing ~70s late while the subscriber catches up on the third filer's data, official 4.41 permanently lost 28,064 of 45,000 tracked files (62.4%) with zero errors. The same single-cursor hole also bites when a slow reader falls off the arrival-bumped aggregated ring back to original-timestamp disk files, and when the master removes a flapping (frozen) peer whose unflushed events still exist.

Bound reads by what every source has provably made visible:

- **Delivery low-watermark**: the aggregator tracks the newest timestamp received per peer stream (events or idle heartbeats — peer streams opt into `ClientSupportsIdleHeartbeat`); in-memory reads hold at the minimum.
- **Flush low-watermark**: each filer reports its flush watermark via a new `flushed_ts_ns` response field (on heartbeats and periodic flush reports; an idle filer reports "now"). Disk passes freeze the minimum before listing the log files and hold at it; day-boundary jumps and metadata-chunk refs are bounded the same way.
- **Original-space disk anchor**: memory cursors live in the ring's bumped (arrival) timestamp space, disk in original timestamps. A reader kicked off the ring resumes the disk pass from the anchor — the newest original-space position proven complete — instead of its bumped cursor: redelivery is within the at-least-once contract, skipping is not.
- **Peer-removal grace**: removal only marks the peer; its watermarks keep participating for 2 × `LogFlushInterval`, so a flap costs nothing and a decommission drops out after the grace.
- **In-flight cap**: events are timestamped before notification work that can block; heartbeat and flush claims are capped below the oldest stamped-but-unappended event.
- **Unbatchable control messages**: heartbeats and flush reports never ride in a batch's `Events` tail, where receivers' nil-notification guard would drop their watermark state.

Both watermarks relax at the settled horizon (2 × `LogFlushInterval`) as a liveness escape; that bounded loss is the only remaining path, and `#10501`'s unproven-crossing counter now fires only for it. Trade-off: aggregated live delivery lags by up to the idle-heartbeat interval when some peers are quiet; `SubscribeLocalMetadata` consumers are unaffected.

# How is the PR tested?

- Unit tests: watermark bookkeeping and removal grace, hold decisions, held-read rewind invariant, in-flight caps, batching rules, crossing accounting. `go test ./weed/filer/ ./weed/server/ ./weed/util/log_buffer/` pass, including the e2e subscription harness from `#10501`.
- Cluster A/B on the staged reproduction above: official 4.41 lost 62.4% silently; with this PR, three independent runs converged to zero loss (45,000/45,000 files, 0 delete ghosts, ~4 minutes from resume) — including one where the master removed the frozen filers mid-test.
- Generated protobuf files match the checked-in generator output, with only the new field's cases added.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

- **New Features**
  - Added accurate delivery and flush progress tracking for metadata synchronization.
  - Added heartbeat and control updates to report synchronization status.
  - Improved metadata stream resumption to prevent skipped or duplicated events.

- **Bug Fixes**
  - Prevented incomplete delivery during peer delays, reconnects, and buffer eviction.
  - Improved metadata gap detection and recovery.
  - Ensured in-progress updates are not reported as fully flushed.

- **Tests**
  - Added comprehensive coverage for watermarks, reconnections, delays, gaps, eviction, and delivery resumption.


chrislusf:
Reviewed this closely - the watermark design holds up, and the code matches the A/B results. I pushed two commits on top for findings from the review.

d44fb7f16: before the ring's first eviction, a below-window cursor was served the earliest retained entry ("nothing evicted yet"), so a chunk-mode subscriber reconnecting right after a filer restart could jump over pre-restart files the bounded pass had withheld - silently, off the counter. The aggregator now marks everything at or below its subscriptions' start as evicted, the idle-wait callback credits the disk anchor only once the run is connected to the ring, and the gap pass gets a real proof (the peer flush low-watermark capped by the pass's listing bound) so it crosses the marked boundary without parking into the 15-minute give-up.

06060f32e: a log file spans past its named minute, and chunk clients apply a shipped file whole - a tail past the hold point could become a persisted client checkpoint beyond what every peer has proven, and a crash in that window resumed past another peer's late-but-in-contract flush. The ref listing now stops a minute plus a flush interval below the hold; the memory pass covers the withheld band, so freshness is unchanged.

One note worth carrying into the description or release notes: during a rolling upgrade, pre-upgrade peers never report flushed_ts_ns, so the flush low-watermark stays unknown and aggregated disk reads trail by the 2-minute horizon until every filer is upgraded. No loss beyond current behavior, and it self-heals once the last peer reports.

PR #10805, opened 2026-08-18
volume: forward fsync=true to replicas in ReplicatedWrite

sunchao6106:
## What

`ReplicatedWrite` reads the `fsync` request parameter for the **local** write but drops it when fanning out to replica volume servers: the replica URL only carried `type`/`ttl`/`ts`/`cm`. So a `?fsync=true` blob upload flushed only the primary's `.dat` to disk — replicas always wrote into the OS page cache without fsync, leaving replica durability dependent on the OS flushing dirty pages.

This PR forwards the `fsync=true` parameter to replica volume servers in the fan-out request. A durable write now means **every replica** has flushed to disk, not just the primary.

## Why

Replication gives logical consistency (all replicas receive the data before the client is acked), but without fsync on replicas the durability is not symmetric: after a power loss, the primary's `.dat` survives while replica `.dat` files can lose recently acked data. For workloads that opt into `fsync=true`, the replicas should honor the same durability contract.

## Behavior

| Request | Before | After |
|---|---|---|
| `?fsync=true` | primary fsyncs, replicas do not | primary + all replicas fsync |
| default (no fsync) | no fsync anywhere | unchanged (zero impact) |
| replica forward (`type=replicate`) | writes locally only | unchanged (no recursion) |

## Correctness

- fsync on a replica runs through the same `writeNeedle2` fsync path (batch worker + `DataBackend.Sync()`), and the replica HTTP response is only sent after `WaitComplete` — so the primary's `DistributedOperation` waiting on replica responses guarantees all replicas flushed before the client ack.
- No server-side default is changed; only requests that already carry `fsync=true` are affected.

## Test

Added `TestReplicatedWriteForwardsFsyncToReplicas` with a mock master gRPC + httptest replica, asserting the fan-out query carries `fsync=true` when requested and omits it otherwise. Full `go test ./weed/topology/` and `go vet` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Replicated writes now preserve the local `fsync` setting on remote replicas.
  * Durable writes are flushed synchronously while storage is shutting down, helping preserve data durability.
  * Failed durable writes during shutdown now return an error without leaving incomplete appended data.
  * Non-durable writes continue using the existing behavior during shutdown.
* **Tests**
  * Added coverage for replica durability forwarding, shutdown writes, and failed flush recovery.

### DIFF
--- a/weed/storage/remote_tier_integration_test.go
+++ b/weed/storage/remote_tier_integration_test.go
@@ -186,7 +186,7 @@ func tierUpVolumeLive(t *testing.T, dir string, vid needle.VolumeId, b *localDir
 	require.NoError(t, err)
 
 	for i := 1; i <= 5; i++ {
-		_, _, _, err := v.writeNeedle2(newRandomNeedle(uint64(i)), true, false)
+		_, _, _, err := v.writeNeedle2(newRandomNeedle(uint64(i)), true, false, false)
 		require.NoError(t, err)
 	}
 
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -80,7 +80,7 @@ type Store struct {
 	DeletedVolumesChan  chan *master_pb.VolumeShortInformationMessage
 	NewEcShardsChan     chan *master_pb.VolumeEcShardInformationMessage
 	DeletedEcShardsChan chan *master_pb.VolumeEcShardInformationMessage
-	isStopping          bool
+	isStopping          atomic.Bool
 	volumeReport        volumeReportState
 }
 
@@ -732,14 +732,14 @@ func (s *Store) deleteExpiredEcVolumes() (ecShards, deleted []*master_pb.VolumeE
 }
 
 func (s *Store) SetStopping() {
-	s.isStopping = true
+	s.isStopping.Store(true)
 	for _, location := range s.Locations {
 		location.SetStopping()
 	}
 }
 
 func (s *Store) IsStopping() bool {
-	return s.isStopping
+	return s.isStopping.Load()
 }
 
 func (s *Store) LoadNewVolumes() {
@@ -760,7 +760,7 @@ func (s *Store) WriteVolumeNeedle(i needle.VolumeId, n *needle.Needle, checkCook
 			err = fmt.Errorf("volume %d is read only", i)
 			return
 		}
-		_, _, isUnchanged, err = v.writeNeedle2(n, checkCookie, fsync, s.isStopping)
+		_, _, isUnchanged, err = v.writeNeedle2(n, checkCookie, fsync, s.isStopping.Load())
 		return
 	}
 	glog.V(0).Infoln("volume", i, "not found!")
--- a/weed/storage/store.go
+++ b/weed/storage/store.go
@@ -760,7 +760,7 @@ func (s *Store) WriteVolumeNeedle(i needle.VolumeId, n *needle.Needle, checkCook
 			err = fmt.Errorf("volume %d is read only", i)
 			return
 		}
-		_, _, isUnchanged, err = v.writeNeedle2(n, checkCookie, fsync && !s.isStopping)
+		_, _, isUnchanged, err = v.writeNeedle2(n, checkCookie, fsync, s.isStopping)
 		return
 	}
 	glog.V(0).Infoln("volume", i, "not found!")
--- a/weed/storage/store_consolidate_index_test.go
+++ b/weed/storage/store_consolidate_index_test.go
@@ -93,7 +93,7 @@ func TestConsolidateVolumeIndexMovesIdxToIdxDir(t *testing.T) {
 	require.NotNil(t, mounted)
 	n := &needle.Needle{Id: types.Uint64ToNeedleId(42), Data: []byte("payload-across-relocate")}
 	n.Checksum = needle.NewCRC(n.Data)
-	_, _, _, err = mounted.writeNeedle2(n, true, false)
+	_, _, _, err = mounted.writeNeedle2(n, true, false, false)
 	require.NoError(t, err)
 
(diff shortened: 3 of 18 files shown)


sunchao6106:
Thanks @greptile-apps for catching the shutdown-window durability gap. Fixed in the latest commit (`storage: fail fast on writes while stopping instead of degrading fsync`).

**What changed**

`Store.WriteVolumeNeedle` previously wrote with `fsync && !s.isStopping`, silently degrading a requested durable write into a non-fsync write while the store was stopping. During the `preStopSeconds` window a stopping replica still served HTTP writes, so it could ack a `fsync=true` write it never flushed to disk.

Now writes are **rejected** while the store is stopping:

```go
if s.isStopping {
    return false, errors.New("volume server is stopping")
}
```

**Effect on the replication path**

- A stopping replica receives the forwarded `?type=replicate&fsync=true` → `WriteVolumeNeedle` fails fast
- The primary's `DistributedOperation` sees the fan-out error → the whole write fails
- The client retries and gets reassigned to a healthy replica (existing `shouldReassignUpload` retry path)

**Test**

Added `TestStoreWriteVolumeNeedleFailsFastWhenStopping` asserting `WriteVolumeNeedle` returns a stopping error while `SetStopping()` is in effect. `go test ./weed/storage/ ./weed/topology/` and `go vet` all pass.

One behavioral note: while a store is stopping, ALL writes to it now fail fast (not just fsync ones). This is the correct semantic — a stopping server should not accept new writes at all — and matches the intent of `SetStopping()` which already triggers the shutdown flow.

sunchao6106:
Good catch on both points. I've adopted the suggestion and pushed an amended commit (`7eaad3593`).

**Change**: the guard is now `isStopping && fsync` — only durable writes fail fast while stopping:

```go
if s.isStopping && fsync {
    return false, errors.New("volume server is stopping")
}
```

**Rationale for keeping non-fsync writes flowing during drain**:
- Non-fsync writes take the `syncWrite` path (direct `pwrite`, no async worker dependency), so they are safe to serve during shutdown.
- They make no durability promise, and the normal close flow (`Close → doClose`) flushes them anyway.
- This preserves the original `fsync && !s.isStopping` semantics for plain writes — the pre-shutdown window keeps accepting in-flight non-fsync traffic, and only the durability contract (fsync) is hardened.

**Test updated**: `TestStoreWriteVolumeNeedleFailsFastWhenStopping` now asserts both sides — fsync write → rejected with a "stopping" error; non-fsync write → passes through the guard (reaches volume lookup instead of the stopping error).

Also addressed the comment nit (now reads "Reject durable writes while stopping instead of silently degrading fsync"). `go test ./weed/storage/ ./weed/topology/` and `go vet` all pass.

chrislusf:
Kept the fan-out commit as-is - that one is a real bug and the forwarding is right. I checked the chain: the replica reads `fsync` off the query in the same `ReplicatedWrite`, and its 201 only comes after `WaitComplete`, so the ack really does mean every replica flushed.

Reworked the second commit. `fsync` on the write path is effectively an "use the async batch worker" switch, which is why it gets masked off while stopping (35f0daa19). Turning that into an error defeats `preStopSeconds`: writes already assigned to this server stop landing, and now that fsync is forwarded, a single draining replica fails the whole fan-out (`MaxAttempts: 1`) - which the filer will not retry either, since "volume server is stopping" matches nothing in `transientErrorMessages`.

So rather than refusing the write, it flushes inline: `syncWrite` takes the fsync flag and does the append plus `DataBackend.Sync()` under `dataFileAccessLock`, truncating the append back off the `.dat` if the sync fails, same as the batch worker. The drain stays writable and the ack still means the data is on disk.

Force-pushed onto your branch. The stopping test moved to `weed/storage`, where the code it exercises lives, and now asserts the write is accepted and actually synced.

chrislusf:
Went through the bot comments.

**Needle map left pointing past the truncated end** (CodeRabbit on `volume_write.go:182`, Greptile on `:180`) - real, fixed. The inline path now grabs the mapping before the write and puts it back after truncating when the fsync fails: the prior offset/size if the write replaced something, a delete if it introduced the needle. Both cases have a regression test that reads the needle back afterwards. The batch worker has the same gap in its own rollback, but that predates this PR, so I left it rather than widening the diff.

**`isStopping` read without synchronization** (CodeRabbit on `store.go`) - real, now an `atomic.Bool` in its own commit. I did not take the second half of that suggestion. A shutdown/write lock spanning `writeNeedle2` would put a lock on the hot write path to close a window that no longer costs anything: if the flag flips right after the check, the write goes down the async batch path, which fsyncs as well. Either way the ack means it is flushed.

**Replica fsync suppressed during shutdown** (Greptile on `store_replicate.go:102`) - that was the bug, and it is what the second commit fixes. A stopping replica flushes inline now instead of dropping the fsync.

**Shutdown rejects non-fsync writes** (Greptile on `store.go`) - stale, and the suggested diff is identical to the code it replaces. That guard is gone; nothing gets rejected while stopping.

**Test error handling** (CodeRabbit on `store_replicate_test.go`) - taken, mostly. The mock replica checks its response write, the mock master reports anything `Serve` returns other than `ErrServerStopped`, and the redundant `lis.Close()` is gone since `Stop` closes the listener it was handed. Skipped the `t.Context()` rewrites - nothing in the tree lints for it and `httptest.NewRequest` reads fine.

Also measured this, since it changes what a durable write costs. Linux VM, 1 master + 2 volume servers, replication 001, `weed benchmark -writeOnly -c 16 -n 20000 -size 1024`, fresh cluster per run: with `-fsync`, 7165 -> 4333 req/s median across 4 rounds and p50 2.0ms -> 3.5ms. Without `-fsync`, unchanged. That is the replica actually flushing, and it scales with copy count, so it is worth a release note. Containerized, so read it as the shape of the change rather than absolute throughput.

PR #10807, opened 2026-08-18
s3: preserve absolute paths with empty buckets folder

zouyixiong:
## What changed

- Preserve an absolute filer path when `filer.options.buckets_folder` is empty.
- Add regression coverage for empty, root, and default bucket paths, including the final object path.

## Why

`path.Join("", bucket)` returns a relative path. When `buckets_folder = ""` is intentionally used to store S3 buckets at the filer root, S3 object paths therefore lose their leading slash.

Besides making existing root-level objects return 404, this breaks stores that derive routing from absolute path components. In particular, the Elasticsearch filer store interprets the second object-path component as the index name, creating `.seaweedfs_<object-prefix>` indices instead of using `.seaweedfs_<bucket>`.

Using `FullPath.Child` retains the existing behavior for `/buckets` while producing `/bucket` for an empty or filesystem-root bucket path.

## Validation

```text
go test ./weed/s3api -run TestBucketDirPreservesAbsoluteFilerPath -count=1
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserved absolute storage paths when accessing buckets and nested objects.
  * Improved handling of empty, root, default, repeated-separator, and dot-component bucket locations.

* **Tests**
  * Added coverage for bucket path handling across supported root-path scenarios.


chrislusf:
`/buckets` is the only allowed value. The option is there due to legacy.

PR #10812, opened 2026-08-18
filer: let a remote mount skip delete propagation

petedodd-pd:

`maybeDeleteFromRemote` propagates every delete under a remote mount, and any failure aborts `doDeleteEntryMetaAndData` before the filer entry is removed. For a read-through mount — where the remote is the system of record and the filer is only a cache — that coupling causes two problems.

**A filer-side directory removal can delete a live remote prefix.** `RemoveDirectory` fires for any directory under a mount whether or not it was ever cached, because the `entry.Remote == nil` guard only covers the file path. The filer can therefore delete remote data it never owned.

**Deletes cannot be made idempotent when the mount credentials are scoped to reads.** GCS answers `403` rather than `404` for a missing object when the caller lacks `storage.objects.delete`, since it will not confirm non-existence to a caller that cannot delete. An object already removed out of band is then indistinguishable from one that is genuinely undeletable, and `DeleteFile`'s `ErrRemoteObjectNotFound` mapping only covers the `404`. Such deletes fail permanently: the S3 gateway returns `500` and the stale filer entry survives until it ages out.

Both are structural for a mount used as a read-through cache, where the application owns the bucket and writes and deletes it directly. There is currently no way to say "this mount is a cache; never write to the remote on delete" — the only alternatives are granting the filer a delete permission it should not have, or accepting permanent `DeleteObject` failures.

Add a per-mount `skip_remote_delete` field on `RemoteStorageLocation` (field 5), following the existing `listing_cache_ttl_seconds` precedent, set by `remote.mount -skipRemoteDelete` and honoured by `maybeDeleteFromRemote`.

The check is a single early return covering both the file and directory paths, placed before remote-client resolution so a skip-delete mount cannot fail on client lookup either. The mount then removes filer metadata only and leaves the remote object untouched.

The default is `false`, so existing mounts keep propagating deletes and behaviour is unchanged unless the flag is passed.

Scope: the field governs the filer's inline propagation only. `filer.remote.sync` is a separate write-back path and is untouched.

# How is the PR tested?

Three unit tests in `weed/filer`:

- `TestDeleteEntryMetaAndData_SkipRemoteDeleteRemovesMetadataOnly` — asserts `DeleteFile` is never invoked and the filer entry is still removed
- `TestDeleteEntryMetaAndData_SkipRemoteDeleteLeavesRemoteDirectory` — the same for `RemoveDirectory`
- `TestRemoteStorageMapping_SkipRemoteDeleteSurvivesMountMappingRoundTrip` — the field survives a `RemoteStorageMapping` proto round-trip and is still set when read back through `FindMountDirectory`

The first two stub a remote client whose delete and remove calls return an error, so they fail if the gate is ever bypassed rather than passing vacuously.

`go build ./weed/...`, `go test ./weed/filer ./weed/shell ./weed/remote_storage/...` and `gofmt` are clean on this branch.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added a `-skipRemoteDelete` option for remote mounts.
  * Local metadata can now be deleted without deleting the corresponding remote file or directory.
  * The setting is preserved when remote storage configurations are saved and restored.

* **Bug Fixes**
  * Deletions no longer fail when remote access is unavailable and remote deletion is disabled.

### DIFF
--- a/weed/filer/filer_lazy_remote.go
+++ b/weed/filer/filer_lazy_remote.go
@@ -126,6 +126,11 @@ func (f *Filer) maybeDeleteFromRemote(ctx context.Context, entry *Entry) (bool, 
 		return false, nil
 	}
 
+	if remoteLoc.SkipRemoteDelete {
+		glog.V(3).InfofCtx(ctx, "maybeDeleteFromRemote: mount %s skips remote delete, removing %s locally only", mountDir, entry.FullPath)
+		return false, nil
+	}
+
 	client, _, found := f.RemoteStorage.GetRemoteStorageClient(remoteLoc.Name)
 	if !found {
 		return false, fmt.Errorf("resolve remote storage client for %s: not found", entry.FullPath)
--- a/weed/filer/filer_lazy_remote_test.go
+++ b/weed/filer/filer_lazy_remote_test.go
@@ -793,6 +793,78 @@ func TestDeleteEntryMetaAndData_DirectoryUnderMountDeletesRemoteDirectory(t *tes
 	require.ErrorIs(t, findErr, filer_pb.ErrNotFound)
 }
 
+func TestDeleteEntryMetaAndData_SkipRemoteDeleteRemovesMetadataOnly(t *testing.T) {
+	const storageType = "stub_lazy_skip_remote_delete"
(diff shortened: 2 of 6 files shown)


petedodd-pd:
dupe of #10811

PR #10816, opened 2026-08-19
rust: honor fsync on the volume server write path

chrislusf:
The Rust volume server ignored `fsync` completely. Nothing parsed the parameter, and `write_volume_needle` -> `write_needle` -> `append_needle` never flushed, so a `?fsync=true` upload was acked straight out of the page cache. #10805 made a Go primary forward `fsync=true` to its replicas, which means a Go primary handing a durable write to a Rust replica now gets the same empty promise, one hop further along.

## What

The upload handler reads `fsync` off the query, the same shape as the existing `type=replicate` probe, and threads it down to `Volume::write_needle`, which flushes the `.dat` before returning. The flush happens under the same lock as the append, so a failed one can only take back its own bytes.

A failed flush rolls the write back rather than leaving it half-applied: the append comes off the `.dat`, and the needle map goes back to the mapping the write replaced, or away entirely if the write introduced it. Without that the index would resolve to an offset past the truncated end. This mirrors what the Go side does.

The replication fan-out needed nothing: `do_replicated_request` already appends `&type=replicate` to the whole original query, so `fsync=true` rides along to replicas as-is.

## Not in this change

Batched writes carry the flag per request, so each durable write in a batch flushes rather than one flush covering the batch. Go gets the batch-wide version by truncating the entire batch when the flush fails, and matching that here means batch-wide rollback; per-request flushing is correct in the meantime, and the queue is opt-in (`SEAWEED_WRITE_QUEUE`, off by default). The write queue module doc claimed it already "syncs once per volume for the entire batch", which was never true - that is corrected.

The `.idx` is not flushed on a durable write. Go does not flush it either; the index rebuilds from the `.dat`.

Deletes still make no durability promise, matching Go.

## Test

Five new tests: a durable write that reads back and still dedups, the rollback restoring a replaced mapping with the old content still readable, the rollback dropping the mapping of a needle that never landed, a durable upload through the handler, and a replica-role write carrying the exact `?fsync=true&type=replicate` query a Go primary sends. Full `cargo test` passes (452), clippy is clean over the changed code, and no line this branch adds has rustfmt drift.

## Summary by CodeRabbit

* **New Features**
  * Upload requests now support `fsync=true` for durable writes before acknowledgment.
  * Durable-write handling is applied consistently to direct, queued, and replicated uploads.
  * Failed flushes preserve existing data and mappings through automatic rollback.

* **Bug Fixes**
  * Improved reliability when durable writes encounter storage errors.

* **Tests**
  * Added coverage for durable uploads, replicated writes, read-back behavior, deduplication, and rollback scenarios.

### DIFF
--- a/seaweed-volume/src/server/grpc_server.rs
+++ b/seaweed-volume/src/server/grpc_server.rs
@@ -2326,7 +2326,7 @@ impl VolumeServer for VolumeGrpcService {
             // Write needle to local volume
             let mut store = state.store.write().unwrap();
             store
-                .write_volume_needle(vid, &mut n)
+                .write_volume_needle(vid, &mut n, false)
                 .map_err(|e| Status::internal(format!("write needle: {}", e)))?;
         }
 
@@ -3987,7 +3987,7 @@ impl VolumeServer for VolumeGrpcService {
         let local_handle = tokio::task::spawn_blocking(move || {
             let mut store = state_clone.store.write().unwrap();
             store
-                .write_volume_needle(vid, &mut n_clone)
+                .write_volume_needle(vid, &mut n_clone, false)
                 .map(|_| ())
                 .map_err(|e| format!("local write needle {} size {}: {}", needle_id, size, e))
         });
@@ -5218,7 +5218,7 @@ mod tests {
                 data_size: "remote-incremental-copy".len() as u32,
                 ..Needle::default()
             };
-            volume.write_needle(&mut needle, true).unwrap();
+            volume.write_needle(&mut needle, true, false).unwrap();
             volume.sync_to_disk().unwrap();
             (
                 std::fs::read(volume.file_name(".dat")).unwrap(),
@@ -5386,7 +5386,7 @@ mod tests {
                 data_size: b"ec-generate".len() as u32,
                 ..Needle::default()
             };
-            volume.write_needle(&mut needle, true).unwrap();
+            volume.write_needle(&mut needle, true, false).unwrap();
             volume.sync_to_disk().unwrap();
         }
 
@@ -5609,7 +5609,7 @@ mod tests {
                 data: payload,
                 ..Needle::default()
             };
-            v.write_needle(&mut needle, true).unwrap();
+            v.write_needle(&mut needle, true, false).unwrap();
             v.sync_to_disk().unwrap();
         }
         let dat_path = {
--- a/seaweed-volume/src/server/handlers.rs
+++ b/seaweed-volume/src/server/handlers.rs
@@ -2580,11 +2580,17 @@ pub async fn post_handler(
         n.set_has_name();
     }
 
+    // A durable write flushes before it is acked. Read it the way Go's
+    // r.FormValue does, off the decoded fields, so a percent-encoded value is
+    // honored here too. ReplicatedWrite forwards the parameter, so a replica
+    // sees it the same way the primary did.
+    let fsync = form_value("fsync").as_deref() == Some("true");
+
     let write_result = if let Some(wq) = state.write_queue.get() {
-        wq.submit(vid, n.clone()).await
+        wq.submit(vid, n.clone(), fsync).await
     } else {
         let mut store = state.store.write().unwrap();
-        store.write_volume_needle(vid, &mut n)
+        store.write_volume_needle(vid, &mut n, fsync)
     };
 
     // Replicate to remote volume servers if this volume has replicas.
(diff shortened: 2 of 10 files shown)


chrislusf:
Both Greptile comments land on the same design mistake, and the fix is to stop making it rather than to patch the rollback.

**Rollback corrupts volume accounting** - correct. `nm.put` runs `metric.on_put` and `nm.delete` bumps the deletion counters, so undoing a rejected write through the ordinary map API adds a second set of numbers instead of reversing the first. An overwrite ended up counted twice in `file_count`; a new needle ended up as one file plus one deletion, inflating both content size and the garbage ratio.

**Failed restore leaves dangling mapping** - also correct. If the truncate succeeded and the restoring `put` then failed, the live map still pointed at the rejected offset, and the previously valid needle became unreadable even though its bytes were untouched.

Rather than reverse the metrics or make the restore fallible-but-critical, the durable path now flushes **between the append and publishing the map entry**. Nothing ever points at bytes that are not down yet, so a failed flush only has to take its own append off the end: the index never moved, the counters never saw the write, and there is no restore to fail. That is also the safer crash order in general - data before index - since the index rebuilds from the `.dat` but not the other way round.

One case this shook out: a dedup hit now flushes too. There is nothing to append, but the write it matched may have been non-durable, so acking without a flush would still be a false promise.

The two tests are rewritten against the new invariant, using a `#[cfg(test)]` seam to inject the flush failure (the same `_for_test` pattern already in this file). They assert the `.dat` length, that the mapping never moved, that the needle a rejected write introduced is not indexed at all, and that `file_count`, `content_size` and `deleted_count` are unchanged - which is the accounting Greptile flagged. I checked both fail if the flush is moved back after the publish.

Full `cargo test` still passes (452).

chrislusf:
Second round. Three new points, all taken.

**Do not acknowledge an fsync write before its index entry is crash-recoverable** - right, and it was the sharpest one. I had claimed in the PR description that the index rebuilds from the `.dat`; it does not. `load()` builds the map from `.idx`, and `check_volume_data_integrity` compares the `.dat` tail against the last `.idx` row, so a `.dat` that runs past it fails `verify_dat_ends_at` and `load()` sets `no_write_or_delete`. Forcing the `.dat` down while leaving the `.idx` row in page cache made that state reachable by design rather than by luck, and the failure mode is not "lose the write" but "volume comes back read only". A durable write now flushes the `.idx` as well, after publishing the row, and is only acked once both are down. It costs a second fsync per durable write; the ack meaning something is worth it.

Worth flagging: Go has this same gap. Its worker syncs `DataBackend` only, and `nm.Sync()` waits for SetStopping or Close. Separate PR.

**Quarantine the volume when rollback truncation fails** - taken. If the `set_len` back to the append offset fails, the rejected record is still on the end, and a later append would bury it mid-file where the tail check cannot see it. That path now sets `no_write_or_delete` and logs at error instead of warn.

**Parse fsync from the decoded query fields** - taken. This file already parses `query_fields` through `serde_urlencoded` and exposes a `form_value` closure that mirrors `r.FormValue`, multipart fallback included, so the raw string match was the odd one out. `fsync` now goes through `form_value`, which is what Go does.

Two notes on that one. `is_replicate` a few lines up still does a raw `p == "type=replicate"` match and has the same blind spot; it is pre-existing and gates throttling and re-replication, so I left it rather than widen this PR. And the integration test I added for `?fsync=%74rue` only proves the write is accepted and readable - the flush is not observable from a separate test crate, since the injection seam is `#[cfg(test)]` inside the lib. The decode itself is pinned by a unit test that asserts the raw match misses the encoded value and the decoded fields do not.

Full `cargo test` passes (455).

chrislusf:
Missed this one before the merge - the index-sync failure comment landed two minutes after my reply and I did not fold it in. It is correct, and it is now #10825.

Short version: #10816 handled a failed `.dat` flush (append comes off the end, nothing was ever published) but not a failed `.idx` flush after the row is published. That path returned the error with the row live and the volume still writable, so the primary served a needle its replicas never saw for a write the client was told had failed - and a lost row on restart took the volume read only anyway, which is the outcome the `.idx` flush was there to prevent.

Rolling the row back is not safe: it is undoing published state on a failing disk, `nm.delete` writes a tombstone through the same `.idx`, and a truncate afterwards would leave a row pointing past the end. So the volume quarantines instead, matching what this PR already does when the post-truncate cannot be done. The replica divergence is still there - the bytes are down and cannot be safely removed - but it is bounded and visible as a read-only volume.

PR #10819, opened 2026-08-19
mount: let a readdirplus stop taking references at a cap

chrislusf:
The memory in #10020 is the inode table: reading `Lookup`'s flat allocation against a measured cost model puts that mount at ~5.6M live inodes averaging ~200-character paths, about 94% of its 2.1GB heap.

The table only shrinks when the kernel returns a reference, so a walk over a large tree grows it with every child it touches, and nothing gives any of it back until the kernel reclaims dentries.

`-maxInodeEntries` stops the speculative half of that. Past the cap a readdirplus still reports every name, but leaves the `EntryOut` zeroed: Linux reads a zero nodeid as "no attributes for this entry" (`fuse_direntplus_link`), takes no reference, and the client looks up the ones it actually needs.

### What it does not do

It is a growth cap on readdirplus, not an eviction policy, and the name should be read that way:

- **Nothing here releases entries.** `delete(i.inode2path, ...)` appears exactly once in `InodeToPath`, inside `Forget`. A kernel FORGET is the only release path, and that arrives when the kernel evicts the dentry — under dcache pressure, on unlink, or at unmount. A machine with RAM to spare never sends one, which is part of why the mount in #10020 sat where it did.
- **LOOKUP is uncapped**, and has to be: a zero nodeid from LOOKUP means ENOENT, so refusing one would be a lie about the file existing. A client that stats every file it listed still puts every one of them in the table.

So this buys "stop paying for attributes nobody asked for", not "hold RSS at N".

Actively shrinking the table means asking the kernel to drop dentries with `EntryNotify(parent, name)` from the invalidation worker — never from a request handler, which is the deadlock `invalidateKernelDirListing` already documents. That wants the `(parent, name)` pair the table does not currently store, so it belongs after phase 2 of #10820 rather than here.

Default 0 keeps today's behavior.

## Summary by CodeRabbit

* **New Features**
  * Added a configurable maximum for tracked inode references when mounting.
  * The limit defaults to unlimited and can be set with the `-maxInodeEntries` option.
  * Directory listings continue to show all entries after the limit is reached, while capped entries omit additional metadata.

* **Bug Fixes**
  * Prevented inode tracking from exceeding the configured limit during enhanced directory listings.

* **Tests**
  * Added coverage verifying directory listings and inode limits behave correctly.

### DIFF
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -21,6 +21,7 @@ type MountOptions struct {
 	concurrentReaders    *int
 	cacheMetaTtlSec      *int
 	cacheDirMaxEntries   *int
+	maxInodeEntries      *int
 	cacheDirForRead      *string
 	cacheDirForWrite     *string
 	cacheSizeMBForRead   *int64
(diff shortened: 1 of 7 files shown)


chrislusf:
Closing this rather than landing it. It names itself a memory cap and is not one: LOOKUP is uncapped and cannot be capped, and nothing here evicts, so the table can still pass the cap and never comes back down. A flag that needs that paragraph of caveats is not worth carrying, especially one that costs a LOOKUP round trip per stat exactly on the walk-a-huge-tree workload that would engage it.

The constant is the thing worth attacking first, and phase 2 of #10820 does it unconditionally with no flag: ~350 B/inode down to ~141, which is 2.0GB to 0.8GB at the scale in #10020. If a ceiling is still wanted after that, the honest version is an eviction sweep over `EntryNotify(parent, name)`, which wants the pair phase 2 makes the table's key.

Nothing is lost: the zero-nodeid mechanism this found is now written up in the design doc as phase 5, and the branch stays put.

PR #10820, opened 2026-08-19
design: key the mount's inode table on parent and name

chrislusf:
The plan behind the #10020 heap work, up for review before the code.

The remaining cost after the two cheap wins is ~125 B fixed per inode plus the full path rounded to a Go size class — at that mount's path lengths, ~350 B across ~5.6M inodes. The path is the larger half and it grows with directory depth.

Proposal: directories keep a materialized full path, files keep only their parent's inode and their own name. There are orders of magnitude fewer directories than files, so no operation has to walk a parent chain — a file's path is one concat off its parent's. That puts per-inode cost at ~141 B independent of depth, roughly 2.0GB down to 0.8GB at that scale.

It also removes a bug that is live today: `MovePath` moves the renamed entry and nothing under it, so after a directory rename every cached child still resolves to its pre-rename path. Keying on the parent removes that by construction.

Four phases, each its own commit, with the table's behavior unchanged until the second. The trade to weigh: `GetPath` currently returns an existing string and allocates nothing, and it runs once per FUSE op; afterwards it costs one allocation.

## Summary by CodeRabbit

* **Documentation**
  * Added a design proposal for reducing mount inode-table memory usage.
  * Documented efficient tracking of files, directories, children, and hard links.
  * Described path lookup, inode retrieval, entry removal, directory renames, and eviction considerations.
  * Included implementation phases, memory targets, and key design invariants.

### DIFF
--- a/design-mount-inode-table.md
+++ b/design-mount-inode-table.md
@@ -1,0 +1,186 @@
+# Design: Keying the mount's inode table on parent and name
+
+Issue #10020 — a mount over 32.7M files sits at 2.1GB RSS after the readdir fixes.
+Its heap profile is 94% one structure: `InodeToPath`.
+
+## Where the bytes are
+
+Measured by populating the table with 1M children and reading `HeapAlloc` around
+it, after the two cheap wins already landed:
+
+    221.5 B/inode at  85-character paths
+    285.6 B/inode at 148-character paths
+
+so roughly **125 B fixed plus the full path rounded to a Go size class**. The
+fixed part is `InodeEntry` (32) and the two maps (~93). Reading the reporter's
+profile against that model — `Lookup` flat / 125 — gives ~5.6M live inodes at
+~200-character paths, ~350 B each, which is the 2.0GB.
+
+The path is the larger half and it grows with directory depth. It is stored once
+and referenced twice: as the `path2inode` key and as `InodeEntry.path`.
+
+## Shape
+
+Directories keep a materialized full path. Files keep only their parent's inode
+and their own name. There are orders of magnitude fewer directories than files,
+so the full paths that remain are a rounding error, and no operation has to walk
+a parent chain: a file's path is one concat off its parent's.
+
+    type childKey struct {
+        parent uint64
+        name   string
+    }
+
+    type InodeEntry struct {   // 32 bytes, one object, no allocation of its own
+        parent  uint64
+        name    string
+        nlookup uint64
+    }
+
+    inodes     map[uint64]*InodeEntry     // every inode
+    children   map[childKey]uint64        // (parent, name) -> inode
+    dirs       map[uint64]*dirState       // directories only; carries the path
+    dirPaths   map[util.FullPath]uint64   // directories only
+    extraLinks map[uint64][]childKey      // hard-linked inodes only
+
+`InodeEntry.name` shares its bytes with the `children` key, exactly as the path
+does with `path2inode` today.
+
+The name has to be copied on the way in. Every name the mount has to hand is a
+slice of something longer — `FullPath.Name()` slices the full path, and a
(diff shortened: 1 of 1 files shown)


chrislusf:
Corrected the rename section. I had it as a live bug; it is not. The filer emits one event per moved entry and the mount turns each into a `MovePath`, so a directory rename is repaired entry by entry rather than left stale. The argument for parent-keying there is cost and robustness — a million cached children is a million table mutations, and correctness depends on every one of those events landing — not a defect.

(The one real gap that turned up alongside it is #10822: remote renames only moved the table for inodes with an open file handle.)

chrislusf:
Folded the review into the doc. Four of these were real gaps in the plan:

**Name ownership** is the one that mattered most, and it undercut the premise as written. Every name the mount has to hand is a slice of something longer — `FullPath.Name()` slices the full path, a readdir's `dirEntry.Name` slices the listed entry's — and a Go string slice keeps its whole backing array alive. Storing one as-is would pin the ~200-byte path it came from, per inode, and the table would cost exactly what it costs today. The doc now says to `strings.Clone` at the point of storage; that is one small allocation per new inode, which is what building the child path costs already.

**extraLinks lifecycle**: `Forget` drops every child key the inode has, not just the primary, and unlinking one link promotes a survivor into `InodeEntry` — the shape `removeOnePath` already has today.

**Parent retention** had a risk noted but no mechanism. `dirState` now carries a child count; `Forget` releases a directory at nlookup 0 **and** count 0. Linux evicts dentries bottom-up so it should never fire, which is the reason to assert it rather than assume it.

**Descendant re-prefixing** had no phase owning it. It is now explicitly part of phase 2, with a regression test through both a local and a subscription-driven rename.

Also stated that phase 3 has to preserve `Lookup`'s identity contract intact — inode reuse, `possibleInode`, the non-hardlink collision probe, `nlookup` only on `isLookup` — with tests before the switch, and labelled the `MovePath` snippet as current behavior since it kept reading as the target design.

chrislusf:
Second round folded in.

**Hard links across the phase boundary** was the real one. Phase 2 reduced `InodeEntry` to a single `(parent, name)` while `extraLinks` only arrived in phase 4, so a hard-linked file would have lost its other names for two commits in between. Since the whole point of the phasing is that each commit is individually correct, `extraLinks` now lands with phase 2 and phase 4 is only the size-class tightening.

**Notifying every alias during eviction** is right and now stated: `EntryNotify` names one dentry, so a hard-linked inode needs one call per name before the entry goes, or the kernel keeps the aliases it was never told about.

**Subscription gaps** — noted, with the direction corrected. A rename this mount never sees strands a stale path either way, so it is not a new exposure, and it is a smaller one: today it strands every cached descendant of the renamed directory, keyed on the parent it strands only the descendant directories, because a file's key never mentioned the prefix. Reconciling against the filer after a gap is separate work this does not buy.

**FORGET counting** was already true and just unstated, so I said it plainly rather than changing anything: the count semantics carry over untouched — subtract the request's count, keep every mapping above zero, clamp a kernel over-decrement as it does today, and only at zero drop the keys.

chrislusf:
Both fixed.

**Subscription-gap scope** — I had it wrong. I claimed parent-keying limits the damage to descendant directories, but a file resolves through `dirs[e.parent].path`, so one stale directory path makes every file beneath it resolve stale too. Fewer table entries hold a stale prefix; the set of affected paths is unchanged. Reworded to say that, rather than claiming a benefit that is not there.

**API boundary** — stated explicitly. `GetInode` keeps taking a path, so its thirteen callers are untouched and only its body changes in phase 2. `Lookup` keeps its path signature through phase 2 and converts with its nine callers in phase 3, so the load-bearing commit does not have to move both halves at once.

PR #10821, opened 2026-08-19
Pin grpc-go to v1.82.0 to avoid a Unix-socket transport deadlock

chrislusf:
Temporary pin while grpc/grpc-go#9330 is open.

grpc-go v1.82.1 broadened the control-buffer throttle to count `registerStream` and `cleanupStream`. Those are per-RPC bookkeeping items that emit no wire traffic, so a connection carrying enough concurrent RPCs latches throttling in both directions at once — both peers stop reading, both `loopyWriter`s block writing to a full socket, and it never recovers. It triggers whenever the socket send buffer is too small to absorb the burst `loopyWriter` emits for the in-flight streams — so it is about buffer size, not transport (TCP forced to a 64KB buffer deadlocks; a Unix socket with a 2MB buffer does not). Local gRPC rides Unix sockets, whose buffers are small and fixed (208KB on Linux, 8KB on macOS), which is why we see it first there; a remote filer over TCP is exposed too wherever send buffers are constrained. We picked this up in `0e78031fa` (the 1.82.0 -> 1.82.1 dependabot bump).

Symptom is `weed mini` wedging under concurrent S3 writes, which is how self-hosted Sentry hits it. Measured with a Sentry-nodestore-shaped load (flat keys, small objects) against Sentry's own mini flags:

| concurrent PUTs | before | after |
|---|---|---|
| 512 | 45 ops/s, 14% failures, 65s stalls | 6,375 ops/s, 0 failures, no stall |
| 2048 | wedged | 2,407 ops/s, 0 failures, no stall |

During a wedge every in-flight PUT sits in `waitOnHeader` on a filer RPC with zero filer handlers running; both transports are parked in `controlBuffer.throttle` and both `loopyWriter`s in `net.(*conn).Write`.

Pinned to v1.82.0, the newest unaffected release — bisected v1.82.0 clean / v1.82.1 deadlocking, on both linux and darwin, with a standalone repro that uses only default gRPC options (in the upstream issue). A `replace` is needed rather than a `require` because rclone v1.75.0 requires the unreleased grpc dev snapshot. `go.sum` picks up only the two v1.82.0 hashes; no `go mod tidy` churn.

Revert once upstream lands a fix. Worth noting connection pooling does not solve this — 8 connections still wedge at c=2048, since it divides stream churn rather than removing the condition. Nor does switching local gRPC to TCP loopback: that only works because Linux TCP defaults are large, not because TCP is immune.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability for concurrent network operations by addressing a potential connection stall under constrained network conditions.
  * Reduced the risk of requests becoming unresponsive when system or network buffers are limited.

* **Maintenance**
  * No new user-facing features or interface changes are included in this release.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -527,3 +527,19 @@ replace github.com/apache/thrift => github.com/apache/thrift v0.23.1-0.202604291
 // through rclone's internxt backend, which calls IsMnemonicValid and NewSeed.
 // cosmos/go-bip39 is a maintained, API-compatible fork.
 replace github.com/tyler-smith/go-bip39 => github.com/cosmos/go-bip39 v1.0.0
+
+// grpc-go v1.82.1 started counting registerStream/cleanupStream toward the
+// control-buffer throttle limit. Those are per-RPC bookkeeping items that emit
+// no wire traffic, so a connection carrying enough concurrent RPCs latches
+// throttling in both directions at once and deadlocks for good: both peers stop
+// reading, both loopyWriters block writing to a full socket.
+//
+// It triggers whenever the socket send buffer is too small to absorb the burst
+// loopyWriter emits for the in-flight streams, so it is about buffer size, not
+// transport: TCP forced to a 64KB buffer deadlocks, a Unix socket with a 2MB
+// buffer does not. Local gRPC rides Unix sockets, whose buffers are small and
+// fixed (208KB on Linux, 8KB on macOS), which is why weed mini wedges under
+// concurrent S3 writes; a remote filer over TCP is exposed too wherever
+// send buffers are constrained. Pin to the last unaffected release until
+// grpc/grpc-go#9330 is fixed.
+replace google.golang.org/grpc => google.golang.org/grpc v1.82.0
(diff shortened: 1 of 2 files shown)


chrislusf:
Blocking this on myself — the pin should not merge as-is. `dependency-review` is right.

v1.82.1 is the fixed version for GHSA-hrxh-6v49-42gf (high), and reading the advisory, the change that causes our deadlock *is* the security fix:

> Earlier mitigations in grpc-go for HTTP/2 Rapid Reset only applied threshold checks to items that directly resulted in control frames being written back to the wire, such as SETTINGS ACKs or server-initiated RST_STREAMs. When a client initiated a rapid flood of stream creation (HEADERS) immediately followed [by resets] ...

That is exactly why `registerStream`/`cleanupStream` started counting toward the throttle. So pinning to v1.82.0 trades our deadlock for a Rapid Reset DoS on every gRPC server we expose. Not a good trade, and I should have checked the advisory before opening this.

Alternative I have working locally, on the current (fixed) grpc: raise `SO_SNDBUF`/`SO_RCVBUF` on both ends of the local Unix sockets. The deadlock needs a send buffer too small to absorb what loopyWriter emits for the in-flight streams — the defaults are 208KB on Linux and 8KB on macOS and never autotune. Forcing 8MB is honored on both, and weed mini then survives c=512 and c=2048 with zero failures and no stalls on unfixed grpc.

Worth noting the Rapid Reset threat model does not apply to these particular sockets — they are same-host, same-process-group, filesystem-permission-gated — so buying headroom there costs no security. But it is still a threshold shift rather than a cure; the real fix is upstream (grpc/grpc-go#9330).

Leaving this as a draft for the record rather than closing, since the analysis is the useful part.

chrislusf:
Superseded by #10824, which fixes the deadlock on the current (patched) grpc instead of pinning back into GHSA-hrxh-6v49-42gf.

PR #10822, opened 2026-08-19
mount: move the inode table when a rename arrives from the cluster

chrislusf:
Found while planning the inode-table work behind #10020.

A rename made by another client in the cluster reaches this mount only as a metadata event. The one `MovePath` on that path sits inside `invalidateOpenFileHandle`, so it runs only for inodes that happen to have an open file handle. Every other inode the kernel is still addressing by nodeid — anything a listing or a stat put in the table — keeps resolving to its pre-rename path, and the next operation on it goes to a path the filer no longer has.

Move the entry for every rename invalidation, not just the ones with a handle. The filer emits one event per moved entry, so a renamed directory's children arrive in turn and follow their parent with no descendant walk. A local rename already moves the table itself in `handleRenameResponse`, and the second move is a no-op.

`TestRemoteRenameMovesInodePaths` fails on master with the pre-rename paths still in the table.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved remote rename handling so files and directories consistently appear at their new locations.
  * Prevented repeated rename notifications from moving open files more than once.
  * Preserved existing target mappings when replayed renames are received.
  * Ensured paths remain correctly tracked when a rename source is no longer available.

* **Tests**
  * Added coverage for directory and child-path updates, repeated notifications, open handles, and missing-source rename scenarios.

### DIFF
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -524,19 +524,18 @@ func (i *InodeToPath) MovePath(sourcePath, targetPath util.FullPath) (sourceInod
 	i.Lock()
 	defer i.Unlock()
 	sourceInode, sourceFound := i.path2inode[sourcePath]
+	if !sourceFound {
+		// Nothing to move: either the source was never visited, or this move
+		// already happened. Whatever is at the target is not ours to take apart.
+		return
+	}
 	targetInode, targetFound := i.path2inode[targetPath]
 	if targetFound {
 		i.removePathFromInode2Path(targetInode, targetPath)
 		delete(i.path2inode, targetPath)
 	}
-	if sourceFound {
-		delete(i.path2inode, sourcePath)
-		i.path2inode[targetPath] = sourceInode
-	} else {
-		// it is possible some source folder items has not been visited before
-		// so no need to worry about their source inodes
-		return
-	}
+	delete(i.path2inode, sourcePath)
+	i.path2inode[targetPath] = sourceInode
 	if entry, entryFound := i.inode2path[sourceInode]; entryFound {
 		entry.replacePath(sourcePath, targetPath)
 		if d := i.dirStates[sourceInode]; d != nil {
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -524,19 +524,20 @@ func (i *InodeToPath) MovePath(sourcePath, targetPath util.FullPath) (sourceInod
 	i.Lock()
 	defer i.Unlock()
 	sourceInode, sourceFound := i.path2inode[sourcePath]
+	if !sourceFound {
+		// Nothing of ours to move: the source was never visited here, or a
+		// redelivery already moved it. Whatever sits at the target is not ours
+		// to take apart on the strength of an absent source, and deciding that
+		// outside this lock would race a concurrent move to the same target.
+		return 0, 0
+	}
 	targetInode, targetFound := i.path2inode[targetPath]
 	if targetFound {
 		i.removePathFromInode2Path(targetInode, targetPath)
 		delete(i.path2inode, targetPath)
 	}
-	if sourceFound {
-		delete(i.path2inode, sourcePath)
-		i.path2inode[targetPath] = sourceInode
-	} else {
-		// it is possible some source folder items has not been visited before
-		// so no need to worry about their source inodes
-		return
-	}
+	delete(i.path2inode, sourcePath)
+	i.path2inode[targetPath] = sourceInode
 	if entry, entryFound := i.inode2path[sourceInode]; entryFound {
 		entry.replacePath(sourcePath, targetPath)
 		if d := i.dirStates[sourceInode]; d != nil {
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -524,18 +524,19 @@ func (i *InodeToPath) MovePath(sourcePath, targetPath util.FullPath) (sourceInod
 	i.Lock()
 	defer i.Unlock()
 	sourceInode, sourceFound := i.path2inode[sourcePath]
-	if !sourceFound {
-		// Nothing to move: either the source was never visited, or this move
-		// already happened. Whatever is at the target is not ours to take apart.
-		return
-	}
 	targetInode, targetFound := i.path2inode[targetPath]
 	if targetFound {
 		i.removePathFromInode2Path(targetInode, targetPath)
 		delete(i.path2inode, targetPath)
 	}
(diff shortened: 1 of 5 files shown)


chrislusf:
Good catch on the double move — it is a real regression this PR would have introduced, and it is worse than a no-op: on the second call `MovePath` found no source but did find the target, tore down the target's mapping, and only then declined the move. The inode came out with no path at all, so `GetPath` returned ENOENT for a file the kernel still holds open.

Fixed as its own commit ahead of the change that makes it reachable: the source check moves to the top, before anything touches the target. `TestMovePathRepeatKeepsTheTarget` fails on master with `inode resolves to "" (2=no such file or directory)`.

I kept the test at the `InodeToPath` level rather than building an open file handle — that is where the defect lives, and the handle only supplies the second caller.

Also rebased onto master to pick up #10818.

chrislusf:
Right, and it means my earlier fix was the wrong half.

Returning early from `MovePath` on a missing source is not safe, because "source missing" covers two different situations. In the repeat-move case the target holds the inode the first move just put there and must not be touched. In the case you describe — a rename from a client that never visited the source — the target holds a *different* inode that the rename genuinely destroyed, and leaving the name resolving to it is what lets a dirty handle flush over whatever took its place. From inside `MovePath` the two are indistinguishable: both see an absent source and a live target.

So I reverted the guard and stopped the double call instead. `invalidateOpenFileHandle` now reports whether it moved, and the handler moves only when it did not. Both paths mark a replaced file's handle deleted, which the no-handle path was not doing at all — that is the protection your scenario was missing.

Two tests, each verified to fail against the specific flaw:

- `TestRemoteRenameMovesOnceWithAnOpenHandle` — without the single-move guard: `inode resolves to "" (2=no such file or directory), want /dir/renamed`.
- `TestRemoteRenameOverAnOpenHandleUnlinksTheTarget` — with the early return restored: `/dir/target still resolves to ... after being renamed over` and `the replaced file's handle can still flush over the renamed source`.

Thanks — the first version would have traded one wrong table state for another.

chrislusf:
Both valid, and together they reverse the position I took in the last round.

**Replay.** Confirmed the dedup ring is 4096 entries (`recentEventDedupWindow`), so a redelivery past that window does reach the handler. A replay finds no source and a live target, unlinks the mapping the first delivery just made — and with the `markReplacedHandleDeleted` I added, marks the *moved* file's handle deleted. That is strictly worse than the stale path this PR set out to fix. Without the guard the new test reports exactly that:

```
/dir/renamed resolves to 0 (found false), want 16863868803248091595
the moved file's handle was marked deleted by the replay
```

**Stale rename against an open handle.** Also right. When the version guard returns early, `RememberPath` never runs, so moving the table underneath left the handle flushing to the pre-rename path — an inconsistency that did not exist before, since previously neither moved. The handle path owns its inode's rename, so the fallback now runs only for an inode with no handle at all, which is the scope this PR actually needed.

The cost is that I give up unlinking a destination the rename replaced when the source was never visited here — the case from the previous round. That is what the mount already did before this branch, and between the two it is the safer default: retaining a stale name costs a wrong lookup, while unlinking the wrong one costs a file's dirty data. I would rather not trade a rare hazard for a worse one, so the narrow fix stands and the unvisited-source case stays untouched.

`TestRemoteRenameReplayLeavesTheTargetAlone` replaces the unlink test from the last round.

chrislusf:
Both valid, both fixed.

**Atomicity.** Right — the `HasPath` check was outside `MovePath`'s lock, so two invalidations for one rename could both observe the source and the loser would unlink what the winner had just placed. That is the same damage the check was meant to prevent, just moved to a narrower window. The decision now happens inside `MovePath` under its own lock, and it reports that nothing moved. That also puts it in one place instead of at each call site.

**The deletion fence.** Also right, and worse than a missing fence: I was setting `isDeleted` bare, while `markHandleDeleted` already exists a few files over and takes exactly the lock the flush reads that flag under — its comment even records the bug from last time (`setting the flag bare raced the flush's own check and resurrected the entry right after the delete`). Mine was a parallel helper next to it. Gone; the caller marks through `markHandleDeleted`.

Marking from the caller also fixes the lock-ordering hazard that reusing it would otherwise create: `invalidateOpenFileHandle` holds the source handle's lock for its whole body, so marking the replaced handle in there would hold two handle locks at once, and two renames crossing each other could deadlock. It returns the replaced inode instead, and the caller marks it holding nothing.

One thing worth flagging: `TestRenameOverExistingMarksReplacedHandleDeleted` failed on the first cut of this, because the invalidation harness wired `invalidateOpenFileHandle` directly while the mount wires `onEntryInvalidation`. The test was passing partly because the harness reached past the real entry point. It wires `onEntryInvalidation` now, and passes on the production path.

chrislusf:
One of these is right and fixes a contradiction between two earlier rounds. The other two I am not taking, with reasons.

**Version guard skips rename bookkeeping — fixed.** This is the mirror image of the finding at 20:49, which said moving the table while the fence skipped `RememberPath` left the handle behind. Both are right, and the resolution is to do neither half conditionally: the fence exists so an old event cannot roll a handle's entry back to stale *content*, and a rename carries no content — it says the name the inode answered to is gone. It is applied ahead of the fence now, and only when `MovePath` reports the source was still ours to move, which is what keeps a replay from remembering a path the handle has already moved past. `TestRenameBelowTheVersionFenceStillMoves` fails without it with `inode resolves to "/dir/file" (OK), want /dir/renamed`.

**Replacement marking loses flush race — not taking it.** The metadata flush has a second, independent fence: it re-resolves the saved path and skips when it no longer maps to the handle's inode (`weedfs_async_flush.go`, `saved path %s/%s no longer maps to this inode`). `MovePath` runs *before* `markHandleDeleted`, so the stronger check lands first — a flush that wins the race to `isDeleted` still has to pass an identity check against a table that has already moved. Getting through needs the flush to have read `GetInode` before `MovePath` ran, which is the pre-existing async-flush-versus-rename window and is not something the marking order can close.

**Absent source preserves overwritten handle — not taking it here.** This is the third round on this exact sub-case, and the two directions have now been proposed alternately: unlinking the destination breaks replayed renames by killing the moved file's handle, and retaining it leaves a stale name. Neither is decidable from inside `MovePath`, because an absent source looks identical in both.

The sound fix is not on this event at all. It belongs on the *destination* invalidation, which carries the new entry: when an entry arrives at a path whose currently mapped inode has a different identity, that inode was replaced, and both the unlink and the handle marking follow from that without guessing. That is a change to a path this PR does not otherwise touch, so I would rather do it separately than keep flipping four lines. Happy to open it if you want it now.

PR #10825, opened 2026-08-19
rust: stop writing when a durable write's index flush fails

chrislusf:
Follow-up to #10816, which merged before this review point was addressed - Greptile raised it two minutes after my reply and I did not catch it in time. Original comment: https://github.com/seaweedfs/seaweedfs/pull/10816#discussion_r-index-sync-failure

## The gap

A durable write flushes the `.dat`, publishes the needle map row, then flushes the `.idx`. #10816 handled a failure of the first flush - the append comes back off the end, nothing was ever published - and quarantines the volume if that truncate cannot be done. It did not handle a failure of the second one. That path returned the error and carried on with the row live and the volume writable, which has two consequences:

- The handler answers 500 and skips replication, while the primary keeps serving the new needle out of its live map. So the primary holds a needle the replicas never saw, for a write the client was told had failed.
- If the unflushed row is lost on restart, the durable `.dat` tail is exactly the state `check_volume_data_integrity` cannot account for, and `load()` sets `no_write_or_delete` anyway - the outcome the `.idx` flush existed to prevent.

## Why quarantine rather than roll back

Taking the row back out means undoing published state on a disk that is already failing. It also cannot be done safely: `nm.delete` appends a tombstone through the same failing `.idx`, and truncating the `.dat` afterwards would leave an `.idx` row pointing past the end - worse than what we started with. Undoing published state through the ordinary map API is also what the accounting review on #10816 rejected.

So the volume stops taking writes, matching what #10816 already does when the post-failure truncate fails. Nothing more is appended past a record whose index is in doubt, and the master routes writes elsewhere once the volume heartbeats read only.

This does not make the replicas consistent - the primary still has a needle they lack. That divergence is inherent once the bytes are down and cannot be safely removed; what changes is that it stays bounded and shows up as a read-only volume rather than quietly accumulating.

## Test

`test_write_needle_failed_idx_sync_quarantines_volume` injects the index flush failure through a `#[cfg(test)]` seam alongside the existing `.dat` one, and asserts the volume goes read only and refuses the next write. I checked it fails without the quarantine. Full `cargo test` passes (456).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved durability for deduplicated writes by synchronizing both data and index updates.
  - If index synchronization fails, the volume now enters read-only mode and rejects subsequent writes.
  - Successfully written data is preserved when a later index synchronization step fails.
- **Tests**
  - Added coverage for synchronization failures, volume quarantine, preserved data, and subsequent write rejection.

### DIFF
--- a/seaweed-volume/src/server/heartbeat.rs
+++ b/seaweed-volume/src/server/heartbeat.rs
@@ -771,6 +771,10 @@ fn collect_volume_snapshot(
     build_heartbeat_with_ec_status(config, &mut store, Vec::new(), true, false).1
 }
 
+/// The heartbeat alone, without the volume snapshot the send loop pairs it
+/// with. Only the tests want it that way; the loop calls
+/// collect_heartbeat_with_snapshot directly.
+#[cfg(test)]
 fn collect_heartbeat(
     config: &HeartbeatConfig,
     state: &Arc<VolumeServerState>,
--- a/seaweed-volume/src/storage/volume.rs
+++ b/seaweed-volume/src/storage/volume.rs
@@ -1537,11 +1537,13 @@ impl Volume {
     // ---- Write ----
 
     /// Write a needle to the volume (synchronous path).
-    /// Write a needle to the volume. With `fsync` the write is only reported
-    /// successful once both the .dat record and the .idx row that indexes it
-    /// are on disk, so an ack survives a crash rather than sitting in the page
-    /// cache. Both flushes happen under the same lock as the append, so a
-    /// failed one can take its own append back off the end and nobody else's.
+    ///
+    /// With `fsync` the write is only reported successful once both the .dat
+    /// record and the .idx row that indexes it are on disk, so an ack survives a
+    /// crash rather than sitting in the page cache. The two flushes fail
+    /// differently: a failed .dat flush happens before anything is published, so
+    /// it takes its own append back off the end and nobody else's, while a
+    /// failed .idx flush cannot be undone and stops the volume taking writes.
     pub fn write_needle(
         &mut self,
         n: &mut Needle,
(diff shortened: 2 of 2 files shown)


chrislusf:
Both taken, one of them only partly - reasons below.

**Update the durable-write comments** - done. `write_needle` now says the ack means both the `.dat` record and the `.idx` row that indexes it are down, and `flush_dat` says it is the first half of a durable write with `flush_idx` doing the second, split so nothing is indexed before the bytes it points at. Those comments predated the `.idx` flush and were stale as written.

**Route an nm.put error through the same quarantine path** - done, and it was a real asymmetry of mine. A failed `nm.put` after the `.dat` flush leaves exactly the durable-but-unindexed record that a failed `.idx` flush leaves, so it now quarantines the same way instead of propagating the error with the volume still writable. Only when `fsync` was asked for; a non-durable write has nothing on disk to protect. No dedicated test - injecting a `put` failure needs a third `#[cfg(test)]` seam and the path is three lines symmetric with the one above it, so I took it on inspection.

**Persist the quarantine across restart** - the finding is real and I reproduced it. A first durable write whose `.idx` flush fails, then a restart that loses the row, reloads as `read_only=false` with a 64-byte `.dat` holding an unindexed record and a 0-byte index. `check_volume_data_integrity` returns early on `idx_size == 0`, so nothing objects.

I have not fixed it here, for two reasons.

It is not something this PR introduced. Go does the same thing - `CheckVolumeDataIntegrity` in `volume_checking.go` has the identical `if indexSize == 0 { return 0, nil }`, and the Rust check mirrors it deliberately. The hole predates the fsync work in both implementations.

And the obvious narrow fix is wrong. "Empty `.idx` plus a `.dat` bigger than the superblock means unindexed data" breaks preallocated volumes: `preallocate_file` extends the `.dat` at creation, so every fresh volume started with `-volumePreallocate` would load read only. Telling a preallocated empty `.dat` from one holding real records means reading the `.dat` to look for a needle after the superblock, which is a new integrity heuristic rather than a guard, and it should be decided for the Go check and the Rust one together rather than bolted onto the port in a failure-path fix.

A durable quarantine marker has the same problem in a different place: it is a new on-disk artifact with no Go counterpart and no defined way for an operator to clear it. Worth doing, worth doing deliberately.

Full `cargo test` passes (456).

chrislusf:
Both lib build warnings are gone; `cargo build --all-targets` is now warning-free. Two commits, since they are unrelated to each other and to the fix.

`NeedleValue` was mine: it came in with `rollback_unflushed_write`, and when the durable path moved to flushing before it publishes, the function went but the import stayed. It slipped through because I had been grepping build output for `^error` by then. Note this means #10816 merged carrying it - the fix here removes it from master.

`collect_heartbeat` is older and not from this work. It has only ever been called from the tests; the send loop calls `collect_heartbeat_with_snapshot`, which it wraps. So a lib build was right to call it dead. Marked `#[cfg(test)]` rather than deleted, since the tests still want the heartbeat without the volume snapshot.

chrislusf:
Both taken, and the second was a real hole rather than a wording problem.

**Dedup skips the durable index flush** - correct. The dedup branch flushed the `.dat` and returned before ever reaching the index flush at the bottom of the function, so a `fsync=true` write that matched an earlier non-durable one was acked with the row that indexes it still in the page cache. That is exactly the false promise the index flush exists to rule out, and the same read-only volume on restart if the row is lost. I had reasoned about the `.dat` half of this case in the earlier round and wrote the comment about it, then missed that the early return skipped the other half.

The dedup path now flushes both files. Rather than repeat the failure handling, the quarantine moved inside `flush_idx`, so it applies wherever the flush is reached instead of only at the one call site that had it inline - which is also what made the omission easy to miss.

`test_dedup_durable_write_flushes_the_index` covers it: non-durable write, then the identical durable one with the index flush rigged to fail, asserting the write is refused and the volume quarantines. Without the fix the dedup path returns `Ok` and it fails.

**Describe the two failure paths accurately** - done, and the doc was worse than reported: my edit last round left the original `Write a needle to the volume (synchronous path).` line above the new block, so there were two first lines. It is one block now and it separates the two cases - a failed `.dat` flush happens before anything is published so it takes its own append back off the end, a failed `.idx` flush cannot be undone and stops the volume taking writes.

`cargo test` passes (457), `cargo build --all-targets` is warning-free.

Unrelated: the earlier CI failures on this PR were an `archive.ubuntu.com` stall during the protoc install - every cancelled job died on the same apt line without reaching a build. The re-run is fully green. #10830 removes that install step entirely.

PR #10827, opened 2026-08-19
mount: index directory state by path

chrislusf:
Phase 1 of #10820. Reopens #10823, which GitHub closed when its base branch went away with #10818; same branch, now based on master.

Every directory-state lookup (`MarkChildrenCached`, `TouchDirectory`, `InvalidateChildrenCache`, the subdir counters, the read-through flags — around 30 call sites) resolved a path through `path2inode` and then an inode through `dirStates`. `path2inode` is the map that holds one full path per inode in the table, and on the mount in #10020 that is ~5.6M entries.

Directories now carry their own materialized path and are indexed by it directly. There are orders of magnitude fewer directories than files, so this index stays small whatever the mount holds.

That is the groundwork for the rest of the plan: once a directory has a path of its own, a file does not need one — its path is its parent's plus its name — and the per-inode cost stops growing with directory depth.

No behavior change. `TestDirStateIndexesStayInStep` asserts the two indexes agree across lookup, rename, unlink and forget.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved directory state tracking during creation, renaming, removal, and cache operations.
  - Kept directory paths synchronized with their corresponding filesystem entries.
  - Improved handling and reporting of paths that are no longer tracked.
  - Prevented stale directory information from affecting newly recreated directories at the same path.

- **Tests**
  - Added coverage for directory and file lookup, renaming, removal, and forgetting.
  - Verified that directory indexes remain consistent throughout the directory lifecycle.
  - Added regression coverage for reused directory paths.

### DIFF
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -18,6 +18,9 @@ type InodeToPath struct {
 	// dirStates holds directory-only readdir-cache state, keyed by inode. An
 	// inode is a directory iff it has an entry here, registered at creation.
 	dirStates map[uint64]*dirState
+	// dirPaths indexes the same states by path, so a directory lookup never
+	// goes through the map that holds one full path per inode.
+	dirPaths map[util.FullPath]*dirState
 }
 
 // InodeEntry exists per inode the kernel references. Directory cache state is
@@ -31,6 +34,7 @@ type InodeEntry struct {
 }
 
 type dirState struct {
+	path              util.FullPath
 	isChildrenCached  bool
 	readDirDirect     bool
 	cachedExpiresTime time.Time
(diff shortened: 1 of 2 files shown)


chrislusf:
Both valid, fixed.

The ownership check is a real bug in this commit: a released directory's state keeps the path it had, so `rmdir /a; mkdir /a` followed by the old inode's FORGET had `dropDirPath` delete the index entry belonging to the *live* `/a`. Everything keyed off it — cached-children, read-through, refresh, subdir counts — then silently no-opped for a directory that was perfectly alive. `dropDirPath` now drops the index only while it is still the state that path resolves to, and `MovePath` goes through the same helper instead of deleting by path itself.

The reverse invariant is now asserted too, and it catches exactly that: without the fix `TestForgetDoesNotDropAReusedDirPath` fails with both `live directory lost its index` and `directory state at /a is missing from dirPaths`.

PR #10829, opened 2026-08-19
topology: refresh oversized mark on every heartbeat

sunchao6106:
## Problem: writable/unwritable flip loop on an oversized volume

A volume that grows past `volumeSizeLimit` keeps **bouncing between writable and unwritable on every heartbeat/assign cycle**, because two independent paths use different facts to decide writability:

- `ensureCorrectWritables` (heartbeat path) checks the **`oversized` bit flag** on the volume location
- `RecordAssign` (write path) checks **`effectiveSize`** (real-time, heartbeat-decayed)

The `oversized` bit is only ever set at **registration time** (`RegisterVolume` → `rememberOversizedVolume`). The delta heartbeat path (`ApplyVolumeChanges`) and the full heartbeat path (`SyncDataNodeRegistration`) never refresh it — so once a volume outgrows the limit, the bit stays stale `false`, and `ensureCorrectWritables` keeps re-adding the volume to `writables` on every heartbeat while `RecordAssign` removes it on every assign.

### Production symptom

```
I0819 10:21:20.613067 volume_layout.go:360 Volume 166140692 reaches full capacity (effective=35551735903, limit=31457280000).
I0819 10:21:20.613049 volume_layout.go:784 Volume 166140692 becomes unwritable
I0819 10:21:20.032460 volume_layout.go:796 Volume 166140692 becomes writable
I0819 10:21:19.860205 volume_layout.go:360 Volume 166140692 reaches full capacity (effective=35551633399, limit=31457280000).
I0819 10:21:19.860192 volume_layout.go:784 Volume 166140692 becomes unwritable
I0819 10:21:19.559009 volume_layout.go:796 Volume 166140692 becomes writable
```

```
E0819 10:18:43.680032 volume_grpc_vacuum.go:72 failed compact volume 166140692: volume 166140692 unexpected new data size: 13922058544 does not match size of content minus deleted: 13933663196
```

Downstream effects:

1. **Volume exceeds the limit**: effectiveSize is an estimate (1 MB per file default) that trails the real on-disk size, and the flip loop keeps the volume writable, so writes continue past the 30 GB limit (actual 35.5 GB).
2. **Vacuum fails**: the volume is still writable during compaction, so `CompactByIndex`'s snapshot disagrees with concurrent writes (`expectedContentSize > dstDatSize`).
3. **Replica divergence**: the three replicas report slightly different sizes (35551436816 / 35551436560 / 35551427488), and the flip loop masks the difference instead of surfacing it.

## Fix

Refresh the `oversized` mark from **each heartbeat's reported size** in both heartbeat paths, mirroring what `RegisterVolume` already does at registration:

- `ApplyVolumeChanges` (delta heartbeats)
- `SyncDataNodeRegistration` (full heartbeats)

New method `VolumeLayout.UpdateOversizedState` wraps the existing `rememberOversizedVolume` with the layout lock. It sets the bit to `true` when the reported size ≥ limit and to `false` when the volume shrank back — so a compacted volume can recover writability, and a grown volume stays out.

This makes both consumers of `AnyOversized` correct again:

- `ensureCorrectWritables` no longer re-adds an oversized volume to `writables`
- `UpdateVolumeSize`'s recovery path (`if vl.vid2location[vid].AnyOversized()`) no longer lets a stale false mark restore a still-oversized volume

## Behavior

| State | Before | After |
|---|---|---|
| Volume grows past limit | writable/unwritable flip every ~4 s; writes continue past limit | stays unwritable; writes reassigned |
| Volume compacts back under limit | (n/a — bit was stale) | oversized mark cleared; volume can recover after `capacityRecoveryDelay` |
| Delta heartbeat (`ApplyVolumeChanges`) | stale bit | refreshed from reported size |
| Full heartbeat (`SyncDataNodeRegistration`) | stale bit (unless re-registered) | refreshed from reported size |

## Tests

`volume_layout_oversized_heartbeat_test.go`:
- `TestUpdateOversizedStateKeepsOversizedVolumeUnwritable`: after `RecordAssign` removes the volume, a heartbeat reporting an oversized size keeps it out of `writables` via `EnsureCorrectWritables`.
- `TestUpdateOversizedStateClearsMarkWhenVolumeShrinks`: a later heartbeat with a normal size clears the mark so the volume is not locked out forever.

`go test ./weed/topology/` and `go vet` pass.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved volume state updates during registration and heartbeat processing.
  * Volumes reaching their hard capacity limit are consistently marked unwritable.
  * Oversized status is refreshed correctly as volume sizes change.
  * Capacity-reduced volumes remain unwritable until the recovery cooldown expires.
  * Crowded volumes can recover to writable status after replicas return and effective capacity is restored.
  * Volumes still at their hard capacity limit remain protected from premature reactivation.

### DIFF
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -637,11 +637,10 @@ func (t *Topology) SyncDataNodeRegistration(volumes []*master_pb.VolumeInformati
 		// Without this, the volume stays visible in volume.list/admin UI yet
 		// LookupVolume returns "volume id not found".
 		if !vl.HasDataNode(v.Id, dn) {
-			// RegisterVolume sets the oversized mark via its defer; the writable
-			// correction runs once, after UpdateVolumeSize below.
-			if !vl.RegisterVolume(&v, dn) {
-				// Dropped with its collection; re-resolve.
-				t.RegisterVolumeLayout(v, dn)
+			for !vl.RegisterVolume(&v, dn) {
+				// Dropped with its collection; the next lookup creates a fresh
+				// one, which the calls below must use too.
+				vl = t.GetVolumeLayout(v.Collection, v.ReplicaPlacement, v.Ttl, diskType)
 			}
 			// Volumes new to the disk map were registered above, so reaching
 			// here means only the lookup index had lost it. Clients were told
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -655,6 +655,7 @@ func (t *Topology) SyncDataNodeRegistration(volumes []*master_pb.VolumeInformati
 			// it is back.
 			newVolumes = append(newVolumes, v)
 		}
+		vl.UpdateOversizedState(&v, dn)
 		if vl.UpdateVolumeSize(v.Id, v.Size, v.CompactRevision) {
 			vl.AdjustActiveVolumeCountAfterRecovery(v.Id)
 		}
@@ -731,6 +732,7 @@ func (t *Topology) ApplyVolumeChanges(changed []*master_pb.VolumeInformationMess
 		if isNew || becameServable {
 			newVolumes = append(newVolumes, vi)
 		}
+		vl.UpdateOversizedState(&vi, dn)
 		vl.EnsureCorrectWritables(&vi)
 		if vl.UpdateVolumeSize(vi.Id, vi.Size, vi.CompactRevision) {
 			vl.AdjustActiveVolumeCountAfterRecovery(vi.Id)
--- a/weed/topology/topology.go
+++ b/weed/topology/topology.go
@@ -613,18 +613,12 @@ func (t *Topology) SyncDataNodeRegistration(volumes []*master_pb.VolumeInformati
 		}
 	}
 	// find out the delta volumes
-	var changedVolumes []storage.VolumeInfo
-	newVolumes, deletedVolumes, changedVolumes = dn.UpdateVolumes(volumeInfos)
+	newVolumes, deletedVolumes, _ = dn.UpdateVolumes(volumeInfos)
 	for _, v := range newVolumes {
 		t.RegisterVolumeLayout(v, dn)
 	}
 	for _, v := range deletedVolumes {
 		t.UnRegisterVolumeLayout(v, dn)
-	}
-	for _, v := range changedVolumes {
-		diskType := types.ToDiskType(v.DiskType)
-		vl := t.GetVolumeLayout(v.Collection, v.ReplicaPlacement, v.Ttl, diskType)
-		vl.EnsureCorrectWritables(&v)
 	}
 	// Update effective sizes for all reported volumes (decay pending estimates).
 	// If decay brings a volume eagerly removed by RecordAssign back under the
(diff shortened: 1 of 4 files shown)


sunchao6106:
Addressed both review comments and pushed `f552d04e6`.

## Greptile: recovery cooldown bypass

The delta path cleared the oversized mark before `EnsureCorrectWritables`, whose add-to-writable branch had no cooldown check — a just-compacted volume was re-added immediately, bypassing `capacityRecoveryDelay`.

Fix: `ensureCorrectWritables` now checks `fullSince` and skips the re-add while the cooldown is pending (`volume_layout.go`). A volume removed for capacity can only recover through `UpdateVolumeSize`'s heartbeat recovery path, which enforces the delay.

## CodeRabbit: full-heartbeat ordering

In `SyncDataNodeRegistration` the mark was refreshed after the writable correction, so a newly oversized volume stayed writable for an extra heartbeat cycle.

Fix: the standalone `changedVolumes` loop is merged into the `volumeInfos` loop, and `EnsureCorrectWritables` now runs **after** `UpdateOversizedState` + `UpdateVolumeSize` in both heartbeat paths — it always sees the freshly refreshed mark.

## Note on the modified existing test

`TestHandlingVolumeServerHeartbeat` used a volume size of **254320** with a test `volumeSizeLimit` of **32768** (32 KiB). That size is *past* the limit — the test only passed because the stale oversized mark hid the oversized state. With the fix, those volumes are correctly kept out of `writables` (6 → 1), which is the intended behavior of this PR.

Since the test's purpose is the add/remove heartbeat flow (not oversized behavior), I sized the volumes down to **30000** — still below the limit, still a change from the initial 25432 (so `changedVolumes` still fires), but no longer accidentally exercising the oversized path. If you'd rather keep the original size and update the assertion to match the new correct behavior instead, I'm happy to do that.

New test `TestEnsureCorrectWritablesHonorsRecoveryCooldown` covers the cooldown window (stays unwritable) and the recovery after it.

`go test ./weed/topology/` and `go vet` pass.

sunchao6106:
Good catch — that is a real gap. Pushed `f8ff1c533`.

**Issue**: after `capacityRecoveryDelay` elapses, `ensureCorrectWritables` could restore a volume whose `effectiveSize` is still past the crowded threshold. The cooldown check only looked at `fullSince`; once the delay passed, it re-added the volume even though `UpdateVolumeSize` had just refused recovery (`effectiveSize > crowdedThreshold` → `setVolumeCrowded` + `return false`).

**Fix**: `ensureCorrectWritables` now checks the crowded mark before re-adding:

```go
if _, crowded := vl.crowded[vid]; crowded {
    return
}
```

A volume `UpdateVolumeSize` just marked crowded stays out of `writables`; only the recovery path (which clears the mark once `effectiveSize` falls under the threshold) brings it back.

**Test**: `TestEnsureCorrectWritablesDoesNotRestoreCrowdedVolume` — `effectiveSize` decays to 10500 (past the 9000 crowded threshold) after a report of 8000, and `ensureCorrectWritables` keeps the volume unwritable past the cooldown.

`go test ./weed/topology/` and `go vet` pass.

sunchao6106:
Hi @chrislusf, just pushed an empty commit (`51a99c83d`) to re-trigger CI — the two FUSE jobs were stuck on an infra flake (apt mirror timeout on `azure.archive.ubuntu.com` during the docker build stage; the previous commit's End to End run passed).

The change itself is only in `weed/topology` (heartbeat oversized-mark refresh + recovery protections), and all code-related checks are green — Greptile and CodeRabbit both approved after their comments were addressed.

Would love your thoughts on the approach when you have a moment:
- Refreshing the oversized mark on every full/delta heartbeat (previously only set at `RegisterVolume`)
- Ordering heartbeat processing as `UpdateOversizedState → UpdateVolumeSize → EnsureCorrectWritables`
- Adding a `capacityRecoveryDelay` cooldown and a crowded-state guard so `ensureCorrectWritables` can't restore a volume that `UpdateVolumeSize` deliberately refused

Happy to adjust if you'd prefer a different direction.

PR #10835, opened 2026-08-19
Add monitoring label to filer servicemonitor

slimes28:
The filer pods are being double-scraped due to 2 services pointing to the same pods. Fixing https://github.com/seaweedfs/seaweedfs/issues/10836

Adding monitoring: "true" to the label selector for the filer service monitor. This label is already present in [filer-service-client](https://github.com/seaweedfs/seaweedfs/blob/master/k8s/charts/seaweedfs/templates/filer/filer-service-client.yaml#L14)

# How is the PR tested?
Rendered the `seaweedfs` chart locally with `helm template` before/after the change and confirmed:
- The filer `ServiceMonitor`'s `spec.selector.matchLabels` now includes `monitoring: "true"`, matching only `<release>-filer-client` (which already carries this label) instead of matching both `<release>-filer` and `<release>-filer-client`.
- `<release>-filer` (the headless service without `monitoring: "true"`) is no longer selected by the ServiceMonitor.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Enabled monitoring discovery for the filer service, allowing supported monitoring systems to collect its metrics.

### DIFF
--- a/k8s/charts/seaweedfs/templates/filer/filer-service-client.yaml
+++ b/k8s/charts/seaweedfs/templates/filer/filer-service-client.yaml
@@ -10,9 +10,6 @@ metadata:
     app.kubernetes.io/managed-by: {{ .Release.Service }}
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/component: filer
-{{- if .Values.filer.metricsPort }}
-    monitoring: "true"
-{{- end }}
 {{- if .Values.filer.annotations }}
   annotations:
     {{- toYaml .Values.filer.annotations | nindent 4 }}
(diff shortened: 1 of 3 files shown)


chrislusf:
Pushed a follow-up commit. The duplicate scrape was real and the selector change is right, but it selected the wrong one of the two services, so I moved the `monitoring: "true"` label from `filer-service-client.yaml` to `filer-service.yaml` instead.

Two reasons:

The ServiceMonitor sets no `jobLabel`, so prometheus-operator derives `job` (and `service`) from the service name. The dashboard this chart ships in `dashboards/seaweedfs-grafana-dashboard.json` queries `job="seaweedfs-filer"` in six expressions across three panels, and its ConfigMap is gated on the same `global.seaweedfs.monitoring.enabled` as the ServiceMonitor. Selecting `-filer-client` would have moved the job to `seaweedfs-filer-client` and left those panels blank.

The headless service also sets `publishNotReadyAddresses: true`; the client service does not. It is the StatefulSet governing service, so it keeps reporting while a filer is starting up or shutting down.

Rendered all three combinations of `filer.metricsPort` and `global.seaweedfs.monitoring.enabled`: exactly one service carries the label when the ServiceMonitor exists, and no label is emitted without a metrics port.

PR #10842, opened 2026-08-19
mount: discard a path-cache insert that raced a purge

chrislusf:
Root-causes the `TestRenameOverExisting` flake from the `mount: windows` run on #10827 — `source survived the rename`, once, passing on re-run of the same commit.

What made it worth chasing is that every SeaweedFS layer is provably synchronous with the rename: `doRename` consumes the filer stream before returning, `handleRenameResponse` runs `MovePath` and the local-store delete in-line (`ApplyMetadataResponseOwned` blocks on the apply worker's `done`), and the adapter purges its path cache after `wfs.Rename` returns. The mount's error log shows nothing at the failure instant. By the time the test called `os.Stat(src)`, the filer, the local store, the inode table, and the adapter cache had all already dropped the name.

What none of that covers is a resolve **in flight** during the rename. The adapter's `walk` resolves a component with a Lookup RPC and inserts the result into the path cache holding no lock:

1. a concurrent opener of `src` misses the cache and issues `Lookup(src)` — succeeds, pre-rename
2. the rename completes; `purgeWithParent(src)` runs — nothing to purge yet
3. the opener's `insert(src, inode)` lands — the cache now holds the name the rename just deleted, for up to the 1s Windows cache TTL
4. `os.Stat(src)` is served the stale entry

The test is single-threaded but the machine is not — an antivirus scan of a just-written file is exactly the background open a CI runner supplies, and the need for a millisecond interleaving matches the once-in-hundreds frequency. The release path already guards its equivalent insert (`stillNames`, "caching it would resurrect an entry that is gone"); the walk had nothing, and its window is RPC-wide.

The cache counts purges now. A resolve snapshots the generation before its lookups; `insert` discards the entry when any purge ran in between, parking the reference in the graveyard — which exists precisely so an in-flight holder keeps a valid inode — so ownership semantics are unchanged in every outcome. Purge bumps unconditionally, since the point is as much the in-flight resolve about to insert a name as anything the map currently holds. All three insert sites pass a snapshot.

Unit tests cover discard-on-purge, discard-under-a-purged-prefix, the reference coming back through the graveyard, and the undisturbed path; the discard tests fail without the guard (`purged name served from an insert that started before the purge`).

The assertion that caught this also discarded its evidence — `!os.IsNotExist(err)` conflates a stale positive with any transient error, and the message dropped both `err` and the `FileInfo`. It now reports what stat returned and whether it persisted 200ms later, so a recurrence indicts a specific layer instead of reading as a mystery.

Verified: `GOOS=windows go build` on the adapter and command packages, package tests with `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file rename handling when source paths remain briefly visible.
  * Prevented stale path information from being reused after targeted cache invalidation.
  * Improved consistency when creating directories and restoring cached paths.
  * Delayed removal of recently discarded path references to support concurrent operations more reliably.

* **Tests**
  * Expanded coverage for cache invalidation, concurrent lookups, path-prefix handling, and delayed cleanup.
  * Added verification that unrelated cache invalidations do not discard valid path information.

### DIFF
--- a/test/winfsp/semantics_test.go
+++ b/test/winfsp/semantics_test.go
@@ -180,7 +180,7 @@ func TestRenameOverExisting(t *testing.T) {
 		// they indict different layers; a second look says whether it persists.
 		time.Sleep(200 * time.Millisecond)
 		fi2, err2 := os.Stat(src)
-		t.Fatalf("source survived the rename: stat=%v err=%v; 200ms later stat=%v err=%v",
+		t.Fatalf("stat of the renamed-away source did not return not-exist: stat=%v err=%v; 200ms later stat=%v err=%v",
 			describeFileInfo(fi), err, describeFileInfo(fi2), err2)
 	}
 }
--- a/test/winfsp/semantics_test.go
+++ b/test/winfsp/semantics_test.go
@@ -175,8 +175,13 @@ func TestRenameOverExisting(t *testing.T) {
 	if string(got) != "new" {
 		t.Fatalf("target holds %q, want the source content", got)
 	}
-	if _, err := os.Stat(src); !os.IsNotExist(err) {
-		t.Fatal("source survived the rename")
+	if fi, err := os.Stat(src); !os.IsNotExist(err) {
+		// Both a stale positive and a transient non-ENOENT error land here, and
+		// they indict different layers; a second look says whether it persists.
+		time.Sleep(200 * time.Millisecond)
+		fi2, err2 := os.Stat(src)
+		t.Fatalf("source survived the rename: stat=%v err=%v; 200ms later stat=%v err=%v",
+			describeFileInfo(fi), err, describeFileInfo(fi2), err2)
 	}
 }
 
@@ -514,4 +519,11 @@ func TestDeleteOnClose(t *testing.T) {
 	if err := closeHandle(h); err != nil {
 		t.Fatalf("close after recreate: %v", err)
 	}
+}
+
+func describeFileInfo(fi os.FileInfo) string {
+	if fi == nil {
+		return "<nil>"
+	}
+	return fmt.Sprintf("{size=%d mode=%v mtime=%s}", fi.Size(), fi.Mode(), fi.ModTime().Format(time.RFC3339Nano))
 }
--- a/weed/mount/winfsp/fs_windows.go
+++ b/weed/mount/winfsp/fs_windows.go
@@ -192,6 +192,10 @@ func (w *WinFS) forget(inode uint64) {
 // to hold the inode longer steals the reference from the cache.
 func (w *WinFS) walk(parts []string) (uint64, fuse.Status) {
 	inode := uint64(rootInode)
+	// Taken before the lookups: a purge racing this walk - a rename or unlink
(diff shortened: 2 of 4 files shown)


chrislusf:
All four were valid, including the one I most needed pointing out.

**The overwritten tests** — right, and worse than a style issue: I created `pathcache_test.go` with a heredoc without noticing the file already existed, which silently destroyed the suite covering expiry, steal ownership, replacement, prefix boundaries, root handling and cacheKey. Restored in full, adapted to the new insert signature, with the race tests added alongside rather than instead.

**Unrelated purges starving resolveAndSteal** — confirmed against the code: four resolve-then-steal attempts and then the open fails with EIO, so a global generation under sustained notifier churn could fail opens of untouched paths — a regression worse than the staleness this PR fixes. Purges are now remembered by key (a bounded window of 128) and only a purge that covers the inserted name discards it; a resolve older than the window is discarded without a check, which costs a retry, not an open. `TestPathCacheUnrelatedPurgeDoesNotDiscard` and `TestPathCacheInsertOlderThanPurgeWindowIsDiscarded` pin both sides.

**The same-call sweep forgetting the discarded reference** — both reports correct, and the commit message's own claim ('the in-flight caller keeps a valid inode either way') was false whenever the discarding insert happened to trip the sweep. The graveyard holds two generations now, so an appended reference always survives the sweep of its own call — which quietly fixes the same boundary case for the pre-existing displaced-entry and purge appends, whose 'sits out one sweep' guarantee had the identical hole. `TestPathCacheDiscardedInsertRestsBeforeForget` fails against the single-generation sweep with 'discarded insert forgotten in the same call'.

**The failure message** — reworded to 'stat of the renamed-away source did not return not-exist', reporting both observations without claiming survival it did not establish.

chrislusf:
Both right, and the overflow one deserved a structural answer rather than a bigger window.

**Overflow starving opens** — the keyed window narrowed how often an insert is discarded, but past 128 purges the discard is blind again, so the EIO cliff had only moved. The real fragility was the shape of `resolveAndSteal`: it cached the final component only to steal it back, making an open depend on an insert surviving whatever races it. It no longer does — a cached entry is still stolen, and anything else is looked up directly with the caller owning the reference from the start. The retry loop and its EIO are gone entirely, so no amount of churn — covered, unrelated, or overflowing the window — can fail an open. Discards now cost at most a cache miss, which is the only currency they should ever spend. The cost of the change is that an open no longer leaves the final component cached; the handle itself serves that path's operations, and Release re-inserts it, so a concurrent second opener pays one extra Lookup.

**The root purge hole** — confirmed: `purge("", true)` clears every entry, but `covers` tested for a `/`-prefixed key that a normalised key never has, so a whole-cache purge covered no in-flight insert at all. Fixed with the empty-key case, and `TestPathCacheRootPurgeCoversEveryInsert` fails without it.

I could not unit-test the `resolveAndSteal` change off-Windows — `fs_windows.go` is platform-tagged — so it is verified by `GOOS=windows` build and vet here, and by the winfsp conformance and mount-windows workflows on this PR.

PR #10868, opened 2026-08-21
S3: source config credentials from the environment, and let the chart point at an existing secret

### LINKED ISSUE #10852: [Helm] Source s3 credentials (admin/read) from an existing Secret keyRef, GitOps-safe
### Chart version
seaweedfs 4.43.0 (Helm v4)

### What I'm trying to do
Have the S3 gateway's **admin/read identities** reuse credentials that already live in an **existing Kubernetes Secret** (managed elsewhere), **without**:
- putting the secret values as literals in `values.yaml` / `--set` (ends up in Git), or
- hand-authoring and maintaining the whole `seaweedfs_s3_config` JSON.

### Current options (chart 4.43.0)
1. `s3.credentials.admin/read.{accessKey,secretKey}` — **literal strings only** (no `valueFrom`/`secretKeyRef`); the template embeds them directly into the generated `seaweedfs_s3_config`.
2. `s3.existingConfigSecret` — reference a Secret that contains the **entire** `seaweedfs_s3_config` JSON (all-or-nothing; I must build/maintain the full JSON).
3. `reuseLegacySecret` + `legacySecretName` — only reuses the chart's **own** generated secret, and only with the chart's fixed key names (`admin_access_key_id`, `admin_secret_access_key`, ...), not an arbitrary secret's keys.

There's no middle ground: "compose the generated config from keys in an existing arbitrary Secret."

### Constraint I understand
The gateway consumes a **file** (`-config=/etc/sw/seaweedfs_s3_config`), so a true k8s `secretKeyRef` (env-only) can't populate inner JSON fields. The chart already resolves existing secrets via `getOrGeneratePassword` -> `lookup`, but `lookup` returns empty under `helm template` / `--dry-run` / Argo CD server-side render, so a naive lookup-based keyRef would silently render empty creds in GitOps diffs.

### Proposal / question
A dry-run/GitOps-safe way to source admin (and read) creds from an existing Secret's keys, e.g. optional:
```yaml
s3:
  credentials:
    admin:
      existingSecret: my-secret
      accessKeyKey: root-user
      secretKeyKey: root-password
```
...or maintainers' preferred design (docs/blessed pattern) if `lookup` fragility makes this undesirable. Happy to help/PR once there's agreement on the approach.

### Use case
Migrating MinIO -> SeaweedFS while **reusing the same credentials on both systems**; the admin key already exists in a k8s Secret, and we'd rather reference it than duplicate it into the chart values or the full identities JSON.

chrislusf:
Fixes #10852.

The chart had no middle ground between "put the keys in values.yaml" and "hand-author the whole `seaweedfs_s3_config` JSON in your own Secret". A `secretKeyRef` cannot fill in fields of a file, and anything built on `lookup` renders empty under `helm template`, `--dry-run` and an Argo CD diff.

So let the file carry the reference instead of the key.

**`weed s3`**: a key in the static config file may be written as `${VAR}` and is resolved from the environment when the file is loaded. Only the static file is expanded, never identities coming from the filer, and a credential whose reference stays unset is dropped with a warning rather than becoming a literal key.

**chart**: `s3.credentials.admin` and `s3.credentials.read` take `existingSecret`, `accessKeyKey` and `secretKeyKey`:

```yaml
s3:
  enabled: true
  enableAuth: true
  credentials:
    admin:
      existingSecret: minio-root
      accessKeyKey: root-user
      secretKeyKey: root-password
```

The generated config then holds `${SEAWEEDFS_S3_ADMIN_ACCESS_KEY_ID}`, the chart wires that variable up from the Secret on the s3, filer and all-in-one pods, and the chart's own `admin_access_key_id` keys are left out since the keys live elsewhere. Nothing is read from the cluster at render time, so a dry run renders exactly what an install applies. The COSI driver parses the config itself and does not resolve the references; that is called out in the docs.

Tested: unit tests for expansion, for the dynamic path staying verbatim, for a literal `$` staying literal, and for the unset-reference drop; a chart render assertion in helm CI; and end to end against a gateway started with the placeholders, where the env-supplied admin key writes, the env-supplied read key reads, a wrong secret gets `SignatureDoesNotMatch`, and the unexpanded placeholder gets `InvalidAccessKeyId`.

## Summary by CodeRabbit

- **New Features**
  - Helm deployments can source S3 admin and read credentials from existing Kubernetes Secrets.
  - Secret key names are configurable, with credentials passed securely through environment variables across supported S3 workloads.
  - Static S3 configurations support environment-variable references for credentials.

- **Bug Fixes**
  - Unresolved or malformed credential references are safely excluded.
  - Literal `${...}` text returned through environment variables is preserved.

- **Documentation**
  - Added guidance for configuration, credential rotation, and compatibility considerations.

### DIFF
--- a/.github/workflows/helm_ci.yml
+++ b/.github/workflows/helm_ci.yml
@@ -59,16 +59,36 @@ jobs:
           echo "S3 deployment renders correctly"
           
           echo "=== Testing S3 credentials from an existing secret ==="
-          helm template test $CHART_DIR \
-            --set s3.enabled=true --set s3.enableAuth=true \
-            --set s3.credentials.admin.existingSecret=minio-root \
-            --set s3.credentials.admin.accessKeyKey=root-user \
-            --set s3.credentials.admin.secretKeyKey=root-password \
-            > /tmp/s3-existing-credentials.yaml
-          grep -q 'accessKey":"${SEAWEEDFS_S3_ADMIN_ACCESS_KEY_ID}' /tmp/s3-existing-credentials.yaml
-          grep -q 'key: "root-password"' /tmp/s3-existing-credentials.yaml
-          ! grep -q "admin_access_key_id:" /tmp/s3-existing-credentials.yaml
-          echo "S3 credentials reference the existing secret"
+          credential_args=(
+            --set s3.credentials.admin.existingSecret=minio-root
+            --set s3.credentials.admin.accessKeyKey=root-user
+            --set s3.credentials.admin.secretKeyKey=root-password
+            --set s3.credentials.read.existingSecret=minio-root
+          )
+          for workload in \
+            "s3.enabled=true,s3.enableAuth=true" \
+            "filer.s3.enabled=true,filer.s3.enableAuth=true" \
+            "allInOne.enabled=true,allInOne.s3.enabled=true,allInOne.s3.enableAuth=true"
+          do
+            helm template test $CHART_DIR --set "$workload" "${credential_args[@]}" > /tmp/s3-existing-credentials.yaml
+            # The identities file names the variables, and each is bound to the key it was pointed at.
+            grep -q 'accessKey":"${SEAWEEDFS_S3_ADMIN_ACCESS_KEY_ID}' /tmp/s3-existing-credentials.yaml
+            grep -q 'secretKey":"${SEAWEEDFS_S3_ADMIN_SECRET_ACCESS_KEY}' /tmp/s3-existing-credentials.yaml
+            grep -q 'accessKey":"${SEAWEEDFS_S3_READ_ACCESS_KEY_ID}' /tmp/s3-existing-credentials.yaml
+            for pair in \
+              "SEAWEEDFS_S3_ADMIN_ACCESS_KEY_ID root-user" \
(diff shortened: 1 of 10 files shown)


chrislusf:
Addressed all three, one commit each.

**Empty environment value accepted as a credential** (greptile) - real, and worse than it looks: a Secret key that exists but is blank would have left an access key that any signature matches. `expandEnvRefs` now counts set-but-empty as unresolved, so the credential is dropped like any other unresolved reference. Test added.

**`allInOne.s3.enableAuth` alone never renders the secret** (coderabbit) - confirmed, and it predates this PR: the all-in-one deployment mounts `<fullname>-s3-secret` whenever any of the three enableAuth flags is set, but the secret template only looked at the s3 and filer flags, so the pod waited on a secret nothing creates. The secret now renders on the all-in-one flag too, and stays off when any `existingConfigSecret` is set, matching how the deployment picks the secret name. I diffed the render across nine flag combinations: that one case is the only behaviour that changed.

**Render check too narrow** (coderabbit) - it only looked at the standalone s3 deployment and at one of the four variables. It now loops over the s3, filer and all-in-one workloads and asserts, for each variable, both the secret it binds to and the key within it, plus the read identity's defaults and the absence of inlined keys. I checked the new assertions fail when the mapping is wrong rather than passing vacuously.

PR #10878, opened 2026-08-22
admin: add visual iam policy editor

mat813:

Add a structured, tabbed editor (Editor / JSON) for creating and editing IAM policies in the admin dashboard, alongside the existing raw-JSON textarea:

- **policies.templ**: per-statement cards for Sid, Effect, Action, Resource/NotResource, and Principal/NotPrincipal. Resource/NotResource and Principal/NotPrincipal are each backed by a two-entry mode dropdown (they're mutually exclusive), sharing one value list per statement. Principal/NotPrincipal supports the standard `{"AWS": [...]}` form and the bare `"*"` wildcard (AWS shorthand for "everyone"); any other shape (a different type key, multiple type keys, or a bare non-`"*"` string/array) stays untouched in the statement's advanced-fields box rather than being guessed at. Any other unmanaged field (Principal/NotPrincipal in an unsupported shape, Condition, or anything else) is preserved verbatim in a per-statement "advanced fields" JSON box so nothing is lost on round-trip. Switching tabs commits and reparses in both directions. Restored the "Use Sample Policy" button, now filling both the structured editor and the JSON tab. The "Validate" button now calls the existing but previously unused `POST /api/object-store/policies/validate` endpoint instead of doing JS-only checks. Per-statement sections are grouped into `<fieldset>`/`<legend>` groups with custom styling so the legend hugs its content instead of stretching full-width.
- Progressive Resource ARN autocomplete: suggests bucket names (and `bucket/*`) first, then once `"bucket/"` is typed, suggests `bucket/*` plus the bucket's direct subfolders, drilling down one path segment at a time as the user types further `"/"` characters.
- Principal/NotPrincipal autocomplete: new `GET /api/principals` endpoint (`weed/admin/dash/principal_suggestions.go`) combining existing S3 user ARNs (`arn:aws:iam::000000000000:user/<name>`) and IAM role ARNs (`arn:aws:iam::role/<name>`, via the existing `FilerRoleStore.ListRoles`) into one flat, cached suggestion list, with `"*"` always offered first. Service accounts aren't listed separately since their ARN is identical to their parent user's.
- New `GET /api/files/list-folders` endpoint (`file_browser_handlers.go`) backing the folder autocomplete: wraps the existing file browser data function and returns just the subdirectory names as JSON, scoped to paths under `/buckets`.
- Action-name suggestions (datalist) for the Action field, sourced from the existing `s3_constants.S3_ACTION_*` constants plus new `s3_constants.S3TABLES_ACTION_*` constants (extracted from the s3tables operation dispatch switch) so the suggestion list can't drift from the strings the engines actually understand.
- `policy_handlers.go`: `ValidatePolicy` now accepts a statement with only `NotResource` set (previously required `Resource`), matching `policy_engine.validateStatement` and the fact the editor makes such statements reachable from the UI; also rejects a statement specifying both `Resource`+`NotResource` or both `Principal`+`NotPrincipal` (by field presence, not just non-empty length).
- Hardening: escapes all interpolated policy fields before rendering the read-only View modal (stored-XSS fix); rejects malformed/non-`"Allow"|"Deny"` Effect values and non-object JSON documents instead of silently defaulting/guessing; the Editor and JSON tabs never silently clobber each other's edits; a save that would drop an unsupported top-level field now asks for confirmation first.
- Tests: `ValidatePolicy` behavior (including the new conflict checks), route registration for the policy API, the `list-folders` and `principals` endpoints, `list-folders` path scoping, and the action-suggestion list's shape.

# How is the PR tested?

Added tests, build, vet, and "played" in the admin interface.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added an API for listing folders in the file browser.
  * Introduced structured and raw JSON policy editors with tabs, synchronization, samples, and action suggestions.
  * Added support for S3 Tables policy actions.
  * Policy views now display statement IDs and support `NotResource`.

* **Bug Fixes**
  * Policy validation now accepts `NotResource`-only statements and rejects conflicting resource fields.
  * Improved validation for invalid paths, missing fields, and malformed policy JSON.
  * Preserved additional policy fields during editing and validation.
  * Added warnings when unsupported fields may be dropped during saving.


mat813:
Thank you for the review and the fixes.
I updated the description with a more comprehensive version, I forgot to update it when I added more features along the way.

PR #10890, opened 2026-08-23
s3api: retry a transient filer failure on metadata listings

### LINKED ISSUE #7221: Transient error on `ListEntries` causes listing multipart uploads fail
### Description

When using SeaweedFS in our cluster, we noticed a bug that transient error on RPC `ListEntries` leads to that listing multipart uploads fails. We expect that transient error on it can be handled by correct retrying mechanism.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22252875/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_list_multipart_uploads`. https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L6475

Transient error can be reproduced by gRPC interceptor.

chrislusf:
Fixes #7221. References #7235.

Carries #10886 by @junkerderprovinz, whose fork rejects maintainer pushes, with the review restructure applied on top: the bounded retry (3 attempts, 100ms doubling backoff) now lives inside `(*S3ApiServer).list` itself rather than being wrapped around the two multipart call sites. The accumulator is a local of `list`, so a replay there is safe for every call site by construction, and all metadata listings get the fix — versioning scans, bucket scans, the owner index, multipart. Classification is `!isFilerNotFound(err) && util.IsTransientError(err)`; the message-based check is required because `DoSeaweedListWithSnapshot` wraps a failed `ListEntries` with `%v`, which drops the gRPC status from the chain, and `IsTransientError` already refuses cancelled contexts.

The original PR's tests carry over unchanged: they script a fake filer client and stream against the real `filer_pb.List` path, covering recovery from a transient `ListEntries` failure, recovery from a mid-stream `Recv` failure, giving up after the bound, and not retrying permission-denied, invalid-argument, or not-found.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of S3 file listings by automatically retrying transient failures.
  * Limited retries to prevent prolonged delays, while preserving non-retryable and not-found errors.
  * Ensured partially received listing results are safely retried without producing incomplete output.

* **Tests**
  * Added coverage for transient request and streaming failures, retry limits, and error handling.

### DIFF
--- a/weed/s3api/filer_multipart.go
+++ b/weed/s3api/filer_multipart.go
@@ -1037,13 +1037,7 @@ func (s3a *S3ApiServer) listMultipartUploads(input *s3.ListMultipartUploadsInput
 		IsTruncated:  aws.Bool(false),
 	}
 
-	// A blip on the way to the filer, either on the ListEntries call or on the
-	// stream receives that follow it, used to fail the whole listing; replay it
-	// a bounded number of times before giving up.
-	uploadsFolder := s3a.genUploadsFolder(*input.Bucket)
-	entries, _, err := listWithRetry(uploadsFolder, func() ([]*filer_pb.Entry, bool, error) {
-		return s3a.list(uploadsFolder, "", *input.UploadIdMarker, false, math.MaxInt32)
-	})
+	entries, _, err := s3a.list(s3a.genUploadsFolder(*input.Bucket), "", *input.UploadIdMarker, false, math.MaxInt32)
 	if err != nil {
 		// A missing .uploads folder normally lists as empty with no error; a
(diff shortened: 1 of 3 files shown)


junkerderprovinz:
Thanks for finishing this properly and for keeping the authorship on it, appreciate that. The repo-wide version wrapping `list()` itself is a cleaner outcome than what I had scoped in #10886.

PR #10894, opened 2026-08-23
debug: DLM suite with the Link change applied (do not merge)

chrislusf:
Experiment arm for the #10885 DLM failures: master + per-test harness logs + a cherry-pick of the Link change. If this goes red while the control (#debug-dlm-logs PR) stays green, the Link change is implicated despite having no visible code path into these tests, and the per-test logs will show the failing cluster. Will close after the comparison.

### DIFF
--- a/test/fuse_dlm/framework_test.go
+++ b/test/fuse_dlm/framework_test.go
@@ -8,6 +8,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"sync"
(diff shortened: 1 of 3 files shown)


chrislusf:
Experiment complete: this arm (master + the Link change) passed the DLM suite, as did the control, so the Link change is exonerated for the #10885 DLM failures — the red runs there are a nondeterministic cluster flake. The per-test log capture lands via #10893 so the next red run finally shows the failing cluster.

Issue #4933, opened 2023-10-23
S3 range off-by-one error


ravriel:
Sponsors SeaweedFS via Patreon https://www.patreon.com/seaweedfs
**Describe the bug**
Requesting a range request starting from the object length returns an incorrect sucess with incorrect Content-Range header.

For example:
Upload a 10 bytes object:
`curl -X PUT http://localhost:8333/app-bucket/0-9.txt -d 0123456789`

Get content from offset 10, which is out of range:
`curl -D - -H 'Range:bytes=10-' http://localhost:8333/app-bucket/0-9.txt`

```
HTTP/1.1 206 Partial Content
Accept-Ranges: bytes
Access-Control-Expose-Headers: Content-Disposition
Content-Disposition: inline; filename="0-9.txt"
Content-Length: 0
Content-Range: bytes 10-9/10
Content-Type: application/x-www-form-urlencoded
Date: Mon, 23 Oct 2023 08:39:45 GMT
Etag: "781e5e245d69b566979b86e28d23f2c7"
Last-Modified: Mon, 23 Oct 2023 08:37:32 GMT
Server: SeaweedFS Filer 30GB 3.57
```

Note the wrong success response `206 Partial Content` and the wrong `Content-Range: bytes 10-9/10`

**System Setup**
- weed command: `.\weed.exe server "-master.volumeSizeLimitMB=1000" -dir=".\data" -s3 "-s3.allowDeleteBucketNotEmpty=false" "-volume.max=50"`
- OS: Windows
- Version: `version 30GB 3.57 0f8168c0c928bba3d2f48b0680d3bdce9c617559 windows amd64`
- default filer,  no `filer.toml`

**Expected behavior**
Expected error 416 like the error returned from any other invalid range.
For example: `curl -D - -H 'Range:bytes=11-' http://localhost:8333/app-bucket/0-9.txt`

```
HTTP/1.1 416 Requested Range Not Satisfiable
...
```

Thanks for great project!

junkerderprovinz:
Hi @ravriel, thanks for the precise repro.

The S3 gateway no longer proxies GET to the filer. Since #7481 (commit ca84a8a7131e2be81ead697c472bf967548d97ec) it reads from volume servers directly and validates the Range header itself in `parseAndValidateRange` in `weed/s3api/s3api_object_handlers.go`:

```go
if startOffset < 0 || startOffset >= totalSize {
[...]

PR #10898, opened 2026-08-23
server: reject a Range start offset equal to the file size

junkerderprovinz:
Related to #4933 (already closed, see below).

## The bug

`parseRange` in `weed/server/volume_server_handlers_helper.go` accepted a start offset equal to `size`, one past the last valid byte of a `size`-byte file (valid offsets are `0` to `size-1`):

```go
i, err := strconv.ParseInt(start, 10, 64)
if err != nil || i > size || i < 0 {
    return nil, errors.New("invalid range")
}
```

`Range: bytes=<size>-` on a `size`-byte file therefore returned `206 Partial Content` with `Content-Range: bytes <size>-<size-1>/<size>` (an inverted, empty range) instead of `416 Range Not Satisfiable`.

## Why this is separate from #4933

#4933 reported the identical symptom against the S3 API and was fixed in #7481, which validates ranges in the S3 gateway's own `parseAndValidateRange` (`weed/s3api/s3api_object_handlers.go`) with the correct `startOffset >= totalSize` check. That fix doesn't touch this function: `parseRange` here is a separate, older implementation used by the filer's own HTTP read endpoint and the volume server's read handlers (shared via `ProcessRangeRequest` in `weed/server/common.go`), so the same class of bug survived on that path. Found while verifying #4933 was still resolved.

## Fix

`i > size` → `i >= size`, matching the equivalent, already-correct check a few lines below for the end of an explicit range (`if i >= size { i = size - 1 }`), and matching the S3 side's fix.

## Testing

New `TestParseRange` (table-driven, covers the boundary the bug was in plus a passing sibling test that already worked) and `TestParseRangeOnEmptyFile` (a 0-byte file has no valid offsets, not even 0). Reverting the one-line fix makes `TestParseRangeOnEmptyFile` fail, confirming the test actually catches the bug. `go build`, `go vet`, and the package's `TestParseRange*` tests are clean.

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected range request handling at the end of files.
  * Invalid ranges starting at the file size are now rejected instead of accepted.

* **Tests**
  * Added coverage for bounded, clamped, suffix, invalid, and empty-file range requests.

### DIFF
--- a/weed/server/volume_server_handlers_helper.go
+++ b/weed/server/volume_server_handlers_helper.go
@@ -62,7 +62,7 @@ func parseRange(s string, size int64) ([]httpRange, error) {
 			r.length = size - r.start
 		} else {
 			i, err := strconv.ParseInt(start, 10, 64)
-			if err != nil || i > size || i < 0 {
+			if err != nil || i >= size || i < 0 {
 				return nil, errors.New("invalid range")
 			}
 			r.start = i
--- a/weed/server/volume_server_handlers_helper_test.go
+++ b/weed/server/volume_server_handlers_helper_test.go
@@ -1,0 +1,55 @@
+package weed_server
+
+import "testing"
+
+func TestParseRange(t *testing.T) {
+	const size = int64(10)
+
+	tests := []struct {
+		name      string
+		rangeSpec string
+		wantErr   bool
+		wantStart int64
+		wantLen   int64
+	}{
+		{name: "start within bounds", rangeSpec: "bytes=0-", wantStart: 0, wantLen: 10},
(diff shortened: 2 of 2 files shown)


chrislusf:
Same bug as #10889, which was already open with the stdlib-parity version — no harm done, the boundary check you landed is correct for single ranges. Rebased #10889 on top of this to cover what the literal check misses: a mixed request like `bytes=10-,0-1` must stay 206 per RFC 9110 (416 only when every range fails), the 416 should carry `Content-Range: bytes */<size>`, and the Rust volume server has the same stale check. Your tests pass unchanged on top.

PR #10900, opened 2026-08-23
master: don't let a dead KeepConnected handler close its successor's channel

### LINKED ISSUE #10892: Master floods a reconnected KeepConnected client with empty responses at wire speed
Three masters, `weed 4.40`. One master is a small quorum-only node in another DC, and the cross-DC link drops for a couple of minutes almost every night. After such a drop the leader starts streaming empty `KeepConnectedResponse` messages to the reconnected follower at wire speed and never stops.

Numbers from our cluster: 10–25 MB/s per affected client (~700k messages/s), one core pinned on the leader per affected stream, one core pinned on the client decoding it. `ss -tin` on the leader showed 675 GB sent to one follower and 1.69 TB to another over ~19 hours, all on a single gRPC connection to port 19333. The quorum node has 1 vCPU, so it was effectively dead from decoding the flood.

tcpdump on the client shows the stream is just this 14-byte unit repeating endlessly — an HTTP/2 DATA frame on stream 1 carrying a gRPC message of length 0:

```
00 00 05 00 00 00 00 00 01 00 00 00 00 00
```

The client log at the exact minute each flood started:

```
I0822 05:13:46 masterclient.go:289 .master masterClient failed to receive from <leader>:9333:
  rpc error: code = Unavailable desc = error reading from server: connection timed out
I0822 05:13:47 masterclient.go:498 .master masterClient reconnection attempt #16
```

What I believe happens, in `weed/server/master_grpc_server.go`:

The old stream dies from a network timeout and the client reconnects before the old handler has exited. `addClient` registers the new channel under the same `clientName`, overwriting the old map entry (nothing closes it).

The old handler's defer then runs `deleteClient`, which closes whatever channel is currently in the map under that name — i.e. the channel of the new, live stream — and deletes the entry.

The new handler's send loop does `case message := <-messageChan:` without checking `ok`. A receive from a closed channel returns nil immediately and forever, so the loop degenerates into `for { stream.Send(nil) }`. It cannot exit: the client is alive and `Send` succeeds.

To be clear, this is just my guess — based on the packet capture and on the fact that the workaround below cured it twice in a row. I also collect load metrics from these nodes into my monitoring, and the episodes are clearly visible on the graphs as a wave. The pattern caught my eye because Proxmox clusters have a similar issue on a flaky link.

Workaround that needs no restarts: kill the flooded TCP connection on the client side (`ss -K dst <leader> dport = 19333`) — the client reconnects cleanly and the leader's spinning goroutine exits on the failed `Send`.

I’m attaching a screenshot of the traffic coming to the node at the moments when I took the dump.
Image

chrislusf:
Fixes #10892.

The reporter's analysis was exactly right. When a client reconnects KeepConnected before the old handler has exited, `addClient` overwrites the map entry under the same client name. The old handler's deferred `deleteClient` then closes whatever channel the map holds under that name, which is the new, live stream's channel. A receive from a closed channel returns nil immediately and forever, and the send loop did not check `ok`, so it degenerated into `for { stream.Send(nil) }`: empty responses at wire speed, one core pinned on each end, until someone kills the TCP connection.

Reproduced against a live master by opening two KeepConnected streams with the same ClientAddress and canceling the first: the surviving stream received 5.5 million empty messages in two seconds. With this change it receives none and stays connected.

The fix is ownership: `deleteClient` closes the channel its own handler registered, and removes the map entry only if it still points to that channel. Closing under the write lock is safe because broadcasters send while holding the read lock and only to channels found in the map. This also closes the old handler's orphaned channel, whose drain goroutine used to leak, and the send loop now treats a closed channel as an exit rather than a message stream.

The cluster membership side of the same defer was already safe: `GroupMembers` refcounts overlapping registrations.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling when a client reconnects using the same name.
  * Prevented stale connections from disrupting newer connections.
  * Ensured closed connections exit cleanly without sending invalid messages.
  * Preserved message delivery to reconnected clients.

* **Tests**
  * Added coverage for reconnect behavior, channel cleanup, and message broadcasts.

### DIFF
--- a/weed/server/master_grpc_server.go
+++ b/weed/server/master_grpc_server.go
@@ -418,7 +418,7 @@ func (ms *MasterServer) KeepConnected(stream master_pb.Seaweed_KeepConnectedServ
 		if req.ClientType == cluster.FilerType {
 			ms.LockRingManager.RemoveServer(cluster.FilerGroupName(req.FilerGroup), peerAddress)
 		}
-		ms.deleteClient(clientName)
+		ms.deleteClient(clientName, messageChan)
 	}()
 
 	// Send volume locations to the client
@@ -481,7 +481,12 @@ func (ms *MasterServer) KeepConnected(stream master_pb.Seaweed_KeepConnectedServ
 	defer ticker.Stop()
 	for {
 		select {
-		case message := <-messageChan:
+		case message, ok := <-messageChan:
+			if !ok {
+				// a closed channel receives nil forever; without this check the
+				// loop would flood the client with empty messages at wire speed
+				return nil
+			}
 			if err := stream.Send(message); err != nil {
 				// The error, not the message: it carries every volume id on a
 				// newly connected node, and formatting a proto that size to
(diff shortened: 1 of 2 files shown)


chrome112:
I'm glad I could help :)

PR #10901, opened 2026-08-23
s3: confine a Lance catalog table location to the caller's own bucket

chrislusf:
The Lance namespace gateway (`handleDeclareTable` / `handleRegisterTable`) takes
the `location` field from the request body, trims a trailing slash, and passes it
into the marker path sink. That location flows into
`TableDataDirFromMetadataLocation`, which joins it under `/buckets` and collapses
any `../` segments, and `writeMarker` then issues a filer `CreateEntry` that
auto-creates every missing parent directory along the way.

Because the field is never confined to the caller's bucket, a location such as
`s3://analytics/../victim/secret` (or one escaping `/buckets` entirely) resolves
outside the caller's namespace, so the fixed-name marker (`.lance-reserved` /
`.lance-deregistered`) — and its parent directories — land in another tenant's
tree. Planting `.lance-deregistered` in a victim's dataset directory even hides
their live table from the catalog.

The Iceberg gateway already confines the identical `location` field
(`parseS3Location` + bucket match + `isValidTablePath`); the Lance gateway did
not. This adds the same confinement: the declared location must be an `s3://`
URI whose bucket is the caller's own and whose path holds no traversal segment,
on both the `declare` and `register` handlers.

Test: `TestLocationOutsideBucketIsRejected` covers cross-bucket, `../`-into-another-bucket,
escape-out-of-`/buckets`, and non-`s3://` locations, asserting each is rejected
and no marker is written.

## Summary by CodeRabbit

* **Bug Fixes**
  * Table declarations and registrations now reject invalid or unsafe storage locations.
  * Locations outside the caller’s bucket, path traversal attempts, non-S3 paths, backslashes, and null characters return a clear 400 invalid-input response.
  * Invalid requests no longer create reserved marker files.

### DIFF
--- a/weed/s3api/lance/handlers_table.go
+++ b/weed/s3api/lance/handlers_table.go
@@ -130,6 +130,9 @@ func (s *Server) handleDeclareTable(w http.ResponseWriter, r *http.Request) {
 	location := strings.TrimSuffix(req.Location, "/")
 	if location == "" {
 		location = tableLocation(bucket, ns, name)
+	} else if err := confineLocation(bucket, location); err != nil {
+		writeError(w, r, http.StatusBadRequest, codeInvalidInput, err.Error())
+		return
 	}
 
 	if err := s.createTable(r, bucket, ns, name, location); err != nil {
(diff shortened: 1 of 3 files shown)


chrislusf:
Addressed the Greptile finding: the register branch of `TestLocationOutsideBucketIsRejected` was asserting on `reservedMarker`, which register never writes. Split the test so each handler checks its real side effect — declare must not write `.lance-reserved` at the escaped path, and register must not remove a pre-planted `.lance-deregistered` from a victim's directory. Verified both branches fail without the confinement and pass with it. CodeRabbit reported no actionable comments.

PR #10946, opened 2026-08-25
filter volume by local or remote storage name

27149chen:
We have over 1 million volumes, most of which are remote volumes. Sometimes we only need to list local volumes, and sometimes we need to list volumes with a specific storage name.

Added two new params in volume list request:
1. remote_storage_name: empty for all (local+remote), `*` for all remote volumes, `aws.*` for storage name starts with "aws.", `aws.gold` for storage name exactly equals to "aws.gold"
2. local_volume_only: if true, return local volumes only

# How is the PR tested?
Tested by ut

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added streaming volume listings with filters for collections, volume IDs, remote storage, and local-only volumes.
  * Added wildcard support for remote-storage filtering.
  * Added configurable caching for remote directory listings.
  * Added fsync decisions and metadata flush watermark reporting.
  * Added controls for EC shard rebuild safety and copy bandwidth limits.

* **Bug Fixes**
  * Improved filtering accuracy for volumes and erasure-coded shards.
  * Disk status responses now report encountered errors directly.


chrislusf:
The filter itself looks right, and the way `local_volume_only` mirrors `default_collection_only` is the correct shape: the empty string is a real remote storage name (the local volumes), so a bool is the only way to ask for it.

I pushed one commit to the branch: the proto files had been copied over wholesale from `weed/pb/` and every `.pb.go` regenerated with a different protoc (v7.35.0 vs v7.35.1) and protoc-gen-go-grpc (v1.6.2 vs v1.5.1). `DiskStatus.error` arriving in `seaweed-volume/proto/volume_server.proto` that way is what broke the Rust build:

```
error[E0063]: missing field `error` in initializer of `DiskStatus`
    --> src/server/grpc_server.rs:3844:32
```

The two CodeRabbit comments were both against those copied-in files, so they go away with it. The change is now 9 files. Please regenerate with the repo's toolchain (`--go_out` only, no `--go-grpc_out`) so the grpc bindings stay put.

A few things left:

**`volumeLike` costs more than it buys.** Six new exported accessors across two storage types, and `EcVolumeInfo.GetRemoteStorageName()` returning a hardcoded `""` is a method on a public type that only makes sense to this one caller. A third scalar argument does the same job with none of that:

```go
func (f VolumeFilter) matches(collection string, id needle.VolumeId, remoteStorageName string) bool
```

with `filter.matches(v.Collection, v.Id, v.RemoteStorageName)` and `filter.matches(ecv.Collection, ecv.VolumeId, "")`. The `""` at the EC call site says out loud what the stub method hides: an EC volume is never tiered.

**`remoteStorageName` is the only unexported field** next to `Collection` and `VolumeId`. Pick one.

**The proto doesn't document the surprise.** "Wildcards are supported" doesn't prepare anyone for `*` excluding the local volumes, which is the whole reason the field is useful, nor for `remote_storage_name` winning when `local_volume_only` is also set. `default_collection_only` carries that note already ("A named collection wins"); this pair deserves the same.

**`NewVolumeFilter` isn't tested for the new fields.** `TestVolumeFilterRemoteStorageNameWildcards` builds `VolumeFilter{remoteStorageName: ...}` by hand, so the request-to-filter mapping is uncovered and `LocalVolumeOnly` is never exercised at all. The table in `TestVolumeFilterSelects` is where it belongs.

PR #10948, opened 2026-08-25
s3: stop one abandoned request from cancelling every concurrent upload

### LINKED ISSUE #10947: s3: a Canceled RPC on a non-cancellable context tears down the shared filer connection, failing every concurrent multipart part
### Summary

A single `Canceled` RPC issued on a **non-cancellable context** tears down the *shared*
cached filer `ClientConn`, which cancels every other in-flight RPC on that connection.
On the S3 gateway this shows up as bursts of failed multipart parts: the S3 client sees
`400 InvalidRequest` for `UploadPart` / `CopyObject`, and a large multipart upload dies
because one unrelated request was cancelled.

The guard that is supposed to prevent exactly this cascade is bypassed on the very path
that needs it.

### Version

- `4.40` (`weed version … 4.40 875cd1f67 linux amd64`)
- The relevant code is **byte-identical in `4.44`** — I diffed
  `weed/pb/grpc_client_server.go` between the two tags; the only change in that file is
  the local-socket buffer work from #10824.

### Setup

- Single node, `weed filer -s3` (S3 gateway in the filer process), master + 2 volume servers
- Filer store: MySQL
- Regular workload: S3 multipart uploads (5 MiB parts) and server-side copies

### Symptom

Filer log, one burst (timestamps trimmed):

```
E … s3api_object_handlers_put.go:832 putToFiler: CreateEntry returned error:
      CreateEntry: rpc error: code = Canceled desc = grpc: the client connection is closing
E … s3api_object_handlers_put.go:891 putToFiler: failed to create entry for
      /buckets/<bucket>/.uploads/<upload-id>/0065_<part>
W … s3api_object_handlers_put.go:896 putToFiler: finalization failed, attempting to cleanup 1 orphaned chunks
E … upload_chunked.go:276 chunked upload failed: assign volume: all filers failed,
      last error: assign volume: rpc error: code = Canceled desc = grpc: the client connection is closing
```

The S3 client receives:

```
operation error S3: UploadPart, https response error StatusCode: 400,
api error InvalidRequest: Invalid Request
```

Frequency over a 72 h window on a moderately loaded node: **139 occurrences of
`the client connection is closing`, spread over 55 distinct minutes**, arriving in
bursts (18 and 28 within a single minute). Every part that happened to be in flight on
that connection fails together.

Ruled out by inspection over the same window:

- filer store (MySQL): **0** errors logged
- master and both volume servers: **0** errors logged
- no container/process restarts (9 days uptime)
- resources idle (filer ~2 GiB of 31 GiB, CPU a few percent)
- not correlated with clients aborting downloads (2 of 17 minutes overlap)

### Analysis

`WithGrpcClient` closes the shared cached connection when
`shouldInvalidateConnection` says so:

```go
executionErr := fn(vgc.ClientConn)
if executionErr != nil && shouldInvalidateConnection(ctx, executionErr) {
    delete(grpcClients, address)
    vgc.Close()      // shared ClientConn
}
```

and for `Canceled` the decision is:

```go
case codes.Canceled, codes.DeadlineExceeded:
    // … Only invalidate while the caller's context is still live,
    // isolating the genuine stale-channel signal.
    return ctx == nil || ctx.Err() == nil
```

The comment states the intent precisely — do not tear the channel down when the
*caller's own* context expired, because that "cancels every other in-flight RPC on it
with 'the client connection is closing', a cascade that turns one slow request into a
flood of failures across all concurrent callers".

The guard tests `ctx.Err() == nil`. For a context that can never be cancelled —
`context.Background()` / `context.TODO()` — `Err()` is **nil forever**, so the guard
always answers "invalidate".

And that is exactly what the S3 put path passes, deliberately:

```go
// weed/s3api/s3api_object_handlers_put.go  (4.44: line 878)
// Use context.Background() to ensure metadata save completes even if HTTP request is cancelled
if err := filer_pb.CreateEntry(context.Background(), client, req); err != nil {
```

There are 5 `context.Background()` call sites in that file alone.

So on the S3 gateway the anti-cascade guard is inverted: any `Canceled` — including the
`Canceled` that a *victim* of a concurrent `Close()` receives — is treated as a genuine
stale-channel signal, and the shared connection is dropped again.

### Suggested fix

A non-cancellable context carries no information about channel health, so it should not
be read as "the channel is stale":

```go
case codes.Canceled, codes.DeadlineExceeded:
    // A context that can never be cancelled (Background/TODO) says nothing about the
    // channel: Err() stays nil forever, so the check below would always answer
    // "invalidate". Callers that deliberately detach from the request (putToFiler)
    // are exactly in this position.
    if ctx == nil || ctx.Done() == nil {
        return false
    }
    return ctx.Err() == nil
```

`Done() == nil` is true precisely for `Background()`/`TODO()` and false for every
cancellable context, so the intended behaviour for real request contexts is unchanged.
The genuine stale-channel case the comment refers to (a peer restart behind a stable L4
endpoint) surfaces as `Unavailable`, which is handled by its own branch and still
invalidates.

An alternative, or an addition: never invalidate when the error *is* the cascade marker
(`grpc: the client connection is closing`) — a caller that was cancelled because someone
else closed the connection is a victim, not a witness of a stale channel.

I am happy to open a PR with the one-line change plus a regression test if the approach
looks right.

chrislusf:
Fixes #10947.

A single abandoned request could close the shared cached filer `ClientConn` and cancel every other RPC riding it, so a large multipart upload died with `rpc error: code = Canceled desc = grpc: the client connection is closing` on unrelated parts, surfacing to the S3 client as `400 InvalidRequest`.

Reproduced in `weed/pb` against a real gRPC server: one caller holds a slow filer RPC, a second unrelated caller fails, and the first dies with the exact error from the issue. Both regression tests fail without their commit.

### A non-cancellable context is no evidence of a stale channel

`shouldInvalidateConnection` only tore the channel down on `Canceled`/`DeadlineExceeded` while the context handed to `WithGrpcClient` was still live. `context.Background()`/`TODO` never expire, so `Err()` stays nil forever and the guard always answered "invalidate" — and `Background` is what almost every caller passes, the S3 gateway included. One request whose RPC rode an abandoned HTTP request context therefore closed the connection for everyone.

Only a cancellable context bounds an RPC attempt, so require one before reading it. This is the fix @arozanov proposed in the issue. The two callers that thread a real per-attempt context (master assign and volume lookup, from #9775) keep the behaviour that guard was written for; a genuinely stale channel (a peer restart behind a stable L4 endpoint) surfaces as `Unavailable` and invalidates on its own branch.

### A bystander of a teardown is not a stale-channel witness

gRPC raises `ErrClientConnClosing` locally, before an RPC reaches the wire, when this process has already closed the `ClientConn`. Every caller that touches a channel during another goroutine's teardown gets it, so reading it as a stale-channel signal lets one teardown re-arm itself across the herd it just cancelled. The cached-connection version check keeps those callers from closing a replacement channel, but the streaming path invalidates by address alone and has no such guard.

### Ending a stream should not drop the peer connection under it

A streaming caller gets its own `ClientConn`, but on **any** error it also dropped the cached non-streaming `ClientConn` every request handler shares with that peer — the recovery path added in #9107 for a peer restart hidden behind a stable L4 endpoint. Any error includes the ordinary ones: a metadata subscription that reached its stop point, a follow callback that refused an event, a caller that gave up.

The S3 gateway follows filer metadata on exactly such a stream and reconnects forever (`followIamChanges`), so each ordinary end of it cancelled every S3 request in flight against the filer. In the reproduction a subscription that ends with a plain `io.EOF` — on a perfectly healthy connection — kills concurrent uploads, which fits the reported bursts better than client aborts do. Invalidation now runs through the same predicate, so a peer that went away (`Unavailable`, `Internal`, transport-level errors) still drops the shared channel immediately.

### Notes

- No Rust counterpart: `seaweed-volume` dials a fresh tonic `Channel` per call and has no shared connection cache to invalidate.
- Verified against `weed/pb/...`, `weed/wdclient`, `weed/operation/...`, `weed/s3api/...` and `weed/filer/...`. No cluster run — this machine is out of disk.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling to avoid unnecessary invalidation after locally closed connections.
  * Prevented unrelated requests and metadata streams from being disrupted when another operation is canceled, times out, or ends.
  * Improved reliability for concurrent requests sharing the same connection.

* **Tests**
  * Added end-to-end coverage for concurrent requests, abandoned operations, and ended metadata streams.
  * Added validation for cancellable and non-cancellable contexts and locally generated connection-closing errors.

### DIFF
--- a/weed/pb/grpc_client_cascade_test.go
+++ b/weed/pb/grpc_client_cascade_test.go
@@ -98,3 +98,46 @@ func TestWithGrpcClient_AbandonedRequestKeepsSharedConnection(t *testing.T) {
 		t.Fatalf("concurrent caller must survive an unrelated abandoned request: %v", concurrentErr)
 	}
 }
+
+func (s *cascadeFilerServer) SubscribeMetadata(req *filer_pb.SubscribeMetadataRequest, stream filer_pb.SeaweedFiler_SubscribeMetadataServer) error {
+	return nil
+}
+
+// TestWithGrpcClient_EndedStreamKeepsSharedConnection covers the other half of
+// seaweedfs#10947: the S3 gateway follows filer metadata on a long-lived
+// SubscribeMetadata stream and reconnects forever. Every ordinary end of that
+// stream used to drop the shared filer ClientConn the request handlers use,
+// firing the same burst of "the client connection is closing" failures.
+func TestWithGrpcClient_EndedStreamKeepsSharedConnection(t *testing.T) {
+	fake, address, dialOption := startCascadeFiler(t)
+
+	var concurrentErr error
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		concurrentErr = WithGrpcClient(context.Background(), false, 0, func(connection *grpc.ClientConn) error {
+			_, err := filer_pb.NewSeaweedFilerClient(connection).LookupDirectoryEntry(context.Background(),
+				&filer_pb.LookupDirectoryEntryRequest{Directory: "/buckets/b", Name: "slow"})
+			return err
+		}, address, false, dialOption)
+	}()
+	<-fake.inFlight
+
+	if err := WithGrpcClient(context.Background(), true, 0, func(connection *grpc.ClientConn) error {
+		stream, err := filer_pb.NewSeaweedFilerClient(connection).SubscribeMetadata(context.Background(), &filer_pb.SubscribeMetadataRequest{})
+		if err != nil {
+			return err
+		}
+		_, err = stream.Recv()
+		return err
+	}, address, false, dialOption); err == nil {
+		t.Fatal("the ended stream should have reported io.EOF")
+	}
+
+	close(fake.release)
+	wg.Wait()
+	if concurrentErr != nil {
+		t.Fatalf("concurrent caller must survive an unrelated stream ending: %v", concurrentErr)
+	}
+}
--- a/weed/pb/grpc_client_cascade_test.go
+++ b/weed/pb/grpc_client_cascade_test.go
@@ -52,10 +52,14 @@ func startCascadeFiler(t *testing.T) (*cascadeFilerServer, string, grpc.DialOpti
 	grpcClients = make(map[string]*versionedGrpcClient)
 	grpcClientsLock.Unlock()
 	t.Cleanup(func() {
-		server.Stop()
 		grpcClientsLock.Lock()
+		cached := grpcClients
 		grpcClients = previous
 		grpcClientsLock.Unlock()
+		for _, connection := range cached {
+			connection.Close()
+		}
+		server.Stop()
 	})
 
 	return fake, listener.Addr().String(), grpc.WithTransportCredentials(insecure.NewCredentials())
--- a/weed/pb/grpc_client_cascade_test.go
+++ b/weed/pb/grpc_client_cascade_test.go
@@ -1,0 +1,100 @@
+package pb
+
+import (
+	"context"
+	"net"
+	"sync"
+	"testing"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+)
+
+// cascadeFilerServer answers a "slow" lookup only once its caller is known to be
+// in flight, so a second caller can fail while the first still holds the shared
+// cached ClientConn.
+type cascadeFilerServer struct {
+	filer_pb.UnimplementedSeaweedFilerServer
+	inFlight  chan struct{}
+	release   chan struct{}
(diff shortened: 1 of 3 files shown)


chrislusf:
Both addressed in c27dee2 and c496e4f.

The cleanup leak is real - the tests restored the previous cache and dropped their own entries without closing them, leaving each `ClientConn`'s transport and reconnect goroutines running past the fake filer. Now detached under the lock, closed after releasing it, then the server stops.

On `ErrClientConnClosing`: agreed that the code-only replacement cannot stand in, since `codes.Canceled` is precisely what this function exists to disambiguate. Keeping the sentinel, with a comment recording why the deprecation notice does not apply. No suppression directive though - the repo has none anywhere, and CI lints with `go vet` only, so adding the first `//nolint` for this would introduce a convention rather than follow one.

PR #10951, opened 2026-08-25
rust volume: search .sdx for read-only volumes instead of holding the index

chrislusf:
Mirrors #10950 into the Rust volume server, which had no counterpart to Go's `SortedFileNeedleMap`.

## The gap

Go loads every read-only volume — `noWriteOrDelete || noWriteCanDelete`, which is every cloud-tiered one — through `SortedFileNeedleMap`: the index stays on disk as a sorted `.sdx`, a lookup is a binary search, and since #10950 no descriptor is held between lookups.

Rust had neither half. `load_index_inmemory` built a full `CompactNeedleMap` in RAM, and a tiered volume (`no_write_can_delete`, so *not* the read-only branch) went down the writable path and pinned an `.idx` append handle on top of that.

Measured on a reloaded tiered volume, per volume:

| | before | after |
|---|---|---|
| `.idx` / `.sdx` descriptors | 1 | 0 |
| index in RAM | whole `.idx` | none |

## Change

- `storage/needle_map/sorted_file.rs` — `SortedFileNeedleMap`: generates `.sdx` when it is older than `.idx`, binary-searches it for `get`, and on `delete` appends a tombstone to the `.idx` tail then marks the `.sdx` record in place. `put` is rejected, as in Go.
- `storage/needle_map/file_pool.rs` — the bounded pool from `needle_map_file_pool.go`, 1024 handles, LRU. Handles are `Arc<File>`, so an eviction cannot close one a borrower still holds; the file closes when the last `Arc` drops.
- `load_index` picks it for both read-only modes, and `set_writable` swaps in a writable map first — Go's `reopenIdxForWrite`, without which the flag flip would leave every write rejected by the sorted map's `put`.

`index_file_size()` answers from the tracked append offset and `sync()` only fsyncs when a tombstone was actually appended, both as in Go.

## Cross-implementation compatibility

The `.sdx` this writes is byte-identical to Go's for the same `.idx` — same sort order, last-write-wins, tombstoned keys dropped entirely — so a volume moved between a Go and a Rust server reads whichever copy is already on disk. `sdx_matches_the_bytes_go_writes` pins the bytes against a fixture generated by `erasure_coding.WriteSortedFileFromIdx` built with `-tags 5BytesOffset`.

## Metrics

Go derives the read-only volume's counters with a reverse walk plus a bloom filter. This computes the same numbers exactly, in one sequential pass and no per-volume allocation: the rows that are the last still-valid row for their key are exactly the entries `.sdx` holds, so `DeletionCounter = rows - live` and `DeletionByteCounter = valid .idx bytes - live .sdx bytes`. `metrics_match_the_go_counters` covers rewrites and tombstones.

## Tests

437 unit tests pass, as do the Go-driven integration tests (`test/volume_server/rust`, run against the Go `weed` binary). New coverage: sorted-map lookup/delete/regeneration, the Go byte-parity fixture, pool eviction-while-borrowed and discard-on-unmount, and `test_remote_volume_index_is_on_disk_and_holds_no_descriptors`, which asserts a writable volume *does* pin its `.idx` before asserting a tiered one holds nothing — so the descriptor probe cannot pass vacuously. `cargo clippy` is clean on the new files.

## Noted, not changed

- `SortedFileNeedleMap::delete` does not move the counters, mirroring Go, where `Delete` skips `logDelete`; the numbers refresh on reload. Worth a Go-side follow-up rather than a divergence here.
- `write_sorted_ecx_from_idx` in `ec_encoder.rs` keeps tombstoned rows where Go's `.ecx` drops them. Pre-existing and out of scope, but the same byte-parity argument applies to `.ecx`.
- EC volumes still pin `.ecx` + `.ecj` + one descriptor per mounted shard on both sides.

## Summary by CodeRabbit

* **New Features**
  * Added disk-backed sorted index support for read-only and cloud-tiered volumes.
  * Added bounded reuse of open index files to reduce resource usage.
  * Added fallback to in-memory indexes when sorted indexes cannot be created.
  * Added support for switching eligible read-only volumes back to writable mode.

* **Bug Fixes**
  * Index reading, rebuilding, and compaction errors are now reported.
  * Improved recovery from incomplete indexes.
  * Improved cleanup and rollback when volume mode changes fail.
  * Corrected sorted-index deletion metrics and handling.

### DIFF
--- a/seaweed-volume/src/server/grpc_server.rs
+++ b/seaweed-volume/src/server/grpc_server.rs
@@ -3575,7 +3575,12 @@ impl VolumeServer for VolumeGrpcService {
                             modified_time: dat_modified_secs,
                             extension: ".dat".to_string(),
                         });
-                        vol.refresh_remote_write_mode();
+                        vol.refresh_remote_write_mode().map_err(|e| {
+                            Status::internal(format!(
+                                "volume {} failed to refresh write mode: {}",
+                                vid, e
+                            ))
+                        })?;
 
                         if let Err(e) = vol.save_volume_info() {
                             return Err(Status::internal(format!(
@@ -3764,7 +3769,15 @@ impl VolumeServer for VolumeGrpcService {
                     if !vol.volume_info.files.is_empty() {
                         vol.volume_info.files.remove(0);
                     }
-                    vol.refresh_remote_write_mode();
+                    // Swaps the read-only sorted map out before the volume is
+                    // published as writable; without it the first write would
+                    // append to the local .dat and then fail to index.
+                    vol.refresh_remote_write_mode().map_err(|e| {
+                        Status::internal(format!(
+                            "volume {} failed to refresh write mode: {}",
+                            vid, e
+                        ))
+                    })?;
 
                     if let Err(e) = vol.save_volume_info() {
                         return Err(Status::internal(format!(
--- a/seaweed-volume/src/server/heartbeat.rs
+++ b/seaweed-volume/src/server/heartbeat.rs
@@ -1578,7 +1578,7 @@ mod tests {
             let (_, volume) = store.find_volume_mut(VolumeId(17)).unwrap();
             volume.set_read_only().unwrap();
             volume.volume_info.files.push(Default::default());
-            volume.refresh_remote_write_mode();
+            volume.refresh_remote_write_mode().unwrap();
         }
 
         let heartbeat = build_heartbeat(&test_config(), &mut store);
@@ -1944,7 +1944,7 @@ mod tests {
             key: "volumes/71.dat".to_string(),
             ..Default::default()
         });
-        volume.refresh_remote_write_mode();
+        volume.refresh_remote_write_mode().unwrap();
 
         let heartbeat = build_heartbeat(&test_config(), &mut store);
 
--- a/seaweed-volume/src/storage/needle_map.rs
+++ b/seaweed-volume/src/storage/needle_map.rs
@@ -942,33 +942,31 @@ impl RedbNeedleMap {
     }
 
     /// Collect all entries as a Vec for iteration (used by volume.rs iter patterns).
-    pub fn collect_entries(&self) -> Vec<(NeedleId, NeedleValue)> {
+    pub fn collect_entries(&self) -> io::Result<Vec<(NeedleId, NeedleValue)>> {
         let mut result = Vec::new();
-        let txn: redb::ReadTransaction = match self.db.begin_read() {
-            Ok(t) => t,
-            Err(_) => return result,
-        };
-        let table = match txn.open_table(NEEDLE_TABLE) {
-            Ok(t) => t,
-            Err(_) => return result,
-        };
-        let iter = match table.iter() {
-            Ok(i) => i,
-            Err(_) => return result,
-        };
+        let txn: redb::ReadTransaction = self
+            .db
+            .begin_read()
+            .map_err(|e| io::Error::other(format!("redb begin_read: {e}")))?;
+        let table = txn
+            .open_table(NEEDLE_TABLE)
+            .map_err(|e| io::Error::other(format!("redb open_table: {e}")))?;
+        let iter = table
+            .iter()
+            .map_err(|e| io::Error::other(format!("redb iter: {e}")))?;
         for entry in iter {
-            if let Ok((key_guard, val_guard)) = entry {
-                let key_u64: u64 = key_guard.value();
-                let bytes: &[u8] = val_guard.value();
-                if bytes.len() == PACKED_NEEDLE_VALUE_SIZE {
-                    let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
-                    arr.copy_from_slice(bytes);
-                    let nv = unpack_needle_value(&arr);
-                    result.push((NeedleId(key_u64), nv));
-                }
+            let (key_guard, val_guard) =
+                entry.map_err(|e| io::Error::other(format!("redb entry: {e}")))?;
+            let key_u64: u64 = key_guard.value();
+            let bytes: &[u8] = val_guard.value();
+            if bytes.len() == PACKED_NEEDLE_VALUE_SIZE {
+                let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
+                arr.copy_from_slice(bytes);
+                let nv = unpack_needle_value(&arr);
+                result.push((NeedleId(key_u64), nv));
             }
         }
-        result
+        Ok(result)
     }
 }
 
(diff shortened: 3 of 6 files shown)


chrislusf:
Both P1s were real. Fixed in d3e0dd563.

**Sorted-index read failures during compaction.** `iter_entries` swallowed the error from `visit_live_entries` and returned however many entries it got. It now returns `io::Result` and compaction propagates it. I confirmed the failure mode before fixing it: with the old code, truncating a `.sdx` under a loaded volume let `compact_by_index` run to completion and commit a short index. `redb`'s `collect_entries` dropped `begin_read` / `open_table` / `iter` / per-entry errors the same way and feeds the same call site, so it became fallible too.

**Tier-down.** Confirmed, and worth noting it is a Rust-only reachable state: Go's tier-down leaves `noWriteCanDelete` set, so the volume stays read-only until a reload or an explicit `volume.mark -writable` — which already goes through `reopenIdxForWrite`. Rust's `refresh_remote_write_mode` clears the flag and publishes a writable volume, which my change turned into a writable volume holding a map that rejects every put.

Rather than patching the one call site, I extracted `reopen_idx_for_write` (the direct mirror of Go's `reopenIdxForWrite`, already needed by `set_writable`) and folded it into `refresh_remote_write_mode`, so the map can never disagree with the mode that function just published. It is a no-op in the other direction — tier-up keeps its in-memory map until reload, as Go does. If the rebuild fails the volume is pinned read-only rather than accepting writes it cannot index. `refresh_remote_write_mode` is now fallible; the four non-test call sites propagate.

**Tests.** `iter_entries_reports_a_truncated_sdx`, `test_compaction_aborts_on_unreadable_sorted_index` (asserts compaction returns `UnexpectedEof` instead of committing), and `test_tier_down_swaps_in_a_writable_needle_map` (runs the tail of the tier-down handler, then writes, reloads, and reads the needle back — so an unindexed append fails the test). I checked each fails with its fix reverted. 440 unit tests plus the Go-driven integration suite pass; clippy clean on the new files.

chrislusf:
Addressed in 3e113dc5f. Three taken, one declined.

**Read-only volume on a read-only mount (coderabbit, major) — valid, fixed.** Reproduced it: with the index directory at `0o555`, `Volume::new` returned `Permission denied` where before the change it mounted and served reads. Two writes were in the way — `File::create` for a missing `.idx`, and `.sdx` generation. The `.idx` is now created only where deletes are allowed, which is also what Go does (`O_RDONLY` on the strictly read-only path), and a sorted map that cannot be built falls back to the in-memory one with a warning. A directory nobody can write now costs memory instead of availability. `test_read_only_volume_mounts_when_the_index_dir_is_not_writable` covers it, and fails with the fallback removed. It skips under uid 0, which ignores the directory mode.

**Batch the `.sdx` scan (coderabbit, nitpick) — valid, fixed.** 1024 entries per read, the batch `idx::walk_index_file` already uses. Kept positional reads rather than switching to a `Read + Seek` cursor: the handle comes from the pool and can be shared with another borrower, so a file position would race between them. The buffer is sized to the entry count, so a small volume does not allocate a full batch. `scan_spans_read_batches` runs 2055 entries through it to cover two full batches and a short tail.

**Gate `GO_SDX_HEX` (coderabbit, minor) — valid, fixed.** `cargo test --no-default-features` did warn on both the constant and the module's `use super::*`, so the whole `go_parity_tests` module is now gated rather than just the test inside it. Both feature configurations build and test warning-free.

**Delete leaves stale lookup state (greptile, P1) — declining, with reasoning.** This is Go's shape line for line: `.idx` first, then the in-place `.sdx` mark, error returned either way. The durable record is right, and neither failure shape leaves a corrupt index:

- nothing written — `.sdx` keeps its old mtime while the `.idx` we just appended to is newer, so the freshness check rebuilds it on the next load;
- partially written — `TOMBSTONE_FILE_SIZE` is `0xffffffff` big-endian, so any prefix of it leaves the sign bit set and the entry already reads back as deleted.

What is left is exactly what you describe: after a write that lands zero bytes, the needle stays readable in this process until the volume reloads. Closing that needs an in-memory mask consulted on every lookup, which is state Go does not have — the divergence is worse than the window, and the fix belongs in Go first so both sides keep matching. I have documented the two failure shapes at the call site since the reasoning is not obvious from the code, and added it to the Go follow-up list alongside `Delete` skipping `logDelete`.

442 unit tests (441 without `5bytes`) and the Go-driven integration suite pass; clippy clean on the new files.

chrislusf:
Both valid, fixed in 32c86de16.

**Writable-flag rollback.** Correct, and the read-only-mount fallback I added last round is what made it reachable — that path installs an in-memory map with no writer, so `attach_idx_writer_if_missing` is the step that fails. `save_vif` had the same hole. All three steps now sit behind one rollback point (`publish_writable`), rather than one of them being guarded and two not.

I kept the flags-first order rather than attaching first: `reopen_idx_for_write` and `load_index` both branch on those flags, so the rebuild has to see the new values. The rollback is what makes that safe.

`test_set_writable_rolls_back_when_the_idx_writer_cannot_attach` simulates both halves of a read-only mount — a `0o555` directory so `.sdx` generation fails and the index falls back to memory, and a `0o444` `.idx` so no writer can attach — then asserts the volume comes back read-only. Worth noting the directory mode alone is not enough: appending to an existing file needs permission on the file, not the directory, so a test that only chmods the directory passes without the fix.

**Torn `.sdx`.** Confirmed, including the mtime detail that makes it bite: truncation bumps the mtime, so the freshness check trusts the shorter file and the entry count floors the last needle away. A torn file is now treated like a stale one and rebuilt from `.idx`, with a hard error if it is still not a whole number of entries afterwards. `torn_sdx_is_regenerated` asserts the truncated file really does look *fresher* than the `.idx` before opening, so it cannot pass for the wrong reason.

Two notes. Go writes `.sdx` in place with `O_TRUNC|O_CREATE|O_WRONLY` where this writes a temporary and renames, so a Go server crashing mid-generation is the realistic way to produce one of these — worth carrying the check back to Go, since it floors the entry count the same way.

Chasing it also turned up an adjacent one you did not mention: `index_file_size` seeded the append offset from the raw `.idx` length, so a torn tail there would have put every subsequent tombstone five bytes off. Appends now start at the last whole entry, and the next tombstone overwrites the torn bytes. Go errors out of `newNeedleMapMetricFromIndexFile` on a non-multiple `.idx` instead, so it never reaches this; flooring keeps the volume mountable and repairs the tail on the first delete. `a_torn_idx_tail_does_not_misalign_appends` covers it.

I checked each new test fails with its own fix reverted. 445 unit tests (444 without `5bytes`) and the Go-driven integration suite pass; clippy clean on the new files.

chrislusf:
All three valid, fixed in 2d272e382.

**Torn `.idx` on the writable path (P1).** Right — the aligned offset I added last round only covers the sorted map's own positional writes. Every writable path appends at EOF, and `set_idx_file`'s recorded size is only used for reporting, so the raw length was never the thing doing damage; `O_APPEND` was. A partial row now goes before any writable writer attaches, in `load_index_inmemory`, `load_index_redb` and `attach_idx_writer_if_missing`. The row is unrecoverable either way — every loader already skips it — so removing it is what keeps the rest of the file parseable. Go refuses to load a volume whose `.idx` is not a whole number of rows; trimming keeps it mountable with everything before the tear intact.

`test_writable_load_trims_a_torn_idx_tail` appends five junk bytes, reloads, writes a second needle, reloads again and reads both back. Without the trim the second needle is lost.

**Delete-capable volumes on an unwritable index dir (P2).** Correct, and it is the case the fallback was added for. The fallback now goes one step further: if the in-memory loader's read-write branch fails too, the volume drops to `no_write_or_delete` and loads read-only. Refusing deletes outright is the honest outcome, since without a writer no tombstone could be recorded anyway, and a remount on a writable directory restores them. `test_delete_capable_volume_falls_all_the_way_back_to_read_only` covers it with a `read_only_can_delete` volume, the same shape every tiered one has.

**Sorted-map delete metrics (P2).** Taking it. I declined the same divergence earlier on the grounds of mirroring Go, but two things you raise change the balance: the counters are what the heartbeat reports, so a tiered volume advertised itself garbage-free until restart, and Rust's other two needle maps do update on delete — the sorted one was the odd one out within Rust, not just against Go.

Worth being precise about the arithmetic, since `+2` on the deletion count looks like a typo. The load-time metric derives Go's counters as `deletions = rows - live`. A delete adds one `.idx` row and removes one live `.sdx` entry, so a reload of the same state reports one more row and two more deletions. The runtime update matches that exactly, and `delete_moves_the_counters_where_a_reload_would` asserts the two agree by deleting, reloading, and comparing — otherwise the numbers would jump at every restart. Deleting an already-deleted needle stays a no-op.

Go skips this update too, which is now the second thing on the Go follow-up list alongside its own missing `logDelete` in `SortedFileNeedleMap.Delete` — same defect, and I would rather fix it there than leave the two drifting. Happy to send that PR.

Each new test fails with its own fix reverted. 448 unit tests (447 without `5bytes`) and the Go-driven integration suite pass; clippy clean on the new files (the `volume.rs` warnings near the edits are pre-existing on master).

PR #10960, opened 2026-08-26
master: cap the reported capacity at what the disks hold

### LINKED ISSUE #10959: weed mount on Windows via WinFsp reports a fixed 1.9 PB volume capacity that does not reflect actual usage
When mounting a SeaweedFS Filer on Windows using WinFsp, the mounted volume always reports a total capacity of approximately 1.9 PB (2,097,152 TB) in Windows Explorer and dir commands. This value does not change as files are written to or deleted from the mount point.

This causes two problems:

The reported capacity is not meaningful for users who want to know how much space is actually available in the cluster.

The used/free space does not update based on actual storage consumption, making it impossible to monitor cluster usage from Windows Explorer.

chrislusf:
Fixes #10959.

## What was happening

`Statistics` answered with `max volume count * volume size limit`: how many
volumes the cluster is *allowed* to place, times how big each is allowed to get.
That is a provisioning number. It says nothing about how much space exists, and
writing or deleting data changes neither term, so it never moves.

`weed mount` feeds that number straight into `statfs`, so on Windows it lands in
Explorer's drive properties and in `dir`, where a capacity that is both wrong and
frozen is impossible to miss.

## Reproduction

A single volume server with 65536 slots, on a 460GB disk with 13GB free:

```
$ weed server -dir=/tmp/d -volume.max=65536 ...
$ # filer Statistics:
TotalSize=2111062325329920 (1966080.00 GiB)   # 1.9PB
UsedSize=0
```

1.9PB of capacity on a disk that holds 460GB. Writing 800MB moves `UsedSize` by
that much and leaves the reported free space visually unchanged, which is the
second half of the report.

## After

```
TotalSize=13714071552 (12.77 GiB)   # what df says is free
UsedSize=0
```

and after writing 800MB:

```
TotalSize=12881190336 (12.00 GiB)
```

The volume servers already send each filesystem's total and free bytes in every
heartbeat -- `command_volume_balance` and the EC balancer read them -- so the
master now bounds its answer by what the disks say is left: `min(slot capacity,
already written + free bytes)`. A disk roomier than its slot allowance keeps
reporting the slot limit, which is the real ceiling there. A volume server that
reports no disk bytes (older build, or a failed statfs) leaves the cluster on the
slot-only answer, so a mixed-version cluster degrades to today's behaviour rather
than to zero.

This is not Windows-specific -- `df` on a unix mount reported the same 1.9PB --
but Windows is where it is in the user's face.

## Second commit

A failed `Statistics` call returned from `StatFs` before filling in the answer,
so a mount whose filer or master was briefly unreachable reported an empty
filesystem instead of the sizes it already had. It now falls through to the last
known values.

## Tests

`TestStatisticsPhysicalCapacity` covers all three cases: nothing reported, disks
tighter than the slots, disks roomier than the slots.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved statistics reporting when filesystem refreshes fail by retaining the latest cached values.
  - Prevented reported cluster capacity from exceeding physical disk capacity.
  - Improved capacity calculations when disk usage data is incomplete or exceeds configured limits.

- **Tests**
  - Added coverage for physical disk capacity and volume-slot capacity handling.

### DIFF
--- a/weed/mount/weedfs_stats.go
+++ b/weed/mount/weedfs_stats.go
@@ -64,8 +64,8 @@ func (wfs *WFS) StatFs(cancel <-chan struct{}, in *fuse.InHeader, out *fuse.Stat
 			return nil
 		})
 		if err != nil {
+			// the last known sizes beat the empty filesystem a bare return reports
 			glog.V(0).Infof("filer Statistics: %v", err)
-			return fuse.OK
 		}
 	}
 
--- a/weed/server/master_grpc_server_statistics_test.go
+++ b/weed/server/master_grpc_server_statistics_test.go
@@ -123,11 +123,11 @@ func TestStatisticsReplicaCopyCount(t *testing.T) {
 	}
 }
 
-// reportDiskBytes has every node report the same filesystem capacity, the way
-// a volume server does in its heartbeat.
-func reportDiskBytes(ms *MasterServer, totalBytes, freeBytes uint64) {
+// reportDiskBytes has the first nodeCount nodes report the same filesystem
+// capacity, the way a volume server does in its heartbeat.
+func reportDiskBytes(ms *MasterServer, nodeCount int, totalBytes, freeBytes uint64) {
 	rack := ms.Topo.GetOrCreateDataCenter("dc1").GetOrCreateRack("rack1")
-	for _, node := range rack.Children() {
+	for _, node := range rack.Children()[:nodeCount] {
 		node.(*topology.DataNode).AdjustDiskUsageBytes(
 			map[string]uint64{"": totalBytes}, map[string]uint64{"": freeBytes})
 	}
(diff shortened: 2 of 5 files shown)


chrislusf:
On the two review points:

**Partial reports hide capacity** — fair, and fixed. `DiskUsages.FreeBytes` now
refuses to answer as soon as a disk holding volume slots reports no bytes, and
the new `Topology.FreeBytes` walks the data nodes and gives up on the whole
cluster if any one of them stays quiet. A cluster part way through an upgrade
from before 4.38 therefore falls back to the slot answer it gets today, rather
than to a sum missing whichever server holds the room. Covered by a new case in
`TestStatisticsPhysicalCapacity`.

**Shared disks count repeatedly** — real, pre-existing, and it errs the safe way.
`Store.CollectHeartbeat` sums total and free per disk *type*, so two `-dir` on
one filesystem report that filesystem twice; the comment there already calls it
out and names per-physical-disk reporting as the fix. Doubling the free bytes
only makes the ceiling looser, so the worst case is the capacity this PR is
narrowing stays wider than it should -- it cannot produce a cluster that reads
as full when it is not. Tightening it means changing what the heartbeat carries,
which is its own change.

Unrelated CI note: `mount: windows` failed on `TestRenameOverExisting`, which is
a known flake -- the identical failure is on master in run 32396055231.
`TestStatfsIsSelfConsistent` passed.

PR #10961, opened 2026-08-26
DIAGNOSTIC (do not merge): reproduce the flaky winfsp rename source

chrislusf:
Not for merge. Runs `TestRenameOverExisting` 300 times on the Windows runner
with the mount at `-v=4`, to catch the intermittent

```
stat of the renamed-away source did not return not-exist
```

seen in #10960's run and on master in run 32396055231, and to record which
layer answers that stat: the mount's own path cache (forgets within a second),
a readdir (reads no per-path cache), or the filer.

Will be closed once the cause is known.

### DIFF
--- a/.github/workflows/mount-windows.yml
+++ b/.github/workflows/mount-windows.yml
@@ -77,7 +77,7 @@ jobs:
 
         function Start-Mount($log) {
           Start-Process -FilePath .\weed.exe `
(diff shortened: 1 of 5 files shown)


chrislusf:
Served its purpose: reproduced both failure modes and produced the -v=4 log that identified the destination clobber. The fix is in #10965. The source-visibility failure is still open and noted there.

PR #10965, opened 2026-08-26
mount: stop a replaced rename destination from flushing over the rename

chrislusf:
A rename that replaces an existing destination could leave the destination
holding its **old** content, on any mount, and reliably enough on Windows to
fail CI.

## What was happening

`WFS.Rename` takes care of the source handle before it renames: it flushes
deferred metadata, marks the handle, and drains a pending async flush. It does
nothing about the destination. But replacing the destination *deletes* the
entry that was there, and a handle still open on it goes on flushing metadata
under that name.

Windows makes this easy to hit. WinFsp runs the cleanup and close that carry
the flush after the application's `CloseHandle` has already returned, so a
handle on the destination is still live when the rename runs, and its flush
lands afterwards. From a `-v=4` mount log of the failure:

```
dir Rename old_entry:{name:"src"} new_entry:{name:"dst" ... inode:16740536293338743416}
DeleteEntry /src
doFlush /dst fh 1521468582993181449            <- handles on the *replaced* dst
/dst saveToStorage 1,6872462993 [0,3)          <- uploads the old dst bytes
flushMetadataToFiler /dst inode 11939747521756968515   <- not the renamed inode
InsertEntry /dst                               <- clobbers the rename
```

The next read of `dst` returned the content the rename was supposed to replace.

## The fix

`Unlink` already solved this shape with `markHandleDeleted`, whose comment says
why the locking matters -- the flag goes up under the handle's flush lock, so a
flush already writing finishes before the caller proceeds and any later one
sees it, where *"setting the flag bare raced the flush's own check and
resurrected the entry right after the delete."*

A rename that replaces an entry deletes it just the same, so `Rename` now calls
the same helper on the destination inode. It is gated to `RenameEmptyFlag`, so
a `RENAME_NOREPLACE` (which fails) or a `RENAME_EXCHANGE` (where the
destination survives) marks nothing, and it is skipped when source and
destination are the same inode.

## Verification

`TestRenameOverExisting` in a 300x loop on the Windows runner: 300/300 with
this, against two runs that reproduced `target holds "old", want the source
content` without it.

## Not fixed here

The same test has a second, unrelated failure mode that predates this and is
still open: after a rename the *source* name stays visible. The added
diagnostics pin it down as the mount's meta cache retaining the entry -- a
listing shows it, it outlives the path cache, and no post-rename write touches
it, so the delete the rename's second stream response carries is not taking
effect. Every attempt to instrument it changes the timing enough to hide it, so
it needs its own investigation. The second commit here makes the next
occurrence report which layer answered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when renaming files over existing entries.
  * Prevented delayed file updates from overwriting destinations after rename or removal.
  * Improved consistency of source and destination paths during rename operations, including cases where entries are no longer immediately visible.
  * Ensured affected file handles are restored correctly if a rename fails.
  * Added diagnostics to help identify delayed visibility of renamed files and directory contents.

### DIFF
--- a/test/winfsp/semantics_test.go
+++ b/test/winfsp/semantics_test.go
@@ -180,9 +180,29 @@ func TestRenameOverExisting(t *testing.T) {
 		// they indict different layers; a second look says whether it persists.
 		time.Sleep(200 * time.Millisecond)
 		fi2, err2 := os.Stat(src)
-		t.Fatalf("stat of the renamed-away source did not return not-exist: stat=%v err=%v; 200ms later stat=%v err=%v",
-			describeFileInfo(fi), err, describeFileInfo(fi2), err2)
+		// A listing reads no per-path cache and the mount's own forgets
+		// within a second, so a name that survives both is back on the filer.
+		listed := dirNames(t, dir)
+		time.Sleep(2 * time.Second)
+		_, errLater := os.Stat(src)
+		t.Fatalf("stat of the renamed-away source did not return not-exist: stat=%v err=%v; 200ms later stat=%v err=%v; past the path cache err=%v; %s lists %v",
+			describeFileInfo(fi), err, describeFileInfo(fi2), err2, errLater, dir, listed)
 	}
+}
+
+// dirNames lists a directory for a failure message, reporting the error in
+// place of the names rather than failing a test that is already failing.
+func dirNames(t *testing.T, dir string) []string {
+	t.Helper()
+	entries, err := os.ReadDir(dir)
+	if err != nil {
+		return []string{"readdir: " + err.Error()}
+	}
+	names := make([]string, 0, len(entries))
+	for _, entry := range entries {
+		names = append(names, entry.Name())
+	}
+	return names
 }
 
 func TestRenameAcrossDirectories(t *testing.T) {
--- a/weed/mount/filehandle.go
+++ b/weed/mount/filehandle.go
@@ -40,7 +40,11 @@ type FileHandle struct {
 	savedName         string // last known file name if inode-to-path state is forgotten
 
 	isDeleted bool
-	isRenamed bool // set by Rename before waiting for async flush; skips old-path metadata flush
+	// deleteEpoch counts the times isDeleted was raised, all of them under the
+	// handle's flush lock. A caller that raised it and then found it had
+	// nothing to delete after all can tell its own mark from a later one.
+	deleteEpoch uint64
+	isRenamed   bool // set by Rename before waiting for async flush; skips old-path metadata flush
 
 	// entryVersionTsNs is the filer log position the handle's entry reflects.
 	// State at or below it must not replace the entry — that rolls it back.
--- a/weed/mount/weedfs.go
+++ b/weed/mount/weedfs.go
@@ -823,7 +823,7 @@ func (wfs *WFS) onEntryInvalidation(invalidation meta_cache.EntryInvalidation) {
 	// the old content back over what took the name. Marked from here, holding
 	// no other handle's lock.
 	if replacedInode != 0 {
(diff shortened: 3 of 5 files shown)


chrislusf:
Both findings were right; fixed in bdafa588f.

**Failed rename poisons handle** — the flag went up before the filer rename,
which can fail. The destination then stays where it was with its handle marked
deleted, dropping its dirty metadata from then on. `markHandleDeleted` now
takes the value to set, so the rename puts the flag back on any path that does
not return OK. Setting it rather than adding a second helper keeps one function
holding the flush lock, which is the part that has to be right.

**Forgotten mapping skips barrier** — the destination was only found through
`inodeToPath`, which Forget drops while the handle is still open, so exactly
the case this guards could slip through. It now falls back to the inode the
entry carries, the way the source side already does. The entry comes from the
sticky-bit check's existing load, hoisted so it is read once rather than twice.

The Windows verification for the original fix stands: `TestRenameOverExisting`
300/300 with it, against two runs that reproduced `target holds "old"` without.

chrislusf:
Right, and fixed in 4a6878f9f — the restore cleared `isDeleted` outright, so an
unlink that marked the same handle in between lost its mark and a later flush
could write the unlinked entry back.

Both places that raise the flag already do so under the handle's flush lock
(`markHandleDeleted`, and the invalidation path in `invalidateOpenFileHandle`,
which holds it for its whole body), so counting the raises there is enough to
tell one caller's mark from another's. `markHandleDeleted` returns the epoch of
its mark and `clearHandleDeleted` lifts it only while that epoch still stands;
a later mark owns the entry's fate and keeps it.

No ownership token or new lock needed — the ordering the flush lock already
provides is what makes the counter sufficient.

chrislusf:
**Rollback loses pending metadata** — correct, and fixed in 4ba999829.
`completeAsyncFlush` ends in `fh.ReleaseHandle()`, so a flush that runs while
the handle is marked skips its metadata *and* takes the handle with it, past
the reach of the restore. The drain now happens before the mark rather than
after: a flush already queued belongs to the destination as it stands, so it
finishes as itself, before the rename has taken anything away.

What is left is much narrower than the pending-flush case: a flush queued in
the window between the drain returning and the mark being taken. If the rename
succeeds, skipping it is correct — the entry is gone. Only a rename that then
fails loses it, and closing that would mean holding the destination handle's
flush lock across the filer rename, which `handleRenameResponse` already takes
for the DLM migration, so it would deadlock.

**Rollback clears newer deletion** — this one was already fixed in 4a6878f9f,
before this review ran; `markHandleDeleted` now returns the epoch of its mark
and `clearHandleDeleted` lifts it only while that epoch still stands.

chrislusf:
On the remaining `Entry.Attributes.Inode` fallback — checked, and I am leaving
it as is.

**Zero** is already handled: the fallback is only consulted when the path
mapping is gone, and the guard below it is `targetInode != 0 && targetInode !=
sourceInode`, so a zero never reaches `markHandleDeleted`.

**Differing from the runtime inode** is real but narrow. `InodeToPath.Lookup`
takes the entry's stored inode as `possibleInode` and uses it verbatim; the one
case it does not is the collision loop, which walks `inode++` until it finds a
free number. So the two differ only when that entry's inode number collided at
allocation time.

Worth being precise about what happens then: the destination handle is not
marked, which is exactly today's behaviour — the barrier does not apply, rather
than applying wrongly. There is no better key available, since handles are
indexed by inode and there is no path-to-handle index to fall back on.

And the same fallback, with the same property, is what the source side of this
function has always used a few lines above (`oldEntry.Attributes.Inode`). I
would rather the destination match it than invent a second rule here; if the
collision case is worth closing it should be closed for both, as its own
change.

PR #10969, opened 2026-08-26
DIAGNOSTIC (do not merge): is the version gate dropping the rename's source removal?

chrislusf:
Not for merge. Chasing the second `TestRenameOverExisting` failure mode, where
the renamed-away source stays visible:

```
past the path cache err=<nil>; S:\winfsp-test-TestRenameOverExisting lists [dst src]
```

A deterministic meta_cache test shows the mechanism exists: in
`applyMetadataResponseLocked` the two halves of a rename event are gated
independently, so when the source's version record outranks the event's `TsNs`
the removal of the source is dropped while the insert of the destination still
lands — leaving exactly `[dst src]` with the correct destination content.

This run logs **only when that gate actually drops something**, so it adds no
output on the common path and does not perturb the timing the way `-v=4` did
(which made the loop pass 300/300 every time).

Carries #10965 so the destination-clobber failure does not mask this one.

### DIFF
--- a/.github/workflows/mount-windows.yml
+++ b/.github/workflows/mount-windows.yml
@@ -127,6 +127,11 @@ jobs:
 
         Start-Mount 'seaweed-mount'
 
+        Invoke-Tests 'rename-loop' @('test','-v','-failfast','-count=400','-timeout','25m','./test/winfsp','-run','TestRenameOverExisting','-mountpoint=S:\')
+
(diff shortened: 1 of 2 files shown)


chrislusf:
Did its job: caught the gate dropping the source removal with record == eventTs on a live entry. Fix in #10973.

PR #10970, opened 2026-08-26
s3: keep verifying the request host when externalUrl is set

### LINKED ISSUE #10967: s3: -s3.externalUrl couples SigV4 host pinning to the advertised endpoint; mixed proxy + direct topologies cannot have both
**Describe the problem**

`-s3.externalUrl` does two jobs at once:

1. **SigV4 verification**: when set, `extractHostHeaderCandidates` (weed/s3api/auth_signature_v4.go) returns its host as the *only* candidate — a request signed under any other Host fails `SignatureDoesNotMatch` regardless of credentials.
2. **Endpoint advertisement**: it is the opt-in for `deriveS3AdvertisedEndpoint` (Iceberg LoadTable FileIO config, #9103) and the primary source for `deriveLanceStorageEndpoint` (Lance storage_options).

In a mixed topology — external clients arriving through a reverse proxy *plus* in-cluster clients dialing the Service directly — the two jobs pull in opposite directions:

- With the flag set to the public URL, every internal client must dial the public name back through the proxy (in our case a plaintext hairpin, since arbitrary pods don't trust the private CA on the TLS listener).
- With the flag unset, verification handles both planes correctly — #10284's candidate walk verifies whatever Host each client actually signed — but the Lance server advertises no storage_options endpoint (the k8s operator passes no `-ip`, so the fallback returns `""`), and object_store clients quietly fall back to real AWS S3.

**System Setup**

- weed 4.44 (`chrislusf/seaweedfs:4.44`) via seaweedfs-operator on Kubernetes (Talos); standalone gateway:
  ```
  weed -logtostderr=true s3 -port=8333 -filer=seaweedfs-filer:8888 -domainName=<public name> -metricsPort=9327 -port.iceberg=8181 -port.lance=9101
  ```
- Envoy (Cilium Gateway API) terminates TLS in front; Host header passes through unmodified
- Internal clients (Loki, a media relay): AWS SDK path-style against the Service name
- External clients: SDK/curl SigV4 against the public name, path-style and virtual-hosted

**Expected behavior**

Some way to advertise a public endpoint without narrowing signature verification to that single host. Two shapes, whichever fits the project's direction:

1. `externalUrl` becomes the **first** candidate rather than the **only** one: fall through to the request-derived candidates when it doesn't match. Same security posture as #10284 — every candidate still needs a valid signature, and the request-derived candidates are already trusted whenever the flag is unset. Smallest diff, but it does relax the current "only the public name is ever valid" behavior, if any deployment relies on that as policy.
2. A separate advertise-only flag (e.g. `-s3.advertisedUrl`) consumed by `deriveS3AdvertisedEndpoint` / `deriveLanceStorageEndpoint`, leaving `-s3.externalUrl`'s strict pin exactly as #8444 defined it.

**Additional context**

- #8444 introduced the pin for topologies where every client rides the proxy; #10284 made the unset path verify each plausible reverse-proxy host and deliberately preserved set-flag behavior; #9103 / #9109 / #10570 are the advertisement side.
- #6697 and #8598 are adjacent proxy-signature reports but different mechanisms (virtual-hosted addressing expectations; path-prefix encoding) — not duplicates of this.
- We currently run the unset-flag configuration in production — both internal Service-dialing clients and external proxied clients verify, confirmed on 4.44 — with the lost Lance advertisement as the accepted cost. Happy to send a PR for either direction.

---
🤖 Drafted with [Claude Code](https://claude.com/claude-code); investigated, reviewed, and submitted by a human.

chrislusf:
Fixes #10967.

`-s3.externalUrl` did two jobs that pull apart in a mixed topology: it is the opt-in for endpoint advertisement (Iceberg `LoadTable` FileIO config, Lance `storage_options`), and once set it was also the *only* host candidate SigV4 verification would try. A deployment with external clients arriving through a reverse proxy and in-cluster clients dialing the Service directly could have one or the other, never both — set the flag and every internal client had to hairpin through the public name, leave it unset and Lance advertised no endpoint, so object_store quietly fell back to real AWS S3.

This takes the first shape the reporter proposed: `externalHost` now *leads* the candidate walk instead of replacing it. The hosts derived from `X-Forwarded-*` or the request `Host` still follow, so a client that reaches the gateway directly verifies against the host it actually signed.

Same security posture as before. Every candidate still needs a signature that verifies under the caller's secret key — the Host header is not a credential, and the request-derived candidates are exactly the ones already trusted whenever the flag is unset.

Reproduced first as a failing case in `TestExternalUrlSignatureVerification`: with `externalUrl` set to the public name, an AWS SDK v2-signed request carrying the in-cluster Service host returned `SignatureDoesNotMatch`. It passes now, and the proxied cases still pass unchanged.

`test/s3/proxy_signature/README.md` had documented the old pin as intended behavior; updated. The `-s3.externalUrl` help text now also mentions the advertisement job, which it never did.

## Summary by CodeRabbit

* **Improvements**
  * S3 signature verification now tries the configured external URL first while retaining fallback host checks.
  * Direct and reverse-proxied clients can authenticate successfully using their respective signed hosts.
  * External URLs are advertised to Iceberg and Lance clients.

* **Documentation**
  * Updated command help, startup messages, architecture guidance, and configuration descriptions to clarify external URL behavior.

### DIFF
--- a/test/s3/proxy_signature/README.md
+++ b/test/s3/proxy_signature/README.md
@@ -26,16 +26,16 @@ nginx (:9000)
     v
 SeaweedFS S3 (:8333, -s3.externalUrl=http://localhost:9000)
     |  externalHost = "localhost:9000" (parsed at startup)
-    |  extractHostHeader() returns "localhost:9000"
+    |  extractHostHeaderCandidates() tries "localhost:9000" first
     |  Matches what AWS CLI signed with
     v
 Signature verification succeeds
 ```
 
-**Note:** When `-s3.externalUrl` is configured, direct access to the backend
-port (8333) will fail signature verification because the client signs with a
-different Host header than what `externalUrl` specifies. This is expected —
-all S3 traffic should go through the proxy.
+**Note:** `-s3.externalUrl` is tried first, not exclusively. A client that
+dials the backend port (8333) directly still verifies against the host it
+actually signed, so a mixed topology of proxied and in-cluster clients works
+with the flag set.
 
 ## Prerequisites
 
--- a/weed/command/filer.go
+++ b/weed/command/filer.go
@@ -158,7 +158,7 @@ func init() {
 	filerS3Options.iamReadOnly = cmdFiler.Flag.Bool("s3.iam.readOnly", true, "disable IAM write operations on this server")
 	filerS3Options.portIceberg = cmdFiler.Flag.Int("s3.port.iceberg", 8181, "Iceberg REST Catalog server listen port (0 to disable)")
 	filerS3Options.portLance = cmdFiler.Flag.Int("s3.port.lance", 9101, "Lance Namespace server listen port (0 to disable)")
-	filerS3Options.externalUrl = cmdFiler.Flag.String("s3.externalUrl", "", "the external URL clients use to connect (e.g. https://api.example.com:9000). Used for S3 signature verification behind a reverse proxy. Falls back to S3_EXTERNAL_URL env var.")
+	filerS3Options.externalUrl = cmdFiler.Flag.String("s3.externalUrl", "", "the external URL clients use to connect (e.g. https://api.example.com:9000). Advertised to Iceberg and Lance clients, and tried first when verifying S3 signatures behind a reverse proxy. Falls back to S3_EXTERNAL_URL env var.")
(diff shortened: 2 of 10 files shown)


chrislusf:
@coderabbitai review

chrislusf:
@coderabbitai full review

mwpastore:
@chrislusf Thank you! 🎉

PR #10972, opened 2026-08-26
VERIFY ONLY (do not merge): 400x rename loop for the source-removal gate fix

chrislusf:
Verification run for the fix in #10965's branch. 400 iterations of TestRenameOverExisting plus the full suite. Will be closed once the result is known.


chrislusf:
400/400 plus the full suite, zero failures. Fix in #10973.

PR #10973, opened 2026-08-26
mount: let a rename remove its source at the source's own version

chrislusf:
The second half of the `TestRenameOverExisting` failure, the one #10965 left
open: after a rename the source name stays visible.

```
stat of the renamed-away source did not return not-exist: ...
past the path cache err=<nil>; S:\winfsp-test-TestRenameOverExisting lists [dst src]
```

## What was happening

`applyMetadataResponseLocked` gates the two halves of a rename event
independently against `entryVersionBlocksLocked`, which reports *whether a
write at this version is already reflected*, with `floor >= tsNs`.

A rename stamps the source and takes its name away at the same log position,
so the removal arrives carrying exactly the version the source already records.
Caught in the act on the Windows runner, logging only when the gate drops
something:

```
gate dropped removal of /winfsp-test-TestRenameOverExisting/src
  eventTs=1787761708173717200 record=1787761708173717200
  floor=1787761708173717200 tombstone=false unversioned=false hasNew=true
```

`record == eventTs` exactly, on a live entry. So the removal of the source was
dropped while the insert of the destination from the same event still applied,
which is exactly `[dst src]` with the destination's content correct. Nothing
else ever removes the name, so it stayed cached: `readdir` listed it, `stat`
answered from it, and it outlived the path cache.

## The fix

A removal asks a different question from a write. An entry still present at
exactly that version has the write reflected but not its removal, so only a
strictly newer record fences a removal out. A `tombstone` is the removal
already reflected and goes on fencing at equality exactly as before -- and that
distinction was already in the code, which is what makes the two cases
separable.

The same log shows the benign side of this: ten earlier drops in that run all
had `tombstone=true`, correctly idempotent. Only the `tombstone=false` ones
were bugs.

## Tests

Two, both verified red without the fix:

- `TestRenameRemovesSourceAtItsOwnVersion` pins the production case -- source
  recorded at exactly the rename event's version, live entry, source must go
  and destination must land.
- `TestRemovalOlderThanTheRecordStillFences` guards the direction this must not
  break: a removal genuinely older than the entry's record is still stale and
  must not apply.

## Windows verification

400 iterations of `TestRenameOverExisting` plus the full suite on the Windows
runner: 400/400, zero failures. The same loop reproduced this failure within
~15 iterations before.

Worth noting for anyone who revisits this: it is timing-sensitive enough that
`-v=4`, and even four extra log lines per iteration, made it pass 300/300 every
time. The gate-fired-only logging above is what made it observable.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved metadata cache handling for file renames and removals.
  - Prevented outdated removal events from deleting newer cached entries.
  - Correctly removes renamed source entries while preserving destination entries.
  - Improved version checks during cache refresh and deletion operations.

- **Tests**
  - Added coverage for rename behavior, outdated removals, and directory refreshes at matching timestamps.

### DIFF
--- a/weed/mount/meta_cache/meta_cache.go
+++ b/weed/mount/meta_cache/meta_cache.go
@@ -497,13 +497,19 @@ func (mc *MetaCache) entryVersionRecordLocked(ctx context.Context, fp util.FullP
 	return int64(util.BytesToUint64(value[:8])), tombstone, unversioned
 }
 
-// entryVersionBlocksLocked reports whether a write at tsNs is already
+// entryVersionBlocksLocked reports whether a change at tsNs is already
 // reflected at fp. The path's version is its own record or, lacking one, the
 // listing floors — which cover names a listing saw present and absent alike.
 // A tombstone fences with no entry present, and a later floor outranks it; a
 // plain record only counts while its entry exists: records linger after a
 // bulk folder wipe and must not fence a recreate.
-func (mc *MetaCache) entryVersionBlocksLocked(ctx context.Context, fp util.FullPath, tsNs int64) bool {
+//
+// removal asks a different question from a write. An entry still present at
+// exactly tsNs has the write at that version reflected but not its removal --
+// a rename stamps the source and takes the name away at the same version --
+// so only a strictly newer record fences one out. A tombstone is the removal
+// already reflected, and fences at tsNs like any other write.
+func (mc *MetaCache) entryVersionBlocksLocked(ctx context.Context, fp util.FullPath, tsNs int64, removal bool) bool {
 	recordTsNs, tombstone, unversioned := mc.entryVersionRecordLocked(ctx, fp)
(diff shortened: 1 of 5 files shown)


chrislusf:
Good catch, and fixed in 70574498d.

Line 407 is inside the `vanished` loop and guards `localStore.DeleteEntry`, so
it removes and should ask about a removal; I had passed `false` there
mechanically to preserve the old behaviour rather than reading what the call
site does. A name recorded at exactly `snapshotTsNs` and absent from the
refreshed listing survived the sweep and stayed cached — the same equality bug
as the rename source, one path over.

Line 365 keeps `false`: it guards `localStore.InsertEntry`, which is a write.

Added `TestSectionRefreshSweepsANameRecordedAtTheSnapshot`, verified red with
`false` and green with `true`.

PR #10986, opened 2026-08-27
s3tables: delete only the location the dropped table owns

chrislusf:
`DeleteTable` derives its recursive purge target from the dropped table's stored `MetadataLocation`. That location comes from the caller on `CreateTable`/`RegisterTable`/`UpdateTable` and is never tied to the table it is stored on, so it can name a path the delete was never authorized for. The existing guards only reject a target outside the table bucket, or one that is an ancestor of the table's own catalog path -- a live table in a sibling namespace passes both, and the recursive delete then takes out its catalog entry and its data files.

A legitimately decoupled location is either a rename source or a leftover the name was reused over. Both have had their catalog attributes stripped, so a surviving metadata marker at the target means the path still belongs to another entry. Refuse those, next to the existing ancestor refusal.

Native `DeleteTable`, Iceberg `DropTable` and Lance `DropTable` all funnel through this path, so they are all covered.

Test covers a table whose metadata location points at a live table in another namespace: the drop is refused and the other table's entry and data survive. The existing decouple tests (rename source, reused name path, colocated) still pass unchanged.

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented deletion of a table when its data path belongs to another active catalog entry.
  - Preserved the associated catalog entry and data when such a conflicting path is detected.
  - Refused deletion when the data path cannot be safely verified.
  - Added coverage to verify that unsafe deletion attempts return an error.

### DIFF
--- a/weed/s3api/s3tables/handler_delete_decouple_test.go
+++ b/weed/s3api/s3tables/handler_delete_decouple_test.go
@@ -94,3 +94,29 @@ func TestDeleteTableColocatedRemovesData(t *testing.T) {
 	assert.Nil(t, fs.Get(GetNamespacePath(renameTestBucket, "ns"), "t"), "colocated table entry must be deleted")
 	assert.Nil(t, fs.Get(GetTablePath(renameTestBucket, "ns", "t"), "metadata"), "colocated table data must be deleted")
 }
+
+// A table whose MetadataLocation points at a sibling that is still a live
+// catalog entry must not be deleted: only the named table was authorized, so a
+// recursive purge of that location would destroy another tenant's table.
+func TestDeleteTableRefusesLiveSiblingDataPath(t *testing.T) {
+	fs, m := startRenameManager(t)
+
+	nsMeta, _ := json.Marshal(namespaceMetadata{Namespace: []string{"attackerns"}, OwnerAccountID: DefaultAccountID})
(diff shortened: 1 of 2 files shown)


chrislusf:
Good catch on the fail-open lookup - `getExtendedAttribute` returns filer/transport errors alongside `ErrAttributeNotFound`, so a transient failure read as "no marker here" and fell through to the recursive delete. Fixed: the check now only proceeds on `ErrAttributeNotFound`/`ErrNotFound` and propagates anything else.

chrislusf:
On the TOCTOU note: the window is real, but it predates this change and I'd rather not close it here.

The recursive delete was already unconditional, and the ancestor guard immediately above has exactly the same check-then-act shape - as does the colocated branch. There is no locking primitive anywhere in `weed/s3api/s3tables` today (no mutex, no filer-side lock); the only concurrency control in the package is the version-token check and the `mutateEntryExtended` CAS, neither of which covers the data path. So making this one call site atomic means introducing locking to the package, which is a design change rather than a fix and shouldn't ride along in this patch.

Exploiting it also needs the caller to land DeleteTable precisely as a victim CreateTable materialises at the attacker-chosen path - the check has to observe no marker and the delete has to hit a live table. That is much narrower than the case this PR closes, where the victim table simply already exists and the delete succeeds every time.

Happy to file it as a follow-up for locking the s3tables mutation paths as a whole, which is where it belongs.

PR #10987, opened 2026-08-27
s3: reject a request that names two operations

chrislusf:
A bucket subresource route matches on the presence of its query key, so `PUT /bucket?policy&tagging` matches both the `policy` route and the `tagging` route and the router serves whichever was registered first -- PutBucketPolicy. The IAM action resolver walks its own list in its own order and gets to `tagging` before the map holding `policy`, so it authorizes the same request as `s3:PutBucketTagging`. An identity delegated only bucket tagging can then write an arbitrary bucket policy.

The same mismatch also reaches PutBucketCors, PutBucketLifecycle, PutBucketVersioning, PutObjectLockConfiguration, the ADMIN-gated PutBucketRequestPayment, and DeleteBucketPolicy / DeleteBucketCors. Pairs that sit inside the resolver's map together resolve nondeterministically, since map iteration order is random.

Keeping the two lists in the same order would work only for as long as both stay in step, and the resolver recognizes fewer keys than the router does. Instead reject the ambiguity outright, alongside the other checks that run before a handler sees the request. Query keys that modify an operation rather than select one -- `versionId`, `partNumber`, `prefix`, `list-type`, the listing markers -- are not in the set and still combine freely, so `?tagging&versionId=`, `?uploadId=&partNumber=`, `?versions&prefix=` and `?attributes&versionId=` are unaffected.

Verified by hand against a running gateway: every ambiguous pair now returns 400, while single-subresource tagging, policy, acl, attributes, uploads, versions and list-type requests still reach their handlers.

## Summary by CodeRabbit

- **Bug Fixes**
  - S3 requests with multiple operation-selecting query parameters are now rejected with a clear HTTP 400 response.
  - Table deletion now prevents removal of data still associated with another active table, protecting catalog entries and table data.

- **Tests**
  - Added coverage for ambiguous S3 request rejection and safeguards against deleting shared table data.

### DIFF
--- a/weed/s3api/s3api_ambiguous_subresource_test.go
+++ b/weed/s3api/s3api_ambiguous_subresource_test.go
@@ -1,0 +1,75 @@
+package s3api
+
+import (
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestAmbiguousSubresource pins the rule that a request may name only one
+// operation. The router picks a handler by registration order and the IAM action
+// resolver picks an action by its own order, so a request carrying two operation
+// subresources gets authorized as one and served as the other.
+func TestAmbiguousSubresource(t *testing.T) {
+	for _, query := range []string{
+		"",
+		"policy=",
+		"tagging=",
+		"acl=&versionId=abc",
+		"tagging=&versionId=abc",
+		"retention=&versionId=abc",
+		"uploadId=xyz&partNumber=3",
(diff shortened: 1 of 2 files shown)


chrislusf:
Rebased onto master. This branch had accidentally been cut from the s3tables branch rather than master, so it was carrying a second, older copy of the `handler_table.go` change from #10986 - which is what conflicted once that PR merged with its follow-up fix.

Dropped that commit; the branch is now just the one commit touching `s3api_path_validation.go` and its test. The two review comments above are against the removed file and were already addressed in #10986.

PR #10988, opened 2026-08-27
s3: cover the directory marker key with object lock

chrislusf:
`PutObject` on a key ending in `/` takes a branch that runs before the versioning and Object Lock handling, and `DeleteObject` removes the same key the unversioned way, before the branches that enforce Object Lock. Between the two, a key the gateway reports as COMPLIANCE-retained can be deleted by anyone with plain delete permission.

The two halves:

- The PUT accepted `x-amz-object-lock-mode` and `x-amz-object-lock-retain-until-date` and stored neither, so an owner could believe a key was retained while nothing recorded it. An invalid mode, or a retain-until date in the past, came back 200 where the same headers on a regular key return 400. Retention set afterwards through `PutObjectRetention` *was* stored on the entry, and `GetObjectRetention` served it back -- only the delete ignored it.
- The delete path takes any key ending in `/`, while a PUT only makes a marker of one up to 1KiB. A larger one is a genuine versioned object with a version ID, so `DELETE ?versionId=` on it was correctly refused while the versionless `DELETE` went through the marker path and dropped the whole history.

Enforcement goes in the marker delete itself, so the single, versioned and multi-object delete paths are all covered; the multi-object path's existing check stays for keys that are not markers. On the PUT side the headers are validated the way the regular path validates them and stored beside the owner the same callback already sets.

Test in `test/s3/retention` covers a retained marker refusing single and batch delete, retention applied through `PutObjectRetention`, invalid lock headers returning an error as they do on a regular key, and an unretained marker still deleting.

One adjacent thing left alone, since it is a behavior change rather than a fix: the PUT branch is gated on `r.ContentLength <= 1024`, which is also true for `-1`, so a chunked upload of any size becomes a marker and is read whole into memory with the listing reporting size 0. Worth its own change.

## Summary by CodeRabbit

* **New Features**
  * Added Object Lock support for directory-marker objects, including retention, legal holds, governance bypass, and bucket-level defaults.
  * Directory markers now honor applicable Object Lock request headers.

* **Bug Fixes**
  * Protected directory markers and retained versions can no longer be deleted.
  * Invalid Object Lock headers are rejected consistently.
  * Deletion reports appropriate access-denied errors for retention and legal hold restrictions.
  * Retained entries in stale version history correctly block deletion.

* **Tests**
  * Expanded coverage for retention, legal holds, batch deletion, version history, invalid headers, and unprotected markers.

### DIFF
--- a/test/s3/retention/s3_object_lock_directory_marker_test.go
+++ b/test/s3/retention/s3_object_lock_directory_marker_test.go
@@ -120,6 +120,44 @@ func TestObjectLockDirectoryMarker(t *testing.T) {
 		require.Error(t, err)
 	})
 
+	// Past 1KiB a trailing-slash key is a real versioned object, and the delete
+	// takes the whole history at once, so an older retained version has to block
+	// it even when the version on top carries no retention of its own.
+	t.Run("a retained version under an unretained one still blocks", func(t *testing.T) {
+		key := "records/history/"
+		body := strings.Repeat("x", 2048)
+
+		first, err := client.PutObject(context.TODO(), &s3.PutObjectInput{
+			Bucket:                    aws.String(bucketName),
+			Key:                       aws.String(key),
+			Body:                      strings.NewReader(body),
+			ObjectLockMode:            types.ObjectLockModeCompliance,
+			ObjectLockRetainUntilDate: aws.Time(retainUntil),
+		})
+		require.NoError(t, err)
+		require.NotNil(t, first.VersionId)
+
+		_, err = client.PutObject(context.TODO(), &s3.PutObjectInput{
+			Bucket: aws.String(bucketName),
+			Key:    aws.String(key),
+			Body:   strings.NewReader(body),
+		})
+		require.NoError(t, err)
+
+		_, err = client.DeleteObject(context.TODO(), &s3.DeleteObjectInput{
+			Bucket: aws.String(bucketName),
+			Key:    aws.String(key),
+		})
+		require.Error(t, err, "the retained version underneath must block the delete")
+
+		_, err = client.HeadObject(context.TODO(), &s3.HeadObjectInput{
+			Bucket:    aws.String(bucketName),
+			Key:       aws.String(key),
+			VersionId: first.VersionId,
+		})
+		assert.NoError(t, err, "the retained version must survive")
+	})
+
 	t.Run("an unretained marker still deletes", func(t *testing.T) {
 		key := "records/plain/"
 		_, err := client.PutObject(context.TODO(), &s3.PutObjectInput{
--- a/test/s3/retention/s3_object_lock_directory_marker_test.go
+++ b/test/s3/retention/s3_object_lock_directory_marker_test.go
@@ -2,6 +2,7 @@ package retention
 
 import (
 	"context"
+	"errors"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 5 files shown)


chrislusf:
The lock lookup diverging from the history is a real gap, thanks. `enforceObjectLockProtections` with an empty versionId resolves the latest version, but the removal below it takes the whole `.versions` directory - so on an object-lock bucket a version retained under an unretained one was still destroyed. Added a commit that enforces against every version the removal covers, plus a test that puts a COMPLIANCE version under a plain one and checks the delete is refused and the retained version survives.

On the other half of that note - a marker with no history still resolves to its own directory entry (the `getLatestObjectVersion` fallback), which is why retention set via PutObjectRetention is read back correctly there.

chrislusf:
Tightened the test as suggested: the retained-marker deletes now assert `AccessDenied` rather than any error, the batch case asserts the reported key and code, and each refusal is followed by a HEAD confirming the marker survived.

Left the two invalid-header PUTs on a bare `require.Error` - that matches `TestObjectLockHeadersErrorCases` right next door, which covers the same validation on regular keys the same way.

chrislusf:
Right again - `getObjectVersionList` drops any entry without `Seaweed-X-Amz-Version-Id`, but the recursive removal takes it regardless, so an unnamed entry slipped past the check I had just added.

Replaced that call with a direct walk of the history directory. Every entry is now visited: a named one goes through `enforceObjectLockProtections` as before, and an unnamed one is checked with `getRetentionFromEntry`/`getLegalHoldFromEntry` and refuses the delete if either is active. Unnamed and unlocked entries still get cleared, which is what this removal exists to do.

Re-ran the original report's PoC and the >1KiB versioned variant against the rebuilt binary: both still return 403, the retained version survives the attempt, and unretained deletes are unaffected.

chrislusf:
Fair - I had made that branch fail closed on purpose, but you are right that it is the wrong shape: governance mode is bypassable by design, and refusing it there made an unnamed governance entry undeletable through this path while the named entry beside it is not. Now it refuses a legal hold and compliance mode unconditionally and honours `governanceBypassAllowed` for governance, matching `enforceObjectLockProtections`.

chrislusf:
Confirmed and fixed - this was the same mistake as the delete side, on the PUT side. `DoMkdir` builds a fresh entry and creates it, so the marker's lock metadata goes with the old one; but the overwrite check resolved the key's *latest version*, which on a key that has grown a history is not the marker at all. An unlocked version then answered for a retained marker and the plain PUT replaced it.

The check now judges the entry the write is about to replace. A versioned write of the same key still adds a version, which is correct - that is not an overwrite.

While fixing it I pulled the retention/legal-hold decision out of `enforceObjectLockProtections` into `enforceObjectLockOnEntry`, so the two callers that already hold an entry (the unnamed history entry, and the marker a PUT replaces) share the same rules instead of repeating them. That also replaces the inline copy I had added earlier.

Verified against a running gateway: plain PUT over a retained marker 403, still 403 after the key has a 2KiB version history, unretained markers still overwrite with 200, and both original PoC variants still 403.

chrislusf:
Both confirmed and fixed.

**Marker deletion checking the latest version.** I had spotted this one myself just before the review landed - it is the delete-side twin of the PUT bug from the previous round, and the same reasoning applies: the delete removes the marker entry, so that entry's own lock is what has to be checked, not whatever the key's latest version happens to be. With a history present, a retained marker sitting under unlocked versions was deletable. Now judged on the entry being removed; the versions underneath are still covered by the walk that follows.

**Marker overwrite race.** Fixed rather than deferred, because unlike the s3tables case there is already a primitive for it here: `withObjectWriteLock`, which the marker delete and the copy, rename and multipart paths all take. The marker PUT was the one write that did not. The check and the mkdir now run inside it, so the entry cannot change in between and marker PUT and marker DELETE serialize against each other.

Verified against a running gateway - retained marker: plain PUT 403, PUT still 403 after a 2KiB version history, DELETE 403. Unretained marker: PUT 200, overwrite 200, DELETE 204. Both original PoC variants still 403.

PR #10992, opened 2026-08-27
shell: remove the directories emptied by volume.fsck's filer entry purge

### LINKED ISSUE #10979: volume.fsck -findMissingChunksInFiler -reallyDeleteFilerEntries leaves empty directories after purging orphan filer entries
When using volume.fsck with the flags -findMissingChunksInFiler and -reallyDeleteFilerEntries to clean up orphaned filer entries (i.e., entries whose underlying volume chunks are missing), the orphaned file entries themselves are successfully deleted, but their parent directories are not automatically removed even if they become empty after the deletion.

Impact: These leftover empty directories accumulate over time, consuming unnecessary metadata space, cluttering the namespace, and requiring manual intervention to clean up.

chrislusf:
Fixes #10979.

`volume.fsck -findMissingChunksInFiler -reallyDeleteFilerEntries` deleted the orphan filer entries but left their parent directories behind, so a namespace slowly accumulated empty directories that had to be cleaned up by hand. The filer's own `EmptyFolderCleaner` does not cover this: it only runs inside buckets that opt in.

fsck now remembers the parent of every entry it purges and, once the purge is done, walks up from each one deleting the directories that are now empty:

- the delete is non-recursive, so the filer itself rejects a directory that still has children — no separate emptiness check to race with a concurrent write
- candidates are collected with all their ancestors and processed deepest first, so a directory emptied by two purged siblings is removed too
- a bucket directory is never deleted (that would drop its whole collection), the walk stops at the scanned root, and a directory that is an S3 object of its own is left alone

Reproduced on a local cluster: uploaded `/orphan/dir/deep/file.txt`, deleted its needle on the volume server, then ran the fsck above. Before, the entry was purged and `/orphan/dir/deep` stayed behind; now it is removed while `/orphan/dir` survives because a live file remains in it. Same for `/sib/a` + `/sib/b` and their shared parent, and for `/buckets/bucket1/dir` while `/buckets/bucket1` is kept.

https://claude.ai/code/session_01BncsNo2RVANCDtdbw96Kfc

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `volume.fsck` cleanup by removing directories that become empty after missing chunks are purged.
  * Preserves protected root, bucket, and scoped paths.
  * Avoids deleting directories that still contain content or were recently modified.
  * Safely prevents removal when directories change during cleanup.

* **Tests**
  * Added coverage for cleanup scope rules, successful deletion tracking, nested empty-directory cleanup, quiet-period handling, and concurrent directory changes.

### DIFF
--- a/weed/shell/command_volume_fsck.go
+++ b/weed/shell/command_volume_fsck.go
@@ -728,18 +728,16 @@ func (c *commandVolumeFsck) purgeEmptyDirectories() {
 	sort.Slice(dirs, func(i, j int) bool { return len(dirs[i]) > len(dirs[j]) })
 
 	for _, dir := range dirs {
-		isKeyObject, lookupErr := c.isDirectoryKeyObject(dir)
-		if lookupErr != nil {
+		entry, _, _, lookupErr := filer_pb.GetEntry(context.Background(), c.env, dir)
+		if lookupErr != nil && !errors.Is(lookupErr, filer_pb.ErrNotFound) {
 			fmt.Fprintf(c.writer, "lookup directory %s: %v\n", dir, lookupErr)
 			continue
 		}
 		// a directory key object is an S3 object of its own
-		if isKeyObject {
+		if entry == nil || entry.IsDirectoryKeyObject() {
 			continue
 		}
-		parent, name := dir.DirAndName()
-		// a non-recursive delete lets the filer reject a directory that is not empty
-		if err := filer_pb.Remove(context.Background(), c.env, parent, name, false, false, false, false, nil); err != nil {
+		if err := c.deleteEmptyDirectory(dir, entry.Attributes.GetMtime()); err != nil {
 			if !strings.Contains(err.Error(), filer.MsgFailDelNonEmptyFolder) {
 				fmt.Fprintf(c.writer, "delete empty directory %s: %v\n", dir, err)
 			}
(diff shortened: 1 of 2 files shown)


chrislusf:
Good catch on the directory-key window — fixed in 0509bcead.

The sweep now passes the mtime the lookup returned as `IfNotModifiedAfter`, so the filer skips any directory that changed between the read and the delete, including one promoted to an S3 object. That also covers the case where the entry appears only after the lookup. Resolution is one second, which is what the field carries.

Added `TestVolumeFsckPurgeEmptyDirectoriesKeepsChangedDirectory`: the stub filer bumps the entry's mtime right after fsck reads it, and the directory has to survive. It fails without the precondition.

chrislusf:
Closed the same-second window in 22740dde4.

The sweep now also skips a directory whose mtime is within the last 5 seconds. With both guards a write after the lookup always carries a later second than the one the delete is conditioned on: either it lands in a later second and `IfNotModifiedAfter` rejects the delete, or it lands in the same second as the directory's existing mtime — which the quiet period already excluded from the sweep. Such a directory is picked up by the next run.

`TestVolumeFsckPurgeEmptyDirectoriesKeepsFreshDirectory` covers it. The remaining assumption is the one the command already makes for -cutoffTimeAgo: the shell host's clock and the filer's are roughly in step.

chrislusf:
Right again — `IfNotModifiedAfter: 0` turns the filer's check off, so a directory whose entry carries no mtime was deleted with nothing holding it. Fixed in e387a245b: such a directory is skipped.

The main purge test now gives its directories a settled mtime and includes a `/nomtime/dir` that has to survive, so the disabled-condition path is covered.

chrislusf:
Taken in 3166b3f12. The race argument held with the strict comparison — a write after the lookup always lands at least `directoryQuietPeriod` seconds past the mtime the delete carries, so it can never share that second — but an mtime sitting exactly on the cutoff second is a boundary the guard may as well fail closed on, and a second of extra patience costs nothing here.

Issue #7220, opened 2025-09-10
Transient error on `LookupDirectoryEntry` leads to `NoSuchBucket` error


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that transient error on RPC `LookupDirectoryEntry` leads to the result that user operation fails and gets `NoSuchBucket` error. This RPC is used in `DeleteBucketHandler`. We expect that transient error on it can be handled by correct retrying mechanism. (In my opinion, `LookupDirectoryEntry` does not have idempotence problem)

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22252547/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run this script.

```python
import boto3


client_config = Config(signature_version='s3v4')
client = boto3.client(service_name='s3',
                    aws_access_key_id=config.main_access_key,
                    aws_secret_access_key=config.main_secret_key,
                    endpoint_url=config.default_endpoint,
                    use_ssl=config.default_is_secure,
                    verify=config.default_ssl_verify,
                    config=client_config)
bucket_name = get_new_bucket_name()
client.create_bucket(Bucket=bucket_name)
client.delete_bucket(Bucket=bucket_name)
```

Transient error can be reproduced by gRPC interceptor.

PR #11000, opened 2026-08-28
s3: distinguish a failed bucket lookup from a missing bucket on HEAD

### LINKED ISSUE #10998: HeadBucket reports NoSuchBucket for a transient lookup failure
### Description

`HeadBucketHandler` in `weed/s3api/s3api_bucket_handlers.go` treats a failed lookup as a missing bucket:

```go
if entry, err := s3a.getBucketEntry(bucket); entry == nil || errors.Is(err, filer_pb.ErrNotFound) {
	s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchBucket)
	return
}
```

Any error leaves `entry` nil, so the first arm fires and the client is told the bucket does not exist, even when the lookup simply failed.

This is the same shape that #7220 reported for `DeleteBucket`. That path no longer has it: `checkBucket` now goes through `getBucketConfig`, which separates the two cases (`weed/s3api/s3api_bucket_config.go`):

```go
entry, err := s3a.getBucketEntry(bucket)
if err != nil {
	if errors.Is(err, filer_pb.ErrNotFound) {
		s3a.bucketConfigCache.SetNegativeCache(bucket)
		return nil, s3err.ErrNoSuchBucket
	}
	glog.Errorf("getBucketConfig: failed to get bucket entry for %s: %v", bucket, err)
	return nil, s3err.ErrInternalError
}
```

`HeadBucket` was not part of that change and still carries the old form. A transient filer error during a HEAD therefore reports 404 rather than 500, which is misleading in itself and also stops clients from retrying, since a 404 is a definite answer.

### Suggested direction

Make `HeadBucketHandler` distinguish `ErrNotFound` from other errors, the way `getBucketConfig` already does.

### Version

Current `master` (d8a189f07).

chrislusf:
`HeadBucketHandler` collapsed every lookup outcome into one arm: `entry == nil || errors.Is(err, filer_pb.ErrNotFound)`. Any error leaves `entry` nil, so a transient filer failure during a HEAD was answered with 404 NoSuchBucket. That is misleading on its own and also stops clients from retrying, since a 404 is a definite answer. Same shape #7220 reported for DeleteBucket, which `checkBucket` has since fixed by going through `getBucketConfig`.

The handler now splits the two cases the way `GetBucketPolicyHandler` and `getBucketConfig` already do: `ErrNotFound` stays 404 NoSuchBucket, anything else is logged and answered 500 InternalError. It keeps the direct `getBucketEntry` read rather than routing through `getBucketConfig`, so a HEAD stays an authoritative filer lookup instead of becoming a cache-first existence answer that could report a stale 200 or 404 while a cross-gateway metadata event is in flight.

Verified with a new table test driving the handler against a fake filer: a lookup failing with a non-NotFound gRPC error yields 500 InternalError, a genuine miss still yields 404 NoSuchBucket, and an existing bucket still yields 200. The test fails on the pre-fix handler with `status = 404, want 500`. `go build ./weed/...` and `go test ./weed/s3api/...` pass.

Fixes #10998

https://claude.ai/code/session_01BjDWtZsCoZY6x4pdDmGWxU

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket status checks to distinguish temporary lookup failures from missing buckets.
  * Returns appropriate responses for existing buckets, unavailable services, and nonexistent buckets.

* **Tests**
  * Added coverage for successful bucket checks, missing buckets, and internal lookup errors.

### DIFF
--- a/weed/s3api/s3api_bucket_handlers.go
+++ b/weed/s3api/s3api_bucket_handlers.go
@@ -561,8 +561,13 @@ func (s3a *S3ApiServer) HeadBucketHandler(w http.ResponseWriter, r *http.Request
 	bucket, _ := s3_constants.GetBucketAndObject(r)
 	glog.V(3).Infof("HeadBucketHandler %s", bucket)
 
-	if entry, err := s3a.getBucketEntry(bucket); entry == nil || errors.Is(err, filer_pb.ErrNotFound) {
-		s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchBucket)
+	if _, err := s3a.getBucketEntry(bucket); err != nil {
+		if errors.Is(err, filer_pb.ErrNotFound) {
+			s3err.WriteErrorResponse(w, r, s3err.ErrNoSuchBucket)
+			return
+		}
+		glog.Errorf("HeadBucketHandler: failed to get bucket entry for %s: %v", bucket, err)
+		s3err.WriteErrorResponse(w, r, s3err.ErrInternalError)
 		return
 	}
 
--- a/weed/s3api/s3api_head_bucket_test.go
+++ b/weed/s3api/s3api_head_bucket_test.go
@@ -1,0 +1,88 @@
+package s3api
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/gorilla/mux"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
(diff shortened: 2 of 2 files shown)


chrislusf:
Good catch on the stale-cache point - the first revision routed HEAD through `getBucketConfig`, which would have turned an authoritative filer lookup into a cache-first answer. Reverted to a direct `getBucketEntry` read and split the error in the handler instead, matching `GetBucketPolicyHandler`. HEAD keeps its fresh-read semantics; only the error classification changes.

Issue #10997, opened 2026-08-27
MasterClient.WithClient waits for a master leader without a context


junkerderprovinz:
### Description

`MasterClient.WithClient` in `weed/wdclient/masterclient.go` uses a background context in two places:

```go
func (mc *MasterClient) WithClient(streamingMode bool, fn func(client master_pb.SeaweedClient) error) error {
	getMasterF := func() pb.ServerAddress {
		return mc.GetMaster(context.Background())
	}
	return mc.WithClientCustomGetMaster(getMasterF, streamingMode, fn)
}
```

`GetMaster` waits until a master is known. With `context.Background()` there is nothing to cut that wait short, so a caller that reaches this while the client has no current master leader blocks for as long as that lasts, rather than failing and letting the layer above decide.

This matters because callers above it do have a request context to offer. A filer serving an S3 request, for instance, ends up here through several paths, and any deadline it sets is applied to the RPC after this call returns, not to the wait inside it. So bounding the RPC does not bound the operation.

I ran into this while adding deadlines to a few master RPCs for a separate set of issues, and the fix there does not cover this case. Threading a context through `WithClient` touches a fair number of call sites, which is why I am filing it rather than attaching it to that change.

### Why this bites harder than it looks

`WaitUntilConnected` itself is written correctly: it selects on `ctx.Done()` in both its wait points. Only the caller is the problem, because `WithClient` hands it `context.Background()`. The loop then sleeps in roughly 200 ms steps until `getCurrentMaster()` is non-empty, and `setCurrentMaster("")` runs after every failed connect cycle, so while the master is down or mid-election the wait has nothing to end it.

That interacts badly with any attempt to bound the calls above it. If a caller adds a deadline to the RPC but the wait for the master address stays unbounded, the caller gives up, its client retries, and each retry creates another handler goroutine that parks in the same wait. Without the deadline one goroutine waits and the client waits with it. With it, the goroutines accumulate. So bounding the RPC without bounding this wait is worse than doing neither, which makes this worth fixing before, not after.

### Suggested direction

Accept a context in `WithClient` and pass it to `GetMaster`, so a caller that already has a deadline can bound the wait for a master as well as the RPC.

### Version

Current `master` (d8a189f07).

Issue #7233, opened 2025-09-15
Timeout on `AllocateVolume` causes Bad Gateway ClientError


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that Bad Gateway ClientError occurs when RPC `AllocateVolume` encounters timeout, during the operation of listing buckets. Here is the error message.

```
ERROR test_s3.py::test_multi_object_delete - botocore.exceptions.ClientError: An error occurred (502) when calling the ListBuckets operation (reached max retries: 4): Bad Gateway
```

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22329583/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_muti_object_delete`. https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L1684

Transient error can be reproduced by gRPC interceptor.

Issue #7232, opened 2025-09-15
Timeout on `CollectionDelete` causes bucket cleanup stuck for one minute


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that bucket cleanup stuck for one minute when RPC `CollectionDelete` from filer encounters timeout. During this process, filer keeps waiting for the completion of `CollectionDelete`, resulting in the cleanup which can be finished in one second costs about one minute.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22329407/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_bucket_list_empty`. https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L106

Transient error can be reproduced by gRPC interceptor. Or you can refer to our artifact for reproduction. https://github.com/qts0312/ChaosRPC/tree/main/examples/seaweedfs

Issue #7228, opened 2025-09-12
Timeout on `Assign` causes operation stuck for one minute


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that when RPC `Assign`, which is invoked by filer, encounters timeout, user operation will stuck. In this case, SeaweedFS keeps waiting for `Assign`, resulting in the operation that can be completed in three seconds stuck for more than one minute.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22291517/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_bucket_list_many`. 
https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L155

Transient error can be reproduced by gRPC interceptor.

PR #11015, opened 2026-08-28
master, filer, s3api: bound the collection RPCs that strand a caller

### LINKED ISSUE #7229: Timeout on `CollectionList` causes operation stuck for one minute
### Description

When using SeaweedFS in our cluster, we noticed a bug that user operation stuck for one minute when RPC `CollectionList` encounters timeout. This is because filer keeps waiting for the response of `CollectionList`, causing operation that can be completed in four seconds stuck for about one minute.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22291741/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_bucket_list_many`. 
https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L155

Transient error can be reproduced by gRPC interceptor.

### LINKED ISSUE #7234: Timeout on `DeleteCollection` causes task stuck for one minute
### Description

When using SeaweedFS in our cluster, we noticed a bug that when RPC `DeleteCollection` from master encounters timeout, the task will stuck for about one minute. The cause is that there is no deadline for this RPC, master keeps waiting for its completion util a high-level shutdown.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22329659/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_bucket_list_many`. https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L155

Transient error can be reproduced by gRPC interceptor.

junkerderprovinz:
Fixes #7234. Fixes #7229. Addresses #7232.

Three issues from the same reporter share one cause: collection RPCs issued with `context.Background()`, so a filer or master that accepts a call and then goes quiet strands the caller until some unrelated higher-level timeout fires.

## What each hop does now

**Master to volume server.** Every `DeleteCollection` in the fan-out is bounded. The per-server contexts derive from `context.WithoutCancel(ctx)`, so an inbound expiry bounds each RPC without truncating the walk. Both passes now run unconditionally: previously a failure in the normal pass returned early and the EC pass never ran, which left EC shards behind with nothing to retry them. The normal error is still returned first, unwrapped, so its status code reaches the caller as before.

**Filer to master.** `CollectionList` and `CollectionDelete` are bounded, and the budget is taken outside `util.Retry` so it covers the whole walk rather than one attempt.

**Gateway to filer.** Same, taken outside `WithFilerClient` so one budget covers the failover walk instead of one per candidate.

## Two things worth calling out separately

**The fan-out sent the same whole-collection delete to the same server thousands of times.** `ListVolumeServers()` returns a node once per replica it holds, while `DeleteCollection` removes the collection from every disk location on the server it reaches. One RPC per server is enough, and `ListEcServersByCollection` already dedupes this way. The normal pass now does too. This is the one hunk that is not strictly about deadlines, and it is easy to drop if you would rather keep the change narrower, but a per-RPC bound is hard to reason about while the same call is issued thousands of times over.

**A failed collection listing no longer fails `PutBucket`.** The result of that listing feeds a `glog.Warningf` and nothing else, yet a transient failure returned 500 and refused the creation. That is the user-visible half of #7229: `test_bucket_list_many` creates buckets. The listing is advisory now, so a transient failure is logged and creation continues. Behaviour when the listing fails is identical to when it returns false, minus the warning. This is an API behaviour change in a request path, so it is a separate commit and easy to review on its own.

## On the wait in front of the RPCs

`MasterClient.WithClient` starts with `GetMaster(context.Background())`, which waits for a master leader with nothing to cut it short. Bounding only the RPC would have made this worse rather than better: the caller gives up, its client retries, and each retry parks another handler goroutine in that wait. So there is a new opt-in `WithClientCtx` used at these two call sites. `WithClient` is untouched and the other callers are unchanged.

That narrows the exposure on two request paths, it does not close the gap. The general fix touches every caller, and I filed it separately as #10997.

## Values

15s for the filer's `CollectionList`, 20s for its `CollectionDelete`, 10s for the gateway's follow-up delete. The follow-up is deliberately shorter: by then the filer has already spent its own budget on the same collection, and that call is only the safety net for when the bucket directory was already gone, so a `DeleteBucket` costs about 25s of the roughly 60s an S3 client waits.

The delete budget bounds the wait, not the work. The master keeps deleting on its own fan-out, so giving up costs the confirmation and not the deletion. A timeout there logs a warning rather than an error for that reason; a genuine refusal still logs an error.

I did not remove the gateway's second `DeleteCollection` call even though it is usually redundant, since dropping it is a behaviour change that does not belong in a timeout change.

## What I am not claiming

For #7232 the stall shortens to roughly a quarter and becomes a clean error instead of a hang. If the master genuinely cannot answer, the cleanup still does not complete promptly, so "addresses" rather than "fixes".

#7228 and #7233 are from the same series but are already bounded on current master, by `operation.Assign`'s shared 30s budget and by `allocateVolumeTimeout`. Both were real at 3.97. They are not part of this change.

## Testing

Six tests, each verified by reverting only its own behaviour and confirming it fails for the right reason. They stand up real gRPC listeners: a wedged volume server, a fake master serving `KeepConnected`, and hanging filers. Covered: each RPC carrying a deadline, the fan-out outliving a cancelled caller, one RPC per server, the EC pass running when the normal pass fails, the EC pass surviving a cancelled caller, and the filer giving up when no master is reachable.

`go build ./...`, `go vet` and the package tests match the base commit; the three failures in `weed/server` and `weed/topology` and the `copylocks` findings in `weed/s3api` test files are all present on `d8a189f07` too.

Not verified: I have no cluster, so this rests on code review plus unit tests at each hop rather than on reproducing the reporter's measurements.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved bucket and collection deletion reliability when requests are cancelled or services are temporarily unavailable.
  * Added bounded timeouts to collection listing, deletion, and volume-server operations.
  * Ensured collection cleanup continues across replicas and erasure-coded volumes.
  * Prevented caller timeouts from incorrectly marking filer services as unhealthy.
* **Tests**
  * Added coverage for cancellation handling, timeout enforcement, failover behavior, and complete collection cleanup.

### DIFF
--- a/weed/filer/filer_delete_collection_test.go
+++ b/weed/filer/filer_delete_collection_test.go
@@ -1,0 +1,241 @@
+package filer
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"os"
+	"testing"
+	"time"
+
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials/insecure"
+
+	"github.com/seaweedfs/seaweedfs/weed/cluster"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
+	"github.com/seaweedfs/seaweedfs/weed/util"
+	"github.com/seaweedfs/seaweedfs/weed/util/log_buffer"
+	"github.com/seaweedfs/seaweedfs/weed/wdclient"
+)
+
+// collectionDeleteMaster is just enough of a master for a MasterClient to
+// consider itself connected, and records every CollectionDelete it is asked for.
+type collectionDeleteMaster struct {
+	master_pb.UnimplementedSeaweedServer
+	calls chan collectionDeleteCall
+}
+
+type collectionDeleteCall struct {
+	name string
+	// budget is the time the RPC arrived with, or 0 when it carried no deadline.
+	budget time.Duration
+}
+
+// KeepConnected is what MasterClient waits on before it reports a master: one
+// response is enough, then the stream idles until the server is torn down.
+func (m *collectionDeleteMaster) KeepConnected(stream master_pb.Seaweed_KeepConnectedServer) error {
+	if _, err := stream.Recv(); err != nil {
+		return err
+	}
+	if err := stream.Send(&master_pb.KeepConnectedResponse{}); err != nil {
+		return err
+	}
+	<-stream.Context().Done()
+	return nil
+}
+
+func (m *collectionDeleteMaster) CollectionDelete(ctx context.Context, req *master_pb.CollectionDeleteRequest) (*master_pb.CollectionDeleteResponse, error) {
+	call := collectionDeleteCall{name: req.Name}
+	if deadline, ok := ctx.Deadline(); ok {
+		call.budget = time.Until(deadline)
+	}
+	m.calls <- call
+	return &master_pb.CollectionDeleteResponse{}, nil
+}
+
+// hookedStore is the stub store with one extra seam: a callback that runs once
+// an entry has actually been removed, so a test can interleave an event -- a
+// client hanging up, say -- between the store write and whatever follows it.
+type hookedStore struct {
+	*stubFilerStore
+	onDeleteEntry func(util.FullPath)
+}
+
+func (s *hookedStore) DeleteEntry(ctx context.Context, p util.FullPath) error {
+	if err := s.stubFilerStore.DeleteEntry(ctx, p); err != nil {
+		return err
+	}
+	if s.onDeleteEntry != nil {
+		s.onDeleteEntry(p)
+	}
+	return nil
+}
+
+// newFilerWithFakeMaster builds a filer backed by the stub store, with its
+// MasterClient connected to a fake master the test can observe.
+func newFilerWithFakeMaster(t *testing.T) (*Filer, *hookedStore, *collectionDeleteMaster) {
+	t.Helper()
+
+	master := &collectionDeleteMaster{calls: make(chan collectionDeleteCall, 4)}
+
+	lis, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("listen: %v", err)
+	}
+	grpcServer := grpc.NewServer()
+	master_pb.RegisterSeaweedServer(grpcServer, master)
+	go func() { _ = grpcServer.Serve(lis) }()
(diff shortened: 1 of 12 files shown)


chrislusf:
Thanks for the write-up, it made this easy to follow.

Since you opened this, #11002 landed and took the filer-to-master half with it: `MasterClient.WithClient` now takes a context at all 35 call sites, `Filer.DoDeleteCollection` takes one, and `FilerServer.CollectionList` passes the request context through. That covers `WithClientCtx`, the `filer_grpc_server.go` hunk and most of `filer_delete_entry.go` here, and it also settles CodeRabbit's note above: the retry is `util.RetryWithBackoff(ctx, ...)`, whose backoff sleep already watches `ctx.Done()`.

Two notes on the parts of `WithClientCtx` that #11002 deliberately did differently, in case they are useful elsewhere: `pb.WithMasterClient` keeps `context.Background()` for the dial on purpose (that context only drives `shouldInvalidateConnection`, and `fn` brings its own RPC context — see #10948), and `DoDeleteCollection`'s call from `DeleteEntryMetaAndData` stays uncancellable rather than `WithoutCancel`-derived.

What is still open is the master's fan-out and the gateway's two RPCs, and the fan-out findings are the valuable half of this PR. I have rebased those onto master as #11026, with the tests, and credited you there. One thing I changed: `doDeleteNormalCollection` still returns at the first server that refuses, so one node being down leaves the collection on every server after it in the list — and with a per-RPC bound in place that now costs a minute before giving up. It keeps going in #11026 and reports the first failure.

Closing this in favour of #11026 unless you would rather rebase and carry it yourself, in which case say so and I will close mine instead.

junkerderprovinz:
#11026 should stand. The early return was wrong for the reason you give, and continuing while still reporting the first failure and leaving the collection in the topology is what I should have written. One residual cost, follow-up material rather than something to change now: the walk is serial with no aggregate bound, so several servers that accept and then go quiet cost a minute each under WithoutCancel, well past the filer's 15s, and a client retry starts another walk.

Issue #10994, opened 2026-08-27
Transient error classification can be steered by user-controlled bucket and object names


junkerderprovinz:
### Description

`util.IsTransientError` classifies by substring against a list of message fragments (`weed/util/retry.go`):

```go
var transientErrorMessages = []string{
	"transport", "connection reset", "connection refused", "broken pipe",
	"unexpected eof", "i/o timeout", "tls handshake timeout", "no such host",
	"no route to host", "network is unreachable", "client.timeout",
	"requesterror", "requesttimeout", "slowdown", "throttling",
	"internalerror", "resourceexhausted", "unavailable",
}
```

The comparison is case-insensitive, and the file's own comment explains why it has to work this way: the gRPC status is often dropped by an intermediate `%v` wrap, so the message is all that is left.

The problem is that several callers classify the error only after wrapping it with a user-controlled path. `DoSeaweedListWithSnapshot` does this at `weed/pb/filer_pb/filer_client.go:158`:

```go
return actualSnapshotTsNs, fmt.Errorf("list %s: %v", fullDirPath, err)
```

`isRetryableListError` in `weed/s3api/filer_util.go` then runs `util.IsTransientError` over that combined string. `fullDirPath` contains the bucket name and the prefix, both of which the client chooses. The same shape exists in `doDeleteEntry` in the same file, which wraps with `parentDirectoryPath` and `entryName` before the error is inspected.

So a bucket named `transport`, or an object under a prefix like `logs/unavailable/`, makes a non-transient failure look transient. A `PermissionDenied` on such a key is then retried rather than reported. The inverse also happens: `isFilerNotFound` substring-matches `"filer: no entry is found in filer store"`, so an object whose name contains that phrase suppresses a retry that should have run.

The impact is bounded today, a small number of extra attempts, but the classification is wrong for a reason the caller controls, and every new retry site inherits it.

I should say plainly that this includes code I contributed: `isRetryableListError` came from #10886, restructured into its current form in #10890. I did not think about the wrapping order at the time.

### The hazard is already documented in the tree

`isFilerNotFound` in `weed/s3api/s3_error_utils.go` carries this note:

```
// Callers must pass errors from paths without user-controlled
// segments that could spoof it.
```

So the spoofing risk was understood for the not-found check. The same reasoning applies to the transient check, which is reached through the same wrapped errors, but the rule was not carried across.

### The delete path is worse than it first looks

[...]

Issue #7224, opened 2025-09-11
Transient error on `DeleteEntry` causes object deletion fail


qts0312:
### Description

When using SeaweedFS in our cluster, we noticed a bug that deleting operation failed when encountered transient network failure. Transient error on RPC `DeleteEntry` causes deleting object fail, and there is no retrying mechanism for it and the object remains. We expect that transient error on it can be handled by correct retrying mechanism.

Here is the concrete call site of this RPC. [call_site.log](https://github.com/user-attachments/files/22278683/call_site_formatted.log)

### Version

3.97

### Reproduction

Install SeaweedFS with `helm` and run Ceph S3 tests `s3tests_boto3/functional/test_s3::test_multi_object_delete`. https://github.com/ceph/s3-tests/blob/6fd748b0e3a0088e0bf32c849865039355326327/s3tests_boto3/functional/test_s3.py#L1684

Transient error can be reproduced by gRPC interceptor.

PR #11016, opened 2026-08-28
s3api: replay a delete the filer's transport dropped, and classify by status code

junkerderprovinz:
References #7204 and #7224.

Both issues bottom out in the same place: `doDeleteEntry` in `weed/s3api/filer_util.go` issues `DeleteEntry` once and turns any failure into an error the client sees. For #7204 that becomes a 500, boto3 resends the DELETE, and the second one finds the bucket already gone and answers `NoSuchBucket`. For #7224 it becomes a per-key `InternalError` inside a 200 response, which no SDK retries, so the object silently survives.

`doDeleteEntry` is the only `DeleteEntry` call in `weed/s3api` proper, so a replay there covers every delete path in both handlers without touching any of the 23 `rm`/`rmObject` call sites.

## What this actually recovers, and what it does not

It recovers a transport blip: a filer restart, a GOAWAY, a single reset stream, anything where the channel is reconnecting underneath the call and a fresh attempt lands on a working transport.

It does not help when a filer is genuinely down. Once the channel is in TRANSIENT_FAILURE every attempt returns `Unavailable` immediately and the handler still emits the 500 or the in-200 `InternalError`. **If you test this with `docker stop filer` you will see no change at all.** In a multi-object delete the loop holds one filer client for the whole batch and `receiveTrackingConn` blocks failover once any reply has arrived, so a filer dying mid-batch means the remaining keys each replay and none of them recover.

That is why this says "references" and not "fixes". The complete answers are different changes: not mapping a transient filer failure to 500, and using a retryable per-key code. Both are behaviour changes I did not want to bundle here.

## The replay is immediate, and bounded anyway

No backoff. What clears a transient failure here is the channel reconnecting, not the passage of time, and a sleep would be paid once per key in handlers that delete up to 1000 entries per request.

Immediate is still not free, though. When a pick finds no ready subconn the picker waits for the balancer to publish one, and `WaitForReady(false)` does not shorten that wait; only the context does. So replaying multiplies that stall by the attempt count. There is a single bounded context shared by the replays to cap it.

The first attempt deliberately keeps the caller's own context. A recursive delete of a large bucket is legitimately slow, and putting a deadline on it here would fail deletes that succeed today. Only the attempts this change adds are bounded, so it cannot make anything slower than it already is.

## Classification is by status code only

The previous shape of this classified by message text through `util.IsTransientError`, which substring-matches. That is spoofable here, because the filer formats the object path into the error before the gateway sees it: `weed/filer/filer_delete_entry.go` wraps with `delete file %s: %v` and `weed/server/filer_grpc_server.go` copies that verbatim into `resp.Error`. A `PermissionDenied` on a key named `logs/transport.log` would have been replayed, and an object named after the not-found sentinel would have suppressed a legitimate replay. Stripping a prefix does not fix it either, since a recursive delete names the child entry it stopped on, which is another client-chosen key.

So this classifies by gRPC status code and never reads message text. `FilerServer.DeleteEntry` has exactly one return statement, `return resp, nil`, and reports failure through `resp.Error`, so everything the classifier sees came from the gRPC stack and carries no client-controlled text by construction.

The consequence is a deliberate narrowing worth seeing plainly: a `resp.Error` whose text happens to read transient is no longer replayed at all. Against current master nothing that works today stops working, since master never replayed anything.

The retryable set is `Unavailable`, `Internal` and `ResourceExhausted`, matching what `pb.shouldInvalidateConnection` treats as a broken channel, less `Aborted`. `Aborted` reports a conflict the filer decided on rather than a channel that broke. `ResourceExhausted` is unreachable today and is there for symmetry; if a filer-side load shed is ever added, an immediate replay would be the wrong answer to backpressure.

## Same class of bug, still open nearby

Two places still match against a message with the object path formatted in, both identical on current master so neither is a regression, but a reader should not take the claim above as broader than it is: `deleteObjectEntry`'s `MsgFailDelNonEmptyFolder` check, and `pb.shouldInvalidateConnection`'s substring fallback. `isRetryableListError` has the same shape and came in with #10886 and #10890, which is my own earlier contribution. I filed the general problem as #10994.

## Testing

Verified by reverting each property separately: removing the replay fails 5 tests, restoring free-text classification fails the classifier tests, removing the context check fails 1, and removing the replay bound fails 1. Several of the other tests pin invariants rather than demonstrating a change, and pass under both the old and new shape; they are there to stop a future edit from reintroducing the spoof.

Not verified: no cluster here, so this rests on unit tests and code reading rather than on reproducing the reporters' setups.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 deletion reliability by automatically retrying transient service failures.
  * Stops retrying when the request is cancelled or the retry time limit is reached.
  * Preserves immediate handling for permanent failures and filer-reported errors.
  * Ensures delete operations use consistent request context and error handling.

### DIFF
--- a/weed/s3api/filer_util.go
+++ b/weed/s3api/filer_util.go
@@ -132,14 +132,30 @@ func deleteObjectEntry(client filer_pb.SeaweedFilerClient, parentDirectoryPath, 
 // second attempt report a failure instead. Worth re-checking if this is ever
 // revisited.
 //
-// The replay is immediate. What clears a transient failure here is the gRPC
-// channel reconnecting underneath the call, not the passage of time, and any
-// wait short enough to sit in a request path is far too short to outlast a
-// filer that is genuinely down. Not sleeping also keeps the cost a property of
-// the delete rather than of the request: DeleteMultipleObjectsHandler deletes
-// up to deleteMultipleObjectsLimit keys and completeMultipartUpload removes one
-// entry per unused part, so a per-delete backoff would be paid once per key.
+// The replay carries no backoff of its own. What clears a transient failure
+// here is the gRPC channel reconnecting underneath the call, not the passage of
+// time, and a sleep would be paid once per key in handlers that delete up to
+// deleteMultipleObjectsLimit entries per request.
+//
+// That does not make the replay free. A pick that finds no ready subconn
+// returns balancer.ErrNoSubConnAvailable, and the picker then waits for the
+// balancer to publish a new one; WaitForReady(false) does not shorten that wait,
+// and its only escape is the context. So while a channel is CONNECTING an
+// attempt blocks rather than failing fast, and replaying would multiply that
+// stall by deleteRetryAttempts. deleteReplayTimeout bounds it instead.
 const deleteRetryAttempts = 3
+
+// deleteReplayTimeout bounds every replay of one delete taken together, so this
+// replay can add at most this much to a request however the channel behaves.
+//
+// The first attempt deliberately keeps the caller's own context: a recursive
+// delete of a large bucket is legitimately slow, and putting a deadline on it
+// here would fail deletes that succeed today. Only the attempts this change
+// adds are bounded, which is what keeps it from making anything worse than it
+// already is. Deriving from the caller's context rather than replacing it means
+// a real request context, once one is threaded in, still cancels the replay and
+// an earlier deadline of its own still wins.
+const deleteReplayTimeout = 2 * time.Second
 
 // isRetryableDeleteRPCError reports whether a failed DeleteEntry call is worth
 // replaying. It classifies by gRPC status code, never by message text.
@@ -159,6 +175,24 @@ const deleteRetryAttempts = 3
 // verbatim). Substring-matching that text would let an object named
 // "transport.log" make a permission denial look transient, and one named after
 // filer_pb.ErrNotFound make a genuine connection reset look authoritative.
+//
+// The set matches what pb.shouldInvalidateConnection treats as a broken
+// channel, less Aborted:
+//
+//   - Unavailable is the ordinary transport failure, and the reason this exists.
+//   - Internal is what grpc-go reports when a server ends a stream without
+//     trailers, which is how an L7 gRPC proxy truncating a reply arrives - the
+//     "error while returning the response" shape issue #7204 describes.
+//   - Aborted is excluded: it reports a concurrency conflict the filer decided
+//     on, not a channel that broke, so replaying it would just lose the race
+//     again.
+//   - ResourceExhausted is included for symmetry with the connection check but
+//     is unreachable today, as nothing filer-side sheds load with it. If a shed
+//     is ever added, an immediate zero-backoff replay is the wrong answer to
+//     backpressure and this line should become a delay instead.
+//
+// Context cancellation arrives as Canceled or DeadlineExceeded and is not
+// replayed: the caller is already gone.
 func isRetryableDeleteRPCError(err error) bool {
 	if err == nil {
 		return false
@@ -168,7 +202,7 @@ func isRetryableDeleteRPCError(err error) bool {
 		return false
 	}
 	switch s.Code() {
-	case codes.Unavailable, codes.ResourceExhausted:
+	case codes.Unavailable, codes.Internal, codes.ResourceExhausted:
 		return true
 	}
 	return false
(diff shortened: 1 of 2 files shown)


chrislusf (commented):
Thanks for this, and for the write-up — the diagnosis is right and the honesty about what it does and does not recover is exactly what I want to read in a PR body. The idea is sound: a delete is idempotent at the filer, so a reply lost in transit should be reissued rather than surfaced. What I disagree with is where the replay sits and what drives it.

I've pushed an alternative at #11022 that keeps your analysis and changes the placement. Details below so you can see the reasoning rather than just the diff.

## It no longer applies

The branch point is #10985. Since then #11003 and #11004 rewrote `doDeleteEntry`, so GitHub reports this as conflicting. That is not just a textual rebase.

## `fail(cause string)` would undo #11004

Every error now goes through `fail`, which formats a string. Against current master that drops the `%w` wrapping and the `filer.DeleteEntryError` call, and `deleteObjectEntry` decides whether to demote a directory marker with `errors.Is(err, filer.ErrNonEmptyFolder)`. That check would never match again and the demote would silently stop happening. Your `TestDoDeleteEntryKeepsItsErrorWrapping` asserts the pre-#11004 string form, which is why it passes on the branch.

## `util.ServerStatus` already gives you the separation

`deleteEntryAttempt` returning `(error, string)` exists to keep the filer's own message away from the classifier. #11004 added `util.ServerStatus` for that: the filer reports its refusals in `resp.Error`, which carries no gRPC status, so it cannot be mistaken for a transport failure even after `%w` wrapping. Checked it:

    status.FromError(wrapped)  ok=true  code=Unavailable  msg="delete entry /buckets/b/transport.log: rpc error: ..."
    errors.As -> GRPCStatus()  servermsg="transport is closing"
    status.Code(wrapped)       Unavailable
    plain resp.Error error     code=Unknown, FromError ok=false

`status.FromError` recovers the code correctly through a wrap but hands back the wrapper's whole text as the message — that is the trap #11004 was about. The code is safe to read, the message never is. So `status.Code` or `ServerStatus` on the wrapped error is enough; no restructuring of the call site is needed.

## The replay runs on the channel that just failed, and delays the thing that recovers

`pb.WithGrpcClient` drops the cached `ClientConn` only after `fn` returns, and `withFilerClientFailover` only advances to the next filer after `fn` returns. Both of your extra attempts happen inside `fn`, so they are stuck with the channel that just failed, and in a multi-filer deployment they hold `deleteReplayTimeout` before a healthy filer is tried. Master fails over immediately. A replay wrapped around `WithFilerClient` gets a fresh dial and walks the failover list again.

## The 2s bound is per key, and the batch is the case being fixed

`deleteReplayTimeout` bounds one delete. `DeleteMultipleObjectsHandler` deletes up to `deleteMultipleObjectsLimit` of them, so in exactly the blocking-picker case the bound exists for, 1000 keys is 2000 seconds. No `WriteTimeout` catches that.

The tree already answered this in #11001: `withFilerRetryBudget` hands one allowance to the whole request, and `DeleteMultipleObjectsHandler` already installs it. `doDeleteEntry` just cannot reach it, because there is no real request context — which is the same `context.Background()` both bots flagged at lines 101 and 116, and the reason it is worth threading for real rather than as a parameter that is always `Background`.

## It is a third retry mechanism

`retryFilerOp` with `isRetryableFilerErr` and a shared allowance is one, `listWithRetry` with `isRetryableListError` is another and sits thirty lines above the new code in the same file — and came in with your #10886. `deleteRetryAttempts` / `deleteReplayTimeout` / `isRetryableDeleteRPCError` is a third shape for the same job. #11022 wires the delete paths into `retryFilerOp` instead, which drops out three bugs in it that only showed up once deletes went through it: a non-empty-folder refusal was retried six times over 3.1s, `Canceled` and `DeadlineExceeded` were matched by sentinel only and so were missed once they had crossed gRPC, and `updateLatestVersionAfterDeletion` was already wrapping `rm` in its own `retryFilerOp`.

## Smaller things

The doc comment on `isRetryableDeleteRPCError` says "Context cancellation arrives as Canceled or DeadlineExceeded and is not replayed" twice, opening and closing. "Master logged every failed delete RPC at V(1)" reads as the master server rather than the branch. And `fail` now logs `resp.Error` at V(1), which master did not.

On greptile's replay-deletes-the-recreation comment: it is real but not new. `withFilerClientFailover` already replays a whole `fn` against the next filer whenever nothing was received, so a delete that committed with its reply lost is already reissued by path alone today. Worth a separate look, not a blocker for either PR.

## What I kept

The framing that the filer's own message is client-controlled text and must never be classified, the argument that `Aborted` is a decision and not a broken channel, and the point that a sleep per key is the wrong shape for a handler that deletes a thousand of them. #11022 credits this PR; happy to keep going here instead if you would rather carry it.

chrislusf:
Closing in favour of #11022, which fixes both issues with the placement described in my review above. Your diagnosis is what it is built on, and the PR body says so.

I have deliberately not touched this branch. If you want to take the approach in #11022 further, or you disagree with any of it, reopen and say so — I would rather land your version than mine.

The three things I kept from here, for the record: the filer's own message is client-controlled text and must never be classified, `Aborted` is a decision rather than a broken channel, and a sleep per key is the wrong shape for a handler that deletes a thousand of them. Thanks for the careful write-up, including the part about what it does not fix — that is the half most PRs leave out.

junkerderprovinz:
No disagreement, and no need to reopen. You are right about fail(): flattening the %w and filer.DeleteEntryError that #11004 added would have stopped errors.Is(err, filer.ErrNonEmptyFolder) from matching in deleteObjectEntry and quietly killed the marker demote, and my test asserted the string rather than the sentinel, which is why it went past me. Reusing retryFilerOp was right too, and the per-request allowance is the answer to the 1000-key batch that my per-key bound was not.

PR #11019, opened 2026-08-28
Let filer.sync move past a chunk the source cluster no longer has

### LINKED ISSUE #11010: filer.sync wedges permanently when vacuum removes a volume it still needs to replay
## Summary

`weed filer.sync` stops making progress — permanently and silently — if it replays a metadata log event whose chunks live in a volume that vacuum has since removed on the source. The process stays alive, the container reports healthy, no further errors are logged, and the replay offset never advances again.

This is reachable whenever the sync is behind (or is started from an older offset) and vacuum runs over the intervening range. Nothing coordinates the two subsystems, so I expect it to be reproducible for anyone running `filer.sync` alongside normal vacuuming.

**Version:** 4.37 (source and target clusters both).

## Symptoms

- Replay position freezes; identical `progressed to <timestamp>` line count before and after a multi-minute observation window
- ~0.00% CPU
- Container `Up`, 0 restarts
- **No new log output at all after the initial handful of errors** — this is what makes it dangerous, since health checks and container state show nothing wrong

Immediately before it stopped:

```
permanent size mismatch replicating 5616,<fid> for /buckets/<bucket>/...
W filer_sink.go:436 skip stale entry /buckets/<bucket>/...
W filer_sink.go:263 replicate entry chunks /buckets/<bucket>/...
volume 5617 not found
```

Confirmed the referenced volumes were genuinely gone from the source master:

```
$ curl "http://<master>:9333/dir/lookup?volumeId=5617"
{"volumeOrFileId":"5617","error":"volume id 5617 not found"}

$ curl "http://<master>:9333/dir/lookup?volumeId=5616"
{"volumeOrFileId":"5616","locations":[{...}]}      # neighbouring volume still present
```

## Root cause

In `weed/replication/repl_util/replication_util.go`, when a chunk's volume cannot be resolved, the `LookupFileId` error propagates straight up:

```go
if err != nil {
    return err
}
```

There is no skip path, no retry budget, and no way to advance the offset past an unresolvable event. The fallback loop over `fileUrls` only applies once a URL list exists, so a lookup failure never reaches it.

The underlying issue is that vacuum and `filer.sync` have incompatible assumptions:

- vacuum reclaims deleted needles with no knowledge of whether a replication consumer still needs them
- `filer.sync` assumes every event in the metadata log is resolvable

There is no back-pressure mechanism — nothing analogous to a PostgreSQL replication slot that pins WAL until the consumer has caught up. So the log can legitimately reference chunks that vacuum has already destroyed, and the consumer has no defined behaviour for that case.

## Reproduction

1. Run `filer.sync` between two clusters
2. Stop it, or start it from an offset well in the past (e.g. `-a.fromTsMs` pointing at the start of the retained log)
3. Let vacuum run on the source and remove a volume referenced within that backlog
4. Start/resume the sync — it advances until it reaches the affected event, then wedges

Retention interacts with this: raising `metaLogRetentionSec` widens the window in which an event can outlive its data, increasing exposure.

## Workaround

Restart with `-a.fromTsMs` set past the affected range:

```
-a.fromTsMs=<ms just after the poison event>
```

In my case the backup had a complete base copy taken after those writes, so skipping the older log range lost nothing, and the outstanding objects replicated within seconds of the restart. That reasoning only holds if you *know* the skipped range is already covered — otherwise this workaround silently drops data.

## Suggested fix

Any of these would be a large improvement over the current behaviour, roughly in order of preference:

1. **Skip and continue.** Treat an unresolvable chunk as a permanent (not transient) failure for that entry: log it at WARN/ERROR, advance the offset, continue. Optionally write skipped paths to a file so operators can reconcile them out-of-band.
2. **A flag to choose the policy**, e.g. `-onMissingChunk=skip|fail`, defaulting to `skip` for `-isActivePassive` backup topologies where liveness matters more than completeness of any single entry.
3. **Fail loudly rather than silently.** If aborting is the intended behaviour, exit non-zero instead of hanging, so supervision and monitoring can react. The current mode — alive, healthy-looking, doing nothing — is the worst outcome for a backup process.
4. **Back-pressure** (largest change): let vacuum consider the minimum offset of registered sync consumers before reclaiming, similar to a replication slot.

Happy to test a patch against my setup.

chrislusf:
Fixes #11010.

`filer.sync` stops checkpointing, permanently, once it replays an event whose chunks live in a volume the source has since vacuumed away. The process stays up and looks healthy while its offset never moves again.

### What happens

Two shapes reach the sink, and both are in the reporter's log:

- **The volume is gone.** `LookupFileId` finds no locations, and `fetchAndWrite`'s `RetryUntil` gate treats that as neither a size mismatch, nor supersession, nor a network error — so it returns `shouldContinue = true` and retries every ~6s forever. The sync job never finishes, so it holds its slot and pins the offset watermark at the event ahead of it. That is the wedge.
- **The needle is gone but the volume is still there.** The volume server answers 404, and `ReadPart` never looked at the response status — so the error page came back as a successful read and its bytes were counted as file content. The entry was then reported as a chunk size mismatch: a corruption claim about data the source had simply lost. A 404 from one replica also ended the search instead of trying the next.

Reproduced against a fake source filer that still holds the entry, unchanged, while its master locates nothing: `fetchAndWrite` was still retrying after five seconds with no way out.

### The fix

- `LookupFileId` returns a typed `ErrVolumeNotFound` instead of formatting a nil error into its message, which was all a caller had to match on.
- `ReadPart` fails on a failure status, closing the response; 404 becomes a not-found error and the next replica gets a turn.
- A chunk the source cannot produce is waited out for a grace period long enough to cover a volume server restart or a master failover, then given up on and marked permanent, instead of being retried forever.
- An entry whose data the source has lost for good is skipped with an error naming it, so replication carries on and the offset advances. Nothing brings those bytes back, and holding the offset for them only stops every later event from ever being checkpointed.

### The wait is per volume, not per chunk

A vacuumed volume took every file it held with it. Timing the grace period per chunk would serialize those waits behind the bounded chunk executor, so one gone volume holding many files would stall the sync for far longer than the grace period — the wedge again, only slower. The wait is tracked per source volume on the sink: the first chunk to find it unlocatable starts the clock, every later chunk inherits it and gives up as soon as it has run out, and a chunk the source does serve clears it. So a gone volume costs one grace period in total, however many files it held.

An incremental sink's dated target keys can't be mapped back to a source path, so supersession can't be checked there at all. Those propagate immediately rather than waiting, and `filer.backup` decides with the event's real source key — the same reasoning `manifestResolveRetryGate` already applies.

### Why the skip is conditional

A volume with no locations reads the same whether it was vacuumed away or every replica is down, and during a cluster-wide outage that answer comes back for *every* chunk — skipping then would drop live files wholesale. So the sink first re-checks a file id the source did serve: if that one has become unresolvable too, the error is held for a retry as before.

Supersession still wins ahead of all of this, so a stale replay whose file was overwritten or deleted is skipped immediately, as it is today.

### Not changed

The reporter pointed at `repl_util/replication_util.go`. That path serves the local/gcs/azure/b2 sinks; `filer.sync` reads through `filersink`'s own `fetchAndWrite`, which is where the loop that never ends lives.

### Tests

- `TestFetchAndWriteStopsOnMissingSourceChunk` — the reproduction, now returning `errSourceChunkMissing` instead of spinning.
- `TestMissingSourceChunkGate` — the grace period, that an unrelated failure doesn't start it, that a second file in the same volume inherits the wait rather than restarting it, and that a served chunk clears it.
- `TestFetchAndWriteMissingSourceChunkUnverifiableSupersession` — no wait when supersession can't be checked.
- `TestOnReplicateChunkErrorMissingSourceChunk` — skip while the source still serves other chunks; hold when it can locate nothing, and when nothing has been served yet to check against.
- `TestReadPartSkipsNotFoundReplica`, `TestReadPartAllReplicasNotFound`, `TestLookupFileIdVolumeNotFound`.

https://claude.ai/code/session_01SRPEP4jRu29FbLSN6bjLaK

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved replication handling for temporarily unavailable source chunks with a per-volume grace period.
  - Stopped retries for chunks confirmed to be permanently unavailable, allowing replication to continue.
  - Improved recovery when source servers restart or experience outages.
  - Correctly handled missing volumes and HTTP 404 responses without treating error pages as chunk data.
  - Added fallback behavior when individual replicas are unavailable.
  - Improved error reporting for source chunks that cannot be retrieved.

### DIFF
--- a/weed/replication/sink/filersink/fetch_write.go
+++ b/weed/replication/sink/filersink/fetch_write.go
@@ -47,23 +47,22 @@ func isSourceChunkMissing(err error) bool {
 	return errors.Is(err, source.ErrVolumeNotFound) || errors.Is(err, util_http.ErrNotFound)
 }
 
-// missingSourceChunkGate tracks how long the source has been unable to produce a
-// chunk across retries, so a lookup race or a restarting volume server is waited
-// out while a vacuumed one is written off instead of retried forever.
+// missingSourceChunkGate decides when to stop retrying a chunk the source cannot
+// produce, so a lookup race or a restarting volume server is waited out while a
+// vacuumed one is written off instead of retried forever. The wait is kept per
+// volume on the sink: a volume that is gone took every file it held with it, and
+// waiting it out once per file would stall the sync for as long as it holds files.
 type missingSourceChunkGate struct {
-	missingSince time.Time
-	gaveUp       bool
+	sink     *FilerSink
+	volumeId string
+	gaveUp   bool
 }
 
 func (g *missingSourceChunkGate) isPermanent(err error) bool {
 	if !isSourceChunkMissing(err) {
-		g.missingSince = time.Time{}
 		return false
 	}
-	if g.missingSince.IsZero() {
-		g.missingSince = time.Now()
-	}
-	if time.Since(g.missingSince) < missingSourceChunkGrace {
+	if time.Since(g.sink.sourceVolumeMissingSince(g.volumeId)) < missingSourceChunkGrace {
 		return false
 	}
 	g.gaveUp = true
@@ -194,7 +193,7 @@ func (fs *FilerSink) replicateOneManifestChunk(ctx context.Context, sourceChunk 
 	// supersession or, when that cannot be checked, after a few attempts.
 	var resolvedChunks []*filer_pb.FileChunk
 	resolveName := fmt.Sprintf("resolve manifest %s", sourceChunk.GetFileIdString())
-	missingGate := &missingSourceChunkGate{}
+	missingGate := fs.newMissingSourceChunkGate(sourceChunk.GetFileIdString())
 	err := util.RetryUntil(resolveName, func() error {
 		rc, e := filer.ResolveOneChunkManifest(ctx, fs.filerSource.LookupFileId, sourceChunk)
 		if e != nil {
@@ -374,7 +373,7 @@ func (fs *FilerSink) fetchAndWrite(sourceChunk *filer_pb.FileChunk, path string,
 
 	transientBackoff := time.Duration(0)
 	_, canCheckSupersession := fs.targetPathToSourcePath(path)
-	missingGate := &missingSourceChunkGate{}
+	missingGate := fs.newMissingSourceChunkGate(sourceChunk.GetFileIdString())
 	var partialData []byte
 	var savedFilename string
 	var savedHeader http.Header
@@ -418,8 +417,7 @@ func (fs *FilerSink) fetchAndWrite(sourceChunk *filer_pb.FileChunk, path string,
 		if err := validateReplicatedReadSize(sourceChunk, len(fullData)); err != nil {
 			return err
 		}
-		servedFileId := sourceChunk.GetFileIdString()
-		fs.lastServedFileId.Store(&servedFileId)
+		fs.sourceServed(sourceChunk.GetFileIdString())
 
 		transferStatus.mu.Lock()
 		transferStatus.BytesReceived = int64(len(fullData))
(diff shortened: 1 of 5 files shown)


chrislusf:
Both bot findings landed on `sourceStillServesChunks`. Addressed in 8f76e9331 and worth stating what is and isn't fixed.

**CodeRabbit — the probe only proved the master had topology, not that a volume server answers.** Correct, and fixed: the probe is now a `ReadPart` of the last chunk the source did serve, with the response closed straight away. If every volume server is unreachable while the master still lists them, the probe now fails and the entry is held instead of written off. `TestOnReplicateChunkErrorMissingSourceChunk/probe_locates_but_cannot_be_read` covers exactly that.

**Greptile — a prolonged outage confined to the one volume can still checkpoint an event whose data was never copied.** That one is real and not fixable at this layer. A volume with no locations is byte-for-byte the same answer whether vacuum removed it or its servers are down, and no RPC the sink can make distinguishes them: the master's lookup returns nothing in both cases, and there is no "this id was deleted" signal to ask for. The only evidence available is time and the health of the rest of the cluster, which is what the grace period and the probe are.

So the residual exposure is: one volume unreachable for longer than the grace period, with the rest of the cluster serving, ends with its files skipped and an ERROR per file naming the path and the chunk. Against that, the alternative is the reported bug — the offset pinned forever, no checkpoint, and a restart replaying into the same wall.

Happy to trade differently if that is the call: raising the grace period is a one-line change, and holding on the no-locations shape while still skipping the 404-from-a-live-volume-server shape (where the evidence *is* positive — the server is up, has the volume, and says the needle is gone) would be a narrow variant. It just would not fix #11010, whose volume was gone outright.

PR #11021, opened 2026-08-28
mount: implement fallocate

### LINKED ISSUE #11013: `posix_fallocate` fails with `EBADF` (Bad file descriptor) if `offset+len` is less than file size after write
# Detailed description

Hello, there seems to be an issue with seaweedfs using `mini` configuration and POSIX interface:

1. Create an empty file.
2. Open the file with `O_WRONLY` flag (`O_RDWR` works as expected).
3. Write some data to file.
4. Try `posix_fallocate` on the file with `offset+len` less than write size.

Last step fails with `Bad file descriptor` errno (NOTE: by definition `posix_fallocate` actually does not set errno and returns the value instead).

# System info

```json
{
        "Hostname" : "ubuntu",
        "StaticHostname" : "ubuntu",
        "PrettyHostname" : null,
        "DefaultHostname" : "localhost",
        "HostnameSource" : "static",
        "IconName" : "computer-vm",
        "Chassis" : "vm",
        "ChassisAssetTag" : null,
        "Deployment" : null,
        "Location" : null,
        "KernelName" : "Linux",
        "KernelRelease" : "7.0.0-30-generic",
        "KernelVersion" : "#30-Ubuntu SMP PREEMPT_DYNAMIC Fri Jul 31 18:22:54 UTC 2026",
        "OperatingSystemPrettyName" : "Ubuntu 26.04.1 LTS",
        "OperatingSystemCPEName" : null,
        "OperatingSystemHomeURL" : "https://www.ubuntu.com/",
        "OperatingSystemReleaseData" : [
                "PRETTY_NAME=Ubuntu 26.04.1 LTS",
                "NAME=Ubuntu",
                "VERSION_ID=26.04",
                "VERSION=26.04.1 LTS (Resolute Raccoon)",
                "VERSION_CODENAME=resolute",
                "ID=ubuntu",
                "ID_LIKE=debian",
                "HOME_URL=https://www.ubuntu.com/",
                "SUPPORT_URL=https://help.ubuntu.com/",
                "BUG_REPORT_URL=https://bugs.launchpad.net/ubuntu/",
                "PRIVACY_POLICY_URL=https://www.ubuntu.com/legal/terms-and-policies/privacy-policy",
                "UBUNTU_CODENAME=resolute",
                "LOGO=ubuntu-logo"
        ],
}
```

# How to reproduce

Tested `weed` versions: `4.44`, `master` (`60893c5ef32bb56247e88ad86a187bc1a948e99c`)

```c
#include <errno.h>
#include <fcntl.h>
#include <stdio.h>
#include <string.h>
#include <sys/stat.h>
#include <sys/types.h>
#include <unistd.h>

int main() {
  int status;
  int fd;

  status = creat("file", S_IRWXU | S_IRWXG | S_IROTH | S_IXOTH);
  printf("CREAT: %d\n", status);
  close(status);

  // bug does not occur with O_RDWR
  status = open("file", O_WRONLY);
  printf("OPEN: %d\n", status);
  fd = status;

  status = write(fd, "Test data!", 10);
  printf("WRITE: %d\n", status);

  // offset+len should be less than write size
  status = posix_fallocate(fd, 0, 1);
  printf("FALLOCATE: %d (%s)\n", status, strerror(status));
}

// ::Expected::
// CREAT: 3
// OPEN: 3
// WRITE: 10
// FALLOCATE: 0 (Success)

// ::Actual::
// CREAT: 3
// OPEN: 3
// WRITE: 10
// FALLOCATE: 9 (Bad file descriptor)
```

Steps:

- Compile the script
- `mkdir /data`
- `weed mini -dir=/data`
- `mkdir /mnt/fstest`
- `weed mount -filer=localhost:8888 -dir=/mnt/fstest`
- `cd /mnt/fstest`
- Run compiled script
- Observe that last operation fails

chrislusf:
Fixes #11013.

`posix_fallocate` on a write-only descriptor failed with `EBADF` whenever the range was already inside the file:

```
CREAT: 3
OPEN: 3
WRITE: 10
FALLOCATE: 9 (Bad file descriptor)
```

`Fallocate` answered `ENOSYS`, which makes the kernel set `no_fallocate` on the connection and return `EOPNOTSUPP` for every mode from then on. glibc falls back to its own emulation, and that fallback `pread`s one byte per block for the part of the range already inside the file, to skip blocks that are allocated. On an `O_WRONLY` descriptor the VFS rejects that read with `EBADF`, which `posix_fallocate` returns. `O_RDWR` avoids it because the `pread` succeeds.

Volume space is assigned when a write is flushed, so there is nothing to reserve up front. The implementation is therefore the size half of `fallocate`:

- a range inside the file is answered OK, untouched
- a range past the end grows the file the way a truncate up would, and the tail reads back as zeros
- `FALLOC_FL_KEEP_SIZE` is a no-op
- any other mode is refused with `ENOTSUP` rather than `ENOSYS`, so the kernel keeps forwarding the modes that do work
- quota and WORM are enforced as on the other size-changing paths

Verified against the reporter's program on a Linux mount, plus the surrounding cases:

```
--- issue repro ---
CREAT: 3
OPEN: 3
WRITE: 10
FALLOCATE: 0 (Success)
--- extended ---
FALLOCATE inside: 0 (Success)          size=10
FALLOCATE grow: 0 (Success)            size=4096
FALLOCATE keep-size: 0                 size=4096
FALLOCATE punch-hole: -1 (Operation not supported)
FALLOCATE grow again: 0 (Success)      size=8192
REOPEN size=8192
READ 16: 54 65 73 74 20 64 61 74 61 21 00 00 00 00 00 00
```

The grow after the refused punch-hole is the check that `ENOTSUP` did not disable `fallocate` for the whole connection.

https://claude.ai/code/session_01XG6sAAdkqTwWffqK5h4rH9

## Summary by CodeRabbit

* **New Features**
  * Added file space allocation support for mounted filesystems.
  * Files can grow when allocation extends beyond their current size.
  * Supports preserving the existing file size with the keep-size option.
  * Unsupported allocation modes are rejected appropriately.
  * No-op allocations avoid unnecessary quota and protection checks.

* **Bug Fixes**
  * Unknown file handles now return an appropriate file-descriptor error.
  * File-size increases through attribute updates are now included in quota accounting.


chrislusf:
On the two points in the latest bot pass, both are deliberate:

**Quota.** SeaweedFS assigns volume space when a write is flushed, so nothing can be reserved ahead of a write — the alternative to honoring fallocate approximately is refusing it, which is the bug. `truncate` up has always grown a file the same way. Deliberately not touching `AddUncommittedBytes` either: that counter is drawn down by `SubtractUncommittedBytes` with the byte counts a flush actually wrote, so charging it for a hole nothing will ever flush would leak until the next periodic quota refresh. The over-quota check that does apply is the one on growth, which is where the later writes get refused.

**Locking.** The handler holds the handle's exclusive `fhLockTable` lock across the entry read and the mutation, which is exactly what `Write` takes — `Attributes.FileSize` and the times are the handle lock's to write, per the note in `GetAttr`. `SetAttr` takes the extra `LockedEntry` lock because it rewrites `Chunks`, which async upload workers append to; this path never touches them. `installAckedEntry` takes the same handle lock, so the entry pointer cannot be swapped between the read and the mutation.

WORM and the quota check now sit after the no-op decision, from the earlier round.

chrislusf:
Correcting my earlier comment: I said the uncommitted counter is drawn down by `SubtractUncommittedBytes` with the byte counts a flush wrote. That is wrong — `SubtractUncommittedBytes` has no callers outside its own unit test. The counter is only ever raised by `Write` and reset wholesale by the periodic quota check, so charging it for a growth is exactly as durable as charging it for a write, and my reason for not charging did not hold.

The half of the review that follows from that is right, and is now fixed in `655f4e5`. `Write` charges the counter by how much the file *grew*, so once fallocate has extended the file the writes that fill that range charge nothing, and the real-time guard never sees the data at all — only the periodic filer refresh does. The growth is now counted where it happens, with a test that fallocates past the end, checks the delta, and checks a second in-range request adds nothing.

Reserving actual storage is still not something this can do: volume space is assigned when a write is flushed, so there is nothing to reserve ahead of one short of writing the zeros. The `posix_fallocate` guarantee is approximated, not met — same as `truncate` up, which has grown a file this way all along. Refusing fallocate instead is what produced the reported `EBADF`.

Note `SetAttr`'s truncate-up path has the same accounting gap; it predates this PR, so I have left it alone rather than widen the change. Happy to fix it in a follow-up.

PR #11036, opened 2026-08-31
get volume topology info without volume details

### LINKED ISSUE #8882: [Feature Request] New master API: Get TopologyInfo
We need to get the addresses of all volume servers. Currently, the only way to do this is through the VolumeList API, but when the number of volumes is very large, this API returns a lot of data. Could you provide a new API that returns TopologyInfo without including specific volume information?

27149chen:
fix https://github.com/seaweedfs/seaweedfs/issues/8882

add a new field `topology_only` in `VolumeListRequest`

# How is the PR tested?

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added an option to retrieve volume topology—including disks and counters—without individual volumes or erasure-coded shards.
  * Reduced response data when only topology information is needed.

* **Bug Fixes**
  * Requests combining topology-only mode with volume selectors are now rejected with a clear error instead of being partially applied.

### DIFF
--- a/seaweed-volume/proto/master.proto
+++ b/seaweed-volume/proto/master.proto
@@ -418,8 +418,10 @@ message VolumeListRequest {
   // Empty and zero take everything. Wildcards are supported.
   string remote_storage_name = 4;
   bool local_volume_only = 5;
-  // Set to true when you only want the topology, not the volumes
-  bool without_volumes = 6;
+  // The topology, its disks and their counters alone, without the volumes
+  // and ec shards. Selecting volumes above contradicts this and is refused.
+  // A master that predates this field ignores it and answers in full.
+  bool topology_only = 6;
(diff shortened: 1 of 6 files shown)


chrislusf:
I pushed two commits onto your branch. The field is now `topology_only`: it shapes the reply rather than selecting volumes, and it also leaves out the ec shards, which `without_volumes` denied. It matches the message's existing `*_only` naming. Combining it with any selector (collection, volume_ids, default_collection_only, remote_storage_name, local_volume_only) is refused with InvalidArgument instead of silently answering nothing, on both VolumeList and VolumeListStream, with tests for both behaviors. The proto comment also notes that an older master ignores the field and answers in full, so a client should tolerate a full reply.

PR #11055, opened 2026-08-31
volume: open volume files with O_NOATIME

chrislusf:
Nothing in the volume server reads the atime of .dat, .idx, .sdx, or the EC files, but every needle read still dirtied the inode: even relatime writes atime on the first read after each write, so an actively written volume paid a metadata write per read/write cycle, and strictatime mounts paid one per read.

Open the long-lived serving handles with O_NOATIME on Linux — .dat load/create, .idx load and reopen, the read-only-volume .idx/.sdx descriptor pool, the integrity-check open, and .ecx/.ecj/.ecd — falling back to a plain open on EPERM, since the flag is refused unless the process owns the file or holds CAP_FOWNER. Other platforms have no per-descriptor equivalent and keep a plain open.

Left alone on purpose: goleveldb opens its .ldb files internally, and one-pass maintenance scans (vacuum, backup, bitrot) cost at most one atime write per file per run.

The second commit mirrors the same change into the Rust volume server.

https://claude.ai/code/session_015uVY4diBgEn3VYQoc2eMuD

## Summary by CodeRabbit

* **Performance**
  * Reduced access-time metadata updates when opening volume, index, journal, and erasure-coding files on supported systems.
  * Added a safe fallback when access-time suppression is unavailable.

* **Compatibility**
  * Preserved existing file access modes, creation behavior, and error handling across supported platforms.
  * Applied consistent file-opening behavior across volume storage, indexing, erasure-coding, and local data workflows.

### DIFF
--- a/seaweed-volume/src/storage/erasure_coding/ec_shard.rs
+++ b/seaweed-volume/src/storage/erasure_coding/ec_shard.rs
@@ -4,6 +4,7 @@ use std::fs::{self, File, OpenOptions};
 use std::io::{self, Write};
 
 use crate::storage::types::*;
+use crate::storage::volume_open::open_volume_file;
 
 pub const DATA_SHARDS_COUNT: usize = 10;
 pub const PARITY_SHARDS_COUNT: usize = 4;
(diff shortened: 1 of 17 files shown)


chrislusf:
Routed the tier-down .dat swap through the no-atime open on both sides, and the zero-length read-only .dat recreate in maybeWriteSuperBlock — all three install long-lived serving handles. The remaining direct opens (volume copy tail check, vacuum, save-to-idx) are one-pass and stay plain.

PR #11071, opened 2026-09-01
volume: an already-deleted EC needle is not a delete failure

chrislusf:
Deleting a needle that is already gone is what the caller asked for. The non-EC paths have always treated it that way; the EC paths report it as a server error.

| Path | non-EC, already deleted | EC, already deleted |
| --- | --- | --- |
| gRPC `BatchDelete` | `304 Not Modified` (`DeleteVolumeNeedle` returns `(0, nil)`) | **`500`** + `"already deleted"` |
| HTTP `DeleteHandler` | `404` from the `ReadVolumeNeedle` pre-check | **`500`**, and `VolumeServerFileWriteFailures` incremented |

`DeleteEcShardNeedle` reads the needle before deleting it, so a repeat delete surfaces `storage.ErrorDeleted`, and neither EC branch had a case for it — it fell through to the generic failure arm. On the HTTP side that also books a no-op as a write failure, inflating the metric on any replayed or duplicated delete.

It is reachable on the gRPC side when `SkipCookieCheck` is set (otherwise the pre-check answers 404 first), and on the HTTP side whenever a delete is retried against an EC volume.

### Why this has stayed quiet

The filer tolerates it already, by string-matching the error text:

```go
if strings.Contains(res.Error, storage.ErrorDeleted.Error()) || res.Error == "not found" {
    continue
}
```

That works, but it makes an error *message* load-bearing — change the wording and the protection disappears silently. This fixes the status at the source; the string match becomes redundant rather than essential.

Note the two branches also disagree on the resulting filer outcome today: non-EC already-deleted counts as `deletionOutcomeSuccess` (empty `Error`), EC as `deletionOutcomeNotFound` (skipped by the string match). Both remove the item from the retry queue and neither retries, so aligning them changes a counter and a log line, not behaviour.

### Tests

Two, one per dispatcher, each pinning the contract rather than the status constant:

- `TestBatchDelete_AlreadyDeletedEcNeedleIsNotAnError`
- `TestDeleteHandler_AlreadyDeletedEcNeedleIsNotAWriteFailure` — also asserts the failure counter does not move

Both mutation-checked. Restoring the old mapping fails them with `status = 500 (error "already deleted")` and `status = 500 ({"error":"Deletion Failed: already deleted"})` respectively.

The fixture mounts an EC volume with one deleted needle. Worth noting for anyone extending it: a runtime EC delete is journalled to `.ecj` and read back into an in-memory set at load — it does *not* rewrite the sealed `.ecx`. A fixture that tombstones the `.ecx` size instead produces `needle not found`, a different error that would pass against a broken fix.

`weed/server` suite green; `go vet` clean.

https://claude.ai/code/session_01P3pE6J2UPFp6G3ksfMV4s1

### DIFF
--- a/weed/server/volume_grpc_batch_delete.go
+++ b/weed/server/volume_grpc_batch_delete.go
@@ -2,11 +2,13 @@ package weed_server
 
 import (
 	"context"
+	"errors"
 	"net/http"
 	"time"
 
 	"github.com/seaweedfs/seaweedfs/weed/operation"
 	"github.com/seaweedfs/seaweedfs/weed/pb/volume_server_pb"
+	"github.com/seaweedfs/seaweedfs/weed/storage"
 	"github.com/seaweedfs/seaweedfs/weed/storage/needle"
 )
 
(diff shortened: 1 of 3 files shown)


chrislusf:
Valid — fixed in 76223613a.

`DiskLocation.Close` does two things the fixture was skipping: it closes the mounted EC volume (releasing its `.ecx`/`.ecj`/shard handles) and `close(l.closeCh)`, which stops the disk-space goroutine `NewDiskLocation` starts. Both otherwise live until the test binary exits.

Registered with `t.Cleanup` immediately after `NewDiskLocation` rather than after the mount, so it also covers a `LoadEcShard` failure. Both tests still pass under `-race`.

PR #11079, opened 2026-09-02
fix(mount): make a concurrent duplicate mkdir fail with EEXIST instead of both succeeding

kisow:

The mount's `Mkdir` sends `CreateEntryRequest` without `OExcl`, 
so the filer treats a concurrent duplicate as an update and both callers get rc=0 — `mkdir(2)` must fail with `EEXIST`. 

Setting `OExcl` alone is not enough with multiple filers: the per-path lock (#9639) is filer-local and the store insert is an upsert, so two mounts streaming to different filers both pass the existence check. 

Measured 18/30 both-success on a 3-filer cluster under a barrier-synchronized cross-node race; 
the losers also saw `EIO` because the stream client collapses a failed create's structured error code into a generic errno.

Set `OExcl` on the mkdir create and map `ErrEntryAlreadyExists` to `EEXIST`. 
Route exclusive creates to a deterministic owner filer — fnv32a of the path over the sorted filer list — the routing contract the per-path lock's comment already assumes; the mutation stream is kept when it already targets the owner, and used as fallback when the owner is unreachable. 

Let `doUnary` hand failed creates to `CreateEntry` so the structured error code survives. 

# How is the PR tested?

`go test ./weed/mount/...` passes. On a live 3-filer cluster, 30 barrier-synchronized cross-node mkdir races of the same fresh path: before, 18/30 ended with both succeeding and the losers got `EIO`; after, 30/30 had exactly one winner and every loser got `EEXIST`. Deployed mount-only against an unmodified 4.40 server — the server prerequisites (#9639 lock, #8767 error codes) are long present.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed directory creation to reliably report an “already exists” error when the target path is duplicated.
  - Improved handling of stream-based creation failures, preserving specific existing-path errors while returning appropriate I/O errors for other failures.
  - Prevented crashes and false success results when creation responses are incomplete or contain unexplained errors.
  - Refreshed parent directory listings after duplicate creation attempts to prevent stale cached results.
- **Tests**
  - Added coverage for duplicate paths, generic failures, missing responses, and unexplained errors.


chrislusf:
This is a feasible approach. However, only Mkdir routes. Create/Mknod/Symlink still stream to the current filer without OExcl, and S3/HTTP creators hit whatever filer they land on. May be fine as a first step. 

This client-side owner-filer routing reinvents server-side route-by-key approach. It is better to do this from the server side.

I would need to create a few PRs to replace this PR. But thanks for the ideas!

chrislusf:
Thanks for this — the diagnosis is right and the 18/30 measurement on a real 3-filer cluster is what made it easy to act on. Two of the three pieces here are keepers. The routing piece I would like to take server-side instead, and I have opened #11108 and #11109 to do that.

## Why not route from the mount

Two reasons, neither obvious from where you were standing.

The ring already exists, and its membership comes from the master as a versioned `LockRingUpdate` (`weed/filer/filer.go:158-164`). `ObjectTransaction` already routes on it with one-hop `is_moved` forwarding (`weed/server/filer_grpc_server.go:284`), and the per-path lock comment you quoted is describing that contract. A mount's `-filer` list is a different, static view, so it picks a different owner than the ring does for the same path — and it only binds clients that opted in. An S3 PUT, a `curl` to the filer, or the Java client will never consult it, and mixed workloads are exactly what multi-filer deployments have.

The other reason is availability. As written, an unreachable owner fails the create outright, so with three filers and one down roughly a third of new mkdirs return EIO while the tree is otherwise fully writable — `mkdir -p` in a build starts failing. Server-side you do not have to make that trade: the master drops the dead filer and the ring reassigns.

I also went looking at whether the store could just do this, since every shared backend has an atomic insert and discards it — `AbstractSqlStore.InsertEntry` catches the duplicate-key error on `PRIMARY KEY (dirhash, name)` and downgrades to UPDATE, `MongodbStore.InsertEntry` is literally `UpdateEntry`, etcd does a bare `Put`. That would work for shared stores and do nothing for leveldb/rocksdb filers, which hold a private store each and cluster through metadata-log replication. Routing covers those too: the owner's own store becomes the arbiter, and its own writes are immediately visible to it.

## What to keep

Once #11109 lands, this PR rebases down to the parts I would have asked for anyway:

- `OExcl: true` on the mkdir create.
- `ErrEntryAlreadyExists` to `EEXIST`, plus the parent children-cache invalidation. That invalidation is a genuinely nice catch — without it the loser's next lookup trusts a children cache that never saw the winner.
- The `doUnary` change. I traced it against the server and it is correct: on a create failure the filer sets both `out.Error` and the nested `CreateResponse.ErrorCode`, and when `fs.CreateEntry` returns a Go error the nested response still carries `Error` with `ErrorCode == OK`, so nothing is swallowed.

`weedfs_exclusive_create.go` and its tests can go.

Three notes on the parts you keep:

1. The `doUnary` fix is the piece that actually turns EIO into EEXIST, and it has no test — while the three new tests all assert the same HRW property. A small table test over `doUnary`'s response handling would be the highest-value test in the PR.
2. The negated condition nested inside `if resp.Error != ""` reads backwards. A `hasCreateResponse(resp)` helper would make the intent obvious.
3. Worth extending the same two lines to `Create`, `Mknod` and `Symlink` for callers that passed `O_EXCL`. That is the case that bites most often in the field — `open(O_CREAT|O_EXCL)` is the lock-file idiom — and it is untouched today.

One hazard to be aware of rather than fix here: if a create lands and the connection then breaks, the retry returns EEXIST to the caller that actually won. Rare, and it needs the filer to compare the stamped `Attributes.Inode` to resolve properly, so it is a separate change.

chrislusf:
I pushed two commits to your branch rather than opening a competing PR — your commits and authorship are untouched, this is a fast-forward on top.

The server-side half landed while this was open: #11108 and #11109 are merged. The filer now resolves an `OExcl` create's ring owner and forwards one hop, so a single filer's per-path lock arbitrates every creator in the cluster. That covers callers who never opted into anything — S3, the filer's own HTTP surface, the Java client — not just mounts.

So `weedfs_exclusive_create.go` is redundant and I removed it. Two reasons it was worth moving rather than keeping: the ring it hashed was the mount's configured `-filer` list, which names a different owner than the master-maintained ring the filer consults, so the two would disagree; and it failed the mkdir outright when its chosen owner was unreachable, where the master's ring reassigns instead. `Mkdir` goes back to `streamCreateEntry`.

I also added the test I mentioned. The `doUnary` change is what actually turns a lost race from EIO into EEXIST and it had no coverage, so I split the two steps into `hasCreateResponse` and `createEntryFromResponse` — testable without a live stream — and read the guard the positive way round instead of negating a type assertion inline.

What survives is your diagnosis and the three things that close the issue: `OExcl`, the `ErrEntryAlreadyExists` to EEXIST mapping, and the parent children-cache invalidation. The PR is now three files. The invalidation in particular was a good catch — without it the loser's next lookup trusts a children cache that never saw the winner.

If you can re-run your 30-race barrier test on a 3-filer cluster against this plus current master, that would be the confirmation worth having. I verified the mount package and the routing on the server side, but I have not reproduced your cross-node race end to end.

One follow-up I deliberately left out to keep this focused on the reported bug: `Create`, `Mknod` and `Symlink` still create without `OExcl`, so `open(O_CREAT|O_EXCL)` — the lock-file idiom — has the same hole. Same two lines each, worth its own PR.

chrislusf:
Addressed the two findings from this round.

## Reject malformed create replies (coderabbit) — valid, fixed

Correct on both halves, and the first is a pre-existing hole this PR moved rather than introduced.

`createEntryFromResponse` read `cr.ErrorCode` without checking the nested response was there. Nothing our filer sends is shaped that way — `handleStreamMutateRequest` always populates it — but the mount is reading off the wire, and a nil there panics the whole mount. Now reported as EIO instead.

The second half is subtler and real: a top-level error whose nested response carried no code fell through every check and returned `cr` with a nil error, so the create read as **success** and the failure vanished. Now the top-level errno is used when the nested response explains nothing.

Both covered: `TestStreamMutateCreateEntryRejectsMissingNestedResponse` and `TestStreamMutateCreateEntryKeepsUnexplainedTopLevelError`.

## Cross-filer mkdir remains unserialized (greptile) — no longer true

This was accurate when written and is not any more. #11109 merged: `FilerServer.CreateEntry` now resolves an `OExcl` create's ring owner and forwards one hop (`filer_grpc_server.go:208`), so both racers converge on one filer's per-path lock regardless of which filer they hit.

The bot was right to flag it against what it could see, though — this branch was based on a master that predated #11109, so neither the reviewers nor CI were evaluating the combination that actually ships. I merged master in (a merge, not a rebase, so the commit history stays intact) and the checks now run against the real thing.

## State

The PR is three files now:

```
weed/mount/weedfs_dir_mkrm.go                 |  17 ++
weed/mount/weedfs_stream_mutate.go            |  30 +-
weed/mount/weedfs_stream_mutate_error_test.go | 136 ++++++++
```

Mount and server packages pass. Still worth having: a re-run of the 30-race barrier test on a 3-filer cluster against this, which is the one thing I cannot verify from here.

Issue #6298, opened 2024-11-28
filer read does not consider the volume replicas info related to remote storage tier


giftz:
**Describe the bug**
Consider below scenario:
- A remote storage tier like s3 is connected to local seaweed cluster
- the s3 bucket is in deep archival mode, which means the retrieval of data from this bucket may take minutes
- seaweed replica mode is set to 001 or 010 to replicate across volume servers or nodes
- after data written to local volume dat file, there will be an external logic to control the data tiering process, which will upload both copies of volume dat to remote s3, then the local volume dat is removed

when there is file read requirement from seaweed client (fuse mount) later, the storage tier logic will try to download one copy of volume dat from remote tier, however, we found that at filer layer, filer read will always get all volume urls, and then read from those urls in order (if first complete, then return).

From current implementation, there is no information for filer layer to understand that, down at volume layer, there is one copy available and another copy is still in remote,  which basically means, even seaweed has storage tier feature, not every component in the system is aware of that.

**System Setup**
- List the command line to start "weed master", "weed volume", "weed filer", "weed s3", "weed mount".
- OS version
- output of `weed version`
- if using filer, show the content of `filer.toml`

Same as in https://github.com/seaweedfs/seaweedfs/issues/6181

**Expected behavior**
filer layer should consider the fact that the volume data could be in remote storage tier, and in multi replica mode, it should try to read from local copy first.

**Screenshots**

**Additional context**
Looks like filer is using `wdclient::GetLocations` to acquire the volume locations, and wdclient is using `KeepConnected` API to get location info from master, however, the `VolumeLocation` returned from `KeepConnected` API does not contain volume info, so filer layer can not have detail volume info including volume tiering state.

Can we add volume info to  `KeepConnected` API so filer layer can get the detail volume info?

PR #11086, opened 2026-09-02
expose whether a volume replica is backed by remote storage, and prefer local replicas

giftz:
#6298 


When a volume has been tiered to remote storage, lookups return its location just like any other, with no indication that the replica is remote-backed. Readers therefore cannot tell a local replica from a remote one and may hammer a remote-backed volume server first — even when a local replica is available and would answer cheaply. 
                                                                                                       
                                                                                                                                              
This PR splits the fix into two commits:  
                                                                                                                                          
1. Add `DataInRemote` to `LookupLocation`, propagate it from the master's volume info, and track it through `wdclient`'s vid map.                                                                           
    The field is opt-in; readers that do not care about it ignore it.                                                                      
2. In `wdclient`, when returning locations for a `vid`, sort local replicas ahead of remote-tier replicas so callers prefer cheap,                                                                          
     local reads. 


# How is the PR tested?
we've tested on a k8s 3 node setup with replica 001, when one volume copy is local, the other is remote (with minutes latency), the read will return immediately.


# Checks

# Checks for AI generated PRs


<!-- This is an auto-generated comment: release notes by coderabbit.ai -->
## Summary by CodeRabbit

* **New Features**
  * Added visibility into whether file and volume locations reside on remote storage.
  * Added reporting for remote volume identifiers and log-buffer flush progress.
* **Improvements**
  * Read requests now prioritize local storage across data centers, using remote replicas as fallback.
  * Remote-only volumes remain supported when no local replica is available.
  * Replica routing updates correctly when storage transitions between local and remote tiers.
* **Tests**
  * Added coverage for local-first routing, remote-only fallback, and storage-tier transitions.
<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/other/java/client/src/main/proto/filer.proto
+++ b/other/java/client/src/main/proto/filer.proto
@@ -576,6 +576,7 @@ message Location {
     string public_url = 2;
     uint32 grpc_port = 3;
     string data_center = 4;
+    bool data_in_remote = 5;
 }
 message LookupVolumeResponse {
     map<string, Locations> locations_map = 1;
@@ -662,6 +663,7 @@ message SubscribeMetadataResponse {
     int64 ts_ns = 3;
     repeated SubscribeMetadataResponse events = 4; // batch of additional events (backlog catch-up)
     repeated LogFileChunkRef log_file_refs = 5;    // log file chunk refs for client direct-read
+    int64 flushed_ts_ns = 6; // local log-buffer flush watermark: everything at or below it is on disk
 }
 message ListMetadataSubscribersRequest {
     repeated string client_types = 1; // optional filter by client type, e.g. "mount"; empty = all
--- a/seaweed-volume/proto/master.proto
+++ b/seaweed-volume/proto/master.proto
@@ -222,6 +222,7 @@ message VolumeLocation {
   uint32 grpc_port = 7;
   repeated uint32 new_ec_vids = 8;
   repeated uint32 deleted_ec_vids = 9;
+  repeated uint32 remote_vids = 10;
 }
 
 message ClusterNodeUpdate {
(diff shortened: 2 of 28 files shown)


chrislusf:
Thanks for this — the tier-transition work on both the delta and full-reconciliation paths is the right shape.

The review turned up a few things I couldn't push here: your fork is organization-owned with maintainer edits off, so I carried your six commits unchanged into #11105 and put the fixes on top. Authorship is preserved.

What changed:

- `findVolumeLocation` dropped a location whenever `GetVolumesById` failed. The nodes answering for an EC volume hold shards, not a volume record, so that emptied the result and 404'd every EC read through the master's HTTP lookup and fid redirect.
- `addLocationToMap` wrote the flipped replica into `entry.locations[i]` in place. `GetLocations` hands out that slice and the caller walks it after the lock is dropped, so this raced `LookupVolumeServerUrl` under `-race`. Copy-on-write now, matching the rest of the file.
- Moving remote vids out of `NewVids` breaks rolling upgrades: a master upgraded ahead of its filers announces a tiered volume only on a field the old client ignores, and the volume disappears from its vid map. Remote vids now go on both lists.
- `ToVolumeLocations` runs on every `KeepConnected`; the tier split made it copy a whole `VolumeInfo` per volume per node. Extended `AppendVolumeIds` to fill the remote list instead.
- Reverted the cross-DC part of `wdclient: prefer local replicas across data-center boundaries`. When the remote tier is in the same region as the replicas, preferring an other-DC local read trades an in-region GET for a WAN round trip. Local now wins inside each DC bucket, DC preference wins overall. Happy to be argued out of this one if your deployment says otherwise.

Please take a look at #11105 — comments there are welcome, and I'll close this once it lands.

https://claude.ai/code/session_01FcSp6quCaxQ1cb3Vx7o9fW

Issue #11087, opened 2026-09-02
filer: fetchWholeChunk does not invalidate stale volume locations


giftz:
### Problem

When a file with chunk manifests is read through the FUSEmount (or any other path that goes through `ChunkGroup.SetChunks`), manifest resolution goes through `filer.fetchWholeChunk`. If the  cached volume locations for the manifest chunk are stale — because the volume has been tiered to remote storage, a volume server has been rolled, or a replica layout has changed — every attempt to read fails with the same stale URL.
The mount retries the same URL across its backoff ladder (observed:6 attempts, ~14s total) and only then surfaces a permanent failure, even when the master already lists a healthy location for the volume.

### Root cause
                                                                                                                                            
`RetriedFetchChunkData` and `retriedFetchChunkDataDirect` were updated by #10156/#10800 to accept a `RefreshUrlsFunc` and re-lookup locations when every known URL fails. `filer.fetchWholeChunk`, however, was missed: it streams via `retriedStreamFetchChunkData`, which has no equivalent invalidation hook, and `fetchWholeChunk` itself returns the raw fetch error without consulting the chunk cache invalidator that the `ChunkGroup` already carries for `ReaderCache`.

PR #11088, opened 2026-09-02
filer: self-heal fetchWholeChunk on stale volume locations

giftz:
#11087 

PR #10156 and #10800 wired cache invalidation into the buffer-based read paths, but chunk-manifest resolution still goes through `fetchWholeChunk`, which returns the raw fetch error on failure. When the cached volume locations are stale — a volume has been tiered to remote storage, a server has rolled, or a replica layout has changed — resolving a large multipart file fails permanently even though a healthy location exists on the master.

The mount hits this path whenever it opens a multipart file with chunk manifests (via `ChunkGroup.SetChunks` → `ResolveOneChunkManifest`→ `fetchWholeChunk`). Every retry against the same stale URL just fails again.

This change threads the `ChunkGroup`'s `cacheInvalidator` through `ResolveChunkManifest` / `ResolveOneChunkManifest` / `fetchWholeChunk` and, on failure, invalidates the cached volume locations, re-looksthem up, and retries once via the existing `retryFetchWithFreshLocations` helper (from #10800). The retry only happens when the fresh locations actually differ from the stale ones, so we never loop against the same broken servers. The streaming `bytesBuffer` is reset before the retry so partial bytes from the failed attempt cannot corrupt the downstream `proto.Unmarshal` of `FileChunkManifest`.

Non-mount callers (`filer` deletion, `s3api`, `shell` fsck/verify/ distribute, replication sinks) pass `nil` for the new invalidator parameter and keep their existing semantics.

# How is the PR tested?
Tests:
- `weed/filer/fetch_whole_chunk_selfheal_test.go`:
- `TestFetchWholeChunkSelfHealOnStaleLocation` — first fetch returns 500 and may stream a partial prefix;
   verifies the invalidator is called, the lookup is re-issued, and the buffer contains only the fresh bytes (i.e. `Reset()` happened).
- `TestFetchWholeChunkNoInvalidatorSkipsRetry` — without an invalidator, the original error surfaces and no re-lookup is issued.
- `TestFetchWholeChunkSameUrlsSkipsRetry` — invalidation that yields the same URL set does not retry, to avoid loops.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file and chunk reads when cached volume locations are stale.
  * Automatically refreshes stale locations and retries failed reads once.
  * Resets buffered data before retrying to prevent corrupted or duplicated output.
  * Prevents repeated retries when refreshed locations remain unchanged.

* **Tests**
  * Added coverage for successful recovery, unavailable cache invalidation, and unchanged stale locations.

### DIFF
--- a/weed/filer/fetch_whole_chunk_selfheal_test.go
+++ b/weed/filer/fetch_whole_chunk_selfheal_test.go
@@ -1,0 +1,184 @@
+package filer
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+	"sync/atomic"
+	"testing"
+
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+
+	"google.golang.org/protobuf/proto"
+)
+
+// staleTrackingInvalidator records invalidation calls and lets the test
+// observe whether fetchWholeChunk triggered a cache drop after a fetch failure.
+type staleTrackingInvalidator struct {
+	invalidations atomic.Int32
+}
+
+func (inv *staleTrackingInvalidator) InvalidateCache(fileId string) {
+	inv.invalidations.Add(1)
+}
+
+// twoStageLookupFn returns staleUrls on the first call and freshUrls on every
+// subsequent call. Mirrors what a vidMapClient does after InvalidateCache: the
+// next LookupVolumeIdsWithFallback queries master and returns current locations.
+type twoStageLookupFn struct {
+	staleUrls []string
+	freshUrls []string
+	calls     atomic.Int32
+}
+
+func (l *twoStageLookupFn) lookup(ctx context.Context, fileId string) ([]string, error) {
+	n := l.calls.Add(1)
+	if n == 1 {
+		return l.staleUrls, nil
+	}
+	return l.freshUrls, nil
+}
+
+// TestFetchWholeChunkSelfHealOnStaleLocation verifies that when the first
+// retriedStreamFetchChunkData attempt fails (stale volume location returning
+// HTTP 500), fetchWholeChunk invalidates the cache, re-looks up, and retries
+// against the fresh location. Without this, mount reads of large multipart
+// files whose manifest chunk's volume has moved would permanently fail.
+func TestFetchWholeChunkSelfHealOnStaleLocation(t *testing.T) {
+	// Build a valid FileChunkManifest protobuf for the fresh response.
+	manifest := &filer_pb.FileChunkManifest{
+		Chunks: []*filer_pb.FileChunk{
+			{FileId: "100,abc", Offset: 0, Size: 8},
+			{FileId: "101,def", Offset: 8, Size: 8},
+		},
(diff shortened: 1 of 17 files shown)


chrislusf:
Reviewed, and the core of this is right — threading the invalidator through manifest resolution is the piece #10156/#10800 missed. I pushed three follow-up commits to `pr-11088-manifest-selfheal` on this repo since your fork is org-owned and I can't push to it:

```
git fetch origin pr-11088-manifest-selfheal
git cherry-pick 8c0bc8a96 e4fcf3166 971a0f910
```

**1. A cancelled read shouldn't cost the location cache.** `retriedStreamFetchChunkData` returns `ctx.Err()` when the caller walks away, and the new code treated that like a dead volume server: invalidate, re-lookup, and hand back `re-lookup chunk X after cache invalidation: context canceled` instead of the real cause. `peer_fetcher` passes a live request context, and FUSE cancels reads all the time. `PrepareStreamContentWithThrottler` already guards this exact case; `fetchWholeChunk` now does too.

**2. The 500 in the test never exercised the `Reset()`.** `ReadUrlAsStream` returns at `r.StatusCode >= 400`, before it ever calls the writer, so no bytes reach `bytesBuffer` on an error status — `TestFetchWholeChunkSelfHealOnStaleLocation` would have passed with the `Reset()` deleted. The `Reset()` is still needed, just for the other case: a 200 that streams a prefix and then dies mid-body. The stale server now does that (partial write, flush, `panic(http.ErrAbortHandler)`), and dropping the `Reset()` makes the test fail on `proto.Unmarshal`. Tests moved into `filechunk_manifest_test.go` alongside the rest of the manifest coverage.

**3. Most of those `nil`s had an invalidator sitting right there.** The filer's read, copy and deletion paths and the log cache all hold the `MasterClient`, and s3api holds the `FilerClient` — both implement `CacheInvalidator` via the embedded `vidMapClient` that backs the very lookup functions they pass in. As written, a stale manifest location still failed permanently on the filer and S3 read paths, which is the same bug the PR title describes. `nil` stays only for the shell tools and replication sinks, whose lookup functions cache in a private map with nothing to invalidate.

Two things I left alone, worth knowing about:

- `NonOverlappingVisibleIntervals` / `ViewFromChunks` still resolve manifests with `nil`, so S3's main range-read path and webdav don't self-heal yet. Threading it there touches the mq and query-engine callers too, and that's a bigger change than this PR should carry.
- `ChunkGroup.SetChunks` resolves with `context.Background()`, so the new cancellation guard is inert there. Pre-existing, not yours.

The red `Test` check is `TestNewLogBufferFirstBuffer` in `weed/util/log_buffer` timing out at 10m — unrelated to this PR, and it passes locally over repeated runs.

chrislusf:
Opened #11107 carrying this plus the three follow-ups, rebased on current master — your commit is preserved as the first one, so the authorship stays with you. Closing this in favor of that; thanks for the fix, the diagnosis in #11087 was spot on.

PR #11091, opened 2026-09-02
avro: field time.Time <> iceberg.date

### LINKED ISSUE #11090: avro: field data_file.partition.<col>: cannot use time.Time with Avro type int
### Environment

- SeaweedFS version: `4.44` (binary reports `version 30GB 4.44 3563738699f29fd1c9efde6fcbf4ba253439cac8 linux amd64`)
- Cluster: 1 master, 1 filer, 1 volume server, 1 admin+worker (all on the same host)
- Iceberg table created via the Doris↔SeaweedFS Iceberg REST catalog integration (Apache Doris writing directly to a SeaweedFS-hosted Iceberg catalog)
- Table partition spec: partitioned by `event_time_day`, a `date`-typed column, via the standard `day(event_time)` (or equivalent identity-on-date) transform

### What happened

The admin's scheduled `iceberg_maintenance` job (lane `iceberg`, hourly detection) runs several maintenance sub-operations per table: `compact`, `rewrite_position_delete_files`, `rewrite_manifests`. The first two occasionally hit ordinary optimistic-concurrency retries ("stale plan: table head changed since planning") when a concurrent writer commits in between planning and commit -- expected and harmless.

`rewrite_manifests`, however, fails **every single time** it runs against this table, with:

```
iceberg maintenance rewrite_manifests failed for <bucket>/<db>/<table>: write merged manifest for spec 0: avro: field data_file.partition.event_time_day: cannot use time.Time with Avro type int
avro: field data_file.partition.event_time_day: cannot use time.Time with Avro type int
```

(exact log lines, from `journalctl -u seaweed@worker.service`, timestamps redacted):

```
W<ts> handler.go:620 iceberg maintenance rewrite_manifests failed for <bucket>/<db>/telemetry_metrics: write merged manifest for spec 0: avro: field data_file.partition.event_time_day: cannot use time.Time with Avro type int
avro: field data_file.partition.event_time_day: cannot use time.Time with Avro type int
```

This repeats on every subsequent scheduled run (hourly `detection_interval_minutes`), indefinitely -- it is not transient. Because `rewrite_manifests` is one of several sub-tasks under the same `iceberg_maintenance` job, a single run of the job can take 20-30+ minutes (multiple sub-tasks, each with its own retry) and reports overall `last_run_status: "error"`, even though `compact` and `rewrite_position_delete_files` may otherwise be succeeding on later attempts.

### Root cause (best guess from the error text; have not located the exact call site)

Per the Iceberg spec, a manifest entry's `partition` struct field for a column using the `date` partition transform must be Avro-encoded as `int` (days since epoch, `logicalType: date`), not as a Go `time.Time`. The manifest-merge code path (`rewrite_manifests`) appears to read the partition value out of an existing manifest as (or convert it to) `time.Time`, then hands that `time.Time` value straight to the Avro encoder for a field whose schema type is `int` -- instead of converting it to the `int32` days-since-epoch representation the schema expects. Every retry hits the identical failure, which is consistent with a deterministic type-handling bug rather than a data or environment issue.

### Impact

- Table data itself is unaffected: hot-tier row counts and Iceberg table row counts for the affected time range match exactly after independent verification (`SELECT COUNT(*)` on both sides), so this is not causing data loss or corruption.
- `rewrite_manifests` is a read-optimization (consolidating many small manifest files into fewer larger ones) -- not required for correctness -- but the job never succeeds and never stops retrying on schedule, so:
  - Manifest file count for affected tables keeps growing unbounded over time (never consolidated), which will eventually degrade query planning performance.
  - Every scheduled run burns worker CPU/memory for the duration of the (failing) sub-task, on a cadence that repeats forever with no way to resolve it operationally.

### Reproduction

1. Create an Iceberg table via the SeaweedFS Iceberg REST catalog (or have an external engine such as Doris create one) partitioned by a `date`-transformed column.
2. Write enough data across at least two partition values that a manifest merge would have something to do.
3. Let the admin's scheduled `iceberg_maintenance` job run (or trigger manifest rewrite maintenance manually if there's a way to do so outside the scheduler).
4. Observe `rewrite_manifests` fail with the error above, every time.

### Workaround being used in the meantime

None yet identified other than tolerating the failing job (data path is unaffected). Looking for a way to disable just the `rewrite_manifests` job type for the `iceberg` lane while keeping `compact`/`rewrite_position_delete_files` enabled, if that's possible via the scheduler config -- happy to hear if there's a supported way to do that already.

Dimonyga:
Fixes #11090.

This PR fixes Iceberg manifest rewrites for tables with date/day partitions written by foreign producers. Some external manifests encode a day partition as the Avro union [date, null]. iceberg-go reads that value as time.Time, while the manifest schema expects an int, so rewrite_manifests fails when it tries to write the merged manifest.

The fix resolves the manifest partition spec before processing entries and normalizes time.Time values for day-transform fields to iceberg.Date (days since the Unix epoch) before serialization.

Regression coverage includes:
- a wire-format test for the foreign [date, null] Avro union;
- an integration test that rewrites three foreign manifests and verifies the merge succeeds.

Validation:
- go test ./weed/worker/tasks/iceberg
- go vet ./weed/worker/tasks/iceberg

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility with Iceberg manifests created by external writers, including different partition-value encodings.
  * Normalized foreign partition values for dates, timestamps, times, and UUIDs during manifest processing.
  * Applied consistent manifest handling across compaction, rewriting, and delete-file workflows.
  * Preserved processing for manifests with unknown partition specifications.

* **Tests**
  * Added coverage for foreign manifest formats, partition-value normalization, and alternate value orderings.
  * Verified normalized manifests can be written successfully.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -154,6 +154,7 @@ require (
 	github.com/tarantool/go-tarantool/v3 v3.0.1
 	github.com/testcontainers/testcontainers-go v0.43.0
 	github.com/tikv/client-go/v2 v2.0.7
+	github.com/twmb/avro v1.7.2
 	github.com/xeipuuv/gojsonschema v1.2.0
 	github.com/ydb-platform/ydb-go-sdk-auth-environ v0.5.2
 	github.com/ydb-platform/ydb-go-sdk/v3 v3.151.1
@@ -269,7 +270,6 @@ require (
 	github.com/substrait-io/substrait v0.87.0 // indirect
 	github.com/substrait-io/substrait-go/v8 v8.1.1 // indirect
 	github.com/substrait-io/substrait-protobuf/go v0.85.0 // indirect
-	github.com/twmb/avro v1.7.2 // indirect
 	github.com/twpayne/go-geom v1.6.1 // indirect
 	github.com/twpayne/go-kml/v3 v3.2.1 // indirect
 	github.com/tyler-smith/go-bip39 v1.1.0 // indirect
--- a/weed/s3api/s3tables/iceberg_manifest.go
+++ b/weed/s3api/s3tables/iceberg_manifest.go
@@ -1,0 +1,84 @@
+package s3tables
+
+import (
+	"bytes"
+	"time"
+
+	"github.com/apache/iceberg-go"
+	"github.com/google/uuid"
+)
+
+const secondsPerDay = int64(24 * 60 * 60)
+
+// ReadManifest parses an Iceberg manifest, restoring partition values that a
+// foreign writer's Avro schema hides from iceberg-go.
+//
+// iceberg-go takes a partition field's logical type from the last branch of its
+// Avro union, so a writer that spells an optional day partition [date, null]
+// rather than [null, date] leaves the value as the time.Time the Avro decoder
+// produced. Rewriting such an entry then fails, either in the manifest writer's
+// partition summaries ("expected type iceberg.Date, got time.Time") or in the
+// encoder itself ("cannot use time.Time with Avro type int"), and a time
+// partition is worse: time.Duration converts to int64 nanoseconds and silently
+// records the wrong value.
+//
+// specs maps partition spec ID to spec the way the table metadata records them.
+// Entries from a manifest written under an unknown spec are returned as read.
+func ReadManifest(m iceberg.ManifestFile, manifest []byte, discardDeleted bool, specs map[int]iceberg.PartitionSpec, schema *iceberg.Schema) ([]iceberg.ManifestEntry, error) {
+	entries, err := iceberg.ReadManifest(m, bytes.NewReader(manifest), discardDeleted)
+	if err != nil {
+		return nil, err
+	}
+	spec, found := specs[int(m.PartitionSpecID())]
+	if !found || schema == nil {
+		return entries, nil
+	}
+
+	partitionType := spec.PartitionType(schema)
+	if partitionType == nil {
+		return entries, nil
+	}
+	for _, entry := range entries {
+		partition := entry.DataFile().Partition()
+		for _, field := range partitionType.FieldList {
+			value, ok := partition[field.ID]
+			if !ok {
+				continue
+			}
+			if normalized, ok := normalizePartitionValue(value, field.Type); ok {
+				partition[field.ID] = normalized
+			}
+		}
+	}
+	return entries, nil
+}
+
+// normalizePartitionValue converts one value the Avro decoder returned for a
+// logical type to the Iceberg representation the manifest writer expects,
+// mirroring what iceberg-go does itself for the unions it does recognize.
+func normalizePartitionValue(value any, fieldType iceberg.Type) (any, bool) {
+	switch v := value.(type) {
+	case time.Time:
+		utc := v.UTC()
+		switch fieldType.(type) {
+		case iceberg.DateType, iceberg.Int32Type:
+			// A day transform reports an int32 result type and an identity
+			// transform on a date column reports date; both hold days.
+			midnight := time.Date(utc.Year(), utc.Month(), utc.Day(), 0, 0, 0, 0, time.UTC)
+			return iceberg.Date(midnight.Unix() / secondsPerDay), true
+		case iceberg.TimestampType, iceberg.TimestampTzType:
+			return iceberg.Timestamp(utc.UnixMicro()), true
+		case iceberg.TimestampNsType, iceberg.TimestampTzNsType:
+			return iceberg.TimestampNano(utc.UnixNano()), true
+		}
+	case time.Duration:
+		if _, ok := fieldType.(iceberg.TimeType); ok {
+			return iceberg.Time(v.Microseconds()), true
+		}
+	case [16]byte:
+		if _, ok := fieldType.(iceberg.UUIDType); ok {
+			return uuid.UUID(v), true
+		}
+	}
+	return nil, false
+}
--- a/weed/s3api/s3tables/iceberg_manifest.go
+++ b/weed/s3api/s3tables/iceberg_manifest.go
@@ -10,37 +10,44 @@ import (
 
 const secondsPerDay = int64(24 * 60 * 60)
 
-// ReadManifest parses an Iceberg manifest, restoring partition values that a
-// foreign writer's Avro schema hides from iceberg-go.
+// ReadManifest parses an Iceberg manifest and converts its partition values
+// while the manifest's own schema still describes them.
 //
-// iceberg-go takes a partition field's logical type from the last branch of its
-// Avro union, so a writer that spells an optional day partition [date, null]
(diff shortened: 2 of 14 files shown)


chrislusf:
Thanks — diagnosis and mechanism are right, I confirmed it in iceberg-go v0.6.0: `getFieldIDMap` reads the logical type from `typ.Branches[len(typ.Branches)-1]` only, so a `[date, null]` union never reaches `convertAvroValueToIcebergType` and the raw `time.Time` from the decoder reaches the writer.

I pushed two commits to this branch rather than asking you for another round.

**The transform is the wrong key.** Same union shape with `identity` on a `date` column still fails, just earlier — in the partition summaries rather than the encoder:

```
error updating field stats for partition 0: event_time_day: expected type iceberg.Date, got time.Time
```

Hive-style `PARTITIONED BY (dt)` is at least as common as `day()`. Identity on `timestamp` fails the same way, and a `time` partition is worse than a failure: `time.Duration` converts to `int64` nanoseconds and silently records the wrong value. Normalization now switches on the value the decoder produced, using the partition field's result type only to tell date from timestamp.

**One operation was fixed out of three.** `compact.go` and `delete_rewrite.go` read entries and write the same partition values back into new manifests, so they break on this table exactly as `rewrite_manifests` does — under the same `iceberg_maintenance` job. Both, plus the detection passes, now read through `s3tables.ReadManifest`, next to the `ReadManifestList` shim that already exists for what foreign writers put on the wire. `where` filtering was affected too: `literalMatchesActual` falls through to `fmt.Sprint`, so a `time.Time` renders as a timestamp and never matches the day you asked for.

Also in there: `go.mod` tidied (`twmb/avro` is a direct dependency now), the fixture builder moved to `s3tablestest` so the shim's own test can use it, `buildTestMetadata` extended with a schema/spec instead of the populate-then-overwrite dance, and the end-to-end test now reads the merged manifest back and checks the days — it passed before even if every partition had been written null. Coverage is table-driven over day, identity-on-date, identity-on-timestamp and identity-on-time; each case asserts the fixture still reproduces the raw decoder value, so the tests fail without the fix (with your exact error).

Two things left:

- Decimal partitions have the same hole (`*big.Rat` where `iceberg.Decimal` is expected). Rebuilding one needs a direct `arrow-go` import for `decimal128`, which isn't worth it here — it fails loudly rather than silently, so I left it.
- Could you attach the `avro.schema` header from one of the failing manifests to #11090? Java's `AvroSchemaUtil.toOption` emits null-first, so which writer produced the reversed union is still unconfirmed. The fix keys on the value, so it holds either way, but there is a worse adjacent path — partition fields without `field-id` props, which iceberg-go drops silently — that this would not touch. Worth an upstream issue too: scanning all branches for a logical type is a one-liner there and retires the shim.

Dimonyga:
Checked this against a real cluster (Doris 4.1.3-rc02, SeaweedFS Iceberg REST catalog + S3 gateway, `telemetry_logs`/`telemetry_metrics` tables, `PARTITION BY LIST (DAY(event_time)) ()` / `event_time_day` day-transform partition).

**The union order on every *input* manifest I checked is already null-first**, not reversed -- checked Doris's own native writer's first-ever manifest for `telemetry_metrics`, a second independent external writer (our own service, `iceberg` Rust crate 0.10.1), and two more Doris-written manifests for `telemetry_logs`. All four show the same shape for the partition field:

```json
"event_time_day": ["null", {"type": "int", "logicalType": "date"}]
```

**What I *did* find, reproducibly: the worker's own `compact-*.avro` output drops `logicalType:"date"` on that same field entirely**, on every compaction I checked (the first `telemetry_metrics` compaction of the day and the most recent one, ~24h apart):

```json
"event_time_day": ["null", "int"]
```

-- a bare `int`, no logical type, while every manifest that fed into those compactions carried it correctly. Notably, sibling map-typed fields in the same output schema (`column_sizes`, `value_counts`, etc.) still correctly carry their own `"logicalType": "map"` -- so this isn't the encoder losing logical types generally, it's specific to the partition sub-schema. That would explain `cannot use time.Time with Avro type int` independent of union order: the reader decodes the *input* partition value as `time.Time` (logical type present on read), then the *output* encoder for the merged manifest has no logical type on that field to encode it back into.

Happy to pull the full raw `avro.schema` header from either file (input vs. compacted output) if useful -- didn't want to dump two multi-KB schemas inline here.

chrislusf:
You're right and I was wrong — thanks for going and looking at the real files. I reproduced your shape and it fails, so the union ordering is not what breaks your table.

**The conversion is lazy, and the writer moves the goalposts before it happens.** iceberg-go converts what the Avro decoder returned on the first `Partition()` call, using the logical types it read from the manifest being parsed — but `ManifestWriter.addEntry` calls `setFieldIDToLogicalTypeMap(w.partFieldIDToType)`, rebinding them to the manifest it is *about to write*, before it makes that call. An entry that nobody looks at between reading and writing therefore converts against the output schema. Your `["null", {"type":"int","logicalType":"date"}]` reads fine, but only if something touches the partition first.

Which lands exactly on the second thing you found. `DayTransform.ResultType()` in iceberg-go returns `Int32Type`, not `DateType`, so the manifest it writes declares `["null","int"]` — I see the same in a local probe, bare `int` for the day field next to a correct `["null",{"logicalType":"date","type":"int"}]` for an identity-on-date field in the same schema. So the rebound map has no date logical type for `event_time_day`, nothing converts, and the `time.Time` reaches the encoder. Your read of the output schema was right; it is the *input* to the failure, not a symptom.

It also explains why only `rewrite_manifests` failed for you. `compactDataFiles` and `rewritePositionDeleteFiles` group entries by `partitionKey(df.Partition())` before writing, and a `where` filter calls `predicate.Matches(spec, entry.DataFile().Partition())` — both force the conversion while the manifest's own logical types are still installed. `rewriteManifests` with no `where` filter never touched a partition, which is the one path that reaches the writer unconverted.

Pushed as a third commit: `s3tables.ReadManifest` now reads every entry's partition, so the conversion happens against the manifest that is being parsed. The normalization stays for what iceberg-go does not recognize on read either — a value-first union, and `time.Duration`, where a wrong `int64` of nanoseconds is written silently instead of failing. Regression coverage now pins the mechanism directly: an entry read but never touched fails to encode, the same entry read through the shim writes, and the end-to-end merge runs over both union orders (both fail without it).

Two follow-ups I'd rather not carry in this PR:

- The bare `int` day partition is worth an upstream issue against iceberg-go. The spec's day transform produces `date`, Java and iceberg-rust write `logicalType: date`, and iceberg-go writes neither that nor a `TimestampNsType` partition stat. Readers resolve partition values through the Iceberg partition type rather than the Avro logical type, so it is not corrupting anything, but it is why our output does not round-trip against its own inputs.
- Same for `getFieldIDMap` taking the logical type from the union's last branch. One line there and this shim can go.

If you can run this branch against `telemetry_metrics`, I'd like to hear that the hourly job goes green before I merge.

PR #11108, opened 2026-09-03
filer: keep a moved key on its prior owner while the ring settles

chrislusf:
ObjectTransaction forwards to the ring owner so one filer's per-path lock arbitrates every writer of a key. But a ring change hands the key over before the new owner has rebuilt the locks the prior owner still holds, so for the length of the cooling-off window both can grant it. `LockRing.PriorOwner` exists for exactly this case and nothing consulted it.

Route to the prior owner while that window is open.

## Resolving the owner

`LockRing.WriteOwner` resolves prior-else-current under a single read lock. Two independently locked reads could otherwise pair a primary from one ring with a prior owner from another, if a `SetSnapshot` landed between them. It shares `PriorOwner`'s body through `priorOwnerLocked`; `HashRing` carries its own mutex, so the nested `Ring.GetPrimary` is safe and the lock order matches `SetSnapshot`'s.

## An unreachable owner fails, it is not re-sent

This is the part worth arguing about, so to be explicit: when the owner cannot be reached the request fails. It is never retried on another filer.

Two reasons. gRPC reports a response lost in transit as `Unavailable`, indistinguishable from a request the owner never saw, so no classification of the error can make a retry safe for a non-idempotent transaction. And unreachability is not evidence of death — an owner unreachable from here may be partitioned and still serving the key to everyone else, so re-sending elsewhere creates exactly the second serialization point this routing removes.

The cost is bounded and self-healing. While a ring change is inside the cooling-off window, keys that moved off a filer that is genuinely down fail rather than being served by their new owner; when the window closes the ring hands them over. A few seconds of `Unavailable` that a client retries through beats a silent double-apply.

## Shape

The owner resolution and the forward move into `writeOwner` / `forwardToWriteOwner` so the next routed RPC reuses them rather than copying the block. `handled=false` means this filer is the serialization point — it owns the key, or there is no ring — and the caller applies locally, exactly as before.

## Tests

`TestLockRing_WriteOwner` covers moved, stable and single-snapshot keys; the `lock_manager` package passes under `-race`. `TestForwardToWriteOwnerNeverTriesASecondFiler` asserts that exactly one filer is contacted and its failure surfaces. `go test ./weed/server/ ./weed/cluster/...` passes.

### DIFF
--- a/weed/cluster/lock_manager/lock_ring.go
+++ b/weed/cluster/lock_manager/lock_ring.go
@@ -163,6 +163,24 @@ func (r *LockRing) GetPrimary(key string) pb.ServerAddress {
 func (r *LockRing) PriorOwner(key string) pb.ServerAddress {
 	r.RLock()
 	defer r.RUnlock()
+	return r.priorOwnerLocked(key)
+}
+
+// WriteOwner returns the filer that should serialize writes to key: the prior
+// owner while a ring change is still within the cooling-off window, otherwise
+// the current primary. Both are read under one lock so the pair cannot come
+// from different rings, which could otherwise name the same filer twice.
+func (r *LockRing) WriteOwner(key string) pb.ServerAddress {
+	r.RLock()
+	defer r.RUnlock()
+	if prior := r.priorOwnerLocked(key); prior != "" {
+		return prior
+	}
+	return r.Ring.GetPrimary(key)
+}
+
+// priorOwnerLocked is PriorOwner's body; the caller holds at least RLock.
+func (r *LockRing) priorOwnerLocked(key string) pb.ServerAddress {
 	if len(r.snapshots) < 2 {
 		return ""
 	}
--- a/weed/cluster/lock_manager/lock_ring_prior_test.go
+++ b/weed/cluster/lock_manager/lock_ring_prior_test.go
@@ -61,3 +61,42 @@ func TestLockRing_PriorOwnerExpires(t *testing.T) {
 		}
 	}
 }
+
+// WriteOwner resolves prior-else-current under one read lock, so the two views
+// can never come from different rings.
+func TestLockRing_WriteOwner(t *testing.T) {
+	r := NewLockRing(5 * time.Second)
+	t.Cleanup(r.WaitForCleanup)
+
+	setA := []pb.ServerAddress{"s1:1", "s2:1", "s3:1"}
+	r.SetSnapshot(setA, 1)
+
+	// One snapshot: no prior owner, so the primary owns every key.
+	if got, want := r.WriteOwner("any"), r.GetPrimary("any"); got != want {
+		t.Fatalf("WriteOwner=%q, want the primary %q", got, want)
+	}
+
+	r.SetSnapshot([]pb.ServerAddress{"s1:1", "s2:1", "s3:1", "s4:1"}, 2)
+
+	var moved, stable string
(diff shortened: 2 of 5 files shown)


chrislusf:
Both bots landed on the same real hole, thanks — fixed in the amended push.

`ObjectTransaction` does return a non-nil Go error for a server-side failure (a store read behind the condition, for instance), and my loop treated any non-nil error as transport and moved to the next candidate. That is precisely the double-evaluation routing exists to prevent.

`forwardToWriteOwner` now advances only when the owner was never reached:

```go
func isOwnerUnreachable(err error) bool {
	if err == nil {
		return false
	}
	s, ok := status.FromError(err)
	if !ok {
		return true // dial or plumbing failure: the request never left
	}
	return s.Code() == codes.Unavailable
}
```

A dial failure comes back from `WithGrpcClient` as a plain wrapped error, so `status.FromError` reports `ok == false` and it retries. An owner that ran the handler and failed answers with a gRPC status — `codes.Unknown` for a handler error — and is surfaced as-is. `DeadlineExceeded` deliberately does not retry: the owner may well have applied the transaction, and re-running it elsewhere is worse than reporting the timeout.

Covered by `TestForwardToWriteOwnerDoesNotRetryAnApplicationError`, which asserts the second candidate is never tried.

chrislusf:
Second round addressed — and both findings pointed at the same thing, so the fallback is gone entirely.

## Unavailable can replay a completed transaction (greptile, coderabbit)

Valid, and it goes deeper than the report. gRPC surfaces a response lost in transit as `Unavailable`, indistinguishable from a request the owner never saw, so no client-side classification can make the retry safe — my `isOwnerUnreachable` was solving an unsolvable problem.

Worse, unreachability is not evidence of death. An owner I cannot reach may be **partitioned** and still serving the key to everyone else. Falling back to the current owner then creates precisely the second serialization point this PR exists to remove. Retrying somewhere else is unsound no matter how the error is classified.

So `forwardToWriteOwner` now resolves one owner and fails if it cannot be reached. That also dissolves the first round"s finding — with no fallback there is nothing for an application error to fall back *to* — and `isOwnerUnreachable` is deleted rather than fixed.

The cost is honest and bounded: while a ring change is inside the cooling-off window, keys that moved off a filer that is genuinely down will fail instead of being served by their new owner. When the window closes the ring hands them over and they work. I would rather have a few seconds of `Unavailable` that a client retries through than a silent double-apply, which is the trade this whole PR is making.

## Resolve both owners from one lock-ring read (coderabbit)

Valid. `GetPrimary` and `PriorOwner` were two independently locked reads, and a `SetSnapshot` landing between them could pair a primary from one ring with a prior owner from another.

Added `LockRing.WriteOwner`, which resolves prior-else-current under a single `RLock` and shares `PriorOwner`"s body through `priorOwnerLocked`:

```go
func (r *LockRing) WriteOwner(key string) pb.ServerAddress {
	r.RLock()
	defer r.RUnlock()
	if prior := r.priorOwnerLocked(key); prior != "" {
		return prior
	}
	return r.Ring.GetPrimary(key)
}
```

`HashRing` carries its own mutex, so the nested `r.Ring.GetPrimary` under `RLock` is safe and the lock order matches `SetSnapshot`"s. `TestLockRing_WriteOwner` covers the moved, stable and single-snapshot cases, and the package passes under `-race`.

`TestForwardToWriteOwnerNeverTriesASecondFiler` now asserts the property directly: exactly one filer is contacted, and its failure surfaces.

PR #11109, opened 2026-09-03
filer: route exclusive and conditional creates to the entry's ring owner

chrislusf:
Stacked on #11108 — review that one first.

## The problem

`CreateEntry` with `o_excl` is a FindEntry-then-Insert. The per-path lock makes that atomic only on the filer running it, and the store's insert is an upsert on every backend, so with several filers two callers both pass the existence check and both report success. `mkdir(2)` then succeeds twice for the same path, which is what #11079 reports and measures (18/30 races on a 3-filer cluster).

The same hole sits under the `condition` precondition. Its proto comment already told callers to route the key's writes to the owner filer themselves — a contract nothing enforced and most callers cannot honour.

## Why not fix it in the store

Every shared store has an atomic insert available and throws it away: `AbstractSqlStore.InsertEntry` catches the duplicate-key error on `PRIMARY KEY (dirhash, name)` and downgrades to UPDATE, `MongodbStore.InsertEntry` is literally `UpdateEntry`, etcd does a bare `Put` where a version-0 txn was available, redis3 does `SET` instead of `SET NX`. Teaching each of them an exclusive insert would work — for shared stores.

It does nothing for `leveldb`, `leveldb2`, `leveldb3` and `rocksdb`. Those filers hold a private store each and cluster through `MetaAggregator` replicating peer metadata logs, so there is no shared constraint to lean on and an exclusive insert is not merely unimplemented but unimplementable. Routing covers them: if every exclusive create for a path lands on one filer, that filer's own store is the arbiter and its writes are immediately visible to itself, so the check is exact even though the cluster is eventually consistent.

It also fails in the wrong direction. A store whose "atomic insert" turns out not to be atomic fails open, silently — precisely the current bug.

## What this does

Resolve the entry's ring owner and forward one hop, bounded by `is_moved`, reusing `forwardToWriteOwner` from #11108. The ring's membership comes from the master, so it tolerates a stale view and reassigns when a filer dies — neither of which a client's configured filer list can do.

Every creator gets this: mount, S3, the filer's own HTTP surface, the Java client. Not only the ones that opted in.

Plain creates are upserts whoever applies them, so they stay local and pay nothing — no extra hop on the write path that matters for throughput.

The route key shares the S3 gateway's namespace (`s3_constants.ObjectWriteRouteKeyPrefix`, whose comment already says the filer is meant to hash the same key), so an object's `ObjectTransaction` and its `CreateEntry` resolve to the same owner and land on that filer's one per-path lock. The name is S3-flavoured for what is now a general filer key; renaming it is a separate cleanup.

## Scope

Routing exclusive creates alone is enough for "at most one exclusive create wins". A concurrent *non*-exclusive upsert on another filer can still clobber the winner's entry, but it cannot make two exclusive creates both succeed, which is the POSIX contract at stake. Routing every write would serialize all namespace mutation through one filer per path — a throughput decision this does not make.

The mount half of #11079 (set `OExcl`, map `ErrEntryAlreadyExists` to `EEXIST`, let the structured code survive `doUnary`) rebases cleanly on top of this and no longer needs its client-side owner ring.

## Tests

New: an exclusive create for a peer-owned path leaves this filer and does not also write locally; a plain create stays local; an `is_moved` create is applied locally even when the ring says otherwise; an owner applies its own exclusive create. Existing `TestCreateEntryOExclSerialized` and `TestCreateEntryConditionEnforced` still pass.

`go test ./weed/server/ ./weed/filer/... ./weed/s3api/...` passes.

## Checks

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved distributed file creation so exclusive and condition-based writes are handled by the correct filer.
  - Prevented duplicate forwarding while ensuring forwarded requests are applied correctly.
  - Improved failure handling when the responsible filer is unavailable, avoiding unintended local writes.
  - Kept standard upsert operations local for faster processing.
  - Ensured object writes consistently use the same routing and locking behavior.
  - Storage synchronization settings are now carried through during volume assignment and uploads.


chrislusf:
Two findings from greptile. One was a real crash, now fixed; the other I am declining, with reasoning.

## Nil response on fallback — valid, fixed

Correct and worse than the report suggests. `CreateEntry` has a *named* return `resp`, pre-set to `&filer_pb.CreateEntryResponse{}`. Assigning `client.CreateEntry(...)` straight into it meant a failed forward stored `nil`, and if the next candidate was this filer the code fell through to the local path, which writes `resp.LogTsNs` — nil dereference, filer panic. A prior owner going away during a ring change is exactly how you reach it.

The forwarded response now lands in a local and is published only on success, so the named return keeps the value the local path expects:

```go
var ownerResp *filer_pb.CreateEntryResponse
handled, forwardErr := fs.forwardToWriteOwner(ctx, entryRouteKey(fullpath), func(owner pb.ServerAddress) error {
	return pb.WithFilerClient(..., func(client filer_pb.SeaweedFilerClient) error {
		forwarded, e := client.CreateEntry(ctx, req)
		if e != nil {
			return e
		}
		ownerResp = forwarded
		return nil
	})
})
```

`TestCreateEntryFallsBackLocallyWhenPriorOwnerIsGone` covers it: a two-snapshot ring that drops the peer, so the key"s prior owner is a dead address and its current owner is this filer. I checked the test earns its keep — against the old code it panics with the exact nil dereference, and passes against the new.

## Client-settable `is_moved` — declining

Accurate as a description, but it is not a new exposure and the suggested framing overstates it.

`is_moved` is the same field, with the same client-settable exposure, that `ObjectTransactionRequest` has carried since routing was introduced. Nothing here widens it.

More to the point, exclusivity is not a security boundary. A client that wants to bypass arbitration does not need this flag: it can send a plain non-exclusive `CreateEntry` to any filer and upsert the path. `o_excl` is a POSIX contract for cooperating callers, not an access control, and the routing makes it hold for callers that ask for it honestly.

Hardening it properly means authenticating that the caller is a peer filer. The gRPC peer address is the source `ip:port`, which will not match the ring"s advertised `host:port`, so it would come down to comparing IPs — fragile under NAT and wrong when several filers share a host. That is a real change to how filers identify each other, worth doing on its own terms for every internal RPC rather than bolted onto this field. The proto comment says clients leave it unset; that is the same contract `ObjectTransaction` runs on.

PR #11114, opened 2026-09-03
mount: name the disk without changing what is mounted

### LINKED ISSUE #11056: disk windows mount rename issue
After renaming the disk, the pictures in the original path can't be seen anymore. For example, the original path was Disk: OKIMAGE/. 

After renaming with a command, you have to use the new name, like IMAGE, and only then can the data in Disk: IMAGE/ be visible. 

All the original data has to be migrated, like OKIMAGE becoming IMAGE/OKIMAGE, and the calling logic gets more complicated because you have to add IMAGE.

chrislusf:
Fixes #11056.

The reporter wanted the Windows disk to be called `IMAGE`. The only thing
that named it was `-filer.path`, so naming it meant mounting `/IMAGE` — and
then everything outside `/IMAGE` was gone from the drive, with the data
needing to move to `/IMAGE/OKIMAGE` and every caller gaining an `IMAGE`
prefix. Their words: "only the data under the A path will be visible in the
virtual disk; everything else won't be visible."

So the mount point names the disk when the whole tree is mounted, ahead of
the filer address. `-dir=\\seaweedfs\Images` labels the disk `Images` with
`-filer.path` left at `/`, and `-dir=/mnt/images` gives macOS the name Finder
already shows for it. A bare drive letter has no name to give, so it still
falls back to the filer address. The mounted path keeps precedence where it
has one, so nothing that is labelled today changes.

Verified: `go test ./weed/command/`, plus `GOOS=windows` and `GOOS=linux`
builds. A live mount was not possible here — macFUSE cannot load its kext on
this machine — so the label itself is covered by the unit tests.

https://claude.ai/code/session_01Q9f8pWBXu1ceJvcQfYRQ7x

### DIFF
--- a/weed/command/mount.go
+++ b/weed/command/mount.go
@@ -201,7 +201,9 @@ var cmdMount = &Command{
 
   Where the platform labels the disk, in Finder and in Explorer, the mounted
   path names it: -filer.path="/Image Disk" shows up as "Image Disk". Mounting
-  the whole tree labels it with the filer address instead.
+  the whole tree takes the name from the mount point instead, so
+  -dir=\\seaweedfs\Images labels the disk "Images" while still mounting
+  everything, and only a bare drive letter falls back to the filer address.
 
   RDMA Acceleration:
   For ultra-fast reads, enable RDMA acceleration with an RDMA sidecar:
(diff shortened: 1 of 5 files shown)


piercashuang:
vendor\weed.exe mount -filer=127.0.0.1:8888 -dir=P: -cacheDirMaxEntries=100000

-dir=\\seaweedfs\Images labels the disk Images with
-filer.path

Using this name conflicts with the original virtual disk drive letter.

PR #11176, opened 2026-09-05
master: honour -volume.fileSizeLimitMB on the master's /submit

### LINKED ISSUE #6748: weed server does not respect -volume.fileSizeLimitMB
When I run `.\weed.exe server -dir=/tmp -volume.fileSizeLimitMB=2048` after having built 9f4be83e885aa397e14e8a73bd7ce18718bfd83e, the actual limit of the file size is still 256MB. The tool acts like this should work and lists it in `weed server --help`. I expect that more than fileSizeLimitMB is not being handled in this path, but did not need it for what I was testing.

junkerderprovinz:

Fixes #6748.

`weed server -volume.fileSizeLimitMB=2048` still refuses anything larger than 256MB, and the reason is not the one the report assumes. The option does reach the volume server. The master does not use it.

An upload posted to the master's `/submit` is buffered by the master before it is forwarded. `submitForClientHandler` passed a hardcoded `256*1024*1024` to `needle.ParseUpload`, so the master rejected a file that the volume server it had just started would have accepted.

The limit is passed in instead of being hardcoded.

`weed master` gains its own `-fileSizeLimitMB`, defaulting to the same 256 that was hardcoded before, so a standalone master behaves exactly as it did. `weed server` and `weed mini` hand their master the value their volume server already receives, so the two halves of a single-binary deployment now agree.

# How is the PR tested?

`weed/server/common_test.go` gains `TestSubmitForClientHandlerFileSizeLimit`, which calls `submitForClientHandler` itself with a 1MB limit: one byte over is answered 400 with the size limit error, one byte under is not. It asserts against the handler that the diff changes rather than against a helper next to it, so a regression to a hardcoded constant fails the test.

`go test ./weed/server/` passes.

# Checks

# Checks for AI generated PRs

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11176" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added a configurable file-size limit for uploads submitted through the master server.
  - The default upload limit is 256 MB and can be adjusted through server configuration.
  - Master, follower, and volume servers now support matching upload limits for consistent behavior.
  - Added configuration options allowing filer and S3 requests to use untrusted remote endpoints.

- **Bug Fixes**
  - Uploads exceeding the configured limit are rejected with a clear client error.
  - Uploads within the configured limit continue to be accepted.


junkerderprovinz:
Both review points are in.

The follower took the fixed 256MB while serving `/submit` exactly as the leader does, so it now takes the same flag. Without that, a cluster raised above 256 accepts an upload through the leader and refuses the identical one through a follower.

The master's flag description now names the `-volume.fileSizeLimitMB` spelling as well, and the under-limit test asserts the request gets past parsing instead of only checking the error message: over the limit answers 400 with the size error, under it answers 500 from the cancelled assignment.

PR #11177, opened 2026-09-05
log_buffer: stop closing a notification channel another reader still holds

### LINKED ISSUE #10810: Idle S3 metadata subscription busy-polls in awaitNotificationOrTimeoutFor, pinning a full core on 4.37 (silent regression of #9007)
### Summary

On 4.37, `weed server` burns a constant ~110% CPU (more than one full core) indefinitely while completely idle: no S3 requests, no network traffic, no disk I/O, and no log output at the default verbosity.

CPU profiling shows the time is spent in the S3 gateway's own metadata subscription spinning inside `LogBuffer.awaitNotificationOrTimeoutFor`, plus the GC pressure that busy loop creates by allocating a `time.Timer` on every iteration.

This looks like the same defect as #9007 (fixed by #9012 for 4.18/4.19). The difference is that in 4.37 it is **silent**: the `Notification timeout (LoopProcessLogData), polling` line only prints at raised verbosity, so at default log level the only symptom is a permanently pinned core.

### Version

```
version 30GB 4.37 c06a2dca8 linux amd64
```

Started as a combined server:

```
weed server -dir=/data -ip=<host> -volume.max=64 -s3 -s3.port=8333 \
    -s3.config=/etc/seaweedfs/s3.json -s3.iam.config=/etc/seaweedfs/iam.json
```

Single node, single filer, no replication, no peers. Filer store is the default `leveldb2`.

### Symptoms

Measured on an instance with 6 days of uptime:

- CPU pinned at **110%**, perfectly constant. Sampling `/proc/<pid>/stat` gives 1105 ticks per 10s with zero variance, so there is no periodic scan behind it.
- Total CPU consumed: **176 core-hours over 144 wall-hours** of uptime, i.e. it has been above one full core continuously since process start.
- **Zero disk I/O**: the cgroup `io.stat` counters moved by 3 write ops in 5 seconds.
- **Zero network traffic** and zero established TCP connections. It is serving nothing.
- Only **24 voluntary context switches in 5 seconds**, so it is not blocking on anything.
- **One log line in 5 minutes** at default verbosity.

### Profile

`perf record -F 199 -g` against the running process. Two independent runs, 20s and 30s:

Run 1 (20s, 5215 samples), cumulative:

```
 58.99%  filer_pb._SeaweedFiler_SubscribeLocalMetadata_Handler
 58.99%  server.(*FilerServer).SubscribeLocalMetadata
 58.63%  log_buffer.(*LogBuffer).LoopProcessLogData
 48.73%  log_buffer.(*LogBuffer).awaitNotificationOrTimeoutFor
 25.41%  runtime.gcBgMarkWorker
 23.73%  time.NewTimer
 21.05%  runtime.scanobject
 18.87%  runtime.selectgo
 14.08%  runtime.mallocgc
 10.55%  runtime.unlock2   (self)
```

Run 2 (30s, 7549 samples), cumulative:

```
 45.93%  runtime.gcBgMarkWorker
 40.98%  server.(*FilerServer).SubscribeLocalMetadata
 40.98%  log_buffer.(*LogBuffer).LoopProcessLogData
 36.60%  runtime.scanobject
 32.78%  log_buffer.(*LogBuffer).awaitNotificationOrTimeoutFor
 16.90%  time.NewTimer
 12.66%  runtime.findObject
 11.12%  runtime.selectgo
```

The two halves are the same problem: the subscription loop accounts for 40-59%, and the GC work accounts for most of the rest because that loop allocates a fresh timer per iteration.

Representative call chain:

```
runtime.unlock2
  runtime.selunlock
  runtime.selectgo
  log_buffer.(*LogBuffer).awaitNotificationOrTimeoutFor
  log_buffer.(*LogBuffer).LoopProcessLogData
  server.(*FilerServer).SubscribeLocalMetadata
  filer_pb._SeaweedFiler_SubscribeLocalMetadata_Handler
  grpc.(*Server).processStreamingRPC
  grpc.(*Server).handleStream
  grpc.(*Server).serveStreams.func2.1
  runtime.goexit.abi0
```

and the allocation path:

```
runtime.(*timer).modify
  time.newTimer
  time.NewTimer
  log_buffer.(*LogBuffer).awaitNotificationOrTimeoutFor
  log_buffer.(*LogBuffer).LoopProcessLogData
  ...
```

`awaitNotificationOrTimeoutFor` is supposed to park until either a notification arrives or the timeout fires. It is returning immediately instead, so the loop runs at full speed: allocate timer, select, return instantly, stop timer, repeat.

### The subscriber is the built-in S3 gateway

The only subscription registered on the affected instance:

```
filer_grpc_server_sub_meta.go:330  + s3@@ local subscribe /etc/ from {Time:<ts> Offset:-2 IsOffsetBased:false} clientId:<id>
```

So the S3 gateway watching `/etc/` for config changes is the client spinning against its own in-process filer.

### It is not purely version-dependent

This is the useful part. I have two instances running the **identical binary** (`4.37 c06a2dca8`), same launch flags, both with the same `s3@@ local subscribe /etc/` subscription registered:

| | Instance A | Instance B |
|---|---|---|
| Version | 4.37 c06a2dca8 | 4.37 c06a2dca8 (identical) |
| `s3@@ local subscribe /etc/` present | yes | yes |
| **CPU** | **110%** | **0.16%** |
| Files stored | ~246,000 | ~3,700 |
| Days present under `/topics/.system/log/` | **36** | **2** |
| Resident memory | 1.39 GiB | 48 MiB |

So the trigger is not the subscription existing, and not the version on its own. The clearest difference is the size of the persisted metadata log: the spinning instance has 36 days of `/topics/.system/log/` accumulated, and the healthy one has 2 days. The affected subscription is also resuming from a timestamp roughly a week older than the current process start, so it is replaying against that backlog.

That matches the 4.19 variant described in #9007, where the rapid polling appears specifically after a failure to resume from persisted logs (`Notification timeout after ResumeFromDiskError, polling`).

### Expected

An idle subscription with no metadata changes should park until notified or until the timeout expires, using effectively no CPU.

### Actual

It polls continuously, consuming a full core forever, with no indication at default log level.

### Notes

- Because the diagnostic log line is above default verbosity, anyone hitting this on 4.37 sees an unexplained pinned core and nothing in the logs. It would help a lot if a busy-poll of this kind logged a rate-limited warning at default level.
- Not fixed by restarting the process: the loop resumes immediately on the next start and has been pinned for the entire uptime of the current process.
- Not related to volume garbage. I ran a vacuum on this instance (reclaimed 3.28 GB) and CPU was unchanged at 110% before and after.

Happy to test a patch or gather any additional profile data on the affected instance.

junkerderprovinz:

Fixes #10810.

The report describes a subscriber burning a full core and points at the polling loop. The loop is not what spins.

`LogBuffer` keeps one notification channel per `subscriberID` and `UnregisterSubscriber` closes it. Two registrations that share a `subscriberID` share that one channel. That happens whenever a client opens a second stream, or an old stream has not yet noticed it was replaced, so the first unregister closes a channel the other reader is still parked on.

A closed channel makes every receive in `awaitNotificationOrTimeoutFor` return immediately. The surviving reader then loops at full speed for the rest of its life, which is the symptom in the report. The notification is also lost for that reader, so it is spinning and unnotified at the same time.

Subscriptions are reference counted.

Registering an existing `subscriberID` returns the same channel and raises the count instead of replacing it. Unregistering lowers the count and only closes the channel when the last holder is gone. `subscribers` and `flushSubscribers` both go from `map[string]chan struct{}` to `map[string]*subscription`, and both use the same pair of helpers so the two maps cannot drift apart.

Nothing else about the polling or notification behaviour changes.

# How is the PR tested?

`weed/util/log_buffer/log_read_test.go` gains `TestLoopProcessLogData_DuplicateReaderLeaving`: two readers register under one `subscriberID`, one unregisters, and the surviving reader must keep blocking instead of returning instantly on a closed channel. It counts the loop iterations rather than trusting a timing assertion.

With only the fix reverted, that reader spins 2,286,180 times in 500ms and the test fails. With the change it stays within the handful of iterations the test allows.

`go test ./weed/util/log_buffer/` passes.

# Checks

# Checks for AI generated PRs

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11177" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue affecting multiple log readers registered under the same name, preventing one reader from disrupting the remaining reader when it disconnects.
  - Prevented excessive processing caused by repeatedly reacting to a closed notification channel.
  - Improved reliability when registering, unregistering, and receiving notifications from duplicate log readers.


junkerderprovinz:
Good catch on the test, that gap was real.

The iteration count alone would have passed just as happily for a reader that had returned as for one that was not spinning. The test now fails if `LoopProcessLogData` returns while its duplicate unregisters, and only then trusts the count.

Reverting the fix and keeping the test still produces about 2.2 million iterations in 500ms, so it is still failing for the right reason.

chrislusf:
can you confirm whether #10810 still reproduces on master before this PR?

junkerderprovinz:
Measured rather than read, and the answer has two halves.

**The defect is still on master, and it reproduces deterministically.** On 331c6c36 I applied only this PR's test to `log_read_test.go`, not the fix. `TestLoopProcessLogData_DuplicateReaderLeaving` fails three times out of three at 2,281,873 / 2,269,674 / 2,291,939 iterations in 500 ms, where 3 are allowed. A probe that builds the same state and then reads `utime+stime` out of `/proc/self/stat` gives 1060, 1061 and 1062 ticks per 10 s, so 106% of a core with no variance across the three samples, and about 46 million loop iterations per 10 s. On this branch at 98fd98b3 the same probe gives 0, 0 and 1 ticks, 40 iterations, and the test passes three times out of three. That is the 110% and the 1105 ticks from #10810, on a machine I can hand you the numbers for: Docker 29.5.3, `golang:1.26-bookworm`, go1.26.8, `--cpus=4 --memory=4g`.

**What I could not reproduce is the route my own report took.** An idle `weed server -s3` in a fresh container burns nothing, and that holds for master as well as for 3b18a635 from 2026-08-19, so an idle run proves neither direction. My instance had 36 days of `/topics/.system/log/` behind it and a subscription resuming against that backlog, which a throwaway container does not have and which I did not try to fake.

Reading the filer path on master, it looks closed: since 813a4b071 (2026-08-02, #10501) the reader keys carry clientId and epoch, and the 2022 dedup guard rejects the second stream outright. I checked that against a running master: a gRPC client opening `SubscribeLocalMetadata` twice with the same client name, id and epoch is refused with `duplicated local subscription detected`, and the server stays at 0.0% after the first one closes. So on current master the S3 gateway is not the way in any more, and that commit predates my report.

The way in that is still open is `weed/mq/topic/local_partition.go:206`, which passes the bare `clientName` as the reader name into `LoopProcessLogData` with no id or epoch. That is the loop and a collidable key together, which is exactly the pairing this PR fixes. `weed/util/log_buffer/log_read_stateless.go:501` registers under `fetch-<startOffset>`, the offset alone, so two concurrent long polls on one offset share a channel and the first to return closes it; no loop there, so the cost is a false "data is ready" rather than a burning core. Both are read, not measured.

So: the fix is worth having on its own terms, since it removes the shared-close entirely rather than making one more key unique, but I should not claim it repairs what I originally reported. On master that path was already closed by someone else.

Issue #10661, opened 2026-08-09
S3: RenameObject accepts x-amz-client-token but does not implement idempotency


chrislusf:
`RenameObject` documents idempotency through `x-amz-client-token`, and the AWS SDKs auto-fill that header on every call (`OperationIdempotencyTokenAutoFill` in aws-sdk-go-v2). SeaweedFS accepts the header and ignores it.

Two gaps against the [documented contract](https://docs.aws.amazon.com/AmazonS3/latest/API/API_RenameObject.html):

- A rename that commits but whose response is lost returns `NoSuchKey` on the SDK's retry, instead of replaying the success. The source is already gone, so the retry cannot succeed on its own.
- Reusing a token with different parameters should return `IdempotentParameterMismatch`; it currently just performs the second rename.

Rejecting the header is not an option — the SDKs always send it, so every SDK caller would break. Accepting and ignoring it is what #10659 ships, with the limitation stated in the PR.

Implementing it properly means durable state the gateways share: token -> normalized request + result, TTL-expired, consulted before the move and written after it. A per-gateway in-memory cache would cover the common retry but behave non-deterministically across a fleet, which is arguably worse than a predictable 404. There is no client-token or idempotency infrastructure in the tree today, so this is a subsystem rather than a patch — and one that would serve any other operation wanting the same guarantee.

PR #11178, opened 2026-09-05
s3: make RenameObject idempotent for a retried request

junkerderprovinz:

Addresses #10661.

`RenameObject` is not idempotent. When a rename commits but its response is lost, the client has no safe move left: resending the identical request returns 404, because the source it names is already gone. A retry that succeeded and a rename that never happened look the same from outside.

This matters most where retries are automatic. An SDK that retries on a dropped connection turns one successful rename into a client-visible failure.

A rename records what it was, on the object it produced.

The destination carries `x-seaweedfs-rename-token` in its extended attributes: the client's token, the source key, and the time. A retried request is then classified against what it finds there.

- Same token, same source, same destination: the rename already committed here, answer 200 and touch nothing.
- Same token, different rename: answer 409 rather than silently doing something the client did not ask for.
- Token older than 24 hours, or none at all: unrelated, behave exactly as before.

The header is an internal one, so it does not leak back out through GET or HEAD. A request that sends no token takes the unchanged path.

A third choice belongs here rather than in the code alone: if the token write fails and the rename commits anyway, the rename goes ahead and only logs. A retry then finds no token, sees the source gone and gets today's 404, which is exactly where a client stands without this feature. Aborting instead would turn a transient metadata write failure into a failed rename that would otherwise have succeeded, so this fails in the direction that cannot make an existing caller worse off. Say so if you would rather the gateway refuse a rename it cannot make idempotent.

Two more limits are worth naming rather than hiding. The token is written to the destination, so an ordinary rename now carries one extra attribute; a copy of a renamed object inherits it, which is why the classification checks the destination key as well and treats a mismatch as unrelated. And the 24 hour window is a deliberate bound, not a correctness guarantee: past it, a retry falls back to today's 404 behaviour instead of a key answering for a request indefinitely.

# How is the PR tested?

`weed/s3api/s3api_object_handlers_rename_test.go` covers the classification directly: a rename that answers its own retry, the same token sent for a different source refused as reused, a different token treated as unrelated, a copy of a renamed object not answering for the original, an expired token, and the assertion that the header is internal so it cannot reach a GET or HEAD response.

`go test ./weed/s3api/` passes.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added idempotent retry support for S3 object rename requests using client tokens.
  * Retrying the same rename can now succeed safely after the original source is no longer available.
  * Rename operations now better support directory and prefix-object destinations.

* **Bug Fixes**
  * Reusing a client token for different rename details now returns a parameter-mismatch error with HTTP 409 Conflict.
  * Invalid, expired, or inconsistent rename metadata is now handled more reliably.


junkerderprovinz:
The status code was a real defect. The description promised Conflict and the mapping returned 400, which tells a client its request was malformed and invites it to give up. It is 409 now, with a test on it, since the status is the part of this a client actually acts on.

On the token write that only logs: that one is deliberate, and the PR body should have said so rather than leaving you to find it. If it fails and the rename commits, a retry finds no token, sees the source gone and gets today's 404, which is exactly where a client stands without this feature at all. Aborting instead would turn a transient metadata write failure into a failed rename that would otherwise have succeeded. I picked the mode that cannot make an existing caller worse off, but it is a judgement call: if you would rather the gateway refuse a rename it cannot make idempotent, say so and I will flip it.

PR #11184, opened 2026-09-05
s3: warn when a lifecycle change leaves fast-path-stamped objects on their old TTL

### LINKED ISSUE #11183: S3 lifecycle: deleting or lengthening an Expiration rule does not protect objects already written under it (per-write volume TTL is irreversible)
### Describe the bug

On a non-versioned bucket, `PutBucketLifecycleConfiguration` with an `Expiration.Days` rule makes every subsequent `PutObject` under the rule's prefix carry a TTL: `lifecycleTTLForObjectWrite` (`weed/s3api/s3api_object_handlers_put.go`) resolves the rule to seconds and `putToFiler` writes it into the filer entry's `Attributes.TtlSec` and onto the chunk uploads, so the volume needles get the same TTL.

From that point the object's lifetime no longer depends on the bucket's lifecycle configuration:

- the volume server refuses the needle once `AppendAt + TTL` has passed (`weed/storage/volume_read.go`, the `HasTtl` check);
- the filer hides the entry once `Crtime + TtlSec` has passed (`weed/filer/filer.go`), and the Redis, TiKV and RocksDB stores expire the entry natively;
- `DeleteBucketLifecycle` / a later `PutBucketLifecycleConfiguration` only replace the stored XML and clear the legacy `filer.conf` day-TTL entries (`ClearBucketLifecycleDayTTLs`); nothing re-evaluates objects that were stamped under the previous rules, and the lifecycle worker (`s3api_internal_lifecycle.go`) only deletes, it never rescues.

`LifecycleTTLResolver`'s own comment already states the constraint for tag-filtered rules ("volume TTL is irreversible — an object that matched at write time would still expire after the tag was removed") and excludes them from the fast path for that reason. Prefix-filtered rules have exactly the same property, but they are on the fast path.

In S3 the lifecycle configuration is evaluated against the current rules: deleting a rule stops future expirations, and raising `Days` from 7 to 30 keeps existing objects for 30 days. Here, objects written under the 7-day rule still disappear at day 7, from both the volume and the filer, after the rule is gone. For anyone using a prefix rule as a retention policy (our case: a video management system keeping camera recordings under `recordings/<camera>/` for N days), the normal operational action "an incident happened, extend retention / remove the rule" silently fails for everything already recorded.

### System Setup

Read from `master` (`weed/s3api/s3api_object_lifecycle_ttl.go`, `s3api_object_handlers_put.go`, `s3api_bucket_handlers.go`, `weed/storage/volume_read.go`, `weed/filer/filer.go`). Non-versioned bucket, prefix-only `Expiration.Days` rule, any filer store.

### Expected behavior

Removing a rule, or lengthening its `Days`, should stop or postpone expiration of objects already written under it, as it does in S3.

### Steps to reproduce

1. `PutBucketLifecycleConfiguration` on a non-versioned bucket: `Expiration.Days=1`, `Filter.Prefix=recordings/`.
2. `PutObject recordings/a.mp4` (the entry now has `TtlSec=86400` and its chunks live on a 1-day TTL volume).
3. `DeleteBucketLifecycle` (or set `Days=30`).
4. After 24 h: `GetObject recordings/a.mp4` returns 404 and the key is gone from `ListObjectsV2`.

### Possible fixes (happy to send a PR for whichever direction fits)

1. **Make the per-write TTL fast path opt-in** (an `s3` option, default off), so that by default the lifecycle worker is the only thing that expires objects and the bucket configuration stays the single source of truth. Cost: the worker's scan instead of free volume-level reclamation; the fast path stays available for people who want it and understand it is a commitment.
2. **Keep the fast path but make configuration changes safe**: when a `PutBucketLifecycleConfiguration` lengthens a rule or `DeleteBucketLifecycle` removes one, have the handler warn (log + response header) that objects stamped under the old rule keep their expiry, and provide a `weed shell` command that re-writes those entries (copy to a non-TTL / longer-TTL volume and clear `TtlSec`) so an operator can rescue them before the old deadline.
3. **Worker-side rescue**: the lifecycle worker already walks the bucket; when it finds an entry whose `Crtime + TtlSec` is earlier than what the current rules dictate, rewrite it before it expires. Most complete, also the most expensive.

1 + the warning half of 2 is the smallest change that removes the data-loss path.

chrislusf:
## Summary
- The per-write TTL fast path (opt-in via `s3.bucket.lifecycle.fastpath`) stamps a volume TTL at `PutObject` time that can't be taken back. When an operator lengthens or removes an `Expiration.Days` rule (or deletes the bucket lifecycle) on a fast-path-enabled bucket, objects already written keep their baked-in TTL and won't be rescued by the change — unlike the default worker-driven path, which re-evaluates the current rules each pass. This is the data-loss direction described in #11183.
- `PutBucketLifecycleConfiguration` and `DeleteBucketLifecycle` now emit a `glog` warning and set `X-Seaweed-Lifecycle-Fastpath-Warning` on the response when the change removes, disables, lengthens, or re-scopes a fast-path-eligible rule.
- Shortening a rule does **not** warn (old objects simply expire later — not data loss). Tag-only and overflow-day rules are never on the fast path and never warn.

This is the **warning half of option 2** from #11183. Option 1 (make the fast path opt-in, default off) is already on `master`; this PR adds the operator-facing signal for the case where the fast path is enabled and a policy change can't retroactively protect already-stamped objects.

#### Test plan
- [x] `go build ./weed/s3api/...`
- [x] `go test ./weed/s3api/` (full suite, 13.4s)
- [x] New unit tests in `s3api_bucket_lifecycle_fastpath_warn_test.go` cover: lengthen (warn), shorten (no warn), remove (warn), delete (warn), unchanged (no warn), no-old-rules (no warn), tag-only rule (no warn), disable (warn), prefix change (warn), overflow days (no warn)
- [ ] Manual: enable fast path on a bucket, put a 7-day rule, put an object, then `DeleteBucketLifecycle` — confirm the `X-Seaweed-Lifecycle-Fastpath-Warning` header and the `glog` warning appear.

Closes #11183 (warning half).

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11184" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Added S3 lifecycle warnings when changes may leave existing objects with outdated expiration behavior.
  * Warnings identify removed, disabled, lengthened, or otherwise modified lifecycle rules.
  * Warning details are returned in the `X-Seaweed-Lifecycle-Fastpath-Warning` response header.

* **Bug Fixes**
  * Warnings are now emitted only after lifecycle changes are successfully applied, preventing misleading warnings when an update fails.

### DIFF
--- a/weed/s3api/s3api_bucket_handlers.go
+++ b/weed/s3api/s3api_bucket_handlers.go
@@ -1183,16 +1183,22 @@ func (s3a *S3ApiServer) PutBucketLifecycleConfigurationHandler(w http.ResponseWr
 	// config removes or lengthens a rule: objects already written keep their
 	// baked-in TTL and won't be rescued by this change (unlike the default
 	// worker path, which re-evaluates the current rules each pass).
+	// Compute the reason before storing, but emit only after the store
+	// succeeds so a failed mutation never carries a warning for a change
+	// that was not applied.
+	var fastpathWarnReason string
 	if cfg, _ := s3a.getBucketConfig(bucket); cfg != nil && cfg.LifecycleTTL != nil {
-		if reason := fastpathConfigChangeLeavesStampedObjects(cfg.LifecycleXML, lifecycleXML); reason != "" {
-			glog.Warningf("PutBucketLifecycleConfigurationHandler %s: %s", bucket, reason)
-			w.Header().Set(fastpathWarningHeader, reason)
-		}
+		fastpathWarnReason = fastpathConfigChangeLeavesStampedObjects(cfg.LifecycleXML, lifecycleXML)
 	}
 
 	if errCode := s3a.storeBucketLifecycleConfiguration(bucket, lifecycleXML, r.Header.Get(bucketLifecycleTransitionMinimumObjectSizeHeader)); errCode != s3err.ErrNone {
 		s3err.WriteErrorResponse(w, r, errCode)
 		return
+	}
+
+	if fastpathWarnReason != "" {
+		glog.Warningf("PutBucketLifecycleConfigurationHandler %s: %s", bucket, fastpathWarnReason)
+		w.Header().Set(fastpathWarningHeader, fastpathWarnReason)
 	}
 
 	writeSuccessResponseEmpty(w, r)
@@ -1222,17 +1228,22 @@ func (s3a *S3ApiServer) DeleteBucketLifecycleHandler(w http.ResponseWriter, r *h
 	// If the per-write TTL fast path is active, every previously-stamped
 	// object keeps its baked-in volume TTL after the config is removed —
 	// deleting the rules does not rescue them (unlike the default worker
-	// path). Warn so the operator can act before the old deadlines.
+	// path). Compute the reason before clearing, but emit only after the
+	// clear succeeds so a failed mutation never carries a warning for a
+	// change that was not applied.
+	var fastpathWarnReason string
 	if cfg, _ := s3a.getBucketConfig(bucket); cfg != nil && cfg.LifecycleTTL != nil {
-		if reason := fastpathConfigChangeLeavesStampedObjects(cfg.LifecycleXML, nil); reason != "" {
-			glog.Warningf("DeleteBucketLifecycleHandler %s: %s", bucket, reason)
-			w.Header().Set(fastpathWarningHeader, reason)
-		}
+		fastpathWarnReason = fastpathConfigChangeLeavesStampedObjects(cfg.LifecycleXML, nil)
 	}
 
 	if errCode := s3a.clearStoredBucketLifecycleConfiguration(bucket); errCode != s3err.ErrNone {
 		s3err.WriteErrorResponse(w, r, errCode)
 		return
+	}
+
+	if fastpathWarnReason != "" {
+		glog.Warningf("DeleteBucketLifecycleHandler %s: %s", bucket, fastpathWarnReason)
+		w.Header().Set(fastpathWarningHeader, fastpathWarnReason)
 	}
 
 	s3err.WriteEmptyResponse(w, r, http.StatusNoContent)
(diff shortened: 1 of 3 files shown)


chrislusf:
Pushed a commit addressing the two issues raised by CodeRabbit, Greptile, and Devin:

**1. Failed mutations retained the warning header** (all three reviewers)
The warning is now computed before the mutation but emitted (log + header) only after `storeBucketLifecycleConfiguration` / `clearStoredBucketLifecycleConfiguration` succeeds. A failed mutation no longer carries a warning for a change that was never applied.

**2. Rule renames produced false "removed" warnings** (all three reviewers)
Replaced the single-key (`fastpathRuleKey`) matching with two-pass greedy matching:
- Pass 1: match by ID. A rule that kept its ID is the same rule; check for filter or days changes.
- Pass 2: for unmatched old rules, match by fast-path predicates (prefix + size). A predicate match with unchanged days is a pure rename — no warning. A predicate match with increased days still warns (lengthened).

Each new rule is consumed by at most one old rule so overlapping rules can't be double-matched. Added regression tests: `TestFastpathWarn_IDOnlyRename_NoWarn`, `TestFastpathWarn_IDRenameAndLengthen_Warns`, `TestFastpathWarn_IDRenameAndShorten_NoWarn`.

**3. Historical expirations lose warnings after fastpath disable** (Devin)
Acknowledged as a known limitation. This edge case (fastpath disabled, then rules changed) requires persisting "ever used fastpath" state on the bucket, which is a larger design change. The warning covers the common case: fastpath is active and the operator changes rules while it's on. The disabled-fastpath-then-change-rules scenario is tracked as a follow-up.

All 13 tests pass; `go build ./weed/s3api/...` is clean.

PR #11190, opened 2026-09-06
s3: gate IAM-cache gRPC RPCs behind admin Bearer auth

chrislusf:
## Summary

The S3 gateway's `SeaweedS3IamCacheServer` (internal gRPC port, default `0.0.0.0:18333`) accepted every IAM-cache RPC with no per-RPC authentication. An unauthenticated network peer that could reach the port could call `PutIdentity` with `Actions:["Admin"]` and write straight into the live `accessKeyIdent` map that the S3 SigV4 path reads — then issue ordinary SigV4-signed requests (signed with their own secret) that resolve to full S3 administrator, gaining read/write/delete over all buckets/objects and the ability to rewrite policies or remove legitimate identities.

#### Test plan

- [x] `go build ./weed/...`
- [x] `go test ./weed/s3api/` (new `TestS3IamCache_*` covers: no metadata, missing/non-Bearer auth header, mis-signed/garbage/expired token, valid token writes the identity, no-key opt-in allows unauthenticated, and each RPC requires auth)
- [x] `go test ./weed/credential/` (new `TestWithIamCacheAdminAuth_*` covers the propagation helper: no-op without key, attaches a verifiable Bearer token with key set)
- [x] `go test ./weed/server/ -run TestIamGrpc` (filer sibling still green)
- [x] `gofmt` clean on all changed files

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11190" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Security**
  * IAM cache administration requests now require a valid Bearer token when signing keys are configured.
  * Requests with missing, invalid, expired, or incorrectly signed credentials are rejected.
  * Deployments without a signing key continue to allow these requests.

* **Bug Fixes**
  * Credential propagation now limits its completion window to the authentication token’s remaining lifetime, preventing token expiry before updates finish.

* **Tests**
  * Added coverage for authenticated and unauthenticated IAM cache administration requests, token validation failures, and token lifetime handling.

### DIFF
--- a/weed/credential/propagating_store.go
+++ b/weed/credential/propagating_store.go
@@ -13,8 +13,11 @@ import (
 	"github.com/seaweedfs/seaweedfs/weed/pb/master_pb"
 	"github.com/seaweedfs/seaweedfs/weed/pb/s3_pb"
 	"github.com/seaweedfs/seaweedfs/weed/s3api/policy_engine"
+	"github.com/seaweedfs/seaweedfs/weed/security"
+	"github.com/seaweedfs/seaweedfs/weed/util"
 	"github.com/seaweedfs/seaweedfs/weed/wdclient"
 	"google.golang.org/grpc"
+	"google.golang.org/grpc/metadata"
 )
 
 var _ CredentialStore = &PropagatingCredentialStore{}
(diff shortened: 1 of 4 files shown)


chrislusf:
Thanks for the reviews. Addressing each:

**Devin Review (expired token on delayed propagation)** — valid; fixed in 5beabd2. `withIamCacheAdminAuth` now runs after `ListClusterNodes` succeeds, immediately before the propagation timeout is derived, so master-client retries can't burn through the default 10s token lifetime before the peer fan-out begins.

**Greptile (JWT purpose unenforced)** — this is a pre-existing property of SeaweedFS's shared signing-key design, not something this change introduces. `checkAdminAuth` here mirrors the filer's `IamGrpcServer.checkAdminAuth` verbatim (same `SeaweedFilerAdminClaims`, same `jwt.filer_signing.key`, same `DecodeJwt`). The filer sibling has the same behavior; tightening the claims discriminator would be a cross-cutting change to the filer IAM service too, so it's out of scope for this parity fix. The change still narrows attack surface: it closes the previously unauthenticated S3 IAM-cache handler.

**CodeRabbit (fail-closed without TLS)** — same parity argument. The filer's `checkAdminAuth` does not require TLS either; the operator knob is the signing key, TLS is opt-in via `grpc.s3` cert config. Rejecting insecure transport when a key is set would diverge from the filer's established model and is a broader behavioral change. Happy to follow up on either of the above as separate PRs if maintainers want the filer and S3 services tightened together.

chrislusf:
**Re: Greptile — JWT purpose remains unenforced**

This is a false positive. The security boundary is the HMAC signing key, not the claims type:

1. **Volume-server JWTs** (`SeaweedFileIdClaims`) are signed with `jwt.signing.key` — a *different* key from `jwt.filer_signing.key` used by `checkAdminAuth`. They fail signature verification and are rejected. Greptile's claim that restricted volume-server tokens can authorize is incorrect.

2. **Filer-server JWTs** (`SeaweedFilerClaims`) are signed with `jwt.filer_signing.key` — the same key. But anyone who can mint a filer-server JWT already possesses `jwt.filer_signing.key` (it's in their `security.toml`), and with that key they can mint an admin JWT directly via `GenJwtForFilerAdmin`. There is no privilege escalation: the capability granted by the key is already "full admin"; the claims type doesn't gate anything.

Adding a claims discriminator would only matter if different token types used *different* keys — they don't. The filer's `IamGrpcServer.checkAdminAuth` (which this PR mirrors) has the same design, and tightening it would be a cross-cutting change to the shared JWT primitive for no real security gain.

**Re: CodeRabbit — propagation deadline vs JWT lifetime (new comment on 5beabd28)**

Valid — fixed in 3e9b2b3. `withIamCacheAdminAuth` now returns the token's lifetime, and `propagateChange` caps the propagation timeout at min(10s, tokenTTL) so the token can't expire mid-fan-out when an operator configures a short `expires_after_seconds`.

Issue #11179, opened 2026-09-05
Rust volume server (weed-volume 0.1.0, tag 4.42) leaks memory proportional to bytes served; allocations resolve into redb's page cache


franchb:
## Problem

The Rust volume server (`weed-volume`) grows without bound in proportion to bytes served and does not return the memory when load stops. On a 60 GB host running 16 instances, sustained bulk ingest drove individual instances to 15–21 GB RSS and the kernel OOM-killed volume servers **27 times in two days**. The same host had run for the previous 40+ days with no OOM kills at all.

The memory is ~100 % anonymous heap, not page cache, and outstanding allocations resolve almost entirely into `redb`.

## Version

- `weed-volume --version` → `weed-volume 0.1.0`
- SeaweedFS release tag **4.42**, prebuilt Rust binary (`weed-volume_*.tar.gz` from the release)
- Linux 7.0.0-22-generic, x86-64
- Index backend: `--index=redb` (the smallest of the redb tiers; `redbMedium`/`redbLarge` not used)

Relevant flags per instance:

```
weed-volume --dir=/srv/hddN/seaweedfs/{a,b} --disk=hdd --tags=warm-ec --index=redb \
  --mserver=<master>:9333 --port=<p> --port.grpc=<g> --dataCenter=<dc> --rack=hddN \
  --max=1152 --minFreeSpace=2TiB --maintenanceMBps=300 --compactionMBps=50 \
  --readBufferSizeMB=8 --ip=<ip>
```

### Memory tracks uptime-under-load, not data size

All 16 instances back comparable 21 TB disks in the same cluster. The only variable is how long each has been up and how much it has served:

| instance | up for | RSS |
|---|---|---:|
| freshly restarted (×4) | 0–2 h, idle | **91 – 246 MB** |
| A | 1.5 h under load | 4,811 MB |
| B | 4.7 h under load | **15,522 MB** |
| C | 17.6 h under load | 9,801 MB + 4,005 MB swap = **13.8 GB** |

A freshly started instance settles at **90–250 MB**. Under bulk ingest, two independent instances both grew at roughly **3.3 GB/hour**.

### It is load-proportional, not time-proportional

Two instances restarted during the ingest window and were still at ~90 MB two hours later, because by then the bulk copy had finished. Idle instances barely grow. Current steady-state growth on a quiet host (0.33 req/s) is +0.17 GB/hour summed across all 16.

### It is anonymous heap, not cache

[...]

PR #11202, opened 2026-09-07
rust volume: insert redb rebuilds in needle-id order

franchb:
> **Stack 1/4.** Follow-ups: #11204 (put/delete in one write transaction), #11203 (quick-repair on durable checkpoints), #11205 (optional `insert_before` bulk load).

## What

Unlink the volume `.rdb` before `Database::create` during `full_rebuild`, collapse last-write-wins from the `.idx`, then insert live keys in needle-id order so redb 4.2.0 packs leaves.

## Why

`Database::create` opens an existing file (`truncate(false)`). The previous rebuild ignored unlink errors and then `table.remove`d only keys that appeared as deletes in the `.idx`. A leftover `.rdb` could keep keys that are no longer in the index.

Inserts also walked a `HashMap`, so they were not in key order. redb 4.2.0 packs btree leaves on ascending inserts; unsorted rebuilds leave the file fragmented.

## How

- Unlink is fatal unless the path is already absent (`NotFound`).
- On `new()` failure or a failed rebuild, drop the map and unlink whatever `.rdb` was left behind.
- Last-write-wins still uses a `HashMap`. Live pairs are then sorted by `NeedleId` and inserted; deleted keys are skipped (`table.remove` is gone — the table is empty after a real unlink).
- Incremental `try_reuse_rdb` is unchanged.

## Tests

- `test_redb_full_rebuild_last_write_wins_from_shuffled_idx`: `.idx` in non-id order (10, 1, 5, overwrite 1, delete 5); live set is `{10, 1@200/200}`, key 5 absent.
- `test_redb_full_rebuild_drops_keys_absent_from_idx`: shrink `.idx` so reuse rejects, then rebuild; a key that existed only in the old `.rdb` must not survive.
- `cd seaweed-volume && cargo test --lib -- needle_map -- --test-threads=1`: 56 passed.

Existing `.rdb` files are unaffected until the next full rebuild.

## Summary by CodeRabbit

* **Bug Fixes**
  * Rebuilding storage indexes now correctly preserves the latest record for each item, even when records are processed out of order.
  * Removed entries are no longer retained after an index is truncated.
  * Rebuilds can continue when removing an existing database file is unsuccessful.
  * Failed rebuilds still clean up incomplete database state to prevent stale or inconsistent data.

### DIFF
--- a/seaweed-volume/src/storage/needle_map.rs
+++ b/seaweed-volume/src/storage/needle_map.rs
@@ -707,7 +707,12 @@ impl RedbNeedleMap {
         }
     }
 
-    /// Full rebuild: delete existing .rdb and rebuild from entire .idx file.
+    /// Full rebuild: unlink any existing .rdb and rebuild from the entire .idx.
+    ///
+    /// `Database::create` opens an existing file (`truncate(false)`). Unlink
+    /// must succeed (or the path must be absent) so the needles table is empty
+    /// before we insert. Inserts are in needle-id order so redb 4.2.0 packs
+    /// leaves.
     fn full_rebuild<R: Read + Seek>(
         db_path: &str,
         reader: &mut R,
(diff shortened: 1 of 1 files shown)


seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
+                } else {
+                    entries.insert(key, Some(NeedleValue { offset, size }));
+                }
+                Ok(())
             })?;
 
-            for (key, maybe_nv) in &entries {
-                let key_u64: u64 = (*key).into();
-                if let Some(nv) = maybe_nv {
+            let mut live: Vec<(NeedleId, NeedleValue)> = entries
```
franchb: **low -- peak rebuild memory is now higher than before this PR, and the `None` entries have become dead weight.**

`entries.into_iter()` cannot free the `HashMap` until the `collect` finishes, so peak memory during a full rebuild is now map + vector + `sort_by_key`'s merge-sort scratch (about half the vector's length), where the previous code held only the map. This is the load path taken for every volume after an unclean shutdown, in code whose whole purpose (#11179) is keeping the needle index off the heap.

Separately, `entries.insert(key, None)` on L741 now exists only to be filtered out three lines below -- the `None` values were there to feed the `table.remove` this PR deletes.

Both collapse into one change: make `entries` a `BTreeMap<NeedleId, NeedleValue>`, with `insert` on a live record and `remove(&key)` on a tombstone. That gives last-write-wins *and* ascending iteration from a single structure -- no `Vec`, no sort scratch, no `None` entries -- so peak memory ends up *below* the pre-PR level while keeping the leaf-packing win measured above.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -1487,6 +1510,107 @@ mod tests {
...
+            NeedleId(5),
+            Offset::default(),
+            TOMBSTONE_FILE_SIZE,
+        )
+        .unwrap();
+        idx_data
+    }
+
+    #[test]
+    fn test_redb_full_rebuild_last_write_wins_from_shuffled_idx() {
```
franchb, line 1561: **low -- both new tests pass against the unmodified pre-PR `full_rebuild`, so they guard nothing.**

I spliced `shuffled_idx_with_overwrite_and_delete`, this test and `test_redb_full_rebuild_drops_keys_absent_from_idx` verbatim onto the pre-PR `needle_map.rs` and ran them: both pass. That is expected -- on Linux the old `let _ = std::fs::remove_file(db_path)` succeeds, so the table is already empty and the old `table.remove` is a no-op. The "leftover `.rdb` keeps stale keys" scenario these tests describe is unreachable unless the unlink fails.

The behavior this PR actually makes observable is the *fatal* unlink, and nothing covers it. A test that makes `remove_file` fail with something other than `NotFound` -- e.g. point `db_path` at a directory -- and asserts `full_rebuild` returns `Err` instead of proceeding onto the leftover file would be a real regression test for the change.

franchb (commented):
Automated review (Claude Code `/code-review high`). Built the head at `89efaed` and ran `cargo test --lib -- needle_map --test-threads=1` -> 54 passed. **No correctness bug found.**

The things that looked suspicious all check out:

- The now-fatal `remove_file` has no realistic regression path -- every case where unlink fails (dir not writable, sticky-bit dir holding a foreign-owned `.rdb`, path is a directory) also fails the subsequent `Database::create`, so nothing that used to mount now fails.
- Dropping `table.remove` for tombstoned keys is safe: after a real unlink the table is empty, so the end state matches the previous code exactly. The resulting row-absent-vs-tombstone asymmetry against the incremental `try_reuse_rdb` path is pre-existing, not introduced here.
- `live.sort_by_key` ordering matches redb's key order (`NeedleId` derives `Ord` over its `u64`, and it is stored as a `u64` key).
- `metrics_from_idx` seeks internally and leaves the reader at 0, so moving it inside the closure is fine; the two non-durable commits revert together on crash to `new()`'s empty state, which forces a clean rebuild on the next load.

**The perf claim holds.** I instrumented `Table::stats()` for a 200k-entry shuffled `.idx` rebuild, before and after:

| | `leaf_pages` | `fragmented_bytes` |
| --- | --- | --- |
| before (unsorted) | 1516 | 2,024,080 |
| after (sorted) | 1031 | 50,916 |

32% fewer leaf pages and ~40x less fragmentation at identical `stored_bytes` -- redb 4.2.0's rightmost-append path does fire. One caveat for the description: the `.rdb` file length is identical at this scale, because redb grows in regions, so file size is not the right metric to quote here; leaf pages / fragmentation is.

Two low-severity notes inline.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
             })?;
 
-            for (key, maybe_nv) in &entries {
-                let key_u64: u64 = (*key).into();
-                if let Some(nv) = maybe_nv {
+            let mut live: Vec<(NeedleId, NeedleValue)> = entries
+                .into_iter()
+                .filter_map(|(k, v)| v.map(|nv| (k, nv)))
+                .collect();
+            live.sort_by_key(|(k, _)| *k);
```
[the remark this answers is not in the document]
franchb: Fixed on this PR: `full_rebuild` now collapses into a `BTreeMap<NeedleId, NeedleValue>` and inserts in that order. No second `Vec`, no `sort_by_key` scratch. Tombstones `remove` the key.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
+                Ok(())
             })?;
 
-            for (key, maybe_nv) in &entries {
-                let key_u64: u64 = (*key).into();
-                if let Some(nv) = maybe_nv {
+            let mut live: Vec<(NeedleId, NeedleValue)> = entries
+                .into_iter()
+                .filter_map(|(k, v)| v.map(|nv| (k, nv)))
+                .collect();
```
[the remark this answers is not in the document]
franchb: Fixed: one `BTreeMap` for last-write-wins and ascending insert. Peak rebuild memory is the map alone.

seaweed-volume/src/storage/needle_map.rs
franchb: Done. `BTreeMap<NeedleId, NeedleValue>`: live `insert`, tombstone `remove`, iterate in id order. `None` entries are gone.

seaweed-volume/src/storage/needle_map.rs
franchb: Kept the two characterization tests (they still lock last-write-wins and leftover keys after a successful unlink). Did not add a directory-as-`db_path` test: unlink is no longer fatal. If unlink fails but the path is writable, the rebuild txn `retain`s the leftover table empty, then inserts. A path that is a directory still fails at `Database::create`.

PR #11204, opened 2026-09-07
rust volume: drop the extra redb read transaction on put/delete

franchb:
> **Stacked on #11202 (2/4).** This branch builds on that PR's sorted rebuild, so the diff here shows both commits until #11202 merges; only the put/delete transaction commit is new. I'll rebase once #11202 lands.

## What

`put` uses `Table::insert()`'s previous value instead of a separate read transaction. `delete` gets then inserts the tombstone in the same write transaction. Truncate the `.idx` row on any failed redb write after the append.

## Why

Each put and each live delete opened a read transaction (`get_internal`) and then a write transaction. That is an extra btree walk and an extra txn per needle. redb 4.2.0 `Table::insert()` already returns the old value. `Table::entry()` is not used: in 4.2.0 `OccupiedEntry` stores only the key and `get()`/`insert()` walk the tree again.

A failed `begin_write` after the `.idx` append used to leave an orphan row; only `commit` failure truncated.

## How

- `packed_to_needle_value` length-checks before unpack so a corrupt blob is `None`, not a panic. `get_internal` and `get_via_table` share it.
- `put`: append `.idx` first, then one `Durability::None` write txn; previous row comes from `insert()`; any redb error after the append truncates.
- `delete`: open the write txn first, `table.get` + `table.insert` of the negated-size tombstone. Missing / already-deleted / wrong-length keys abort with no `.idx` write. Live delete still writes the `.idx` tombstone before the table mutation; insert/commit failure truncates.
- `idx_file_offset` and `writes_since_checkpoint` still advance only after a successful commit.

## Tests

- `test_redb_put_overwrite_metrics_without_prior_get`: overwrite metrics without a prior `get`.
- `test_redb_delete_wrong_length_value_is_absent_not_panic`: a 2-byte value deletes as `None` with no counter bump.
- Existing put/delete/double-delete/checkpoint tests still pass.
- `cd seaweed-volume && cargo test --lib -- needle_map -- --test-threads=1`: 56 passed.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11204" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved storage consistency when updating or deleting entries.
  - Rebuilds now correctly reflect the latest index state, including removing entries no longer present.
  - Invalid stored values are handled safely instead of causing incorrect deletion behavior.
  - Updated metrics now remain accurate after overwriting entries.

- **Reliability**
  - Rebuild failures no longer leave behind potentially corrupted partial databases.
  - Rebuild processing is now deterministic, improving recovery and data integrity.

### DIFF
--- a/seaweed-volume/src/storage/needle_map.rs
+++ b/seaweed-volume/src/storage/needle_map.rs
@@ -57,6 +57,15 @@ fn unpack_needle_value(bytes: &[u8; PACKED_NEEDLE_VALUE_SIZE]) -> NeedleValue {
     }
 }
 
+fn packed_to_needle_value(bytes: &[u8]) -> Option<NeedleValue> {
+    if bytes.len() != PACKED_NEEDLE_VALUE_SIZE {
+        return None;
+    }
+    let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
+    arr.copy_from_slice(bytes);
+    Some(unpack_needle_value(&arr))
+}
+
 // ============================================================================
 // NeedleMapMetric
 // ============================================================================
@@ -689,16 +698,7 @@ impl RedbNeedleMap {
         key_u64: u64,
     ) -> io::Result<Option<NeedleValue>> {
         match table.get(key_u64) {
-            Ok(Some(guard)) => {
-                let bytes: &[u8] = guard.value();
-                if bytes.len() == PACKED_NEEDLE_VALUE_SIZE {
-                    let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
-                    arr.copy_from_slice(bytes);
-                    Ok(Some(unpack_needle_value(&arr)))
-                } else {
-                    Ok(None)
-                }
-            }
+            Ok(Some(guard)) => Ok(packed_to_needle_value(guard.value())),
             Ok(None) => Ok(None),
             Err(e) => Err(io::Error::new(
                 io::ErrorKind::Other,
@@ -814,30 +814,35 @@ impl RedbNeedleMap {
         }
 
         let key_u64: u64 = key.into();
-        let nv = NeedleValue { offset, size };
-        let packed = pack_needle_value(&nv);
+        let packed = pack_needle_value(&NeedleValue { offset, size });
 
-        // Read old value for metric update
-        let old = self.get_internal(key_u64)?;
-
-        let txn = Self::begin_write_no_fsync(&self.db)?;
-        {
-            let mut table = txn.open_table(NEEDLE_TABLE).map_err(|e| {
-                io::Error::new(io::ErrorKind::Other, format!("redb open_table: {}", e))
+        let old = match (|| -> io::Result<Option<NeedleValue>> {
+            let txn = Self::begin_write_no_fsync(&self.db)?;
+            let old = {
+                let mut table = txn.open_table(NEEDLE_TABLE).map_err(|e| {
+                    io::Error::new(io::ErrorKind::Other, format!("redb open_table: {}", e))
+                })?;
+                let prev = table.insert(key_u64, packed.as_slice()).map_err(|e| {
+                    io::Error::new(io::ErrorKind::Other, format!("redb insert: {}", e))
+                })?;
+                prev.and_then(|g| packed_to_needle_value(g.value()))
+            };
+            txn.commit().map_err(|e| {
+                io::Error::new(io::ErrorKind::Other, format!("redb commit: {}", e))
             })?;
-            table
-                .insert(key_u64, packed.as_slice())
-                .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("redb insert: {}", e)))?;
-        }
-        if let Err(e) = txn.commit() {
-            self.truncate_idx_to_offset();
-            return Err(io::Error::new(io::ErrorKind::Other, format!("redb commit: {}", e)));
-        }
+            Ok(old)
+        })() {
+            Ok(old) => old,
+            Err(e) => {
+                self.truncate_idx_to_offset();
+                return Err(e);
+            }
+        };
+
         if self.idx_file.is_some() {
             self.idx_file_offset += NEEDLE_MAP_ENTRY_SIZE as u64;
         }
         self.writes_since_checkpoint = self.writes_since_checkpoint.saturating_add(1);
-
         self.metric.on_put(key, old.as_ref(), size);
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


franchb (commented):
Review of the redb `put`/`delete` transaction change.

The core refactor looks correct to me: reusing the `AccessGuard` returned by `Table::insert` is semantically equivalent to the old `get_internal` read, tombstone construction (`old.offset`, `Size(-old.size.0)`) is unchanged, the `.idx` tombstone still precedes the table mutation, and `idx_file_offset` / `writes_since_checkpoint` still advance only after a successful commit. Borrow/drop ordering in both rewritten functions is sound (the `Table` is dropped before `commit`). The new `truncate_idx_to_offset()` on `begin_write`/`open_table`/`insert` failure is a genuine fix, and it is safe because the only `IdxFileWriter` impl is `std::fs::File` (unbuffered, `truncate_to` = `set_len`). `full_rebuild` dropping the `table.remove()` branch is fine since the `.rdb` is unlinked first, so the table is always empty.

Three non-blocking notes inline.

Verified by building the branch and running `cargo test --lib -- needle_map` (56 pass).

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
+                } else {
+                    entries.insert(key, Some(NeedleValue { offset, size }));
+                }
+                Ok(())
             })?;
 
-            for (key, maybe_nv) in &entries {
-                let key_u64: u64 = (*key).into();
-                if let Some(nv) = maybe_nv {
+            let mut live: Vec<(NeedleId, NeedleValue)> = entries
```
franchb: `entries` and `live` are both alive at this point: `into_iter().collect()` does not release the `HashMap` allocation until the `Vec` is fully built, so peak memory during rebuild roughly doubles. On a volume with 10M needles that is ~400 MB of `HashMap` plus ~240 MB of `Vec` held simultaneously — in the code path whose entire reason to exist (#11179) is bounded index memory. The old code streamed straight from the `HashMap` into the table.

Making `entries` a `BTreeMap<NeedleId, Option<NeedleValue>>` and iterating it directly fixes this and also removes both the `live` Vec and the `sort_by_key` — a `BTreeMap` is already in needle-id order.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
     fn full_rebuild<R: Read + Seek>(
         db_path: &str,
         reader: &mut R,
         idx_size: u64,
         version: Version,
         cache_bytes: usize,
     ) -> io::Result<Self> {
-        let _ = std::fs::remove_file(db_path);
-        let mut nm = RedbNeedleMap::new(db_path, cache_bytes)?;
+        match std::fs::remove_file(db_path) {
```
franchb: This now returns any `remove_file` error other than `NotFound`, and both `load_from_idx` call sites in `volume.rs` propagate it with `?`, so a failed unlink fails the whole volume load.

Master's `let _ = remove_file(...)` tolerated the error and rebuilt in place, which was still correct for every key present in the `.idx` (live entries inserted, deleted ones `table.remove()`d) — only keys entirely absent from the `.idx` leaked. Concrete case: a data directory with the sticky bit where the `.rdb` is owned by a different uid after a root -> service-user permission drift. `unlink` returns EPERM while `Database::create` on the existing file would still succeed: master serves the volume, this branch refuses to open it.

Falling back to clearing the needles table when the unlink fails would keep the stale-key fix without the availability regression.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -1487,6 +1514,107 @@ mod tests {
...
+            NeedleId(5),
+            Offset::default(),
+            TOMBSTONE_FILE_SIZE,
+        )
+        .unwrap();
+        idx_data
+    }
+
+    #[test]
+    fn test_redb_full_rebuild_last_write_wins_from_shuffled_idx() {
```
franchb, line 1556: All four new tests pass unmodified on `upstream/master` — I grafted them onto b82cb05d7 and got 56 passed — so none of them discriminates this change:

- `test_redb_full_rebuild_last_write_wins_from_shuffled_idx` (this one) — insert order is a performance property the assertions cannot observe.
- `test_redb_full_rebuild_drops_keys_absent_from_idx` (L1581) — master already does `let _ = remove_file(db_path)` before `RedbNeedleMap::new`, so key 99 is gone there too.
- `test_redb_put_overwrite_metrics_without_prior_get` (L1639) — master's `put` did its own `get_internal`, so "without a prior get" is not observable from outside.
- `test_redb_delete_wrong_length_value_is_absent_not_panic` (L1790) — master's `delete` went through `get_internal`, which already length-checked; the panic the name guards against was never reachable.

The one behavior this PR genuinely introduces — hard-failing the load when the unlink errors — has no coverage.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +707,80 @@ impl RedbNeedleMap {
...
-        })?;
+        };
 
-        // Write all live entries to redb in a single transaction
-        let txn = Self::begin_write_no_fsync(&nm.db)?;
-        {
-            let mut table = txn.open_table(NEEDLE_TABLE).map_err(|e| {
-                io::Error::new(io::ErrorKind::Other, format!("redb open_table: {}", e))
+        let result = (|| -> io::Result<()> {
+            let mut entries: HashMap<NeedleId, Option<NeedleValue>> = HashMap::new();
```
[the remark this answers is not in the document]
franchb: Not taking disk-backed staging. Rebuild now uses a `BTreeMap` (see #11202) so there is one ordered last-write-wins structure and no second live `Vec`. External sort is YAGNI for this path.

seaweed-volume/src/storage/needle_map.rs
franchb: Fixed on #11202 and restacked here: `BTreeMap`, no overlapping `HashMap`+`Vec`+stable sort.

seaweed-volume/src/storage/needle_map.rs
franchb: Unlink is best-effort now. `EPERM` (sticky bit / foreign owner) no longer fails the volume load. If unlink fails, `table.retain(|_, _| false)` clears leftover keys in the rebuild txn, then sorted insert. Skip-`remove` stays valid because the table is empty either way.

seaweed-volume/src/storage/needle_map.rs
franchb: The characterization tests stay as live-set / leftover-key locks. They were never meant to fail on master. The put/delete tests still lock overwrite metrics and wrong-length-as-absent after the txn change. Fatal-unlink coverage was dropped because unlink is no longer fatal (see the sticky-bit fallback).

PR #11205, opened 2026-09-07
rust volume: optional redb insert_before bulk load

franchb:
> **Stacked on #11203 (4/4).** This branch builds on the quick-repair change, so the diff here shows the full stack until that PR merges; only the `insert_before` commit is new. I'll rebase once #11203 lands.

## What

Behind a default-off cargo feature `redb-experimental-cursor`, `full_rebuild` bulk-loads with redb's experimental `CursorMut::insert_before`. The production binary stays on sorted `table.insert()`. CI unit tests run both feature settings.

## Why

redb 4.2.0 can pack even tighter with `insert_before` on an empty table, but the API is gated on `experimental_cursor` (and therefore `experimental-api-5`). That should not ship in the default volume-server binary.

## How

- `seaweed-volume` feature `redb-experimental-cursor = ["redb/experimental_cursor"]` is not in `default`.
- Default `full_rebuild` still inserts live keys in needle-id order via `table.insert()`.
- With the feature: `upper_bound_mut(Unbounded)` then ascending `insert_before`, then `close()`. `UnorderedKey` fails the rebuild (no `insert()` fallback); the existing error path drops the map and unlinks the `.rdb`.
- `experimental-api-5` drops inherent `'static ReadOnlyTable::get`. Lookups keep using `ReadableTable::get`; the match result is bound so the guard does not outlive the table. Lookup semantics are unchanged.
- CI: `rust-unit-tests` runs `cargo test` and `cargo test --features redb-experimental-cursor`. No `--features` on any `cargo build --release`.

## Tests

- Default: `cd seaweed-volume && cargo test --lib -- needle_map -- --test-threads=1` — 56 passed (cfg test compiled out).
- Feature on: `cargo test --lib --features redb-experimental-cursor -- needle_map -- --test-threads=1` — 57 passed, including `test_redb_full_rebuild_insert_before_matches_shuffled_live_set`.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11205" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved storage rebuild behavior, including correct handling of overwritten and deleted records.
  - Improved recovery when storage files are missing or rebuild operations encounter errors.
  - Fixed update and delete handling for invalid or mismatched stored values.
  - Improved checkpoint repair behavior for durable commits.

- **Tests**
  - Added coverage for rebuild consistency, cursor-based processing, overwrite metrics, and invalid deletion scenarios.
  - Added automated testing for the experimental cursor configuration.

### DIFF
--- a/.github/workflows/rust-volume-server-tests.yml
+++ b/.github/workflows/rust-volume-server-tests.yml
@@ -62,6 +62,9 @@ jobs:
       - name: Run Rust unit tests
         run: cd seaweed-volume && cargo test
 
+      - name: Run Rust unit tests (redb experimental cursor)
+        run: cd seaweed-volume && cargo test --features redb-experimental-cursor
+
   rust-integration-tests:
     name: Rust Integration Tests
     runs-on: ubuntu-22.04
--- a/.github/workflows/rust-volume-server-tests.yml
+++ b/.github/workflows/rust-volume-server-tests.yml
@@ -63,7 +63,7 @@ jobs:
         run: cd seaweed-volume && cargo test
 
       - name: Run Rust unit tests (redb experimental cursor)
-        run: cd seaweed-volume && cargo test --features redb-experimental-cursor
+        run: cd seaweed-volume && cargo test --features redb-experimental-cursor --lib storage::needle_map
 
   rust-integration-tests:
     name: Rust Integration Tests
--- a/seaweed-volume/src/storage/needle_map.rs
+++ b/seaweed-volume/src/storage/needle_map.rs
@@ -489,13 +489,16 @@ impl RedbNeedleMap {
     /// much of the .idx the table reflects, so a reload replays only the
     /// tail appended after it. When `sync_idx` is true the .idx is fsynced
     /// first (see [`checkpoint`](Self::checkpoint)).
+    /// Durable commit uses `set_quick_repair(true)` so the next open after a
+    /// crash does not scan the whole file.
     fn begin_checkpoint(&self, sync_idx: bool) -> io::Result<redb::WriteTransaction> {
         if sync_idx {
             self.sync()?;
         }
-        let txn = self.db.begin_write().map_err(|e| {
+        let mut txn = self.db.begin_write().map_err(|e| {
             io::Error::new(io::ErrorKind::Other, format!("redb begin_write: {}", e))
         })?;
+        txn.set_quick_repair(true);
         if self.idx_file.is_some() {
             let mut meta = txn.open_table(META_TABLE).map_err(|e| {
                 io::Error::new(io::ErrorKind::Other, format!("redb open meta: {}", e))
--- a/seaweed-volume/src/storage/needle_map.rs
+++ b/seaweed-volume/src/storage/needle_map.rs
@@ -13,6 +13,9 @@ use std::io::{self, Read, Seek, Write};
 use std::path::Path;
 use std::sync::atomic::{AtomicI64, AtomicU64, Ordering};
 
+#[cfg(feature = "redb-experimental-cursor")]
+use std::ops::Bound;
+
 mod compact_map;
 pub mod file_pool;
 mod idx_metric;
@@ -570,14 +573,18 @@ impl RedbNeedleMap {
         let meta = txn
             .open_table(META_TABLE)
             .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("redb open meta: {}", e)))?;
-        match meta.get(META_IDX_SIZE) {
+        // experimental-api-5 drops inherent ReadOnlyTable::get ('static guard).
+        // ReadableTable::get guard borrows `meta`; bind the match so the
+        // temporary Result is dropped before `meta`.
+        let result = match meta.get(META_IDX_SIZE) {
             Ok(Some(guard)) => Ok(Some(guard.value())),
             Ok(None) => Ok(None),
             Err(e) => Err(io::Error::new(
                 io::ErrorKind::Other,
                 format!("redb get meta: {}", e),
             )),
-        }
+        };
+        result
     }
 
     /// Load from an .idx file, reusing an existing .rdb if it is consistent.
@@ -767,11 +774,38 @@ impl RedbNeedleMap {
                     })?;
                 }
 
-                for (key, nv) in &entries {
-                    let key_u64: u64 = (*key).into();
-                    let packed = pack_needle_value(nv);
-                    table.insert(key_u64, packed.as_slice()).map_err(|e| {
-                        io::Error::new(io::ErrorKind::Other, format!("redb insert: {}", e))
+                #[cfg(not(feature = "redb-experimental-cursor"))]
+                {
+                    for (key, nv) in &entries {
+                        let key_u64: u64 = (*key).into();
+                        let packed = pack_needle_value(nv);
+                        table.insert(key_u64, packed.as_slice()).map_err(|e| {
+                            io::Error::new(io::ErrorKind::Other, format!("redb insert: {}", e))
+                        })?;
+                    }
+                }
+                #[cfg(feature = "redb-experimental-cursor")]
+                {
+                    let mut cursor = table
+                        .upper_bound_mut(Bound::<u64>::Unbounded)
+                        .map_err(|e| {
+                            io::Error::new(
+                                io::ErrorKind::Other,
+                                format!("redb upper_bound_mut: {}", e),
+                            )
+                        })?;
+                    for (key, nv) in &entries {
+                        let key_u64: u64 = (*key).into();
+                        let packed = pack_needle_value(nv);
+                        cursor.insert_before(key_u64, packed.as_slice()).map_err(|e| {
+                            io::Error::new(
+                                io::ErrorKind::Other,
+                                format!("redb insert_before: {}", e),
+                            )
+                        })?;
+                    }
+                    cursor.close().map_err(|e| {
+                        io::Error::new(io::ErrorKind::Other, format!("redb cursor close: {}", e))
                     })?;
                 }
             }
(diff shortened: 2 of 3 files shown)


seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +717,107 @@ impl RedbNeedleMap {
...
-                    })?;
-                } else {
-                    // Entry was deleted — remove from redb if present
-                    table.remove(key_u64).map_err(|e| {
-                        io::Error::new(io::ErrorKind::Other, format!("redb remove: {}", e))
+            let mut live: Vec<(NeedleId, NeedleValue)> = entries
+                .into_iter()
+                .filter_map(|(k, v)| v.map(|nv| (k, nv)))
+                .collect();
+            live.sort_by_key(|(k, _)| *k);
```
franchb: **`full_rebuild` roughly doubles peak transient memory.**

`entries.into_iter().filter_map(..).collect()` keeps the whole `HashMap<NeedleId, Option<NeedleValue>>` allocation alive while `live` fills — hashbrown frees the table only when the `IntoIter` is dropped, i.e. after `collect()` returns — and `sort_by_key` is a *stable* sort, so it allocates another ~n/2 scratch buffer on top.

Rough per-entry cost: map ~29 B + vec 24 B + sort scratch ~12 B. A 10M-needle volume whose `.rdb` is missing or stale goes from ~290 MB transient at startup to ~600+ MB, and several volumes rebuilding concurrently on one server can OOM — which is awkward inside a stack whose stated purpose is bounding redb memory (#11179).

One-token fix for a third of it: `sort_unstable_by_key`. `HashMap` keys are unique, so stability buys nothing here. Overlapping the map and the vec (e.g. draining in chunks) needs a small restructure if you want the rest.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -707,57 +717,107 @@ impl RedbNeedleMap {
...
-            .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("redb commit: {}", e)))?;
-
-        nm.save_idx_size_meta(idx_size)?;
-        nm.metric = metrics_from_idx(reader, version)?;
+            txn.commit()
+                .map_err(|e| io::Error::new(io::ErrorKind::Other, format!("redb commit: {}", e)))?;
 
-        Ok(nm)
+            nm.save_idx_size_meta(idx_size)?;
+            nm.metric = metrics_from_idx(reader, version)?;
```
franchb: **A valid, already-committed `.rdb` gets unlinked when only metric computation fails.**

`metrics_from_idx` runs *after* `txn.commit()` and `save_idx_size_meta`, so any error it returns falls into the new

```rust
Err(e) => { drop(nm); let _ = std::fs::remove_file(db_path); Err(e) }
```

arm. A transient read error on the `.idx` (EIO, or a short read on a network mount) at this point throws away a correct, fully-built database, and the next mount pays a second full rebuild.

`metrics_from_idx` seeks independently (`needle_map/idx_metric.rs:86`), so hoisting it above the write transaction is a safe reorder and makes the cleanup arm mean what it says.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -1487,6 +1555,128 @@ mod tests {
...
+        assert!(reloaded.get(NeedleId(1)).unwrap().is_some());
+        assert!(
+            reloaded.get(NeedleId(99)).unwrap().is_none(),
+            "full_rebuild must not keep keys that are not in the .idx"
+        );
+    }
+
+    #[cfg(feature = "redb-experimental-cursor")]
+    #[test]
+    fn test_redb_full_rebuild_insert_before_matches_shuffled_live_set() {
```
franchb: **This test adds no coverage, and doesn't exercise what the feature changes.**

It asserts a strict subset of `test_redb_full_rebuild_last_write_wins_from_shuffled_idx` (same fixture, minus the `v1.offset` check), and that test already runs under the feature in the new CI job — so as written it's dead weight.

More importantly, 5 keys fit in a single leaf, so nothing here exercises node splits, which is exactly where `CursorMut::insert_before` differs from `table.insert` (cross-page pending-insert buffering) and where a `close()` / poison failure would surface.

Worth replacing with a few-thousand-key shuffled rebuild that asserts the full live set reads back, and compares `WriteTransaction::stats()` `leaf_pages()` / `fragmented_bytes()` against the `table.insert()` path — that hits splits *and* actually substantiates the "packs tighter" claim, which is currently unmeasured.

.github/workflows/rust-volume-server-tests.yml
```diff
@@ -62,6 +62,9 @@ jobs:
       - name: Run Rust unit tests
         run: cd seaweed-volume && cargo test
 
+      - name: Run Rust unit tests (redb experimental cursor)
+        run: cd seaweed-volume && cargo test --features redb-experimental-cursor
```
franchb: **This re-runs the whole suite when only one module is feature-sensitive.**

Only `storage::needle_map` changes behaviour under the flag, but this step recompiles redb, `seaweed-volume` and every test binary under a different feature resolution and then runs everything — in a job that already does `cargo build --release` + `cargo test` under `timeout-minutes: 15`.

The two feature sets also produce separate fingerprints under a single `hashFiles('Cargo.lock')` cache key, so the cached `target/` roughly doubles.

```yaml
run: cd seaweed-volume && cargo test --features redb-experimental-cursor --lib storage::needle_map
```

gives identical coverage.

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -57,6 +60,15 @@ fn unpack_needle_value(bytes: &[u8; PACKED_NEEDLE_VALUE_SIZE]) -> NeedleValue {
...
 }
 
+fn packed_to_needle_value(bytes: &[u8]) -> Option<NeedleValue> {
+    if bytes.len() != PACKED_NEEDLE_VALUE_SIZE {
+        return None;
+    }
+    let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
+    arr.copy_from_slice(bytes);
+    Some(unpack_needle_value(&arr))
+}
```
franchb, line 70: **Three call sites still inline what this helper replaces.**

`save_to_idx` (~L1069), `ascending_visit` (~L1100) and `collect_entries` (~L1128) each still carry the

```rust
if bytes.len() == PACKED_NEEDLE_VALUE_SIZE {
    let mut arr = [0u8; PACKED_NEEDLE_VALUE_SIZE];
    arr.copy_from_slice(bytes);
    let nv = unpack_needle_value(&arr);
```

block verbatim. The PR migrated four sites and missed these three, so they're now the only places where the length invariant is spelled out by hand — which is the thing this helper exists to centralize.

franchb (commented):
Automated review pass (high effort) on this PR — five findings, one medium, four low. Nothing blocking; the `full_rebuild` memory one is the only one I'd call out as worth doing before merge, since it cuts against the stated goal of the #11179 stack.

**Checked and cleared**, for the record:

- Cursor usage matches redb 4.2.0's own doc example verbatim (`upper_bound_mut(Bound::<u64>::Unbounded)` + ascending `insert_before` + `close()`). `NeedleId` derives `Ord` over a `u64`, so `sort_by_key` ordering agrees with redb's key ordering, and `UnorderedKey` can't fire on a freshly-unlinked (empty) table. An early `?` out of the loop drops the cursor, which poisons the transaction, so a lost insert can't be committed.
- `full_rebuild` dropping deleted keys rather than storing negative-size tombstones matches Go's `doLoading` *and* this crate's own `CompactNeedleMap::delete_from_map`. `try_reuse_rdb`'s incremental replay is the divergent side — pre-existing, not this PR.
- `set_quick_repair` genuinely takes effect: every non-checkpoint commit is `Durability::None`, so the last on-disk commit is always a quick-repair one.
- The new write-transaction placement in `delete()` is sound — an aborted transaction deregisters via `TransactionGuard::drop` → `end_write_transaction`, so a delete-miss leaks no tracker entry. It trades a read transaction for a begin+abort write transaction on the miss path and holds the write lock across the `.idx` append, but `delete` takes `&mut self`, so there's no added contention.
- `put()` using `table.insert`'s returned old value is equivalent to the removed `get_internal`, and the new error handling is strictly better: `begin_write` / `open_table` / `insert` failures now truncate the orphan `.idx` row, which the old code only did for commit failures.
- `let result = match ..; result` is load-bearing under edition-2021 tail-expression temporary scoping, not redundant.
- The experimental feature never reaches a shipped artifact — no `--all-features` anywhere, and every release/container workflow uses default or `--no-default-features`.

<sub>🤖 Automated review via [Claude Code](https://claude.com/claude-code)</sub>

seaweed-volume/src/storage/needle_map.rs
```diff
@@ -785,30 +845,35 @@ impl RedbNeedleMap {
...
-            self.truncate_idx_to_offset();
-            return Err(io::Error::new(io::ErrorKind::Other, format!("redb commit: {}", e)));
-        }
+            Ok(old)
+        })() {
+            Ok(old) => old,
+            Err(e) => {
+                self.truncate_idx_to_offset();
+                return Err(e);
+            }
```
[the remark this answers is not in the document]
franchb: Not taking this. Puts use `Durability::None`; the rustdoc about a commit that may already have become durable is the Immediate/2PC case. Truncating the `.idx` row on a failed `None` commit is the pre-PR `put` behavior and keeps a later checkpoint from recording an unreflected append. Failures before `commit()` still truncate.

seaweed-volume/src/storage/needle_map.rs
franchb: Fixed on #11202 (`BTreeMap`) and restacked. No overlapping map+vec, no stable-sort scratch.

seaweed-volume/src/storage/needle_map.rs
franchb: `metrics_from_idx` now runs before the rebuild write txn. It seeks on its own. A metric-read error unlinks only the new empty `.rdb`, not a committed rebuild.

seaweed-volume/src/storage/needle_map.rs
franchb: Replaced the 5-key clone with `test_redb_full_rebuild_insert_before_reads_back_thousands_of_shuffled_keys`: 4000 keys in reverse id order, overwrite 1, delete 2, then assert the live set (including a full `ascending_visit` count). No `leaf_pages` / `fragmented_bytes` assertion — `insert_before` vs already-sorted `insert()` is equal or only slightly better, so a numeric packing check would flake or assert nothing.

.github/workflows/rust-volume-server-tests.yml
franchb: CI extra step is now `cargo test --features redb-experimental-cursor --lib storage::needle_map`.

seaweed-volume/src/storage/needle_map.rs
franchb: `save_to_idx`, `ascending_visit`, and `collect_entries` now go through `packed_to_needle_value` (on #11204, restacked here).

PR #11228, opened 2026-09-08
admin: allow insecurely binding to any IP if -allowInsecureNoAuth is set

AlessandroZanatta:
#11185 removed the possibility to expose the admin UI without mTLS or an admin password. 

However, this breaks deployments that don't need either of those. For example, one may prevent network access to the admin UI if not to authorized users, or use forward auth from an IdP to perform authentication on a gateway.

This PR is adding the possibility to opt-out of the new protection, allowing the user to run the admin UI in an "insecure" configuration.

I've added a simple CLI flag to the admin options, called `allowInsecureNoAuth`, which allows to bypass the admin password and mTLS checks and start the admin UI anyway, without those. This flag is an opt-in, thus defaulting to the new behaviour introduced by #11185. When set, the flag also prints a warning to the user, though I think the flag name is pretty self-explanatory.

# How is the PR tested?
I don't believe there are tests that can be extended to verify this behaviour, but correct me if I missed them. Anyway, I don't think there is a need to add tests for a very simple CLI flag that is only used locally in a single file.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

* **New Features**
  * Added an explicit option to allow unauthenticated admin access on non-loopback addresses.
  * Displays a clear warning when this insecure configuration is enabled.
  * Documents the security implications of enabling the option.

* **Bug Fixes**
  * Non-loopback unauthenticated admin access remains blocked unless explicitly permitted.

### DIFF
--- a/weed/command/admin.go
+++ b/weed/command/admin.go
@@ -51,17 +51,17 @@ type AdminOptions struct {
 	readOnlyUser     *string
 	readOnlyPassword *string
 	// nil for callers other than runAdmin (e.g. `weed mini`)
-	allowInsecureNoAuth *bool
-	dataDir             *string
-	icebergPort         *int
-	lancePort           *int
-	urlPrefix           *string
-	metricsHttpPort     *int
-	metricsHttpIp       *string
-	debug               *bool
-	debugPort           *int
-	cpuProfile          *string
-	memProfile          *string
+	allowInsecureBind *bool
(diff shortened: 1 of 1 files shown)


AlessandroZanatta:
Go build action has failed. Looks like a transient error when trying to download remote packages, force pushing to re-run it!

PR #11229, opened 2026-09-08
filer/postgres: create filemeta table on startup via createTable config

### LINKED ISSUE #383: postgres database table is not created with the flag createTable = true
## Summary
I want to use postgres as external database so I configure the relevant secret as it is possible since https://github.com/seaweedfs/seaweedfs-operator/issues/344 with the value : 
```yaml extract of my decoded secret
    [leveldb2]
    enabled = false
    [postgres]
    enabled = true
    createTable = true
    username = "myuser"
    password = "mypassword"
    database = "seaweed"
    hostname = "postgres-rw.mynamespace.svc.cluster.local"
    port = 5434
```
## Current behaviour
the table is not created in seaweed database even if the parameter createTable is set to true
Consequently, filer pod is in crash loop back with the following message : 
```
F0904 08:31:14.509854 filer.go:173 read filer.store.id= : kv get: ERROR: relation "filemeta" does 
not exist (SQLSTATE 42P01)                                                                        
```
When I create the table manually, the filer pod starts without error.

## Expected behaviour
the parameter must create the table when its value is true

chrislusf:
## Summary

The `postgres` filer store hardcoded `CreateTableSqlTemplate` to empty and never created the `filemeta` table, unlike `postgres2`/`mysql2`/`sqlite` which all create it during `Initialize`. Users had to create the table manually or the filer would crash loop with:

```
F filer.go:173 read filer.store.id= : kv get: ERROR: relation "filemeta" does not exist (SQLSTATE 42P01)
```

Reported in seaweedfs/seaweedfs-operator#383 — the user set `createTable = true` under `[postgres]` but the flag was silently ignored.

### Changes (one commit per logic change)

1. **`filer/postgres: create default filemeta table on startup`** — Read the `createTable` config option (same as `postgres2`), default to `DefaultCreateTableQuery` when unset, and execute `CREATE TABLE IF NOT EXISTS` on the default table after the connection pool is established. `SupportBucketTable` stays `false` so per-bucket table creation remains a no-op; only the shared `filemeta` table is created, via direct `ExecContext` since `AbstractSqlStore.CreateTable` short-circuits without bucket support.

2. **`filer/postgres: accept boolean createTable = true/false`** — viper reads a TOML boolean as the string `"true"`/`"false"` via `GetString`, so `createTable = true` was being used as a SQL template and failed. Add `ResolveCreateTableQuery` to normalize the value: `true` and empty select the default template, `false` disables table creation, anything else is a custom template. Both `postgres` and `postgres2` now use it, and both skip the `CREATE TABLE` call when the resolved template is empty.

3. **`scaffold: document createTable option for postgres filer store`** — Replace the commented-out `CREATE TABLE` SQL in the `[postgres]` scaffold with a `createTable` config hint, matching the `[postgres2]` section.

#### Test plan
- [x] `go build ./weed/...`
- [x] `go test ./weed/filer/postgres/... ./weed/filer/postgres2/... ./weed/filer/abstract_sql/...`
- [x] `go vet ./weed/filer/postgres/... ./weed/filer/postgres2/...`
- [x] New unit test `TestResolveCreateTableQuery` covers empty/true/false/custom
- [ ] Manual: start filer with `[postgres] createTable = true` and verify `filemeta` table is created

Closes seaweedfs/seaweedfs-operator#383

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11229" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * PostgreSQL filer setup now supports configurable metadata and bucket table creation.
  * Choose the default schema, provide a custom SQL template, or skip automatic table creation.
  * PostgreSQL storage initialization applies the selected table-creation behavior during setup, including support for restricted database roles.

* **Documentation**
  * Updated configuration guidance to explain supported table-creation options and defaults.

### DIFF
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -106,8 +106,9 @@ enableUpsert = true
 upsertQuery = """INSERT INTO `%s` (`dirhash`,`name`,`directory`,`meta`) VALUES (?,?,?,?) AS `new` ON DUPLICATE KEY UPDATE `meta` = `new`.`meta`"""
 
 [postgres] # or cockroachdb, YugabyteDB
-# Set to true to auto-create the filemeta table, or provide a custom CREATE TABLE template.
-# createTable = true
+# createTable = true          # auto-create filemeta with the default schema
+# createTable = false         # skip table creation (default; for restricted DB roles)
+# createTable = """CREATE TABLE IF NOT EXISTS "%s" (...)"""  # custom template
 enabled = false
 hostname = "localhost"
 port = 5432
--- a/weed/command/scaffold/filer.toml
+++ b/weed/command/scaffold/filer.toml
@@ -106,13 +106,8 @@ enableUpsert = true
 upsertQuery = """INSERT INTO `%s` (`dirhash`,`name`,`directory`,`meta`) VALUES (?,?,?,?) AS `new` ON DUPLICATE KEY UPDATE `meta` = `new`.`meta`"""
 
 [postgres] # or cockroachdb, YugabyteDB
-# CREATE TABLE IF NOT EXISTS filemeta (
-#   dirhash     BIGINT,
-#   name        VARCHAR(65535),
-#   directory   VARCHAR(65535),
-#   meta        bytea,
-#   PRIMARY KEY (dirhash, name)
-# );
+# Set to true to auto-create the filemeta table, or provide a custom CREATE TABLE template.
+# createTable = true
 enabled = false
 hostname = "localhost"
 port = 5432
--- a/weed/filer/abstract_sql/abstract_sql_store.go
+++ b/weed/filer/abstract_sql/abstract_sql_store.go
@@ -32,6 +32,7 @@ type AbstractSqlStore struct {
 	DB                     *sql.DB
 	KvDB                   *sql.DB
 	SupportBucketTable     bool
+	SkipDDL                bool
 	dbs                    map[string]bool
 	dbsLock                sync.Mutex
 	RetryableErrorCallback func(err error) bool
@@ -40,7 +41,7 @@ type AbstractSqlStore struct {
 var _ filer.BucketAware = (*AbstractSqlStore)(nil)
 
 func (store *AbstractSqlStore) CanDropWholeBucket() bool {
-	return store.SupportBucketTable
+	return store.SupportBucketTable && !store.SkipDDL
 }
 func (store *AbstractSqlStore) OnBucketCreation(bucket string) {
 	store.dbsLock.Lock()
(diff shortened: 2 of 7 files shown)


chrislusf:
Addressed the review feedback in the latest commits:

**postgres opt-in (Greptile/Devin P1):** `ResolveCreateTableQuery` now returns empty for an unset `createTable`, so the `postgres` store only creates the `filemeta` table when `createTable` is explicitly `true` or a custom template. Existing deployments with restricted DB roles and pre-provisioned tables are unaffected — no DDL runs unless the user opts in.

**postgres2 `false` handling (Greptile/Devin):** `createTable = false` now skips the default `filemeta` table creation on startup. Per-bucket `CreateTable` calls are unaffected — they are a runtime requirement of the bucket-aware store. Users who need to suppress all DDL should use the `postgres` (non-bucket) store with `createTable` unset.

chrislusf:
Addressed the remaining review feedback (CodeRabbit + Devin on postgres2 `createTable = false`):

**Per-bucket table creation with `false`:** `SupportBucketTable` is now set to `false` when `createTable` is explicitly `false`. Since `AbstractSqlStore.CreateTable` already returns `nil` immediately when `SupportBucketTable` is false, this disables ALL table creation — neither the default `filemeta` table nor per-bucket tables are created. The template stays empty, no DDL runs, and all data routes to the pre-provisioned `filemeta` table (same behavior as the `postgres` non-bucket store). This honours the explicit opt-out for restricted DB roles.

The scaffold comment already documents all three modes (true/false/custom) for the `[postgres]` section.

chrislusf:
Revised the `createTable = false` handling based on the latest feedback (Devin + Greptile on bucket routing):

**Preserved bucket routing:** `SupportBucketTable` stays `true` when `createTable = false` — per-bucket routing continues to map `/buckets/<name>` paths to their per-bucket tables. Previously setting it to `false` would route all entries to `filemeta`, hiding objects in pre-provisioned per-bucket tables.

**Suppressed DDL without disabling routing:** `GetSqlCreateTable` now returns an empty string when `CreateTableSqlTemplate` is empty (for both postgres and mysql SQL generators), and `AbstractSqlStore.CreateTable` skips the `ExecContext` call when the SQL is empty. So no DDL runs — neither for the default `filemeta` table nor for per-bucket tables — but bucket routing is fully preserved. Users who pre-provisioned per-bucket tables will have their data routed correctly.

chrislusf:
Addressed the DROP TABLE concern (Devin on bucket deletion):

**Added `SkipDDL` flag to `AbstractSqlStore`:** Independent of `SupportBucketTable`, this flag suppresses both `CREATE TABLE` and `DROP TABLE` execution. `CanDropWholeBucket` now returns `false` when `SkipDDL` is set, so bucket deletion falls back to row-by-row metadata deletes instead of dropping the whole table. `postgres2` sets `SkipDDL = true` when `createTable = false` — bucket routing is preserved, no DDL runs (neither CREATE nor DROP), and bucket deletion removes metadata without dropping externally managed tables.

PR #11233, opened 2026-09-08
fix(volume): handle faulty storage media (Go + Rust)

### LINKED ISSUE #11227: SeaweedFS does not handle faulty storage media
Storage R/W failures in Seaweed are propagated upstream as any other error type and never handled appropriately, even when the root cause is clearly faulty hardware. As a result, Seaweed always tries to serve data from faulty media, _even when healthy replicas / shards are available_:

```
I0805 06:39:06.290571 common.go:109 response method:DELETE URL:/123456,xxxxxxxxxxxxxxxxxx with httpStatus:500 and JSON:{"error":"Deletion Failed: locate in local ec volume: FindNeedleFromEcx: ecx file xxxxxx read at 2946440: read /data/123456.ecx: input/output error"}
I0805 06:39:06.290519 common.go:130 error JSON response status 500: Deletion Failed: locate in local ec volume: FindNeedleFromEcx: ecx file xxxxxxx read at 2946440: read /data/123456.ecx: input/output error
```

This behavior makes managing Seaweed clusters large enough that HDD/SSDs failures happen regularly particularly painful, as there's no way to detect/automate mitigations without manually inspecting the nodes hosting affected volumes.

Seaweed should address this by having servers bubble up R/W errors so these can be reacted upon:

  * Volume read/write attempts on broken media should be logged, and the affected replicas and/or EC shards automatically unmounted.
  * Any R/W failure on storage should be reflected in metrics and exported for monitoring.
  * Any R/W failure on storage should be reflected in `/healthz`, or similar, enabling the construction of liveness probes.

chrislusf:
## Summary

Closes #11227.

SeaweedFS previously propagated storage read/write failures upstream as ordinary errors without treating them as faulty hardware. This PR adds detection and handling of persistent storage-media failures in both the Go and Rust volume servers, mirroring the existing volume-replica EIO quarantine semantics:

- **EC shard EIO tracking + auto-unmount** (commit 1): Extract the volume EIO tracker into a reusable `IoErrorTracker` and add the same tracking to `EcVolume`. Sustained EIO on `.ecx` lookups or `.ecd` shard reads now unmounts the EC volume in the heartbeat (without deleting files) so the master re-replicates from healthy peers. Mirrored in Rust.
- **Storage IO error metrics** (commit 2): Add a `storage_io_error_total` counter incremented on every EIO recorded by the volume or EC shard tracker, and an `io_quarantine` gauge labelled by kind (`volume`/`ec_shard`) reflecting the count of replicas suppressed in the heartbeat. Mirrored in Go and Rust.
- **healthz reflects storage IO errors** (commit 3): Add `Store.HasIoQuarantine` (Go) / `Store::has_io_quarantine` (Rust) and have `/healthz` return 503 when any local volume or EC shard is quarantined due to sustained storage-media EIO, so a load balancer can drain a server whose underlying media is faulty. Mirrored in Go and Rust.

### Design notes

- A single EIO does not immediately remove or suppress a replica.
- Consecutive EIOs must reach a tolerance threshold (`3`, matching the existing volume tracker).
- A successful or non-EIO operation resets the active streak.
- Once heartbeat quarantines a replica, quarantine is sticky until an explicit operator action (`MarkVolumeWritable`); a stray successful read clears the streak but does not silently put a known-bad replica back into rotation.
- Quarantined replicas are not announced to the master and become no-write/no-delete.
- Local data is never physically deleted merely because the disk temporarily fails — the disk may be transiently bad and this could be the last good copy.
- EC shards are handled consistently with regular volumes, including read failures and heartbeat/unmount behavior.
- The Go `erasure_coding` package cannot import the parent `storage` package (import cycle), so EC tracker state lives directly on `EcVolume` with exported accessors.

#### Test plan
- [x] `go test ./weed/storage/` passes.
- [x] `go build ./weed/...` passes.
- [x] `cargo build` (seaweed-volume) passes.
- [ ] CI green.
- [ ] Review feedback addressed.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11233" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added storage I/O error tracking for volumes and erasure-coded shards.
  - Replicas with repeated storage I/O failures are quarantined and excluded from active heartbeat reporting for re-replication.
  - Health checks now return `503 Service Unavailable` when local replicas are quarantined, allowing traffic draining.
  - Added metrics for storage I/O errors and quarantined volumes or shards, including separate volume and shard counts.

### DIFF
--- a/seaweed-volume/src/metrics.rs
+++ b/seaweed-volume/src/metrics.rs
@@ -3,8 +3,8 @@
 //! Mirrors the Go SeaweedFS volume server metrics.
 
 use prometheus::{
-    self, Encoder, GaugeVec, HistogramOpts, HistogramVec, IntCounterVec, IntGauge, IntGaugeVec,
-    Opts, Registry, TextEncoder,
+    self, Encoder, GaugeVec, HistogramOpts, HistogramVec, IntCounter, IntCounterVec, IntGauge,
+    IntGaugeVec, Opts, Registry, TextEncoder,
 };
 use std::sync::Once;
 
@@ -171,6 +171,23 @@ lazy_static::lazy_static! {
         &["mode"],
     ).expect("metric can be created");
 
+    /// Counter of storage read/write EIO errors on volumes and EC shards.
+    /// Mirrors Go's VolumeServerStorageIoErrorCounter.
+    pub static ref STORAGE_IO_ERROR_COUNTER: IntCounter = IntCounter::new(
+        "SeaweedFS_volumeServer_storage_io_error_total",
+        "Counter of storage read/write EIO errors on volumes and EC shards.",
+    ).expect("metric can be created");
+
+    /// Number of volumes quarantined due to storage IO errors.
+    /// Mirrors Go's VolumeServerIoQuarantineGauge.
+    pub static ref IO_QUARANTINE_GAUGE: IntGaugeVec = IntGaugeVec::new(
+        Opts::new(
+            "SeaweedFS_volumeServer_io_quarantine",
+            "Number of volumes or EC shards quarantined due to storage IO errors.",
+        ),
+        &["kind"],
+    ).expect("metric can be created");
+
     // ---- Legacy aliases for backward compat with existing code ----
 
     /// Total number of volumes on this server (flat gauge).
@@ -283,6 +300,8 @@ pub fn register_metrics() {
             Box::new(SCRUB_LAST_TIME_SECONDS.clone()),
             Box::new(SCRUB_VOLUME_FAILURES.clone()),
             Box::new(SCRUB_SHARD_FAILURES.clone()),
+            Box::new(STORAGE_IO_ERROR_COUNTER.clone()),
+            Box::new(IO_QUARANTINE_GAUGE.clone()),
             // Legacy metrics
             Box::new(VOLUMES_TOTAL.clone()),
             Box::new(DISK_SIZE_BYTES.clone()),
--- a/seaweed-volume/src/server/grpc_server.rs
+++ b/seaweed-volume/src/server/grpc_server.rs
@@ -3334,9 +3334,14 @@ impl VolumeServer for VolumeGrpcService {
         while bytes_read < total_size {
             let chunk_size = std::cmp::min(BUFFER_SIZE_LIMIT, total_size - bytes_read);
             let mut buf = vec![0u8; chunk_size];
-            let n = shard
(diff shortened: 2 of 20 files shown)


chrislusf:
Addressed review 5147386333 in commit b72ad5d72:

1. **EIO streak reset moved to end of full write** — the success-side reset in `append_needle` (after `write_all` only) is now moved to the end of `do_write_request`, after `flush_dat`/`flush_idx` complete. A successful `write_all` followed by a failed `fsync` no longer resets the counter before the EIO is recorded, so repeated fsync EIOs now accumulate toward the quarantine threshold.

2. **Windows storage-media error codes** — `check_read_write_error` now recognizes `ERROR_CRC` (23) and `ERROR_IO_DEVICE` (1117) in addition to Unix `EIO` (errno 5), via a platform-aware `isStorageIoError` helper. Mirrored in both Go (`io_error.go` + `io_error_windows.go`/`io_error_other.go` build-tagged files) and Rust (`is_storage_io_error`).

Local validation: Go `./weed/...` build + `go test ./weed/storage/` pass; Rust `cargo test` 510 passed, 0 failed.

chrislusf:
Addressed Devin Review 5148063028 and CodeRabbit 5148126797 in commit 094d2047e:

**Devin Review:**
1. **Checkpoint media failures never accumulate** — `maybe_checkpoint_index` now returns `bool` indicating whether the checkpoint succeeded. The success-side EIO reset in `do_write_request` only fires when the checkpoint succeeded, so a checkpoint media failure is no longer erased by the unconditional reset.
2. **Successful deletes preserve failure streaks** — `do_delete_request` now clears the EIO streak after a successful delete (tombstone append + index update + checkpoint), mirroring `do_write_request`.

**CodeRabbit:**
1. **Unconditional reset after checkpoint** — same fix as Devin #1 above.
2. **Hard-coded EIO value** — `is_storage_io_error` now uses `libc::EIO` on Unix instead of a hard-coded `5`, with the comparison compiled only under `#[cfg(unix)]`.
3. **Windows ECX read path** — added a `seek` + `read_exact` fallback for `#[cfg(not(unix))]` in `find_needle_from_ecx`, so the buffer is no longer zeroed on Windows. Read errors are propagated through `check_read_write_error`.

Local validation: Go `./weed/...` build + `go test ./weed/storage/` pass; Rust `cargo test` 510 passed, 0 failed.

PR #11238, opened 2026-09-09
[Volume] Validate record counts after volume copy

### LINKED ISSUE #11224: [Volume] Validate received and deleted counts after volume copy
# Problem Description

`checkCopyFiles` currently validates only the sizes of the `.idx` and `.dat` files. It does not verify the received and deleted record counts, as noted in the existing TODO at [volume_grpc_copy.go:272](https://github.com/seaweedfs/seaweedfs/blob/master/weed/server/volume_grpc_copy.go#L272).

Image

Volumes with matching file sizes but different record counts may not be detected as inconsistent.

# Scope

- Compare the received and deleted counts after copying a volume.
- Return an error when the counts do not match.
- Add tests for matching counts, mismatches, empty volumes, and deleted records.
- Do not modify storage formats or public APIs.

# Impact Radius

Only affects volume copy validation. Normal file reads and writes are not affected.

# Tests

`go test ./weed/server/...`

ssshr-66:
Fixes [#11224](https://github.com/seaweedfs/seaweedfs/issues/11224)

Volume Copy only validates `.dat` and `.idx` file sizes. Volumes with matching file sizes but different live or deleted Needle counts may still be reported as successfully copied.

- Compare live and deleted Needle counts between the source and target volumes.
- Run strict count validation for read-only source volumes.
- Unmount the target volume and return an error when counts differ.
- Add table-driven tests for empty volumes, matching counts, and count mismatches.
- Do not modify storage formats, protobufs, public APIs, or normal read/write paths.

# How is the PR tested?

- `go test ./weed/server/...`
- `go test -race ./weed/server/...`
- `go vet ./weed/server/...`

# Checks

- [x] All AI code review comments have been addressed. No more comments to fix if reviewed again. Reviewer may request additional Gemini and Copilot reviews.

# Checks for AI generated PRs

- [x] The PR is kept as minimal as possible.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11238" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Replica copying now validates source and target file and deleted-record counts when reliable status snapshots are available.
  - Replicas with count mismatches are rejected before becoming available for routing.
  - Volumes are announced only after mount validation succeeds.
  - Failed validation leaves volumes unmounted and prevents them from being advertised as available.
  - Copy validation now uses the operation’s request context for status checks.

- **Tests**
  - Added coverage for stable count validation and mount behavior when validation succeeds or fails.


weed/storage/store.go
```diff
@@ -967,11 +967,30 @@ func (s *Store) MarkVolumeWritable(i needle.VolumeId) error {
 }
 
 func (s *Store) MountVolume(i needle.VolumeId) error {
+	return s.mountVolume(i, nil)
+}
+
+// MountVolumeWithValidator loads a volume, validates it before announcing it
+// to the master, and unloads it when validation fails. This keeps an invalid
+// newly copied replica out of the master's routable volume set.
+func (s *Store) MountVolumeWithValidator(i needle.VolumeId, validator func(*Volume) error) error {
```
chrislusf, line 976: Please don't add a new exported `Store` method here. `MountVolumeWithValidator` exists only to thread a validator into the mount path, and it widens the public API surface for a single internal caller.

Preferred: change `MountVolume` to take a variadic validator so the single public name is preserved and all existing callers stay source-compatible:

```go
func (s *Store) MountVolume(i needle.VolumeId, validators ...func(*Volume) error) error {
    // ... existing load ...
    for _, v := range validators {
        if err := v(vol); err != nil {
            location.UnloadVolume(i)
            return err
        }
    }
    // announce to master
}
```

Then the copy handler calls `vs.store.MountVolume(id, countValidator)` and every other call site (`MountVolume(id)`) is unchanged. No `mountVolume` helper, no new symbol. If you prefer to keep the announce-before-validate ordering guarantee another way, validate-then-`UnmountVolume` in the copy handler also avoids a new name, at the cost of a brief announce window.

weed/server/volume_grpc_copy.go
```diff
@@ -50,11 +50,20 @@ func (vs *VolumeServer) VolumeCopy(req *volume_server_pb.VolumeCopyRequest, stre
...
+	var sourceVolumeStatusAfterCopy *volume_server_pb.VolumeStatusResponse
 	var dataBaseFileName, indexBaseFileName, idxFileName, datFileName string
 	var hasRemoteDatFile bool
 	err := operation.WithVolumeServerClient(true, pb.ServerAddress(req.SourceDataNode), vs.grpcDialOption, func(client volume_server_pb.VolumeServerClient) error {
 		var err error
-		volFileInfoResp, err = client.ReadVolumeFileStatus(context.Background(),
+		sourceVolumeStatus, err = client.VolumeStatus(stream.Context(), &volume_server_pb.VolumeStatusRequest{
+			VolumeId: req.VolumeId,
+		})
+		if err != nil {
```
chrislusf, line 62: This pre-copy `VolumeStatus` is only used to detect count drift during the copy (via `copyCountsStable`). If it fails, the whole copy now aborts where it previously would have succeeded — a new failure mode for a value that's purely advisory.

Better: on pre-copy status error, log and set `sourceVolumeStatus = nil`. `copyCountsStable(nil, ...)` already returns false, so validation is skipped gracefully. Reserve hard failures for the post-copy status (line ~202), which is the one actually used for the comparison.

weed/server/volume_grpc_copy.go
```diff
@@ -50,11 +50,20 @@ func (vs *VolumeServer) VolumeCopy(req *volume_server_pb.VolumeCopyRequest, stre
...
+			return fmt.Errorf("read volume status failed, %w", err)
+		}
+
+		volFileInfoResp, err = client.ReadVolumeFileStatus(stream.Context(),
```
chrislusf, line 66: Switching `ReadVolumeFileStatus` from `context.Background()` to `stream.Context()` is an unannounced behavior change bundled into this PR: a client disconnect now cancels the status read where it didn't before. If this is intentional, call it out in the description and apply it consistently to the other `context.Background()` calls in this file; if not, keep `context.Background()` here to keep the diff focused on the count-validation fix.

weed/server/volume_grpc_copy.go
```diff
@@ -300,6 +329,22 @@ func checkCopyFiles(originFileInf *volume_server_pb.ReadVolumeFileStatusResponse
...
+	if origin.FileCount != targetFileCount {
+		return fmt.Errorf("target file count [%d] is not same as origin file count [%d]", targetFileCount, origin.FileCount)
+	}
+	if origin.FileDeletedCount != targetDeletedCount {
+		return fmt.Errorf("target deleted count [%d] is not same as origin deleted count [%d]", targetDeletedCount, origin.FileDeletedCount)
+	}
+	return nil
+}
+
+func copyCountsStable(before, after *volume_server_pb.VolumeStatusResponse) bool {
```
chrislusf, line 342: Two things here:

1. **Description mismatch:** the PR description says "Run strict count validation for read-only source volumes," but the code checks count *stability* between before/after snapshots, never `IsReadOnly`. Stability is actually the better proxy (a read-only flag can flip without count changes — your own test covers this), so the description should say "stable source volumes," not "read-only source volumes."

2. **Known limitation worth a comment:** for a writable source being written during the copy, before/after counts differ and validation is skipped entirely. That's a reasonable safety tradeoff (you can't validate a moving target), but it means this fix does *not* close the bug for in-flight writable volumes — only quiescent/read-only ones. #11224 doesn't restrict itself to read-only volumes, so please add a comment near `copyCountsStable` noting this limitation explicitly.

chrislusf (commented):
Thanks for the fix — validating live/deleted counts after a copy is a real gap in #11224, and the `checkCopyCounts`/`copyCountsStable` helpers are clean and well-tested. A few things to address before merge, the main one being the new public `Store` method.

ssshr-66:
Received. I’ll check and fix it.

PR #11249, opened 2026-09-09
Mount req with collection

shaofeng66:
# What problem are we solving?
issue #11248 

# How are we solving the problem?
add an optional collection field into VolumeMountRequest to shortcut the iteration
rust mirror is AI generated.

# How is the PR tested?

# Checks
- [ ] I have added unit tests if possible.
- [ ] I will add related wiki document changes and link to this PR after merging.
- [ ] All AI code review comments have been addressed. No more comments to fix if reviewed again. Reviewer may request additional gemini and copilot reviews.

# Checks for AI generated PRs
- [ ] I have reviewed every line of code.
- [ ] The PR is kept as minimum as possible. Large PRs would not be accepted.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11249" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Volume mount requests can now specify a collection.
  * Volumes are located more directly when a collection is provided, improving mounting for collection-specific volume files.
  * Volume copying and staged volume adoption preserve the specified collection during mounting.

PR #11252, opened 2026-09-09
fix(volume): Harden Volume Copy Validation and Failure Handling

### LINKED ISSUE #11250: [Volume] Follow up on volume copy validation issues after #11238
Related PR: #11238

## Problem Description

PR #11238 added live and deleted record count validation for Volume Copy. After the PR was merged, follow-up review identified several issues that still need to be addressed:

- A failure in the pre-copy `VolumeStatus` request aborts the entire copy, although this status is only used to detect count changes.
- A failure in the post-copy `VolumeStatus` request currently disables count validation, allowing an unvalidated target volume to be mounted and advertised.
- The implementation adds a new exported `Store` method, `MountVolumeWithValidator`, for a single internal caller.
- `ReadVolumeFileStatus` was changed from `context.Background()` to the request context without being part of the original issue scope.
- The PR description does not clearly document that validation is skipped when a writable source changes during the copy.

## Proposed Changes

- Treat pre-copy `VolumeStatus` failures as advisory and continue the copy without strict count validation.
- Treat post-copy `VolumeStatus` failures as copy failures instead of silently skipping validation.
- Reuse the existing `MountVolume` API with an optional validator, or use another approach that does not add a new exported `Store` method.
- Keep unrelated context behavior unchanged unless it is intentionally applied consistently.
- Document the limitation that count validation is skipped for moving writable sources.
- Add regression tests for these cases.

## Impact Radius

This only affects the Volume Copy validation and target volume publication path.

It does not modify storage formats, protobuf definitions, normal file reads or writes, or the replication data format.

## Tests

- Pre-copy status failure does not unnecessarily abort the copy.
- Post-copy status failure prevents the target volume from being mounted or advertised.
- Count mismatches are detected before the target is announced.
- Existing `MountVolume` callers remain compatible.
- Writable sources that change during copying are handled without false validation failures.

ssshr-66:
fix [#11250](https://github.com/seaweedfs/seaweedfs/issues/11250)

After #11238, Volume Copy still had several failure-handling issues. A failure when reading the source status before copying could abort a valid copy, while a failure after copying could allow an unvalidated volume to be mounted and announced.

- Treat pre-copy status failures as advisory and skip strict count validation.
- Fail and clean up when post-copy status retrieval fails.
- Validate the target volume before announcing it to the master.
- Reuse the existing `MountVolume` flow instead of adding a new exported API.
- Restore the original context behavior for `ReadVolumeFileStatus`.
- Add regression tests for failure handling, cleanup, and mount validation.

# How is the PR tested?

- `go test ./weed/server/... ./weed/storage/...`
- `go test -race ./weed/server/...`
- `go vet ./weed/server/... ./weed/storage/...`

# Checks

# Checks for AI generated PRs

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11252" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Volume copying now continues when the initial source status cannot be read, while still failing safely if the final status is unavailable.
  * Failed validation during volume mounting no longer announces or leaves the volume mounted.
  * Copy failures clean up mounted volumes and partial copy artifacts more reliably.
  * Existing destination replicas are preserved when no destination space is available.

* **Improvements**
  * Destination selection can replace an existing replica when the location has reached its slot limit.
  * Volume mounting validates data before announcing it.
  * Copy validation better accounts for status changes during concurrent operations.


chrislusf:
Addressed the review comments in `f90cafa0d`:

**Devin Review — "Failed replacement deletes healthy replica" (BUG/Major):** Valid. When the initial `VolumeStatus` check became advisory, `VolumeCopy` could delete an existing replica before `FindFreeLocation` reserved a destination. If the target was then full, the copy failed with the original replica already destroyed. Fixed by reordering: `FindFreeLocation` now runs before `DeleteVolume`, so a full target fails safely with the existing replica intact. Added `TestVolumeCopyKeepsExistingReplicaWhenDestinationFull` as a regression test.

**CodeRabbit — "Keep MountVolume source-compatible" (Major):** Not applicable. The `MountVolume(i, validators ...)` signature was introduced in this PR itself (never released), and the `collection *string` parameter was added by #11249 on master. The merge combines both into `MountVolume(i, collection *string, validators ...)`, and all in-tree callers are updated. There are no external callers of the unreleased `validators`-only signature to break.

chrislusf:
Addressed the latest review comments in `d10da3cc7`:

**Devin Review — "Existing replica blocks its replacement" (BUG/Major):** Valid. Moving `FindFreeLocation` before `DeleteVolume` meant a location at its `MaxVolumeCount` limit could not replace its sole replica — the slot was still occupied. Fixed by extending `FindFreeLocation` with a variadic `replaceVid` parameter (backward-compatible): the location holding the volume being replaced gets +1 free slot. `VolumeCopy` passes the volume ID, so destination selection succeeds before the existing replica is deleted. Added `TestVolumeCopyReplacesReplicaAtSlotLimit` covering a single-slot location replacing its only replica.

**CodeRabbit — "Reserve destination capacity before deleting" (Major):** Not addressed in this PR. `FindFreeLocation` has never reserved capacity atomically — concurrent copies selecting the same location is a pre-existing race, not a regression from this PR. An atomic reservation API is a larger change better suited to a separate PR.

chrislusf:
**CodeRabbit — "Keep MountVolume compatible with its callers" (outside-diff, Major):** False positive. `VolumeMountRequest.Collection` is `*string` in the protobuf (`optional string collection = 2` generates `Collection *string`), so `vs.store.MountVolume(needle.VolumeId(req.VolumeId), req.Collection)` in `volume_grpc_admin.go` is already type-correct. `go build ./weed/server/...` passes.

PR #11263, opened 2026-09-10
fix(vacuum): stop comparing compact size against the live needle map

Koalk:
CompactByIndex's post-copy integrity check compares bytes written to the new `.cpd` against `v.nm.ContentSize()-DeletedSize()` — the volume's live needle map, which keeps mutating for as long as the volume stays writable during the copy. Any write landing after the point-in-time index snapshot (`oldNm`) was loaded makes the live map's tally exceed what was actually copied, and the check aborts compaction with `unexpected new data size`, even though `CommitCompact`'s `makeupDiff` exists specifically to reconcile writes that land mid-copy afterward. On a volume under sustained write traffic this can fail every vacuum cycle, since there's no guarantee a compaction window will ever see zero writes.

This fix tallies the expected live size from `oldNm` — the same frozen snapshot the copy loop actually reads from — instead of the live map. This keeps the check's original intent (a destination file smaller than what should have been copied still signals real data loss, e.g. a failed flush) while removing the false positive from ordinary concurrent traffic.

Added `TestCompactByIndex_ConcurrentWriteDoesNotFailIntegrityCheck`, which injects a write via `ProgressCallback` partway through the copy loop to reproduce the race; it fails with the exact `unexpected new data size` error against the old check and passes against the fix.

A follow-up commit fixes a second issue caught by review (Devin, Greptile, and CodeRabbit independently flagged the same bug): unreadable needles are dropped before reaching the `expectedLiveBytes` tally, so it already excludes them, but the check was also subtracting `skippedDataBytes` from the threshold — a double-subtraction that loosened the check's margin by that same amount. The comparison is now extracted into `exceedsExpectedCompactedSize` and the subtraction removed, with `TestExceedsExpectedCompactedSize` pinning the threshold to `expectedLiveBytes` alone.

Verified with `go test ./weed/storage/...` and `go build ./...`.

# Checks

# Checks for AI generated PRs

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved compaction integrity checks to accurately reflect data present when compaction begins.
  - Prevented compaction from failing when live writes occur during the copy phase.
  - Improved detection and reporting when compacted data exceeds the expected live content size.
  - Ensured writes made during compaction remain available after the operation completes.

- **Tests**
  - Added regression coverage for concurrent writes during compaction.
  - Added coverage for exact-size, oversized, and undersized compaction results.

### DIFF
--- a/seaweed-volume/src/storage/volume.rs
+++ b/seaweed-volume/src/storage/volume.rs
@@ -94,6 +94,12 @@ fn is_skippable_needle_read_error(e: &VolumeError) -> bool {
     }
 }
 
+/// Reports whether the compacted .dat is short of the live bytes the
+/// pre-compaction index snapshot expected.
+fn exceeds_expected_compacted_size(expected_live_bytes: u64, dst_dat_size: u64) -> bool {
+    expected_live_bytes > dst_dat_size
+}
+
 /// Returns true for I/O errors that indicate faulty storage media, not
 /// transient/network failures. On Unix this is EIO; on Windows it covers
 /// ERROR_CRC and ERROR_IO_DEVICE, which the kernel returns for failing disks.
@@ -3592,6 +3598,7 @@ impl Volume {
 
         let mut skipped_needles: u64 = 0;
         let mut skipped_data_bytes: u64 = 0;
+        let mut expected_live_bytes: u64 = 0;
         for (id, offset, size) in entries {
             // Progress callback
             if !progress_fn(offset.to_actual_offset()) {
@@ -3655,6 +3662,11 @@ impl Volume {
                 }
             }
 
+            // Tally the live bytes from the frozen snapshot this loop copied
+            // from, not the live needle map. Unreadable needles return before
+            // this point, so no further skipped-byte adjustment is needed.
+            expected_live_bytes += size.0 as u64;
+
             // Write needle to destination
             let bytes = n.write_bytes(version);
             dst.write_all(&bytes)?;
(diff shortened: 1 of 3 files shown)


chrislusf (commented):
Independent review with verification — no blocking issues.

**Verification performed**
- `go test ./weed/storage/...` passes with the fix, including the two new tests and the existing `TestCompactByIndex_DropsDanglingNeedle`.
- Ran `TestCompactByIndex_ConcurrentWriteDoesNotFailIntegrityCheck` against master's pre-PR vacuum logic: it fails with exactly the described error (`volume 1 unexpected new data size: 3848 does not match size of content minus deleted: 4288` — the 440-byte gap is the mid-copy write), confirming the test genuinely reproduces the race. It passes with the fix.
- CI is fully green (Build, Test, Go Vet incl. 32-bit, Vacuum/EC/FUSE integration, CodeQL).
- All prior review comments (Devin, Greptile, CodeRabbit) flagging the double-subtraction are resolved in d503fe3.

**Correctness**
- `expectedLiveBytes` is tallied only for needles that pass the read + TTL-expiry checks, so it matches exactly what was written — apples-to-apples against `dstDatSize`. The check (`expectedLiveBytes > dstDatSize`) preserves the original intent: a destination smaller than what should have been copied still signals real data loss (e.g. a failed flush), while removing the false positive from concurrent traffic.
- Removing the `v.nm != nil` guard is safe — the new block no longer dereferences `v.nm`, and `CompactByIndex` already rejects a nil `v.nm` upstream.
- `skippedDataBytes` is retained for the warning log, which is the right place for it.

**Minor nits (non-blocking)**
- The 8-line block comment in the integrity-check block is on the verbose side; it could be trimmed to 2-3 lines without losing the key snapshot-vs-live-map rationale.
- The `else if v.nm.DeletedSize() > v.nm.ContentSize()` warning branch was dropped. Acceptable since the check no longer depends on live-map metrics, but it does remove an observability signal for that edge case — not worth restoring given the new approach.

No issue references are introduced in the new code/comments, and `exceedsExpectedCompactedSize` is a sensibly named new helper with no existing equivalent to reuse.

PR #11268, opened 2026-09-11
mount: rebuild expired directory cache on entry lookup

### LINKED ISSUE #11262: mount: metadata lookups do not rebuild expired directory cache, causing N LookupEntry RPCs
## Environment

- SeaweedFS: 4.46
- Filer store: TiKV 8.5.8, 3 nodes
- Mount command:

```bash
weed mount -dir=/mnt/seaweedfs -writebackCache -filer=10.111.0.15:8888
```

## Dataset

```bash
aws s3 --no-sign-request sync s3://open-images-dataset/validation ./validation
rsync -a ./validation/ /mnt/seaweedfs/buckets/validation/
```

12,577,596 KiB across 41,620 files in a single directory.

## Reproduction

Immediately after remounting:

```text
$ time ls --color /mnt/seaweedfs/buckets/validation/ >/dev/null

real	0m2.395s
user	0m0.075s
sys	0m0.299s

$ time ls --color /mnt/seaweedfs/buckets/validation/ >/dev/null

real	0m0.185s
user	0m0.073s
sys	0m0.112s

$ time ls --color /mnt/seaweedfs/buckets/validation/ >/dev/null

real	0m4.029s
user	0m0.119s
sys	0m1.102s
```

Wait longer than `cacheMetaTtlSec` (60 seconds by default):

```text
$ time ls --color /mnt/seaweedfs/buckets/validation/ > /dev/null

real	1m35.072s
user	0m0.406s
sys	0m1.936s

$ time ls --color /mnt/seaweedfs/buckets/validation/ > /dev/null

real	1m35.049s
user	0m0.341s
sys	0m1.998s
```

Remounting restores the fast path. Increasing `-cacheMetaTtlSec` only delays the same performance cliff.

Without color, listing the directory consistently takes less than one second:

```text
$ time ls --color=never /mnt/seaweedfs/buckets/validation/ >/dev/null

real	0m0.037s
user	0m0.022s
sys	0m0.015s
```

## Actual

After the TTL expires, `IsChildrenCached` returns false. The kernel can still reuse the directory listing cached through `FOPEN_CACHE_DIR | FOPEN_KEEP_CACHE`, so `ReadDir` does not run again and `EnsureVisited` is not called.

A colored `ls` then requests metadata for every directory entry. Each request reaches `Lookup`, which calls `lookupEntry`. Since the parent directory is no longer considered cached, `lookupEntry` issues one `LookupEntry` RPC per file without rebuilding the directory cache.

For 41,620 files, the listing time increases from under 4 seconds to about 95 seconds. Repeating the same command remains slow.

With `weed -v=4`, the slow listing produces one filer lookup per entry:

```text
weedfs.go:622 lookupEntry fetching from filer /buckets/validation/10d357d3c6c4c45e.jpg
weedfs.go:622 lookupEntry fetching from filer /buckets/validation/10d6a655bb5b2831.jpg
weedfs.go:622 lookupEntry fetching from filer /buckets/validation/10d7b9072acf2f7a.jpg
```

## Expected

When metadata lookups arrive for entries from a kernel-cached directory whose mount metadata cache has expired, the directory cache should be rebuilt once through `EnsureVisited`. The entry lookups should then be served from the rebuilt cache instead of issuing one `LookupEntry` RPC per file.

chrislusf:
## Summary

Fixes #11262.

After `cacheMetaTtlSec` elapses, `IsChildrenCached` returns false but the kernel can still serve the directory listing from its page cache (`FOPEN_CACHE_DIR | FOPEN_KEEP_CACHE`). `ReadDir` therefore never runs and `EnsureVisited` is not called, so a metadata-heavy listing (e.g. `ls --color`) reaches `Lookup` for every entry. Since the parent is no longer considered cached, `lookupEntry` issues one `LookupEntry` RPC per file — for the reporter's 41,620-file directory, listing time jumps from ~4s to ~95s and stays there.

This rebuilds the directory cache once through `EnsureVisited` when a lookup arrives for an entry in a directory whose cache has merely TTL-expired (distinct from never-cached / invalidated / evicted / read-through). Subsequent lookups then hit the rebuilt cache. The singleflight in `EnsureVisited` deduplicates concurrent rebuilds, and after the first rebuild the directory is cached again with a fresh TTL, so repeating the command is fast.

### Failure handling
A rebuild can fail (e.g. the filer listing is transiently unavailable). Two competing concerns arise:
- Re-attempting the full rebuild (with backoff retries) on **every** later lookup adds heavy delay to each lookup while the outage lasts.
- Fully clearing the expired state would disable lookup-triggered rebuilds, so if the kernel keeps serving the cached listing (no `ReadDir`), lookups stay on the slow per-entry RPC path even after the filer recovers.

The fix records the last failed rebuild attempt and gates retries on a 30s cooldown (`expiredDirRebuildCooldown`). During the outage, lookups fall through to the per-entry RPC without re-attempting the rebuild; once the cooldown elapses and the filer has recovered, the next lookup rebuilds the cache. Oversized (read-through) directories are not rebuilt here — they are left for `ReadDir` to handle, as before.

### Changes (one logic change per commit)
- `test`: reproduction test asserting 0 `LookupEntry` RPCs and 1 `ListEntries` after TTL expiry (fails before the fix: 3 `LookupEntry`, 0 `ListEntries`).
- `mount`: add `InodeToPath.ShouldRebuildExpiredDir` (TTL-expired + cooldown) and `MarkRebuildAttempt` to distinguish an expired cached directory from an uncached/invalidated/evicted/read-through one and back off after failures.
- `mount`: in `lookupEntry`, when the directory is not cached but its cache is only expired, rebuild it via `ensureDirectoryVisited` before the cache-hit check; on a non-oversized failure, record the attempt for the cooldown.
- `mount`: wait for the target inode's pending async flush before the rebuild, so the rebuilt cache does not capture pre-flush filer metadata and bypass the wait that guards the filer-fallback path.
- `test`: cover concurrent rebuild deduplication (singleflight) and the rebuild-cooldown-after-failure behavior.

#### Test plan
- [x] `go test ./weed/mount/ -run TestLookupRebuildsExpiredDirectoryCache` fails before the fix, passes after
- [x] `go test ./weed/mount/ -run TestLookupRebuildDeduplicatesConcurrentRebuilds`
- [x] `go test ./weed/mount/ -run TestLookupRebuildCooldownAfterFailure`
- [x] `go test ./weed/mount/ -race -count=5 -run 'TestLookupRebuild...'`
- [x] `go test ./weed/mount/ ./weed/mount/meta_cache/`
- [x] `go build ./...`

## Summary by CodeRabbit

- **New Features**
  - Directory metadata caches are now automatically rebuilt when they expire, improving lookup performance and reducing repeated per-entry requests.
  - Concurrent lookups share a single cache rebuild, avoiding duplicate work.
  - Failed rebuilds are retried after a brief cooldown, while immediate repeated attempts are suppressed to improve reliability during temporary listing failures.

### DIFF
--- a/weed/mount/inode_to_path.go
+++ b/weed/mount/inode_to_path.go
@@ -34,19 +34,21 @@ type InodeEntry struct {
 }
 
 type dirState struct {
-	path              util.FullPath
-	isChildrenCached  bool
-	readDirDirect     bool
-	cachedExpiresTime time.Time
-	lastAccess        time.Time
-	lastRefresh       time.Time
-	subdirCount       int32 // tracked in-memory for POSIX directory nlink
+	path               util.FullPath
+	isChildrenCached   bool
+	readDirDirect      bool
+	cachedExpiresTime  time.Time
+	lastAccess         time.Time
+	lastRefresh        time.Time
+	lastRebuildAttempt time.Time
+	subdirCount        int32 // tracked in-memory for POSIX directory nlink
 }
 
 func (d *dirState) resetCacheState() {
 	d.isChildrenCached = false
 	d.readDirDirect = false
 	d.cachedExpiresTime = time.Time{}
+	d.lastRebuildAttempt = time.Time{}
 }
 
 // appendPaths appends every path the inode is reachable by, primary first.
@@ -315,6 +317,7 @@ func (i *InodeToPath) MarkChildrenCached(fullpath util.FullPath) {
 	now := time.Now()
 	d.lastAccess = now
 	d.lastRefresh = now
+	d.lastRebuildAttempt = time.Time{}
 	if i.cacheMetaTtlSec > 0 {
 		d.cachedExpiresTime = now.Add(i.cacheMetaTtlSec)
 	}
@@ -331,6 +334,34 @@ func (i *InodeToPath) IsChildrenCached(fullpath util.FullPath) bool {
 		return d.cachedExpiresTime.IsZero() || time.Now().Before(d.cachedExpiresTime)
 	}
 	return false
+}
+
+// ShouldRebuildExpiredDir reports whether a directory was fully cached, its
+// TTL has elapsed, and enough time has passed since the last failed rebuild
+// attempt to retry. Distinct from IsChildrenCached returning false for a
+// never-cached, invalidated, evicted, or read-through directory: those leave
+// isChildrenCached clear, while a plain TTL expiry keeps it set.
+func (i *InodeToPath) ShouldRebuildExpiredDir(fullpath util.FullPath, cooldown time.Duration) bool {
+	i.RLock()
+	defer i.RUnlock()
+	d := i.dirStateOf(fullpath)
+	if d == nil || !d.isChildrenCached {
+		return false
+	}
+	if d.cachedExpiresTime.IsZero() || time.Now().Before(d.cachedExpiresTime) {
+		return false
+	}
+	return d.lastRebuildAttempt.IsZero() || time.Since(d.lastRebuildAttempt) >= cooldown
+}
+
+// MarkRebuildAttempt records that a rebuild was attempted at now, so
+// ShouldRebuildExpiredDir backs off for the cooldown before retrying.
+func (i *InodeToPath) MarkRebuildAttempt(fullpath util.FullPath, now time.Time) {
+	i.Lock()
+	defer i.Unlock()
+	if d := i.dirStateOf(fullpath); d != nil {
+		d.lastRebuildAttempt = now
+	}
 }
 
 func (i *InodeToPath) HasInode(inode uint64) bool {
--- a/weed/mount/weedfs.go
+++ b/weed/mount/weedfs.go
@@ -3,6 +3,7 @@ package mount
 import (
 	"bytes"
 	"context"
+	"errors"
 	"math/rand/v2"
 	"os"
 	"path"
@@ -576,6 +577,12 @@ func (wfs *WFS) maybeLoadEntry(fullpath util.FullPath) (*filer_pb.Entry, entryVe
 	return entry.ToProtoEntry(), version, fuse.OK
 }
 
+// expiredDirRebuildCooldown limits how often a lookup re-attempts a failed
+// rebuild of an expired directory, so a transient listing failure does not
+// trigger a full rebuild (with backoff retries) on every later lookup while
+// still allowing recovery once the filer is healthy again.
+const expiredDirRebuildCooldown = 30 * time.Second
+
 // lookupEntry looks up an entry by path, checking the local cache first.
 // Cached metadata is only authoritative when the parent directory itself is cached.
 // For uncached/read-through directories, always consult the filer directly so stale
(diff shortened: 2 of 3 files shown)


chrislusf:
Thanks for the review feedback. Both points addressed in the newly pushed commits:

- **Failed rebuilds delay every lookup** (Devin): `lookupEntry` now clears the TTL-expired state when a rebuild fails with a non-oversized error, so later lookups fall through to the per-entry RPC instead of re-attempting the full rebuild (with backoff) on every call. Oversized dirs are already marked read-through by `ensureDirectoryVisited`. Covered by `TestLookupFallsThroughWhenRebuildFails`.
- **Concurrent rebuilds remain untested** (Greptile): Added `TestLookupRebuildDeduplicatesConcurrentRebuilds`, which runs concurrent lookups into the same expired directory behind a gated listing and asserts they share one `ListEntries` via the `EnsureVisited` singleflight.

chrislusf:
Restructured the failure handling in response to the Greptile P1 ("Transient Failures Disable Rebuilds") and the earlier Devin flag ("Failed rebuilds delay every lookup"). Instead of clearing the expired state on failure, the rebuild attempt is now recorded and gated by a 30s cooldown (`expiredDirRebuildCooldown`):

- During a transient listing outage, later lookups skip the rebuild and fall through to the per-entry RPC (no backoff-retry storm on every lookup).
- Once the cooldown elapses and the filer has recovered, the next lookup rebuilds the cache — recovery does not depend on `ReadDir` running.
- Oversized dirs remain read-through and are not retried via lookup.

Covered by `TestLookupRebuildCooldownAfterFailure` (immediate retry suppressed; rebuild resumes after cooldown + recovery). History was rebased into 4 commits (one logic change each).

chrislusf:
Addressed the CodeRabbit concern about pending async flushes: `lookupEntry` now waits for the target inode's pending async flush before calling `ensureDirectoryVisited` when rebuilding an expired parent cache. This prevents the rebuilt cache from capturing pre-flush filer metadata and returning it via the cache-hit path (which bypasses the wait that guards the filer-fallback path). `waitForPendingAsyncFlush` is a no-op when no flush is pending, so ordinary rebuilds are unaffected. Added as a separate commit.

chrislusf:
Thanks for the follow-up reviews. Reasoning through the new points:

**Devin @607 / @613 (cooldown race):** The recommended fix — claiming `lastRebuildAttempt` *before* calling `ensureDirectoryVisited` — would regress the core scenario this PR fixes. The marker is set only on failure, so during a *successful* rebuild concurrent lookups see no marker, pass `ShouldRebuildExpiredDir`, and join the `EnsureVisited` singleflight; once the rebuild publishes, they hit the rebuilt cache (0 per-entry RPCs). If the marker were pre-claimed, those same concurrent lookups would see the cooldown as active, skip the rebuild, and fall through to one `LookupEntry` RPC each — recreating the N-RPC behavior the issue reports (worst for large directories where the rebuild is in flight the longest). The current design keeps the success path optimal and uses the cooldown only to back off after failures. The race window is the few instructions between the singleflight completing a failed rebuild and the first waiter's `MarkRebuildAttempt`; its only effect is that one rebuild attempt may start slightly earlier during a sustained outage (the singleflight still dedups all concurrent callers). I believe this is the right tradeoff, but I'm happy to switch to a generation/claim scheme if a maintainer prefers it.

**CodeRabbit @test:241:** The test does verify singleflight joining, via `lookupCalls == 0` together with the gate. The gate holds the rebuild open, so the directory stays uncached and no failure marker is set; any lookup that did *not* join the in-flight rebuild would fall through to a per-entry `LookupEntry` RPC. Asserting `lookupCalls == 0` proves none did — they all joined the singleflight and then hit the cache once the rebuild published. The 50ms hold is generous for goroutine scheduling and only needs to cover that, since the gate keeps the rebuild in flight.

PR #11270, opened 2026-09-11
fix: write the new key when a remote-synced file is renamed

### LINKED ISSUE #11261: filer.remote.sync: renaming a replicated file deletes it from the remote and never writes the new key
## Summary

Renaming a file inside a remote-mounted directory removes its object from the remote and never writes one at the new key. The local file is fine, its `RemoteEntry` still says it is synced, and nothing is logged at default verbosity. The only durable copy is gone.

Two components each behave reasonably on their own; the combination loses the object.

## Reproduction

With a remote configured and mounted, and `filer.remote.sync -dir=/buckets` running:

```
# 1. write, and wait for the sync to upload it
weed filer.copy ./probe.bin http://filer:8888/buckets/b/src/
#    log: create b2/<bucket>/b/src/probe.bin

# 2. confirm it is really there
rclone lsjson remote:<bucket>/b/src/probe.bin        # present

# 3. rename it
echo 'fs.mv /buckets/b/src/probe.bin /buckets/b/dst/probe.bin' | weed shell

# 4. within a few seconds
rclone lsjson remote:<bucket>/b/src/probe.bin        # []
rclone lsjson remote:<bucket>/b/dst/probe.bin        # []
```

`filer.remote.sync` logs nothing at all for step 3.

Ran it twice, 4096 and 2048 byte files. The bucket has versioning on, and the non-current versions are still listed at the right sizes and upload timestamps, so the objects were genuinely uploaded and then deleted rather than never written.

## What each side does

**The filer deletes the remote object.** `filer_delete_entry.go` calls `maybeDeleteFromRemote` (`weed/filer/filer_lazy_remote.go:119`) for any entry under a remote mount with `entry.Remote != nil`. A rename deletes the source entry, so this fires. It logs at V(3), which is why the sync's log is silent — the sync is not the deleter.

**`filer.remote.sync` skips writing the new key.** A rename arrives as a single update event: `directory` = old parent, `newParentPath` = new parent, and `newEntry` carrying the **inherited** `RemoteEntry` from the source. In `filer_remote_sync_dir.go`:

```go
oldDest := toRemoteStorageLocation(..., util.NewFullPath(resp.Directory, message.OldEntry.Name), ...)
dest    := toRemoteStorageLocation(..., util.NewFullPath(message.NewParentPath, message.NewEntry.Name), ...)
if !shouldSendToRemote(message.NewEntry) {
    glog.V(2).Infof("skipping updating: %+v", resp)
    return nil        // <-- returns here
}
...
if !proto.Equal(oldDest, dest) {
    glog.V(0).Infof("delete %s", ...)   // the rename handling, never reached
```

`shouldSendToRemote` returns false because the inherited `RemoteEntry` has `RemoteMtime >= Attributes.Mtime`. But that `RemoteEntry` describes the **old** key. The code immediately below the early return is exactly the delete-old/write-new that a rename needs, and it is unreachable for the case it was written for.

## Suggested fix

Only take the skip when the destination key is unchanged, since a path change always needs a write:

```go
if proto.Equal(oldDest, dest) && !shouldSendToRemote(message.NewEntry) {
    glog.V(2).Infof("skipping updating: %+v", resp)
    return nil
}
```

Happy to send this as a PR if the shape looks right.

One related note: even with this fixed there is a brief window where neither key exists, because the filer's delete is synchronous with the rename while the sync's write is asynchronous. That may be acceptable, but it is worth deciding deliberately rather than inheriting.

This is a different path from #11139 / #11140 and #11146, which were about the metadata-only branch for an unchanged in-place file. Here the path itself changes and the entry never had a `RemoteEntry` for its new key.

Seen on 4.46; the same early return is present in 4.41 and on master.

chrislusf:
## Summary

Renaming a file inside a remote-mounted directory deleted its object from the remote and never wrote one at the new key (#11261). The local file was fine and its `RemoteEntry` still claimed it was synced, so the only durable copy was gone and nothing was logged at default verbosity.

Two components combined to lose the object:
- The filer deletes the remote object synchronously during the rename (`maybeDeleteFromRemote` fires for the source entry).
- `filer.remote.sync` skipped writing the new key. A rename arrives as a single update event whose `NewEntry` carries the `RemoteEntry` inherited from the source. `shouldSendToRemote` returns false for it (`RemoteMtime >= Mtime`), so `processUpdateEvent` hit the early skip return. The delete-old/write-new handling just below that return is exactly what a rename needs, but it was unreachable for the case it was written for.

The skip is only valid when the destination key is unchanged. Guard it with `proto.Equal(oldDest, dest)` so a path change always falls through to the rename handling.

Review feedback also surfaced two related edge cases, both addressed:
- A **remote-only** entry (no local chunks/content, data lives only on the already-deleted remote object) would upload EOF and create a zero-byte object at the new key. A guard skips the upload for a path change on a remote-only entry.
- A failed delete of the old key was silently swallowed, potentially leaving both keys. Non-multipart delete errors are now returned so `MetadataProcessor` retries — except `ErrRemoteObjectNotFound`, which the filer's synchronous delete makes the expected result on GCS (whose delete is not idempotent, unlike S3/Azure).

### Commits
- `refactor: extract update event handling into processUpdateEvent` — pulls the update branch out of the event closure so the skip logic can be exercised with stub clients. No behavior change.
- `test: reproduce remote sync rename dropping the new key` — runs a rename event with an inherited `RemoteEntry` through `processUpdateEvent`; fails before the fix.
- `fix: write the new key when a remote-synced file is renamed` — guard the skip with `proto.Equal(oldDest, dest)`.
- `fix: skip empty upload when renaming a remote-only entry` — guard a path change on a remote-only entry so it does not upload a zero-byte object.
- `fix: propagate old-key delete errors except already-deleted` — return non-multipart `DeleteFile` errors so the event retries; treat `ErrRemoteObjectNotFound` as success since the filer already deleted the source.

Note: even with this fixed there is a brief window where neither key exists, because the filer's delete is synchronous with the rename while the sync's write is asynchronous (called out in the issue). The remote-only rename remains fundamentally lossy because the filer deletes the source object before the sync can copy it; a complete fix for that case needs filer-side remote copy/move support, which is out of scope here.

Fixes #11261.

#### Test plan
- [x] `TestRenameWithInheritedRemoteEntryWritesNewKey` — fails before the fix, passes after
- [x] `TestRenameRemoteOnlyEntrySkipsEmptyUpload` — guards the zero-byte edge case
- [x] `TestRenameDeleteOldKeyFailureReturnsError` — non-multipart delete failure retries
- [x] `TestRenameDeleteOldKeyNotFoundStillWrites` — GCS `ErrRemoteObjectNotFound` does not block the write
- [x] `go test ./weed/command/` relevant suite passes
- [x] `go build ./weed/...`, `go vet ./weed/command/`, gofmt clean

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed remote synchronization for renamed files received as updates.
  - Renamed entries are now written under the new remote key while the previous key is removed.
  - Improved handling of inherited remote metadata during rename operations.
  - Prevented remote-only files without local content from being uploaded as empty files.
  - Synchronization now reports remote deletion failures, while ignoring already-missing remote objects so valid destination writes can continue.

### DIFF
--- a/weed/command/filer_remote_sync_dir.go
+++ b/weed/command/filer_remote_sync_dir.go
@@ -261,6 +261,10 @@ func processUpdateEvent(
 		}
 		glog.V(0).Infof("never replicated, uploading %s", remote_storage.FormatLocation(dest))
 	}
+	if !proto.Equal(oldDest, dest) && !filer.HasData(message.NewEntry) && message.NewEntry.IsInRemoteOnly() {
+		glog.V(0).Infof("skip uploading renamed remote-only entry %s: content is only on the deleted remote object", remote_storage.FormatLocation(dest))
+		return nil
+	}
 	glog.V(2).Infof("update: %+v", resp)
 	if !proto.Equal(oldDest, dest) {
 		glog.V(0).Infof("delete %s", remote_storage.FormatLocation(oldDest))
--- a/weed/command/filer_remote_sync_dir.go
+++ b/weed/command/filer_remote_sync_dir.go
@@ -268,8 +268,13 @@ func processUpdateEvent(
 	glog.V(2).Infof("update: %+v", resp)
 	if !proto.Equal(oldDest, dest) {
 		glog.V(0).Infof("delete %s", remote_storage.FormatLocation(oldDest))
-		if err := client.DeleteFile(oldDest); err != nil && isMultipartUploadFile(resp.Directory, message.OldEntry.Name) {
-			return nil
+		if err := client.DeleteFile(oldDest); err != nil {
+			if isMultipartUploadFile(resp.Directory, message.OldEntry.Name) {
+				return nil
+			}
+			if !errors.Is(err, remote_storage.ErrRemoteObjectNotFound) {
+				return err
+			}
 		}
 	}
 	remoteEntry, writeErr := retriedWriteFile(client, filerSource, message.NewParentPath, message.NewEntry, dest)
--- a/weed/command/filer_remote_sync_dir.go
+++ b/weed/command/filer_remote_sync_dir.go
@@ -213,59 +213,70 @@ func (option *RemoteSyncOptions) makeEventProcessor(remoteStorage *remote_pb.Rem
 			return client.DeleteFile(dest)
 		}
 		if message.OldEntry != nil && message.NewEntry != nil {
-			if isMultipartUploadFile(message.NewParentPath, message.NewEntry.Name) {
-				return nil
-			}
-			// Skip updates to internal version paths
-			if isVersionedPath(message.NewParentPath, message.NewEntry.Name, message.NewEntry.IsDirectory) {
-				glog.V(2).Infof("skipping update of internal version path: %s/%s", message.NewParentPath, message.NewEntry.Name)
-				return nil
-			}
-			oldDest := toRemoteStorageLocation(util.FullPath(mountedDir), util.NewFullPath(resp.Directory, message.OldEntry.Name), remoteStorageMountLocation)
-			dest := toRemoteStorageLocation(util.FullPath(mountedDir), util.NewFullPath(message.NewParentPath, message.NewEntry.Name), remoteStorageMountLocation)
-			if !shouldSendToRemote(message.NewEntry) {
-				glog.V(2).Infof("skipping updating: %+v", resp)
-				return nil
-			}
-			if message.NewEntry.IsDirectory {
-				return client.WriteDirectory(dest, message.NewEntry)
-			}
-			if isMetadataOnlyUpdate(resp.Directory, message) {
-				remoteEntry, err := liveRemoteEntry(option, message.NewParentPath, message.NewEntry)
-				if errors.Is(err, filer_pb.ErrNotFound) {
-					glog.V(2).Infof("skipping updating deleted entry: %+v", resp)
-					return nil
-				}
-				if err != nil {
-					return err
-				}
-				if remoteEntry != nil {
-					glog.V(2).Infof("update meta: %+v", resp)
-					return client.UpdateFileMetadata(dest, message.OldEntry, message.NewEntry)
-				}
-				glog.V(0).Infof("never replicated, uploading %s", remote_storage.FormatLocation(dest))
-			}
-			glog.V(2).Infof("update: %+v", resp)
-			if !proto.Equal(oldDest, dest) {
-				glog.V(0).Infof("delete %s", remote_storage.FormatLocation(oldDest))
-				if err := client.DeleteFile(oldDest); err != nil && isMultipartUploadFile(resp.Directory, message.OldEntry.Name) {
-					return nil
-				}
-			}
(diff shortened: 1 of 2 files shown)


chrislusf:
Thanks for the reviews. Addressed in the latest push:

**@coderabbitai (delete error handling):** Good catch. `b5e3405b0` now returns non-multipart `DeleteFile(oldDest)` errors before `retriedWriteFile`, so `MetadataProcessor` retries the event instead of silently continuing and leaving both keys. The multipart exception is preserved. Added `TestRenameDeleteOldKeyFailureReturnsError`.

**@greptile-apps (remote-only rename breaks reads):** The source object does not remain at the old key — the filer's `moveSelfEntry` calls `DeleteEntryMetaAndData` → `maybeDeleteFromRemote` for the old path synchronously during the rename (`weed/server/filer_grpc_server_rename.go:326`, `weed/filer/filer_delete_entry.go:200`), so the source is already gone by the time the sync processes the event. The reporter confirmed this in #11261 (the old key is gone after rename). So for a remote-only rename the data is lost regardless of what the sync does — the filer deleted the only copy before the sync can act. The guard in `8a8b3c89e` is harm reduction: it avoids replacing the file with a misleading zero-byte object. A complete fix for remote-only renames needs filer-side remote copy/move support (copy before delete), which is out of scope for this PR.

**@devin-ai-integration (remote-only empty upload):** Addressed in `8a8b3c89e` — a path change on a remote-only entry (`!filer.HasData && IsInRemoteOnly`) skips the upload instead of writing EOF. Added `TestRenameRemoteOnlyEntrySkipsEmptyUpload`.

chrislusf:
Good catch on the GCS issue — that was a real regression in the previous version of the delete-error change. The filer deletes the source remote object synchronously during rename (`maybeDeleteFromRemote`), so the sync's `DeleteFile(oldDest)` is redundant and the object is already gone. GCS reports that as `ErrRemoteObjectNotFound` (unlike S3/Azure, whose deletes are idempotent), which would have pinned the sync offset and never written the new key.

Fixed in `85a0d67a9`: `ErrRemoteObjectNotFound` is now treated as a successful deletion and the write proceeds. Other non-multipart delete errors are still returned so the event is retried. Added `TestRenameDeleteOldKeyNotFoundStillWrites` covering the GCS-shaped case.

impuls42:
Running this on a live cluster since it merged (the 4.46 image with `weed` rebuilt from master, B2 as the remote tier). Both paths do what they say:

```
20:31:04 delete b2/<tier>/.../src/acc.bin
20:31:04 create b2/<tier>/.../dst/acc.bin
20:31:38 skip uploading renamed remote-only entry b2/<tier>/.../ro2/ro.bin: content is only on the deleted remote object
```

Checked against the bucket with a never-renamed control alongside, so the "absent" readings are real. The destination's `RemoteEntry` comes out freshly written — ETag equal to the file's md5, `RemoteMtime` the new upload — rather than inherited from the source. The remote-only rename leaves nothing at either key instead of a zero-byte object.

That empty-upload guard is the part we would have shipped wrong on our own. Thanks for catching it.

PR #11271, opened 2026-09-11
fix(s3api/iam): avoid transient AccessDenied from full reloads on single IAM file changes

### LINKED ISSUE #11259: Full S3 IAM reload on any single identity/policy change causes transient AccessDenied for unrelated clients
**Version**: 4.44-enterprise (`chrislusf/seaweedfs-enterprise:4.44`)

### What we're seeing

We have several independent S3 credentials configured (one per application/tenant), each stored as a separate identity/policy file under `/etc/iam/identities` and `/etc/iam/policies` and refreshed on its own schedule by an external secrets-management tool. None of these credentials rotate or change value on refresh in the vast majority of cycles - the refresh just re-writes the same file.

Whenever any *one* of these files changes (or is even just re-written with the same content), the S3 gateway logs a full IAM reload, e.g.:

```
auth_credentials_subscribe.go:91 IAM change detected in /etc/iam/identities, reloading configuration
auth_credentials.go:2275 Loading S3 API configuration from credential manager
auth_credentials.go:2301 Successfully loaded S3 API configuration from credential manager
```

This reload appears to reload the *entire* IAM configuration (all identities/policies), not just the one file that changed. Because several of our credentials happen to refresh within the same rough time window, we see bursts of dozens of these reload cycles firing back-to-back within the same second.

### Impact

During these reload bursts, we've observed unrelated S3 clients (in our case, an Apache Flink job writing checkpoints via the S3A connector) occasionally receive a `403 AccessDenied` on a request using credentials that were **not** the ones being changed/reloaded at that moment:

```
java.nio.file.AccessDeniedException: s3a://<bucket>/<path>: getFileStatus on s3a://<bucket>/<path>:
com.amazonaws.services.s3.model.AmazonS3Exception: Access Denied. (Service: Amazon S3; Status Code: 403; Error Code: AccessDenied)
```

The credentials involved are valid before and after the event - this looks like a timing collision with the reload itself (a request landing mid-reload), not an actual permission/credential problem.

### Question

Is this expected behavior? We're not sure whether:
- the reload is intended to be scoped to the single changed identity rather than the full IAM config, or
- concurrent requests are expected to be safe across a reload and this is a bug, or
- this is a known tradeoff that isn't planned to change.

We don't need a fix - if this is a known limitation or not something on the roadmap, that's a perfectly fine answer. Mainly wanted to flag the behavior in case it's useful, since it's fairly easy to hit with multiple independently-refreshing credentials.

chrislusf:
## Summary

Fixes #11259.

A single identity/policy/service-account/group file change triggered a **full synchronous IAM reload** that re-read every IAM file from the filer. Two problems caused unrelated clients to receive transient `AccessDenied`:

1. **Incomplete snapshot accepted.** When an external secrets tool rewrites a file, a reload that reads it mid-rewrite sees empty or partially-written content. The identity, policy and service-account loaders silently skipped such files (`continue`), so the snapshot was missing entries that still existed on disk. The atomic swap then installed an incomplete identity set while `isAuthEnabled` stayed on, denying unrelated clients mid-reload. The group loader and the read-error paths already failed the snapshot in this situation (a skipped entry reads as deleted); the empty-content and unmarshal-failure paths did not.

2. **Reload storm.** `onIamConfigChange` did one full synchronous reload per metadata event. Several independently-refreshing credentials can rewrite their files within the same second, producing dozens of back-to-back full reloads, each widening the window where a mid-rewrite file is observed.

### Changes (one commit per logical change)

**Commit 1 — `fix(s3api/iam): fail config snapshot on empty or malformed IAM files`**
- Identity, policy, service-account and group loaders now return an error on empty content or JSON unmarshal failure (and service-account read errors), instead of silently skipping. A transient mid-rewrite fails the reload and preserves the last known-good state, rather than dropping entries.
- Adds reproduction tests in `weed/credential/filer_etc/filer_etc_identity_test.go` covering empty/malformed identity files and empty managed-policy files, plus a valid-files baseline.

**Commit 2 — `fix(s3api/iam): coalesce burst IAM config reloads through the reload queue`**
- `onIamConfigChange` now routes every IAM config change through the existing coalescing reload queue (`scheduleReload`/`reloadRetryLoop`) instead of doing a synchronous reload per event. A burst of N events collapses into one reload (plus one tail reload for events arriving mid-reload). `scheduleReload` gains a `reason` argument for the existing log line; the retry loop already retries failed reloads, so the per-event failure handoff is removed.
- `newTestS3ApiServerWithMemoryIAM` now wires up the queue and retry loop; tests that asserted on the synchronous reload poll via `waitForIdentity`/`waitForIdentityGone`. Adds `TestOnIamConfigChangeCoalescesBurstReloads` showing 50 events coalesce into ≤3 reloads.

### Preserved behavior
- Static identities and the static/dynamic merge semantics are unchanged.
- The one-way authentication-enablement behavior (`updateAuthenticationState` enables auth once identities exist; it does not disable during a reload) is unchanged.
- Serialized reloads (`reloadMu`) and serialized advanced IAM policy sync (`iamManagerSyncMu`) are unchanged.

#### Test plan
- [x] `go test ./weed/credential/filer_etc` (new reproduction tests fail before commit 1, pass after)
- [x] `go test ./weed/credential/...`
- [x] `go test ./weed/s3api` (incl. `TestOnIamConfigChangeCoalescesBurstReloads`, `TestFailedReloadRetriesUntilSuccess`)
- [ ] CI (race build) on the PR

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11271" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid, empty, malformed, or unreadable identity, policy, group, and service-account files now cause configuration loading to fail with descriptive errors.
  * Non-JSON auxiliary files are ignored during multi-file IAM configuration loading and policy listing.
  * IAM configuration changes reload asynchronously and more reliably, with rapid bursts consolidated to reduce unnecessary reloads.
  * Empty identity, group, and service-account names are rejected.

* **Tests**
  * Expanded coverage for validation, successful loading, asynchronous updates, reload retries, and reload consolidation.

### DIFF
--- a/weed/credential/filer_etc/filer_etc_group.go
+++ b/weed/credential/filer_etc/filer_etc_group.go
@@ -30,6 +30,9 @@ func (store *FilerEtcStore) loadGroupsFromMultiFile(ctx context.Context, s3cfg *
 			if entry.IsDirectory {
 				continue
 			}
+			if !strings.HasSuffix(entry.Name, ".json") {
+				continue
+			}
 
 			var content []byte
 			if len(entry.Content) > 0 {
--- a/weed/credential/filer_etc/filer_etc_group.go
+++ b/weed/credential/filer_etc/filer_etc_group.go
@@ -52,6 +52,9 @@ func (store *FilerEtcStore) loadGroupsFromMultiFile(ctx context.Context, s3cfg *
 			if err := json.Unmarshal(content, g); err != nil {
 				return fmt.Errorf("failed to unmarshal group %s: %w", entry.Name, err)
 			}
+			if g.Name == "" {
+				return fmt.Errorf("group file %s has empty name", entry.Name)
+			}
 			// Merge: overwrite existing group with same name or append
 			found := false
 			for i, existing := range s3cfg.Groups {
--- a/weed/credential/filer_etc/filer_etc_group.go
+++ b/weed/credential/filer_etc/filer_etc_group.go
@@ -42,23 +42,24 @@ func (store *FilerEtcStore) loadGroupsFromMultiFile(ctx context.Context, s3cfg *
 				content = c
 			}
 
-			if len(content) > 0 {
-				g := &iam_pb.Group{}
-				if err := json.Unmarshal(content, g); err != nil {
-					return fmt.Errorf("failed to unmarshal group %s: %w", entry.Name, err)
+			if len(content) == 0 {
+				return fmt.Errorf("group file %s is empty", entry.Name)
+			}
+			g := &iam_pb.Group{}
+			if err := json.Unmarshal(content, g); err != nil {
+				return fmt.Errorf("failed to unmarshal group %s: %w", entry.Name, err)
+			}
+			// Merge: overwrite existing group with same name or append
+			found := false
+			for i, existing := range s3cfg.Groups {
+				if existing.Name == g.Name {
+					s3cfg.Groups[i] = g
+					found = true
+					break
 				}
-				// Merge: overwrite existing group with same name or append
-				found := false
-				for i, existing := range s3cfg.Groups {
-					if existing.Name == g.Name {
-						s3cfg.Groups[i] = g
-						found = true
-						break
-					}
-				}
-				if !found {
-					s3cfg.Groups = append(s3cfg.Groups, g)
-				}
+			}
+			if !found {
+				s3cfg.Groups = append(s3cfg.Groups, g)
 			}
 		}
 		return nil
--- a/weed/credential/filer_etc/filer_etc_identity.go
+++ b/weed/credential/filer_etc/filer_etc_identity.go
@@ -92,6 +92,9 @@ func (store *FilerEtcStore) loadFromMultiFile(ctx context.Context, s3cfg *iam_pb
 			if entry.IsDirectory {
 				continue
 			}
+			if !strings.HasSuffix(entry.Name, ".json") {
+				continue
+			}
 			hasIdentities = true
 
 			var content []byte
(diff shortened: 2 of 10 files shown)


chrislusf:
Thanks for the review. Good catch — addressed in 28a2c998.

The multi-file loaders unmarshal every entry in an IAM directory, so a non-JSON auxiliary file (README, `.DS_Store`, a migration backup like `identity.json.old`) would hit the new empty/malformed errors and reject the whole snapshot. Only `*.json` files are IAM objects (SeaweedFS writes identities/policies/service-accounts/groups as `<name>.json`, and other call sites already gate on the `.json` suffix), so each loader now skips non-`.json` entries at the top of the loop, before reading content. Auxiliary files are ignored; empty/malformed `.json` files still fail the snapshot.

Added `TestLoadConfigurationIgnoresNonJsonAuxiliaryFiles` covering a `.txt`, a `.json.old` backup in the identities dir, and a `.md` in the policies dir.

chrislusf:
Both follow-up review points addressed in 872d493cb.

**Phantom policies in `ListPolicyNames`** (Devin): `ListPolicyNames` listed every regular entry in the policies directory, including non-JSON auxiliary files, while `GetPolicy` cannot retrieve them. It now applies the same `.json` suffix filter as the loader, so the list only exposes retrievable policies. Added `TestFilerEtcStoreListPolicyNamesSkipsNonJsonAuxiliary`.

**Empty identifiers** (CodeRabbit): `json.Unmarshal` accepts `{}` and unknown fields. The identity/group loaders merge by the decoded `Name` (not the file name), so a `{}` file could install an empty-key record and displace a real one; the service-account loader accepted an empty `Id`. Each loader now validates `Identity.Name`, `Group.Name`, and `ServiceAccount.Id` (via `validateServiceAccountId`) after unmarshal and fails the snapshot on empty identifiers. Added empty-identifier regression tests for all three record types.

PR #11282, opened 2026-09-12
s3: fail closed on unsupported bucket-policy condition operators

chrislusf:
## Summary

The S3 bucket-policy condition engine failed open on unrecognized condition operators. `EvaluateConditions` logged a warning and `continue`d past an unsupported operator, leaving no condition to fail, so it returned `true` and made a conditional `Allow` statement unconditional. This could let an unauthenticated request bypass a non-matching condition (e.g. a `StringEquals: aws:username` guard) and read private objects, simply by using an operator the engine does not recognize.

This is split into three focused commits:

1. **Support the IgnoreCase string operators** (`StringEqualsIgnoreCase`, `StringNotEqualsIgnoreCase`, `StringLikeIgnoreCase`, `StringNotLikeIgnoreCase`) in the S3 policy engine. These are valid AWS operators that the IAM policy engine already supports; adding them here means legitimate policies evaluate correctly instead of falling into the unsupported-operator path.
2. **Reject policies with unsupported condition operators at upload/parse time.** `validateStatement` now reuses `GetConditionEvaluator` to reject unknown operators, so a typo or unsupported key is refused when the policy is stored rather than silently skipped later.
3. **Fail closed at evaluation time.** `EvaluateConditions` and `EvaluateConditionsLegacy` now return `false` on an unsupported operator instead of skipping it, so an unrecognized operator fails the condition block (the statement does not match) rather than granting access. This mirrors the fail-closed posture of the IAM policy engine.

The three changes are independent: (1) makes valid operators work, (2) catches unknown operators at the entry point, and (3) is defense in depth so an unknown operator can never widen access if it reaches evaluation.

#### Test plan
- [x] `go test ./weed/s3api/policy_engine/` — new tests for the IgnoreCase evaluators, validation rejection, and fail-closed behavior
- [x] `go test ./weed/s3api/...` — no regressions
- [x] `go build ./weed` — builds cleanly
- [ ] CI

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11282" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added case-insensitive S3 policy conditions for equality and wildcard matching.
  - Unsupported or invalid policy conditions now fail closed instead of being ignored.

- **Bug Fixes**
  - Admin dashboard links and file browser controls now use valid HTTP addresses, including bracketed IPv6 hosts.
  - Improved error handling during volume reads and maintenance operations, preserving underlying error details.

- **Tests**
  - Added coverage for case-insensitive conditions, unsupported operators, and policy validation.

### DIFF
--- a/weed/admin/view/app/admin.templ
+++ b/weed/admin/view/app/admin.templ
@@ -3,6 +3,7 @@ package app
 import (
     "fmt"
     "github.com/seaweedfs/seaweedfs/weed/admin/dash"
+    "github.com/seaweedfs/seaweedfs/weed/pb"
 )
 
 templ Admin(data dash.AdminData) {
@@ -173,7 +174,7 @@ templ Admin(data dash.AdminData) {
                                     for _, master := range data.MasterNodes {
                                         <tr>
                                             <td>
-                                                <a href={templ.SafeURL(fmt.Sprintf("http://%s/ui/index.html", master.Address))} target="_blank">
+                                                <a href={templ.SafeURL(fmt.Sprintf("http://%s/ui/index.html", pb.ServerAddress(master.Address).ToHttpAddress()))} target="_blank">
                                                     {master.Address}
                                                     <i class="fas fa-external-link-alt ms-1 text-muted"></i>
                                                 </a>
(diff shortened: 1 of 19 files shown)


chrislusf:
Closing in favor of #11283, which rebases the same changes onto the current upstream master (this branch was mistakenly based on a divergent fork master, producing an noisy diff).

PR #11293, opened 2026-09-13
filer: reject remote-storage confs that dial blocked endpoints at load (SSRF)

chrislusf:
## Summary

The filer's lazy-fetch / lazy-list / remote-delete paths resolve remote storage clients by name from `FilerRemoteStorage.storageNameToConf` and dial them via `remote_storage.GetRemoteStorage`, which bypasses the SSRF deny-list the volume server (`BuildGuardedRemoteStorageClient`) and the filer's own direct-read path already apply.

A `RemoteConf` planted under `/etc/remote` with a loopback / private / IMDS S3 endpoint is reloaded into `storageNameToConf` on the next metadata-change event, then dialed on the next cache miss — server-side request forgery from the filer.

### Fix

Apply the volume server's SSRF deny-list at conf **load** time — the single chokepoint that populates `storageNameToConf`:

- Add `RemoteStorageConfValidator`, injected into `FilerRemoteStorage` by the filer server (the filer package cannot import the server package). A conf that fails validation is dropped from `storageNameToConf`, so the name-based client resolution on the lazy paths returns "not found" instead of dialing the blocked endpoint.
- Add `ValidateRemoteConfForLoad` in `weed_server`, which mirrors `BuildGuardedRemoteStorageClient`'s gcs credential + endpoint checks (`validateRemoteEndpoint` via `guardedRemoteClient`) without building a client. `allowUntrusted` skips the check, mirroring the volume server opt-out (`-filer.allowUntrustedRemoteEndpoints`).
- The filer server injects the validator at construction.

A conf whose type dials a fixed provider host (no caller-supplied endpoint) passes; only caller-influenced endpoints are denied. This is defense-in-depth alongside the existing per-dial guard on the direct-read path.

#### Test plan
- [x] `go test ./weed/filer/` — added `TestLoadRemoteStorageConfigurationsAndMapping_RejectsBlockedEndpoint` (loopback conf dropped at load, valid conf loaded)
- [x] `go test ./weed/server/` — added `TestValidateRemoteConfForLoad` (loopback rejected, `allowUntrusted` accepts, fixed-host provider passes, nil conf no-op)
- [x] `go build ./weed/filer/ ./weed/server/` clean
- [x] existing `TestBuildGuardedRemoteStorageClient` / `TestRemoteEndpointGuardCoversS3CompatibleSiblings` still pass

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11293" target="_blank">
  <picture>
[...]

PR #11294, opened 2026-09-13
filer: apply SSRF guard to the lazy-remote fetch/list/delete paths

chrislusf:
## Summary

The filer's lazy-remote metadata paths (`maybeLazyFetchFromRemote`, `maybeLazyListFromRemote`, `maybeDeleteFromRemote`) resolved their remote-storage client through the unguarded shared cache (`remote_storage.GetRemoteStorage`), bypassing the `BuildGuardedRemoteStorageClient` chokepoint that #10990 wired into the volume `FetchAndWriteNeedle`, filer `streamFromRemote` and s3 `openRemoteStream` dial paths.

On a default (unauthenticated) filer this lets an unauthenticated request coerce the filer into issuing attacker-directed requests to arbitrary internal hosts / cloud metadata endpoints:
- the **fetch** path surfaces the HEAD result (reachability, size, etag, last-modified) to the caller — an internal port/host scanning + object-metadata oracle
- the **delete** path issues outbound DELETE-shaped requests to internal hosts (write-side SSRF)
- the missing rebinding-safe dialer also makes a mount whose hostname flips to an internal IP reachable (DNS-rebinding) with no filer write

This applies two layers of defense-in-depth:

**Layer 1 — load-time validation** (from #11293): a `RemoteStorageConfValidator` injected into `FilerRemoteStorage` rejects a `RemoteConf` whose endpoint resolves to a blocked address (loopback / private / IMDS) at `/etc/remote` reload, before it enters `storageNameToConf`. The lazy paths resolve clients by name, so a rejected conf yields "not found" instead of dialing the blocked endpoint. Fail-fast with a clear log line.

**Layer 2 — build-time guarded client** (original to this PR): a `RemoteStorageClientBuilder` hook on `Filer` routes the lazy fetch/list/delete through `BuildGuardedRemoteStorageClient` (the same endpoint deny-list + DNS-rebinding-safe pinned dialer + gcs credential check the sibling streaming paths use). This catches DNS-rebinding — a hostname that resolves to a public IP at load but flips to an internal IP at dial time — which the load-time check alone misses.

Both layers honor the existing `-filer.allowUntrustedRemoteEndpoints` opt-out for operators whose remote storage legitimately sits on a private network.

#### Changes (one commit per logic change)
- add a `RemoteStorageClientBuilder` hook on `Filer` + conf-only lookups on `FilerRemoteStorage`, and route the lazy fetch through it
- route the lazy directory listing through the guarded client
- route the lazy remote delete through the guarded client
- wire `BuildGuardedRemoteStorageClient` + `AllowUntrustedRemoteEndpoints` into the filer at startup
- regression test: a rejecting builder makes the lazy fetch return no entry without reaching the remote
- skip remote client for local-only lazy deletes (review feedback)
- build the guarded client inside the singleflight (review feedback)
- guard-rejection tests for list + delete + local-only
- decouple the build from the first caller's cancellation (review feedback)
- reject blocked-endpoint confs at load time (from #11293): `RemoteStorageConfValidator` + `ValidateRemoteConfForLoad` + load-time drop + tests

#### Test plan
- [x] `go build ./weed/...`
- [x] `go test ./weed/filer/ ./weed/server/` (including existing lazy-remote + guarded-client tests)
- [x] end-to-end: an unauthenticated filer with a planted `/etc/remote/evil.conf` whose S3 endpoint is a loopback target — before this change the filer dials it (HEAD /b/probe, 200 reflected); after this change the conf is rejected at load (`reject remote storage conf /etc/remote/evil.conf: ... resolves to loopback address`) and the dial is refused (404, target untouched)
- [x] `-filer.allowUntrustedRemoteEndpoints` still bypasses both layers for operators with private remote storage

#### Reconciliation note
This PR supersedes #11293 (closed). #11293's load-time validation is cherry-picked as commit `df9e3128c` on top of the build-time guarded client builder, combining both approaches into one defense-in-depth fix. The unrelated iceberg/s3tables commits that were on #11293's branch are not included.

## Summary by CodeRabbit

* **Security**
  * Added guarded remote-storage connections with endpoint and DNS-rebinding protections.
  * Blocked remote-storage endpoints are rejected during configuration loading.
  * Added an option to allow untrusted remote endpoints when explicitly configured.
  * Hostname validation is deferred until connection time to avoid disrupting valid mounts during temporary DNS issues.

* **Reliability**
  * Improved remote fetching, listing, and deletion with safer connection handling.
  * Remote operation failures no longer issue remote requests.
  * Local-only files can still be deleted without contacting remote storage.


chrislusf:
Thanks for the reviews. Pushed three follow-up commits addressing each point:

**Local-only file becomes undeletable under a rejected mount** (CodeRabbit, Devin) — fixed in `77d1e0e`. `maybeDeleteFromRemote` now checks `!entry.IsDirectory() && entry.Remote == nil` before constructing the remote client, so a local-only file under a mount whose endpoint the guard rejects still deletes its metadata. Directories and remote-backed files still go through the guard.

**Guarded client built before the singleflight** (Greptile, Devin) — fixed in `292b543`. The lazy fetch and list paths now build the guarded client inside their singleflight blocks, so concurrent same-key requests no longer each allocate a fresh SDK client + HTTP transport when only one remote operation runs. The per-request guarded-client model itself matches the sibling `streamFromRemote` / `openRemoteStream` / `FetchAndWriteNeedle` paths (the rebinding-safe dialer re-validates the resolved IP on every dial), so caching guarded clients across requests would diverge from that established pattern and is out of scope for this fix.

**Guard coverage incomplete** (Greptile) — fixed in `589408b`. Added rejection tests for the list path (no remote call), the delete path on a remote-backed file (delete blocked), and the local-only file case (still deletes). The fetch rejection test from the earlier commit remains.

Re-verified end-to-end: an unauthenticated filer with a planted loopback remote conf still returns 404 and does not dial the internal target; `-filer.allowUntrustedRemoteEndpoints` still bypasses the guard.

chrislusf:
Addressed the cancellation concern (Greptile P1) in `83c68a0`: the guarded-client build inside both singleflight closures now uses `context.WithoutCancel(ctx)`, so the first caller canceling during endpoint DNS validation no longer publishes a false not-found to concurrent valid callers. This matches the list path's existing decoupling for the remote operation itself.

chrislusf:
@coderabbitai re: the finite-context suggestion on `83c68a0` — the build context is intentionally `context.WithoutCancel(ctx)` to prevent the first caller's cancellation from poisoning concurrent valid callers (the Greptile P1). The build is still bounded for the network-touching part: `BuildGuardedRemoteStorageClient` → `validateRemoteEndpoint` creates its own `context.WithTimeout(ctx, 2*time.Second)` for endpoint DNS validation (weed/server/volume_grpc_remote.go:70,163), so that check cannot hang regardless of the caller's context. The remaining client construction is local config/credential parsing without network I/O, matching the sibling `streamFromRemote` / `openRemoteStream` / `FetchAndWriteNeedle` paths which also pass the request context to the same builder. No change needed.

chrislusf:
Addressed both new review points on the load-time validation:

**Transient DNS failures disable mounts** (Greptile P1, Devin 🟡) — fixed in `70d5d4e`. `ValidateRemoteConfForLoad` now uses `validateRemoteEndpointForLoad`, which applies only the static parts of the SSRF deny-list (scheme, IMDS hostnames, IP-literal blocked addresses, gcs credentials) without resolving hostnames. DNS resolution is left to `BuildGuardedRemoteStorageClient` at dial time, which re-resolves and re-validates with the rebinding-safe dialer anyway. So a transient DNS failure during `/etc/remote` reload no longer drops a working mount from the live map — the conf passes at load and the build-time guard catches it at dial if it truly resolves to a blocked address. Fail-fast is preserved for statically-blocked confs (loopback IPs, IMDS hostnames).

**Configured GCS mounts discarded** (Devin 🔴) — this is a pre-existing restriction from the CVE-2026-73080 remediation, not a new regression. `BuildGuardedRemoteStorageClient` (the build-time layer, already used by the volume/filer/s3 streaming paths) calls the same `checkGcsCredentials` and rejects file-path creds. The existing comment in `checkGcsCredentials` notes operators should use the `GOOGLE_APPLICATION_CREDENTIALS` env var for file-path credentials — the inline-JSON restriction is intentional (prevents the server from reading arbitrary caller-supplied file paths). The load-time layer applies the same check for consistency, so both layers reject file-path creds the same way. No change needed.

chrislusf:
Addressed the standard AWS S3 regression (Devin 🔴) in `bcc9b1b`: `guardedRemoteClient` now returns `ok=false` for empty S3-compatible endpoints, so `BuildGuardedRemoteStorageClient` and `ValidateRemoteConfForLoad` fall through to the shared unguarded cache instead of rejecting with "remote endpoint is empty". An empty endpoint is not caller-supplied — the AWS SDK derives the regional endpoint from the region — so there is nothing for the SSRF guard to validate. This also fixes the same pre-existing issue on the sibling streaming read paths (volume `FetchAndWriteNeedle`, filer `streamFromRemote`, s3 `openRemoteStream`) that already used the guarded builder. Added a test asserting both the load-time validator and the build-time builder accept a standard AWS S3 config.

chrislusf:
Refined the empty-endpoint fix (Devin 🟡 "Empty provider endpoints target AWS"): `guardedRemoteClient` now returns `ok=false` (fall through to unguarded) only for `type=s3` with an empty endpoint — the only S3-compatible type that legitimately derives its endpoint from the region. Other S3-compatible types (aliyun, b2, etc.) with empty endpoints are misconfigurations: the AWS SDK would derive an AWS endpoint, misdirecting operations. Those now enter the guard and are rejected with "remote endpoint is empty", matching the pre-existing behavior. Added a test asserting aliyun with an empty endpoint is rejected by both the load-time validator and the build-time builder, while standard AWS S3 (empty endpoint) is accepted.

PR #11296, opened 2026-09-13
filer: accept gcs credentials file paths in the guarded remote client builder

chrislusf:
## Summary

`checkGcsCredentials` rejected all filesystem paths, so a gcs mount configured with `remote.configure -gcs.appCredentialsFile` (which stores a path in `GcsGoogleApplicationCredentials`) was rejected by `BuildGuardedRemoteStorageClient` with "gcs credentials must be inline JSON". This broke existing gcs mounts on the volume `FetchAndWriteNeedle`, filer `streamFromRemote`, and s3 `openRemoteStream` read paths that use the guarded builder.

Read and validate the file content instead of rejecting the path, mirroring what the gcs client itself does in `MakeWithHTTPClient`:
- `loadGcsCredentialsContent` returns the credential JSON for either inline JSON or a filesystem path (reading the file)
- `checkGcsCredentials` uses it to validate the credential type before any client is built — a path that doesn't exist or doesn't contain valid gcs credentials is still rejected
- `guardedRemoteClient` uses it to extract the token exchange URL for the SSRF deny-list, so the rebinding-safe dialer still guards the token endpoint

#### Test plan
- [x] `go build ./weed/...`
- [x] `go test ./weed/server/` (including updated `TestBuildGuardedRemoteStorageClient`)
- [x] A non-credentials file path (`/etc/hostname`) is rejected with "reject remote credentials"
- [x] A file containing valid gcs service-account JSON passes both `checkGcsCredentials` and `BuildGuardedRemoteStorageClient`

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11296" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - GCS credential file paths are now resolved consistently when configuring remote storage connections.
  - Unreadable or missing credential files return a clear, consistent error.
  - Credential-related errors no longer expose sensitive filesystem paths or filenames.
  - Valid credential files continue to be accepted for remote storage configuration.


chrislusf:
Addressed all three review comments in `21f7ce0`:

**Tilde paths rejected** (Greptile P1, Devin 🟡) — `loadGcsCredentialsContent` now calls `util.ResolvePath(creds)` before `os.ReadFile`, matching the gcs client's own behavior in `MakeWithHTTPClient` (line 108). A documented `~/service-account-file.json` path now resolves correctly.

**Remote credentials paths expose server files** (Devin 🟥) — the `os.ReadFile` error includes the file path (`open /etc/passwd: no such file or directory`), which would expose file existence to a caller who planted a conf with an arbitrary path. Replaced the wrapped error with a generic sentinel `errGcsCredentialsUnreadable` ("gcs credentials file is not readable or does not contain valid credentials") so the path is not reflected in the error. The credential type validation still runs on the file content, so a path that doesn't contain valid gcs credentials is rejected before any client is built.

Added tests asserting:
- `/etc/hostname` (non-credentials file) is rejected and the error does not leak the path
- A nonexistent file is rejected and the error does not leak the path
- A valid gcs service-account JSON file passes both `checkGcsCredentials` and `BuildGuardedRemoteStorageClient`

PR #11315, opened 2026-09-14
s3api: configurable trusted-proxy allowlist for aws:SourceIp (#11302)

### LINKED ISSUE #11302: s3api X-Forwarded-For needs to be respected based on trusted whitelist
Commit b88156f removed support for X-Forwarded-For. In an environment where there is a cluster in a private network, and a s3 proxy with a public IP and proxying traffic into the seaweedfs cluster, it dismisses the support of IP ACL per user account in seaweedfs.

Similar to the idea of RemoteIPTrustedProxy in Apache HTTP Server, I suggest to have a config in security.toml to configure "s3 trusted proxies" in style of:
[s3.trusted_proxies]
white_list = "10.10.10.0/24"     # comma separated; bare IPs or CIDRs

chrislusf:
## Summary

Commit b88156f (#11231) removed implicit trust of `X-Forwarded-For` / `X-Real-Ip` headers because, without a configurable allowlist, any caller behind a private-looking peer could spoof `aws:SourceIp` and bypass `IpAddress` / `NotIpAddress` policy restrictions. This restores forwarded-header support behind an explicit, operator-configured allowlist, as proposed in #11302.

Adds a `[s3.trusted_proxies]` section to `security.toml`:

```toml
[s3.trusted_proxies]
white_list = "10.10.10.0/24" # comma separated; bare IPs or CIDRs
```

When the direct TCP peer is in the allowlist, `X-Forwarded-For` is walked right-to-left skipping trusted hops (then `X-Real-Ip`) to determine `aws:SourceIp`. When the peer is not trusted (or no allowlist is configured), the direct peer address is used — matching AWS S3 semantics and the safe default from b88156f.

The allowlist is applied to all three `aws:SourceIp` evaluation paths:
- bucket policy engine (`PolicyEngine`)
- IAM attached-policy engine (`IdentityAccessManagement.iamPolicyEngine`, persisted across rebuilds)
- IAM role/session policy middleware (`S3IAMIntegration`)

Reloaded on `SIGHUP` alongside the JWT signing keys. Also settable via `WEED_S3_TRUSTED_PROXIES_WHITE_LIST`.

### Commits (one per logic change)
1. Add `policy_engine.TrustedProxies` allowlist helper + tests
2. Wire `TrustedProxies` into the bucket/IAM policy engine (`ExtractConditionValuesFromRequest` → method)
3. Wire `TrustedProxies` into the IAM role/session policy middleware
4. Load `[s3.trusted_proxies]` from `security.toml` and wire to all engines + scaffold docs

#### Test plan
- [x] `go build ./...`
- [x] `go test ./weed/s3api/... ./weed/iam/... ./weed/security/...`
- [x] `go vet ./weed/s3api/...` (only pre-existing warnings in untouched test files)
- [x] New unit tests for `TrustedProxies` (trusted/untrusted peer, XFF right-to-left, X-Real-Ip, all-trusted leftmost, nil/empty allowlist, unix socket)
- [x] Regression tests: trusted-proxy path through `PolicyEngine` and `S3IAMIntegration`
- [x] Existing regression tests still pass (forwarded headers ignored without an allowlist)

Closes #11302.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11315" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added trusted reverse-proxy configuration for S3 and IAM source-IP policy conditions.
  - Forwarded client IP headers are honored only for requests from configured trusted proxies.
  - Trusted-proxy settings support application configuration and environment variables.
  - Configuration changes apply during startup and reloads.

- **Bug Fixes**
  - Improved handling of invalid proxy addresses and malformed forwarded headers.

- **Tests**
  - Added coverage for proxy matching, forwarded headers, CIDR ranges, and invalid or missing request addresses.


chrislusf:
Thanks for the review feedback. Pushed two follow-up commits addressing the issues:

**Resolved:**
- **Reload race** -- trustedProxies is now stored behind sync/atomic.Pointer in PolicyEngine and S3IAMIntegration (matching the existing Guard pattern), so SIGHUP swaps cannot race with concurrent request handlers. The IdentityAccessManagement copy is already guarded by iam.m.
- **Bare IPv6 normalization** -- NewTrustedProxies now parses bare entries via net.ParseIP and stores the canonical ip.String(), so 2001:0db8::1 matches a peer rendered as 2001:db8::1. Unparseable bare entries are logged and skipped.
- **Malformed XFF chain bypass** -- the right-to-left walk now stops at the first non-empty unparseable hop (instead of skipping it) and only falls back to the leftmost valid IP when the chain was well-formed. A malformed hop no longer masks a forged IP to its left.

**Not addressed (by design):**
- **Removed allowlist survives reload** -- this is a pre-existing limitation of the Viper MergeInConfig-based reload mechanism: removing a key from security.toml and sending SIGHUP does not clear the previously merged value. This affects all security.toml keys reloaded this way (e.g. JWT signing keys have the same behavior). To clear the allowlist, set white_list = "" explicitly (which produces an empty TrustedProxies and restores the safe direct-peer default). Fixing the merge semantics is a broader change to the config-loading layer and is out of scope for this PR.

go test -race passes for the affected packages.

PR #11323, opened 2026-09-15
expose /metrics on master, volume and filer HTTP ports

chrislusf:
## Summary
- Adds a `/metrics` Prometheus endpoint to the main HTTP port of each server type (master, volume, filer, S3), using the same `promhttp.HandlerFor(stats.Gather, ...)` pattern as the admin server's own `/metrics` endpoint.
- This allows the admin server (or any Prometheus scraper) to collect per-server metrics without requiring operators to configure a separate `-metricsPort` on every server.
- Placed alongside existing `/healthz` and `/status` endpoints on each server's internal/admin mux.

## Why
The admin server already knows every server's HTTP address (from topology + `ListClusterNodes`). With `/metrics` on the main port, the admin can scrape each server directly to power monitoring charts — no extra config, no Push Gateway dependency.

## Changes
- `weed/server/master_server.go`: `/metrics` on master router (alongside `/healthz`)
- `weed/server/volume_server.go`: `/metrics` on volume server's `adminMux` (same port as `/status`)
- `weed/server/filer_server.go`: `/metrics` on filer's `defaultMux` (same port as `/healthz`)
- `weed/s3api/s3api_server.go`: `/metrics` on S3 `apiRouter` (alongside `/status`)

#### Test plan
- [ ] Start `weed master`, `weed volume`, `weed filer`, `weed s3` and `curl http://<host>:<port>/metrics` on each — verify Prometheus text exposition output
- [ ] Verify existing endpoints (`/healthz`, `/status`, `/dir/status`) still work
- [ ] Verify `go build ./...` passes

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11323" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **New Features**
  - Added Prometheus metrics endpoints at `/metrics` for the S3 API, filer, master, and volume servers.
  - Metrics endpoints accept HTTP GET requests and expose collected server metrics for monitoring.


chrislusf:
Thanks — this finding was valid and is fixed in a4e5755.

I confirmed empirically that the S3 registry carries tenant identifiers: 8+ S3 metric families are labelled with `bucket`, e.g. `SeaweedFS_s3_request_total{bucket="...",code="200",type="GET"}`, plus `bucket_size_bytes`, `bucket_object_count`, `bucket_traffic_*`. Serving the shared registry on the client-facing S3 listener let any client that could reach the port enumerate buckets and their traffic with no IAM check.

The `/metrics` route is now removed from the S3 API router. S3 already has a dedicated `-metricsPort` for this.

Master, volume and filer keep the route: they are internal cluster services and their metrics carry no tenant identifiers (collection names at most, not bucket-level traffic).

Verified after the change: `GET /metrics` on the S3 port returns 404, while master, volume and filer still return 200. Downstream monitoring still gets S3 series, because a combined `weed server -s3` process shares one Prometheus registry with the filer.

chrislusf:
Closing in favour of #11326, which takes the opposite approach.

The review feedback here was right, and on reflection the concern is broader than the S3 case that was flagged: the filer and volume HTTP ports are also client-facing (they serve file reads and writes), so putting the shared Prometheus registry on service ports exposes telemetry to clients on those too, not just S3.

#11326 instead has each server advertise the `-metricsPort` listener it already supports, via the paths each server type already reports on (`KeepConnectedRequest`/`ListClusterNodes` for filer and S3, `Heartbeat`/`DataNodeInfo` for volume servers, `GetMasterConfiguration` for masters). A central scraper can then discover the dedicated endpoint, and **no metrics are served on any service port**.

Verified on a live cluster: all four server types advertise the port, and `/metrics` returns 404 on the filer, S3 and volume service ports.

PR #11324, opened 2026-09-15
admin: collect metrics from advertised endpoints, store and chart them

chrislusf:
## Summary
The data + rendering layer for the admin monitoring pages: an in-memory metrics store, a scraper that collects from the metrics endpoints the cluster advertises, and a server-side SVG chart renderer.

> **Stacked on #11326**, which is what makes each node's `-metricsPort` discoverable. (This PR previously sat on #11323, which was closed in favour of that approach.)

## How collection works
- Every 15s the admin builds its target list from the ports nodes advertise: masters via `GetMasterConfiguration`, volume servers via `DataNodeInfo`, filer and S3 via `ListClusterNodes`.
- **Nodes started without `-metricsPort` advertise 0 and are skipped**, so nothing is ever fetched from a client-facing service port.
- Targets are **deduplicated by address**, because a combined `weed server` advertises one listener for all of its components. Series are attributed by metric name, which already identifies the component.
- The admin's own maintenance/worker metrics are read straight from its local registry, with no network round trip.

## Rates and quantiles
Raw counters and histograms are process-lifetime cumulative, so charting them directly is meaningless. Counters become per-second rates and histograms become p50/p95/p99, both computed from the delta against the previous scrape:
- Counter resets (process restarts) are detected and skipped rather than plotted as a huge negative spike.
- Quantiles are interpolated within the matching bucket, as Prometheus does, from **bucket deltas** so they reflect the interval rather than all history.

## Changes
- `metrics_store.go` — bounded ring buffer (240 samples), series keyed by source/metric/labels and retaining labels for label-aware queries
- `metrics_scraper.go` — target discovery, dedup, Prometheus text parsing via `expfmt`, local registry gather
- `metrics_derive.go` — counter rates and histogram quantiles
- `chart.go` — inline SVG renderer with axes, gridlines, multi-series, legends, thresholds and unit formatters. No JS chart library.
- `go.mod` — promotes `prometheus/common` from indirect to direct

#### Test plan
- [x] `go build ./...`
- [x] `go test ./weed/admin/...`
- [x] Unit tests for histogram quantiles, counter-rate resets, bucket-delta mismatches, endpoint resolution and ring bounding
- [x] Live cluster: series collected from the advertised port only; 404 on every service port


chrislusf:
Re: **Scraper Uses Wrong Ports** — correct as of the branch state when reviewed, and now resolved by stacking. `/metrics` is served on each server's main HTTP port by #11323, which is now this PR's base branch, so the addresses the scraper uses do serve the endpoint.

Two caveats worth stating explicitly:
- **S3 is intentionally excluded.** #11323 originally added `/metrics` to the S3 API router too, but that leaks bucket names to any S3 client (8+ S3 metric families carry a `bucket` label), so it was removed. S3 metrics remain available only on the dedicated `-metricsPort`.
- Nodes that do not serve `/metrics` are handled gracefully: the scrape error is logged at V(1) and the affected charts render a "collecting data" placeholder rather than failing the page.

chrislusf:

PR #11325, opened 2026-09-15
admin: monitoring pages for cluster metrics

chrislusf:
## Summary
Adds the **MONITORING** section to the admin UI: an overview page plus five detail pages, server-rendered with inline SVG charts (no JS chart library).

- `/monitoring` — health, throughput, latency, errors, maintenance
- `/monitoring/volume-servers` — health table plus request rate, p99, disk usage, errors
- `/monitoring/filers` — request rate, filer/store p99, in-flight, replication lag
- `/monitoring/s3` — operation rate, 4xx/5xx, p99
- `/monitoring/masters` — heartbeat rate, volume creation, leader changes, placement mismatches
- `/monitoring/workers` — queue depth, slots, completion rate, connected workers

> **Stacked on #11324**, which is stacked on #11326.

## Verified on a live cluster
`weed server -filer -s3 -metricsPort=9988` plus `weed admin`, with filer and S3 read/write/404 traffic across several scrape cycles:

- Volume servers 4/4 charts, S3 3/3, filers 3/4, masters 3/4 populated
- Real values: 2.40 req/s, 394 µs p99, 90% disk usage (matching the host)
- Each detail page shows **one** section, not four duplicates, confirming endpoint dedup in combined mode
- Without `-metricsPort`: pages render, all charts show the placeholder, and the notice explains how to enable collection

## Bugs found and fixed by that verification, each in its own commit
1. **Volume read/write charts empty** — the request counters label `type` with the HTTP method, so `type=read`/`write` matched nothing.
2. **Latency always displayed `0 ms`** — SeaweedFS latencies are frequently sub-millisecond, so a fixed millisecond format rounded every tick to zero. Now steps down to µs.

## Fixes from review feedback on the earlier revision
3. **Master gauges were overcounted** — under-replicated/writable/crowded are cluster-wide values only the leader maintains, so they are read from the leader alone. A demoted master keeps serving stale values.
4. **Sample timestamps were discarded** — samples now carry their scrape time. `reduce` buckets by timestamp, `diskUsagePct` joins used/all on timestamp (dropping cycles that captured only one side), `RenderChart` lays series out on a shared time axis and breaks the line across gaps, and `LatestValue` picks the newest timestamp.

## Notes for reviewers
- Rates and counts are summed across servers; latency quantiles take the **worst** server, since quantiles cannot be summed.
- Disk usage is derived from `SeaweedFS_volumeServer_resource{type=used|all}`.
- Charts show a "collecting data" placeholder until two scrapes exist, since rates and quantiles need a delta.
- The `layout_templ.go` diff is large only because inserting a nav section renumbers templ's generated string literals.
- Generated templ output uses the version pinned in `go.mod` (v0.3.1020); unrelated generated files were left untouched.

#### Test plan
- [x] `go build ./...`
- [x] `go test ./weed/admin/...`
- [x] All six pages return 200 with real scraped data
- [x] Unit tests for timestamp alignment, the disk-usage timestamp join, and newest-sample selection


chrislusf:
Thanks — all three findings were valid and are fixed, each in its own commit. Re-verified on a live cluster afterwards.

**Master gauges are overcounted** — fixed. `under_replicated_volumes`, `volume_layout_writable` and `volume_layout_crowded` are now read from the leader only via a new `leaderSource()` helper, rather than summed across masters. As you noted, a demoted master keeps serving its last values, so summing could both double-count and resurrect stale ones.

**Sample timestamps are discarded** — fixed, and this was the most consequential of the three. Samples now carry their scrape time (`Point{T, V}`) instead of bare floats:
- `reduce` buckets strictly by timestamp.
- `diskUsagePct` joins used/all **on timestamp** and drops any cycle that captured only one side, so it can no longer divide values from different scrapes.
- `RenderChart` lays every series out on a shared time axis built from the union of timestamps, and breaks the path across gaps rather than drawing a straight line through missing time.
- `LatestValue` picks the newest timestamp rather than the last slice element.

Covered by new tests: `TestChartAlignsSeriesByTimestamp`, `TestDiskUsagePctJoinsOnTimestamp`, `TestLatestValueUsesNewestTimestamp`.

**Custom S3 ports break scraping** — fixed by removing the guess entirely. You're right that `-s3.port.grpc` is independent, so reversing the +10000 default was wrong. Related: the S3 `/metrics` route was also dropped from the client-facing port in the base PR, because 8+ S3 metric families carry a `bucket` label. So S3 is no longer a scrape target at all. Instead `SeaweedFS_s3_*` is matched by metric name across every scraped source — the name already identifies the component, and a combined `weed server -s3` process exports these from the filer's registry. Verified: with the S3 port returning 404 for `/metrics`, all three S3 charts still populate. The page is now cluster-wide and states explicitly when data is unavailable.

chrislusf:

PR #11326, opened 2026-09-15
servers: advertise the configured metricsPort so it can be discovered

chrislusf:
## Summary
Lets a central scraper (the admin server, in follow-up PRs) discover each node's Prometheus endpoint, by having every server advertise the `-metricsPort` it was started with.

Replaces #11323, which took the opposite approach of adding `/metrics` to each server's main HTTP port. That was rejected on review because it puts telemetry on client-facing ports — and the concern generalises beyond the S3 case that was flagged, since the filer and volume HTTP ports serve file reads and writes to clients too. **No metrics are exposed on any service port here.**

## The problem
Every server already has a `-metricsPort` Prometheus listener, but nothing advertises it, so it cannot be discovered:

- `StartMetricsServer` no-ops when the port is `0`, which is the default
- `ListClusterNodes` returns only address, version, timestamp, DC and rack
- `KeepConnectedRequest` carries nothing about it either

The pre-existing `metrics_address` / `metrics_interval_seconds` fields do **not** help: they carry the Prometheus **push gateway** that servers push to (`-metrics.address` is documented as `"Prometheus gateway address <host>:<port>"`), not a scrape target.

## Changes
`metrics_port` is added to the paths each server type already reports on:

| Server | Path |
| --- | --- |
| Filer, S3, broker | `KeepConnectedRequest` → `ListClusterNodes` |
| Volume | `Heartbeat` → `DataNodeInfo` (via `VolumeList`) |
| Master | `GetMasterConfigurationResponse` |

Two notes on how it is wired:
- `MasterClient` gains `SetMetricsPort` rather than an eighth constructor parameter, so the callers with no metrics listener are untouched.
- The volume server takes it as a constructor argument because its heartbeat goroutine starts inside `NewVolumeServer`, so a post-construction setter would race the first heartbeat. Similarly it is set on `DataNode` in the master's heartbeat handler rather than threaded through `GetOrCreateDataNode`, which has ~80 test callers.

In combined `weed server` there is one process, one shared registry and therefore one metrics listener, so every component advertises that same port. Consumers are expected to deduplicate by address and attribute series by metric name, which already identifies the component.

## Verified on a live cluster
Ran `weed server -filer -s3 -metricsPort=9977` and queried the master:

```
filer  address=127.0.0.1:8788.18788  metrics_port=9977
s3     address=127.0.0.1:18733       metrics_port=9977
master metrics_port=9977             (push gateway metrics_address="")
volume address=127.0.0.1:8780        metrics_port=9977
```

`GET /metrics` returns 200 on `:9977` and 404 on the filer, S3 and volume service ports. Without `-metricsPort`, every node advertises `0`.

#### Test plan
- [x] `go build ./...`
- [x] `go test ./weed/cluster/ ./weed/topology/ ./weed/storage/ ./weed/wdclient/ ./weed/server/`
- [x] All four server types advertise the port through the master on a live cluster
- [x] No service port serves `/metrics`

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11326" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **New Features**
  * Prometheus metrics listener ports are now propagated and advertised for master, volume, filer, and S3 services.
  * Cluster and node information now reports each service’s metrics port when enabled.
  * Shared server configuration consistently applies the configured metrics port across services.
  * Metrics port details are also synchronized during node registration and heartbeat updates.
* **Bug Fixes**
  * Services can now start safely when no metrics port is configured.


chrislusf:
CI caught a real bug in the first push, now fixed in 2f26d57.

`weed mini` builds `MasterOptions` directly and never set `metricsHttpPort`, so reading it in `toMasterOption` dereferenced a nil pointer and crashed startup. That is what the S3 filer-group and delete-regression suites hit, since both start the cluster with `weed mini`.

Two changes:
- Point mini's master, volume and filer at its single `-metricsPort` listener, the same way `weed server` does. `weed mini` is one process with one shared registry, so one listener serves everything and each component should advertise that port.
- Treat an unset port as disabled in `toMasterOption`, so a partially initialised `MasterOptions` cannot panic this way again.

Verified locally by reproducing the panic with the fix reverted (`weed mini -dir=... -s3.port=...` → `invalid memory address`) and confirming a clean start with it applied, with master and S3 both answering.

chrislusf:

PR #11347, opened 2026-09-16
fix(ec): encode drops tombstoned needles, last-wins replay

franchb:
P0-2 from franchb/seaweedfs#10.

write_sorted_ecx_from_idx sorted by (key,offset) then kept highest offset with no deleted check; decode appends tombstones (K,0,-1) after live rows, so re-encode resurrected deletes. Go readNeedleMap replays file order last-wins, deleted never reach .ecx.

Changes:
- Replay .idx file-order into HashMap (last-wins), drop size.is_deleted() regardless of offset, sort survivors by key.

Tests: test_encode_drops_tombstone_last_wins (tombstone-last dropped, recreate-live kept with offset/size assert). cargo test ec_ 156 PASS, --lib 570 PASS. Refs upstream #9880/#9953 behavior.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11347" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved erasure-coded index generation to correctly handle deleted entries.
  * Prevented deleted records and zero-offset entries from appearing in generated indexes when they are the latest update.
  * Preserved records that are recreated after deletion with their current live data.


franchb:
Review feedback addressed in 213d78e: filter is now size.is_deleted() || offset.is_zero(), matching Go readNeedleMap (ec_encoder.go:747) and CompactNeedleMap::load_from_idx (needle_map.rs:251). Added (K,0,Size(0)) regression case.

PR #11348, opened 2026-09-16
fix(ec): BatchDelete cookie fail-closed via locate_data geometry

franchb:
P0-3 from franchb/seaweedfs#10.

journal_delete_with_cookie located cookie via actual_offset / first-local-shard-size with fall-through to journal on any miss: (a) shard 9 holder deletes shard-3 needle with wrong cookie (bypass), (b) legacy block_size==0 geometry hits wrong bytes → valid delete 500. Go DeleteEcShardNeedle always verifies via remote-capable path.

Changes:
- Locate via locate_ec_shard_needle_interval + interval_to_shard_id_and_offset, read local shard header, fail-closed InvalidData on empty intervals / non-local shard / read error / mismatch. cookie==0 still skips (orphan cleanup). Caller mapping unchanged (Err→500).
- Follow-up out of scope: peer-held shards via distributed reader (needs async/store context); this P0 fails closed instead of bypassing.

Tests: test_journal_delete_wrong_cookie (red: Ok bypass, green: Err cannot verify + no .ecj). cargo test ec_volume 57 PASS, --lib 570 PASS.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11348" target="_blank">
  <picture>
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened deletion validation for erasure-coded data by checking the correct shard and rejecting non-local or invalid requests.
  * Improved handling of missing, mismatched, unreadable, and incomplete deletion metadata.
  * Ensured incomplete metadata reads no longer pass validation.
  * Preserved cookie-check bypass behavior for authorized cleanup operations.


franchb:
Review feedback addressed in 19124ce:

- Devin (skip-cookie): BatchDelete EC call site now passes Cookie(0) when skip_cookie_check=true, so the callee gate actually skips (orphan cleanup on non-primary holders works).
- Greptile (short-read): byte count now required ==4, else 'incomplete header' fail-closed (new truncated-shard test). No .ecj append on either error path.

franchb:
Follow-up in a7a014d: cookie header read now loops to a full 4 bytes (spurious partials retried), EOF/truncation still fail closed with 'incomplete header' and no journal append.

PR #11352, opened 2026-09-16
rust volume: document every unsafe block and stop mutating the process env in tests

franchb:
## What was wrong

Two separate things, both about `unsafe` hygiene in the Rust volume server.

**Undocumented production blocks.** Three `unsafe` blocks carried no `// SAFETY:`
comment at all: `libc::fallocate` in `storage/volume.rs`, `libc::sysinfo` in
`server/memory_status.rs`, and `libc::statvfs` in `storage/disk_location.rs`.
Nothing enforced the convention either — `clippy::undocumented_unsafe_blocks` is
a `restriction` lint, allow-by-default, and appeared nowhere in either crate, so
the next `unsafe` block could land just as bare.

**Tests mutating the process environment.** `config.rs`'s test module called
`std::env::set_var`/`remove_var` through `with_temp_env_var` and
`with_cleared_security_env`, and justified the `unsafe` with a comment saying
every caller holds `process_state_lock()`. That mutex only serialises the
fourteen annotated tests in `config.rs`. The same lib test binary runs the
`grpc_server.rs` tests, which bind a `TcpListener`, dial loopback and drive a
multi-thread tokio runtime; tonic, hyper, rustls and the AWS SDK all read the
environment lazily on those threads. Those are exactly the concurrent readers
Rust 2024 made `set_var` unsafe for, and no test-local mutex can reach them.
`restore_env_var` had no SAFETY comment at all.

`#[serial]`/`serial_test` would **not** fix this. It serialises annotated tests,
which is precisely what the mutex already did — the racing readers are in other
modules and are not annotated. The fix has to remove the unsafety, not schedule
around it.

## What changed

- `seaweed-volume/Cargo.toml` and `seaweed-worker/Cargo.toml` turn on
  `clippy::undocumented_unsafe_blocks = "warn"` next to the existing
  `needless_update = "allow"` policy line. The worker's only `unsafe`
  (`crates/core/build.rs`, setting `PROTOC`) already has a SAFETY comment, so
  the lint costs nothing there and guards the future.
- A `// SAFETY:` comment immediately above every remaining block: the three
  production ones and the three test-only `libc::geteuid()` guards in
  `volume.rs`. `malloc_tuning.rs` and both `build.rs` files already had theirs.
  At the `fallocate` site the SAFETY paragraph follows the pre-existing
  `FALLOC_FL_KEEP_SIZE = 1` line inside the same comment block rather than
  starting it — clippy scans the whole contiguous block, so that satisfies the
  lint and keeps the flag's explanation where it was.
- `memory_status.rs` and `disk_location.rs` each have their block narrowed to
  the `zeroed()` and the libc call, so every comment sits beside the operation
  it justifies and the arithmetic is outside the block. Both restructure the
  conditional into an early `return` on failure (`return None` and
  `return (0, 0)`) so the success path falls through — a control-flow change
  only: the same inputs, the same casts in the same multiplication order, and
  the same values returned on both paths.
- `config.rs` no longer reads the environment implicitly. A single lookup type,
  `type EnvLookup<'a> = &'a dyn Fn(&str) -> Option<OsString>`, is threaded from
  the public entry points down to every reader: `HOME`, `USERPROFILE`, the
  twenty-four `WEED_*` security keys and `SEAWEED_WRITE_QUEUE`. `parse_cli` and
  `parse_security_config` keep their signatures and pass `process_env`, a thin
  wrapper over `std::env::var_os`. `resolve_config` becomes
  `resolve_config_with_env` (still private, one caller), and
  `parse_security_config_with_env`, `resolve_security_config_path`,
  `home_dir_from_env` and `apply_env_overrides` each take the lookup.
- The tests pass a lookup built by a `fake_env` helper over a `HashMap`, or
  `&|_| None` for "nothing set". `with_temp_env_var`, `restore_env_var`,
  `with_cleared_security_env` and every `unsafe` in the module are gone.
- `process_state_lock()` stays, with a smaller and honest job: `set_current_dir`
  is safe but still process-global, so tests that move the working directory are
  serialised against the ones that read it. Tests naming an explicit config file
  never reach that search and no longer take the lock.

## What did not change

- No production behaviour, including at the two narrowed `unsafe` sites: the
  restructured conditionals return the same values on both paths. The same env
  keys are read, in the same order, with the same precedence — including the
  `WEED_GRPC_CA` / `WEED_GRPC_VOLUME_CA` fallback and Go's Viper naming
  convention. `env_string` reproduces `std::env::var(key).ok()` exactly: absent
  and non-UTF-8 both read as unset.
- No public API. `parse_cli()` and `parse_security_config(path: &str)` keep
  their signatures, so `main.rs`'s startup and SIGHUP-reload call sites are
  untouched. The `_with_env` variants are private.
- No test was added or removed (30 in `config.rs`, 569 in the lib, as before),
  and `std::env::args()`, `std::env::temp_dir()` and `std::env::current_dir()`
  still read the real process state — they are not variable lookups and nothing
  in the suite mutates them.
- Clippy is still not a CI gate: the step in
  `.github/workflows/rust-volume-server-tests.yml` is commented out. The lint is
  a local convention that `cargo clippy --all-targets` enforces before a push.

## How it was verified

From `seaweed-volume/`, on stable 1.98.1:

- RED, with the lint on and before the fixes: `cargo clippy --all-targets`
  reported all eight undocumented blocks — `config.rs:1220`, `config.rs:1231`,
  `server/memory_status.rs:39`, `storage/disk_location.rs:1131`,
  `storage/volume.rs:4650`, `:7296`, `:7355`, `:7502`.
- GREEN: `cargo clippy --all-targets` and
  `cargo clippy --all-targets -- -W clippy::undocumented_unsafe_blocks` are both
  silent. `cargo fmt --check` clean.
- `cargo test`: 569 + 5 + 1 + 28 passed, 0 failed. Run three times back to back
  with identical counts — the full suite is what exercises the config tests
  alongside the socket-binding gRPC tests, so it is the run that would show an
  env-related flake. `cargo test config::` (30 tests) also run three times as a
  targeted check.
- `cargo test --lib disk_location` (19 passed) and `cargo test --lib config::`
  (30 passed) as targeted checks on the two rewritten regions.
- `cargo build --release --no-default-features` (4-byte offsets) and
  `cargo test --features redb-experimental-cursor --lib storage::needle_map`
  (55 passed), since `volume.rs` is touched.
- From `seaweed-worker/`: `cargo clippy --all-targets` clean across all three
  member crates with the workspace lint on.

Part of franchb/seaweedfs#10 (P3) — closes the two unsafe-checker checkboxes.

## Summary by CodeRabbit

- **Reliability**
  - Improved handling when system memory or disk statistics cannot be read, allowing the service to continue with safe fallback values.

- **Configuration**
  - Environment-based configuration parsing is now more consistently handled, including security and write-queue settings.

- **Documentation**
  - Added safety documentation around low-level system operations to improve maintainability and diagnostics.


franchb:

PR #11353, opened 2026-09-16
rust volume: derive has_remote_file instead of mirroring it

franchb:
### What was wrong

`Volume` carried `pub has_remote_file: bool` next to `pub volume_info: PbVolumeInfo`. The bool was only ever the answer to `!volume_info.files.is_empty()`: outside the two constructors, which set it to `false`, the single writer was `refresh_remote_write_mode`, whose first line recomputed it from the list.

Because both fields were public, keeping them in agreement was a convention, not an invariant. Every site that touched `volume_info.files` — `load_vif` (two paths), the tier-up handler, the tier-down handler and its rollback — had to remember to call `refresh_remote_write_mode` afterwards. A caller that forgot would leave the volume advertising a write mode its `.vif` contradicts, or serving a remote `.dat` through a writable needle map.

Go keeps a flag too — `weed/storage/volume.go:37 hasRemoteFile atomic.Bool` — but it is unexported, and exactly two places in the package set it. The `.vif` load path stores it alongside the volume info it was computed from (`volume_tier.go:22-23`, where `MaybeLoadVolumeInfo` derives it from a non-empty files list at `volume_info/volume_info.go:63-67`), and `SwapDataBackend` stores it together with the data backend under `dataFileAccessLock`, precisely so the flag and the backend cannot disagree (`volume.go:306-313`). Nothing else can move it.

The Rust port bound its flag to `volume_info.files` and then left both the flag and the info public, which is the half of that design that went missing: any caller could edit one without the other. The shape this PR gives Rust matches Go's — one in-module load path (`load_vif`) plus one narrow entry point for everyone else (`update_remote_files`) — and goes one step further by deriving the answer instead of storing it, so the two cannot drift even inside the module.

### What changed

- The field is gone. `pub fn has_remote_file(&self) -> bool` returns `!self.volume_info.files.is_empty()`, so the two can no longer drift.
- `volume_info` is private, with `pub fn volume_info(&self) -> &PbVolumeInfo` for the read-only callers (`VolumeInformationMessage` assembly, the tier-up duplicate-backend scan, the tier-down mtime lookup, tests).
- `pub fn update_remote_files(&mut self, f: impl FnOnce(&mut Vec<PbRemoteFile>)) -> Result<(), VolumeError>` applies the edit and then refreshes the derived write mode and the needle map. It is the only way to change the reference list from outside `storage::volume`; the tier-up push, the tier-down removal and its rollback, and the tests now go through it.
- `refresh_remote_write_mode` has no caller left outside the module, so it is private. Inside, it is reached from `load_vif`'s two whole-struct assignments and from `update_remote_files`.

### What did not change

- The refresh logic itself: same flag derivation, same `reconcile_index_mode` call, same "pin read-only on failure" behaviour.
- The order of operations in both tier handlers, including the cancellation checks, the `.vif` save, the fsyncs and the remote-object delete.
- The tier-down rollback semantics. It still snapshots the removed reference before the refresh runs, restores it on failure, and re-refreshes on the error path *unconditionally* — the restoring `update_remote_files` call runs with a no-op closure when there was nothing to restore, matching the old code, which re-ran `refresh_remote_write_mode` whether or not it had re-inserted a reference.
- `update_remote_files` is not transactional, and does not pretend to be: on error the closure has already been applied and the volume is pinned read-only, which is exactly why the tier-down rollback exists. The doc comment says so.

### How it was verified

A new unit test, `test_update_remote_files_refreshes_the_derived_write_mode`, pushes a `RemoteFile` through `update_remote_files` and asserts `has_remote_file()` plus `is_no_write_can_delete()` flip on and `is_no_write_or_delete()` stays off, then removes it and asserts all three go back — the invariant the field used to depend on a convention for.

From `seaweed-volume/`: `cargo fmt --check`, `cargo clippy --all-targets` (both clean), `cargo test` (570 + 5 + 1 + 28 passed, 0 failed), `cargo build --release --no-default-features` (4-byte offsets), `cargo test tier` (7 passed, covering both tier-move handlers and the cancellation path), and `cargo test --features redb-experimental-cursor --lib storage::needle_map` (55 passed).

Refactor only: no behaviour change intended, and no `.vif` or wire format change.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when switching volumes between local and remote storage modes.
  - Ensured index and write behavior remain consistent after tier-up or tier-down operations.
  - Improved recovery when a storage-mode transition fails, preserving a safe read-only state when necessary.

- **Refactor**
  - Consolidated remote-file state handling to keep volume metadata and operational behavior synchronized.


franchb:

One note: this overlaps #11357 in the tier-move handlers by design. They are independent changes, so either order works — whichever lands second I will rebase.

PR #11355, opened 2026-09-16
rust volume: typed errors for store compaction so gRPC can answer NotFound

franchb:
### What was wrong

The vacuum entry points on `Store` returned `Result<_, String>`: `compact_volume`,
`commit_compact_volume`, `cleanup_compact_volume` and `delete_collection`. The gRPC layer had
nothing to branch on, so `VacuumVolumeCompact`, `VacuumVolumeCommit`, `VacuumVolumeCleanup` and
`DeleteCollection` all answered `Status::internal` no matter what happened. A vacuum loop that
races a volume being moved or deleted got the same code back as one that hit a failing disk, and
`weed shell` could only tell the two apart by matching on the message text. `VacuumVolumeCheck`
right next door already answers `NotFound` for the same condition, so the RPCs in this family
were inconsistent with each other.

Go's `weed/storage/store_vacuum.go` has the same flat-`error` shape, but Go's shell never
inspects those messages — nothing under `weed/shell/` matches on "is not found during" or "not
enough free space" — so tightening the Rust codes is safe. The messages do change wording; see
"Behaviour change" below.

Three smaller things in the same code:

- `Store::check_compact_volume` had zero callers anywhere in the crate. `VacuumVolumeCheck`
  computes the garbage level from its own `find_volume`, so the helper was dead duplication.
- `compact_volume` looked the volume up, then looked it up a second time with `.unwrap()` purely
  to dodge a borrow while computing the required space.
- `ascending_visit` on `CompactNeedleMap`, `RedbNeedleMap`, `SortedFileNeedleMap` and the
  `NeedleMap` dispatch hard-coded a `String` visitor error, while `CompactMap::ascending_visit`
  underneath them was already generic over the error type.

### What changed

`VolumeError` gains two variants:

- `VolumeNotFound(VolumeId)` — `#[error("volume id {0} is not found")]`. The existing `NotFound`
  is needle-level and carries no payload, so the compaction and collection API needs its own
  variant to keep the id in the message an operator reads. The rest of `Store` still answers
  `NotFound` for a missing volume; the migration is incremental.
- `InsufficientSpace { vid, required, free }` —
  `#[error("not enough free space: required {required}, free {free}")]`. The `vid` field is kept
  for callers that inspect the error; the message leaves it out because every call site already
  prefixes `compact volume {vid}: `.

The four store functions return `Result<_, VolumeError>`. `check_compact_volume` is deleted.
`compact_volume` binds the disk index and the size estimate from a single `find_volume`, so the
second scan and the `unwrap()` are gone and the sizes describe the same lookup the free-space
check is made against.

`server/mod.rs` gets `impl From<VolumeError> for tonic::Status`:

| variant | code |
| --- | --- |
| `NotFound`, `VolumeNotFound` | `not_found` |
| `ReadOnly` | `failed_precondition` |
| `InsufficientSpace` | `resource_exhausted` |
| `AlreadyExists` | `already_exists` |
| everything else | `internal` |

The `ReadOnly` and `AlreadyExists` rows are part of the mapping but no RPC wired up in this PR can
reach them today — they are there so the table is complete for the handlers that move over next.

The message is the error's `Display`. A `status_with_context(context, err)` helper keeps the
code from that mapping and prefixes the RPC's own context, so the four call sites still produce
`commit compact volume 7: ...`, `cleanup volume 7: ...`, `compact volume 7: ...` and
`delete collection pics: ...` the way the storage layer used to format them in place.

Follow-up note for whoever migrates the next handler: `VolumeDelete` still maps ad-hoc at
`grpc_server.rs:1563-1572`, and it maps `NotEmpty` → `failed_precondition`. The `From` impl has no
`NotEmpty` row, so that row has to be added *before* `delete_volume` is switched over, or a
non-empty volume would start answering `INTERNAL`.

`ascending_visit` is now `<F, E>` on all four types. Three signatures bound `E: From<String>` —
`RedbNeedleMap`, `SortedFileNeedleMap` and the `NeedleMap` dispatch that fans out to them — so
their own read failures (`redb iter: ...`, a short `.sdx`) still reach the caller. The in-memory
`CompactNeedleMap` needs no bound. Four visitors that never fail now name their error type
explicitly, because there is nothing left to infer `E` from: `Infallible` for the in-memory walk in
`iter_entries`, `String` for the three test visitors that go through a bounded signature.

### Behaviour change

Two user-visible parts. First, these RPCs now return codes other than `Internal`:

- `VacuumVolumeCompact` / `VacuumVolumeCommit` / `VacuumVolumeCleanup` on a volume this server
  does not have → `NOT_FOUND` (was `INTERNAL`).
- `VacuumVolumeCompact` when the disk cannot hold the rewritten copy → `RESOURCE_EXHAUSTED`
  (was `INTERNAL`).
- Everything else, including all I/O failures, still → `INTERNAL`.

Second, the message wording changes. The store-side phrases are gone:

| was | is now |
| --- | --- |
| `volume id 7 is not found during compact` | `compact volume 7: volume id 7 is not found` |
| `volume id 7 is not found during commit compact` | `commit compact volume 7: volume id 7 is not found` |
| `volume id 7 is not found during cleaning up` | `cleanup volume 7: volume id 7 is not found` |
| `not enough free space to compact volume 7. Required: X, Free: Y` | `compact volume 7: not enough free space: required X, free Y` |

Every message still names the volume and still says what failed, but "during compact" /
"during commit compact" / "during cleaning up" and the `Required: X, Free: Y` capitalisation no
longer appear. Anything grepping for those exact phrases needs updating; nothing in this repo does
— see the verification section.

### What did not change

No storage format, no on-disk layout, no wire schema. The `internal` fallback means any error not
listed above is reported exactly as before. `Status::internal` is still what `check_maintenance`,
the is-stopping gate and every other pre-store check produce. No storage-layer file gained a
`tonic` dependency: the `From` impl lives under `server/`.

Only these four RPCs are wired through the new mapping. Other handlers that surface a
`VolumeError` keep their ad-hoc conversions; moving them over is mechanical follow-up work and is
deliberately not in this PR, which is scoped to the compaction API that the `Result<_, String>`
signatures were blocking.

### Out of scope

`remote_storage/s3_tier.rs` keeps its `String` errors — retyping it belongs with the P2
`ObjectNotFound` / `block_on_tier_future` fixes. `needle.rs:661,712` and `ttl.rs:78` keep theirs
too: their P1 non-ASCII panic fixes land first, and retyping afterwards avoids two PRs editing the
same lines. `endpoint_guard.rs` stays as it is — those are validation predicates and a string is
the right shape for them.

### How it was verified

From `seaweed-volume/`, on stable 1.98.1:

- `cargo fmt --check` — clean.
- `cargo clippy --all-targets` — clean.
- `cargo test` — 573 lib + 5 bin + 1 + 28 integration tests, 0 failures.
- `cargo test --features redb-experimental-cursor --lib storage::needle_map` — 55 passed.
- `cargo build --release --no-default-features` (4-byte offsets) — builds.

Four new tests. The first two were written before the change and confirmed failing against
master:

- `storage::store::tests::test_compaction_of_missing_volume_is_volume_not_found` — compact,
  commit and cleanup of an unmounted vid all return `VolumeError::VolumeNotFound`, and its
  `Display` still names the id.
- `server::grpc_server::tests::test_vacuum_volume_commit_missing_volume_is_not_found` —
  `VacuumVolumeCommit` for a vid this server does not have answers `Code::NotFound` and still
  names the volume. On master this fails with `Internal`.
- `storage::store::tests::test_compact_then_commit_reclaims_a_deleted_needle` — the happy path
  the not-found test cannot reach: writes three needles, deletes one, compacts and commits, and
  asserts the volume shrinks, stays mounted, and reports 2 live / 0 deleted. This is what covers
  the folded lookup, the free-space check and the mutable re-lookup.
- `server::tests::test_volume_error_maps_to_grpc_code` — every row of the mapping table, the
  `internal` fallback, and that `status_with_context` preserves the code while prefixing.

`grep -rn "is not found during\|not enough free space" weed/` returns only
`weed/storage/store_vacuum.go` (four lines, the Go server's own producers). Nothing under
`weed/shell/` matches either string, so no shell command depends on the old wording.

## Summary by CodeRabbit

- **Bug Fixes**
  - Volume operations now return accurate status codes, including “Not Found,” “Failed Precondition,” “Resource Exhausted,” and “Already Exists.”
  - Error messages now include relevant operation context and volume details.
  - Missing volumes during compaction, commit, cleanup, vacuum, or deletion are no longer reported as generic internal errors.
  - Compaction now clearly reports insufficient disk space and can reclaim space from deleted data while keeping volumes mounted.
  - Vacuum operations consistently return clear errors while preserving successful responses and status reporting.


franchb:

chrislusf:
Pushed a review commit on top (`4019ddaaa`):

- `vacuum_volume_check` now returns `VolumeError::VolumeNotFound` like its siblings — its `"not found volume id"` wording differed from the variant's message
- trimmed the rationale comments to match the file's style; no GitHub issue references in code
- Go mirror: `store_vacuum.go` now wraps `ErrVolumeNotFound` and a new `ErrInsufficientSpace`, and the vacuum + `DeleteCollection` handlers map them through `volumeStatusError` (generalized from `volumeDeleteStatusError`) — so a Go volume server answers `NotFound`/`ResourceExhausted` exactly like the Rust one now does. Same wire behaviour regardless of which implementation serves the volume.

chrislusf:
Two follow-up commits pushed:

- `5b636942c` — addressed the CodeRabbit note: Go vacuum/admin handlers now wrap lower-level errors with `%w` (`compact volume %d`, `commit compact volume %d`, `cleanup volume %d`, `delete collection %s`) so `volumeStatusError` still sees the sentinel chain while callers get operation context — matching what the Rust `status_with_context` adds. `VolumeDelete`/`CheckCompactVolume` left as-is per the review.
- `f76db39c0` — `VolumeError::NotEmpty` now maps to `FailedPrecondition` in the shared `From<VolumeError> for Status` (parity with Go's `ErrVolumeNotEmpty` mapping), and `volume_delete` uses `status_with_context` instead of its own match.

PR #11356, opened 2026-09-16
rust volume: keep the EC shard-location map, its refresh time and stale mark under one lock

franchb:
**What was wrong.** Each `EcVolume` caches where its EC shards live, when it last
learned that from the master, and whether a failed read has since disproved it.
Those three pieces were three separate fields under three separate locks —
`RwLock<HashMap<ShardId, Vec<String>>>`, `Mutex<Option<Instant>>`,
`Mutex<bool>` — with nothing tying them together.

- `merge_shard_locations` published the merged map, released the write lock, and
  only then took the second lock to stamp the refresh time.
- Both readers took the two guards one after the other:
  `scrub_ec_volume_distributed`'s under-lock snapshot and `build_snapshot` on the
  distributed-read path.

So a reader landing between the two writes paired a freshly merged map with the
*previous* lookup's timestamp. That pair is exactly what `needs_refresh` judges,
which means the read aged a brand-new map against an old clock and went back to
the master for a lookup that had just completed. Go keeps the same state in one
struct behind one `ShardLocationsLock sync.RWMutex`
(`weed/storage/erasure_coding/ec_volume.go`), so there is no window there.

`replace_shard_locations` documented itself as moving the cache "from old → fresh
in a single observable step" while doing the same two-section write. It, plus
`set_shard_locations` and `get_shard_locations`, had zero callers anywhere in the
crate.

**What changed.**

- One `ShardLocationCache { locations, refreshed_at, stale }` behind a single
  `RwLock`, on a now-private `EcVolume::shard_locations` field. Every Go-parity
  doc comment moved onto the new struct and its fields.
- `merge_shard_locations` upserts the reply, stamps the refresh time and clones
  the result in one write section.
- `shard_locations_snapshot() -> (HashMap, Option<Instant>)` replaces both
  two-guard readers: map and time now come from one read section and always
  describe the same refresh.
- `mark_shard_locations_stale()` / `claim_shard_locations_refresh()` move the
  mark's read-and-consume onto the cache, so it is judged under the lock that
  guards the map rather than a lock of its own.
- The three zero-caller accessors are deleted. With the field private, the
  map/time/mark invariant can no longer be sidestepped from outside the module.
- The two test-only seeding sites in `grpc_server.rs` now call
  `merge_shard_locations`, which already produces the state they were assembling
  by hand.

**What did not change.** The freshness rule. `needs_refresh` keeps its 11s / 7min
/ 37min thresholds and still judges the caller's snapshot — the map that caller
will actually read from — rather than whatever happens to be cached by the time
the claim runs; only the stale mark is read from under the new lock.

This is a deliberate divergence from Go, and it is worth naming: Go's
`cachedLookupEcShardLocations` (`weed/storage/store_ec.go:709-716`) takes
`ShardLocationsLock` and judges the *live* `len(ShardLocations)` and
`ShardLocationsRefreshTime` under it, alongside the mark. Here the decision keeps
using the snapshot the caller already captured under the store lock, because the
read path then proceeds on exactly that snapshot: judging a live map a concurrent
merge had just replaced would report "fresh, no lookup needed" while the read
still serves from its own older copy. That is unchanged behaviour from before this
PR — the split locks never let the decision see a live map either — so the
divergence is pre-existing, not introduced here. What did move under the map's
lock is the mark's read-and-consume.

The master lookup, the completeness guard in `write_back_shard_locations`, the
per-shard upsert semantics and the stale-mark-goes-back-on-failure behaviour are
all untouched. No on-disk format, protocol or public behaviour moves.

**How it was verified.** From `seaweed-volume/`: `cargo fmt --check`,
`cargo clippy --all-targets` (both clean), bare `cargo test` (571 + 5 + 1 + 28
passing, 0 failed — the lib suite goes 569 -> 571, the two new tests being purely
additive), plus `cargo test ec_` (157) and `cargo test store_ec` (27).
Two new unit tests in `ec_volume.rs` pin the API contract rather than the
atomicity itself: one that a merge's map and its refresh time come back from a
single snapshot call and that a second merge re-stamps the time, one that the
stale mark is consumed by exactly the claim that acts on it and survives a claim
that declines. Both fail against master by not compiling — there is no call that
returns both halves there — so they are a regression fence, not a demonstration
of the race. The atomicity is structural: one private field, one lock, one
section per operation, and no way to write the map without the timestamp.

## Summary by CodeRabbit

- **Reliability**
  - Improved consistency when tracking erasure-coded shard locations during volume scrubbing and snapshot operations.
  - Reduced the risk of stale or partially updated shard-location information being used during maintenance activities.

- **Refactor**
  - Consolidated shard-location state handling to provide more coordinated updates and reads.
  - Updated internal test setup to use the same shard-location management behavior as production workflows.


franchb:

PR #11375, opened 2026-09-17
s3: commit multipart upload and remove .uploads atomically; purge completed uploads metadata-only

### LINKED ISSUE #10663: [s3] Cleaning uploads can cause corruption [LOW RISK]
When running `s3.clean.uploads` which is enabled by default in the admin script worker (weed scaffold -config=master), it cleans files under `.uploads` and deletes their contents as well. A successful upload takes those chunks and then does a metadata delete on those files so that there is only one referent per file chunk. If an error occurred when metadata-deleting the `.uploads` files, the admin script would delete the chunks from the completed file as the partial files reference them. The error is simply logged only at V(1+) and ignored. This is low risk because the metadata file deletion would have to fail or the s3 server would have to go offline right before it deletes the directory. 

The easiest way to fix this is to look at the `SeaweedFSUploadId` while pruning files in `.uploads` to determine if the multipart chunks assembled into a complete file. If such a file exists, simply do a metadata delete instead.

The other one is to use the filer's `ObjectTransaction` request to apply the following under a lock when completing an upload
1. Put the new version in with all the chunks
2. Delete old metadata
3. Update the latest version

Everything runs in order, no roll back, exits on the first failure, and things are committed in order, and observable partial states are completely safe. If step 2 fails, the upload is failed and the stray version is cleaned up. If the stray version fails to be cleaned up then it can also cause complications but that is pre-existing and out of scope for this issue.

Example patch 
<details>

```diff
diff --git a/weed/s3api/s3api_multipart_finalize_txn.go b/weed/s3api/s3api_multipart_finalize_txn.go
new file mode 100644
--- /dev/null
+++ b/weed/s3api/s3api_multipart_finalize_txn.go
@@ -0,0 +1,139 @@
+package s3api
+
+import (
+	"time"
+
+	"github.com/seaweedfs/seaweedfs/weed/glog"
+	"github.com/seaweedfs/seaweedfs/weed/pb"
+	"github.com/seaweedfs/seaweedfs/weed/pb/filer_pb"
+	"github.com/seaweedfs/seaweedfs/weed/s3api/s3err"
+)
+
+// routedMultipartFinalize commits a completed multipart upload in ONE
+// ObjectTransaction on the owner filer, instead of three separate RPCs:
+//
+//  1. PUT     <object>.versions/<versionFile> (the finished object)
+//  2. DELETE  <bucket>/.uploads/<uploadId>   (metadata only)
+//  3. RECOMPUTE_LATEST on <object>.versions   (publish it as current)
+//
+// Why this exists rather than extending routedVersionedFinalize: that helper is
+// shared with CopyObject and createDeleteMarker, neither of which has an upload
+// directory or writes its entry inside the transaction. Keeping the multipart
+// shape here leaves those call sites untouched.
+//
+// Why the upload directory is removed with IsDeleteData=false: a multipart
+// object's chunks ARE the part chunks. Completion references the fids, it does
+// not copy bytes, so freeing them would destroy the object being published.
+// After this transaction the upload directory no longer exists, so
+// s3.clean.uploads has nothing to mistake for an abandoned upload.
+//
+// ORDER MATTERS, and not for the usual reason. ObjectTransaction holds one
+// per-path lock for the whole request (filer_grpc_server.go:305-307), which
+// gives isolation from other writers of the same object — but the mutation loop
+// applies each mutation in turn and returns on the first error with NO rollback
+// and no store-level transaction (filer_grpc_server.go:333-339). So a partial
+// application is reachable, and the order decides which partial states exist:
+//
+//	PUT, DELETE, RECOMPUTE   (this order)
+//	  fail at 1 -> nothing happened; the client can retry the completion
+//	  fail at 2 -> version file written but not current, upload directory still
+//	               there: BOTH reference the chunks, so nothing is unreferenced.
+//	               The caller rolls the version file back and the state is exactly
+//	               stock's "upload not completed" -- fully retryable.
+//	  fail at 3 -> version exists but is not current; the read-path heal and the
+//	               background reconciler both promote it
+//
+// Two orderings are wrong, for opposite reasons:
+//
+//	RECOMPUTE first leaves "object published, upload directory still present",
+//	which is exactly the bug this is fixing.
+//
+//	DELETE first leaves, between mutations 1 and 2, chunks referenced by NOTHING.
+//	The per-path lock does not help there: volume.fsck computes orphans from its
+//	own filer scan and never consults that lock, so a concurrent
+//	`volume.fsck -reallyDeleteFromVolume` could classify the parts as orphans and
+//	delete them, and mutation 2 would then publish an object over freed needles.
+//	Narrow (two store operations) and gated behind an expert-only flag and fsck's
+//	5h -cutoffTimeAgo, but it is a window stock does NOT have: stock keeps the
+//	parts referenced by .uploads/<id> until after the object is committed.
+//	Putting the PUT first removes it entirely.
+//
+// None of the reachable partial states loses a byte the client was told was
+// durable, and none leaves a chunk unreferenced.
+func (s3a *S3ApiServer) routedMultipartFinalize(
+	owner pb.ServerAddress,
+	bucket, object string,
+	useInvertedFormat bool,
+	versionDir, versionFileName string,
+	chunks []*filer_pb.FileChunk,
+	decorate func(*filer_pb.Entry),
+	uploadsDir, uploadID string,
+) s3err.ErrorCode {
+
+	// Same shape filer_pb.MkFile builds (filer_client.go:272-283); the owner
+	// applies collection/replication/TTL defaults itself in applyObjectMutation.
+	now := time.Now().Unix()
+	versionEntry := &filer_pb.Entry{
+		Name:        versionFileName,
+		IsDirectory: false,
+		Attributes: &filer_pb.FuseAttributes{
+			Mtime:    now,
+			Crtime:   now,
+			FileMode: uint32(0770),
+			Uid:      filer_pb.OS_UID,
+			Gid:      filer_pb.OS_GID,
+		},
+		Chunks: chunks,
+	}
+	if decorate != nil {
+		decorate(versionEntry)
+	}
+
+	mutations := []*filer_pb.ObjectMutation{
+		{
+			Type:      filer_pb.ObjectMutation_PUT,
+			Directory: versionDir,
+			Name:      versionFileName,
+			Entry:     versionEntry,
+		},
+	}
+	if uploadsDir != "" && uploadID != "" {
+		mutations = append(mutations, &filer_pb.ObjectMutation{
+			Type:         filer_pb.ObjectMutation_DELETE,
+			Directory:    uploadsDir,
+			Name:         uploadID,
+			IsDeleteData: false,
+			IsRecursive:  true,
+		})
+	}
+	mutations = append(mutations,
+		s3a.latestPointerRecompute(bucket, object, useInvertedFormat, "", true),
+	)
+
+	// With no ring view there is no owner to route to, but ObjectTransaction
+	// only requires LockKey: an empty RouteKey makes the receiving filer apply
+	// the mutations locally under its own per-path lock instead of forwarding
+	// (filer_grpc_server.go:277). That is no weaker than the three unlocked RPCs
+	// this path used to make, and it still gets the ordering guarantee, so the
+	// upload directory can never outlive the published object.
+	routeKey := ""
+	if owner != "" {
+		routeKey = s3a.objectRouteKey(bucket, object)
+	}
+	req := &filer_pb.ObjectTransactionRequest{
+		LockKey:   s3a.toFilerPath(bucket, object),
+		RouteKey:  routeKey,
+		Mutations: mutations,
+	}
+	resp, err := s3a.objectTxnOnFiler(owner, req)
+	switch {
+	case err != nil:
+		glog.Errorf("routedMultipartFinalize: %s/%s upload %s on %s: %v", bucket, object, uploadID, owner, err)
+		return s3err.ErrInternalError
+	case resp.Error != "":
+		glog.Errorf("routedMultipartFinalize: %s/%s upload %s: %s", bucket, object, uploadID, resp.Error)
+		return s3err.ErrInternalError
+	default:
+		return s3err.ErrNone
+	}
+}
diff --git a/weed/s3api/filer_multipart.go b/weed/s3api/filer_multipart.go
--- a/weed/s3api/filer_multipart.go
+++ b/weed/s3api/filer_multipart.go
@@ -236,6 +236,7 @@
 	manifestsReferenced     bool                  // failed rollback left an entry holding newManifestChunks
 	supersededPartManifests []*filer_pb.FileChunk // part-entry blobs replaced by flattening; deleted after commit
 	metadataOnlyCleanup     bool                  // deleteEntries share chunks with the live object; keep their data
+	uploadDirRemoved        bool                  // .uploads/<id> was removed inside the finalize transaction
 }
 
 func completeMultipartResult(r *http.Request, input *s3.CompleteMultipartUploadInput, etag string, entry *filer_pb.Entry) *CompleteMultipartUploadResult {
@@ -662,8 +663,9 @@
 			versionMtime := time.Now().Unix()
 			amzAccountId := r.Header.Get(s3_constants.AmzAccountId)
 
-			// Create the version file in the .versions directory
-			if err := s3a.mkFile(versionDir, versionFileName, completionState.finalParts, func(versionEntry *filer_pb.Entry) {
+			// How the version entry is filled in. On the routed path this runs
+			// inside the finalize transaction; on the fallback it runs in mkFile.
+			decorateVersionEntry := func(versionEntry *filer_pb.Entry) {
 				if versionEntry.Extended == nil {
 					versionEntry.Extended = make(map[string][]byte)
 				}
@@ -712,9 +714,6 @@
 				}
 				versionEntry.Attributes.FileSize = uint64(completionState.offset)
 				versionEntry.Attributes.Mtime = versionMtime
-			}); err != nil {
-				glog.Errorf("completeMultipartUpload: failed to create version %s: %v", versionId, err)
-				return s3err.ErrInternalError
 			}
 
 			// Construct entry with metadata for caching in .versions directory
@@ -737,21 +736,44 @@
 			// Pass entry to cache its metadata for single-scan list efficiency
 			// Route the pointer flip to the owner (off the lock) via
 			// RECOMPUTE_LATEST; the just-written version file is the newest.
-			if owner != "" {
-				if code := s3a.routedVersionedFinalize(owner, *input.Bucket, *input.Key, useInvertedFormat); code != s3err.ErrNone {
+			// objectTxnOnFiler needs either a routed owner or a filerClient to
+			// pick a filer from; only the initialization/testing case where both
+			// are absent still needs the legacy three-RPC sequence.
+			if owner != "" || s3a.filerClient != nil {
+				// The transaction removes the whole upload directory, so the
+				// superseded part entries -- whose chunks the object does NOT
+				// reference -- have to be freed first or they leak.
+				for _, deleteEntry := range completionState.deleteEntries {
+					if err := s3a.rm(uploadDirectory, deleteEntry.Name, !completionState.metadataOnlyCleanup, true); err != nil {
+						glog.Warningf("completeMultipartUpload cleanup %s upload %s unused %s : %v", *input.Bucket, *input.UploadId, deleteEntry.Name, err)
+					}
+				}
+				// One transaction: drop the upload directory, write the version
+				// entry, publish it. See routedMultipartFinalize on why the order
+				// is what it is.
+				if code := s3a.routedMultipartFinalize(owner, *input.Bucket, *input.Key, useInvertedFormat,
+					versionDir, versionFileName, completionState.finalParts, decorateVersionEntry,
+					s3a.genUploadsFolder(*input.Bucket), *input.UploadId); code != s3err.ErrNone {
 					if rollbackErr := s3a.rollbackMultipartVersion(versionDir, versionFileName); rollbackErr != nil {
 						glog.Errorf("completeMultipartUpload: failed to rollback version %s for %s/%s after routed finalize error: %v", versionId, *input.Bucket, *input.Key, rollbackErr)
 						completionState.manifestsReferenced = true
 					}
 					return code
 				}
-			} else if err := s3a.updateLatestVersionInDirectory(*input.Bucket, *input.Key, versionId, versionFileName, versionEntryForCache); err != nil {
-				if rollbackErr := s3a.rollbackMultipartVersion(versionDir, versionFileName); rollbackErr != nil {
-					glog.Errorf("completeMultipartUpload: failed to rollback version %s for %s/%s after latest pointer update error: %v", versionId, *input.Bucket, *input.Key, rollbackErr)
-					completionState.manifestsReferenced = true
-				}
-				glog.Errorf("completeMultipartUpload: failed to update latest version in directory: %v", err)
-				return s3err.ErrInternalError
+				completionState.uploadDirRemoved = true
+			} else {
+				if err := s3a.mkFile(versionDir, versionFileName, completionState.finalParts, decorateVersionEntry); err != nil {
+					glog.Errorf("completeMultipartUpload: failed to create version %s: %v", versionId, err)
+					return s3err.ErrInternalError
+				}
+				if err := s3a.updateLatestVersionInDirectory(*input.Bucket, *input.Key, versionId, versionFileName, versionEntryForCache); err != nil {
+					if rollbackErr := s3a.rollbackMultipartVersion(versionDir, versionFileName); rollbackErr != nil {
+						glog.Errorf("completeMultipartUpload: failed to rollback version %s for %s/%s after latest pointer update error: %v", versionId, *input.Bucket, *input.Key, rollbackErr)
+						completionState.manifestsReferenced = true
+					}
+					glog.Errorf("completeMultipartUpload: failed to update latest version in directory: %v", err)
+					return s3err.ErrInternalError
+				}
 			}
 
 			// For versioned buckets, all content is stored in .versions directory
@@ -936,13 +958,15 @@
 	}
 
 	if completionState != nil {
-		for _, deleteEntry := range completionState.deleteEntries {
-			if err := s3a.rm(uploadDirectory, deleteEntry.Name, !completionState.metadataOnlyCleanup, true); err != nil {
-				glog.Warningf("completeMultipartUpload cleanup %s upload %s unused %s : %v", *input.Bucket, *input.UploadId, deleteEntry.Name, err)
-			}
-		}
-		if err := s3a.rm(s3a.genUploadsFolder(*input.Bucket), *input.UploadId, false, true); err != nil {
-			glog.V(1).Infof("completeMultipartUpload cleanup %s upload %s: %v", *input.Bucket, *input.UploadId, err)
+		if !completionState.uploadDirRemoved {
+			for _, deleteEntry := range completionState.deleteEntries {
+				if err := s3a.rm(uploadDirectory, deleteEntry.Name, !completionState.metadataOnlyCleanup, true); err != nil {
+					glog.Warningf("completeMultipartUpload cleanup %s upload %s unused %s : %v", *input.Bucket, *input.UploadId, deleteEntry.Name, err)
+				}
+			}
+			if err := s3a.rm(s3a.genUploadsFolder(*input.Bucket), *input.UploadId, false, true); err != nil {
+				glog.V(1).Infof("completeMultipartUpload cleanup %s upload %s: %v", *input.Bucket, *input.UploadId, err)
+			}
 		}
 		if len(completionState.supersededPartManifests) > 0 {
 			s3a.deleteOrphanedChunks(completionState.supersededPartManifests)
```

</details>

Additional details from AI:
<details>

**The bug, in three facts**
- A multipart object's chunks are the part chunks — completion references the fids, it doesn't copy bytes. Say this first; everything else follows from it.
- Completion therefore removes the upload directory metadata-only: s3a.rm(genUploadsFolder(bucket), uploadId, false, true) at filer_multipart.go:951. Quote the line — the false is the whole point.
- That call is best-effort: its error is logged at V(1) and ignored, after the 200 is already decided. Any transient filer error or a crash there strands .uploads/<uploadId>/ permanently, and s3.clean.uploads then purges it with chunk deletion.

**What the patch does**
- New routedMultipartFinalize sends one ObjectTransactionRequest with three mutations: DELETE .uploads/<uploadId> (metadata-only), PUT the version entry, RECOMPUTE_LATEST.
- Why a new function instead of extending routedVersionedFinalize: that helper is shared with CopyObject and createDeleteMarker, which have no upload directory. Their call sites and file are untouched.
- The unrouted fallback (owner == "") keeps the old mkFile + updateLatestVersionInDirectory sequence; the entry decoration is hoisted into a closure both paths share.

- Patch reuses the `ObjectMutation.entry`: it's a full filer `Entry`, and the PUT branch runs FromPbEntry → CreateEntry. The header comment about "data-bearing writes" is about writing chunk bytes; the needles already exist.
- A big object won't blow up the request. MaybeManifestize folds lists above ManifestBatch = 10000 into manifest chunks.
- `deleteEntries` got moved ahead of the transaction because the transaction removes the whole directory metadata-only, so superseded duplicates and ETag-mismatched/empty parts still in there would leak their chunks instead of being freed. Note the consequence: a failed completion now frees those duplicates permanently — harmless, since `sortEntriesByLatestChunk` would discard them on a retry anyway.

**Evidence**:
1. A/B on the window: `.uploads` children stays 1 through the whole gap on stock, drops to 0 before it on the patch.
2. Crash mid-window: both builds commit the object; only stock leaves the directory.
3. `s3.clean.uploads -timeAgo=0`, then needle probe — `direct=404 / readDeleted=200` on the stock object's fids, `200` on the patched one. 
4. After vacuum: `HEAD 10485760` vs `GET wrote 0 bytes / IncompleteRead`. This is the line that lands.
5. `weed/s3api` passes `-count=1` on a pristine tree before and after.

**Scope**:
- Closes the leftover-directory class only.
- The concurrent-abort race survives: an abort landing before the transaction still frees the parts. Abort keys on uploadId, `withObjectWriteLock` on (bucket, object) — different keys, never mutually exclusive. That's a separate issue, and the fix there is routing abort through a transaction on the same LockKey.
- Also worth a sentence: the same destructive delete has three call sites, and only one is the client — `abortMultipartUpload`, `lifecycleAbortMPU`, and `s3.clean.uploads`. The unattended ones are why this matters.

**Repro (see description at top of file):**
```bash
#!/bin/bash
# Reproduce: s3.clean.uploads destroys a completed object's data.
#
# Runs entirely on a STOCK weed binary. No patches, no instrumentation.
#
#   ./repro-clean-uploads-dataloss.sh /path/to/weed [attempts]
#
# THE BUG
#   A multipart object's chunks ARE the part chunks -- completion references the
#   fids, it does not copy bytes. Completion therefore removes the upload
#   directory metadata-only (filer_multipart.go, `s3a.rm(..., false, true)`), and
#   that removal is best-effort: its error is logged at V(1) and ignored. Any
#   filer error or a crash in that window leaves .uploads/<uploadId>/ behind
#   while the object is committed and serving. s3.clean.uploads then purges the
#   leftover with `?recursive=true` and NO skipChunkDeletion, freeing the live
#   object's needles.
#
# HOW THIS HITS THE WINDOW WITHOUT PATCHING ANYTHING
#   The window -- object published, upload directory not yet removed -- is small:
#   measured at ~2.2ms on a loopback single-node cluster with the default leveldb
#   filer store. It widens with anything that slows a filer metadata write, so a
#   remote/SQL filer store, a busy machine, or slower storage all make this
#   EASIER to hit, not harder.
#
#   Landing a SIGKILL in ~2ms needs a sampler faster than the window. A shell
#   poll loop cannot do it: each iteration spawns curl and python, ~80ms. This
#   script uses one Python process with a persistent HTTP connection polling the
#   .versions pointer every 2ms. Each attempt is then roughly a coin flip, and
#   attempts are cheap, so it converges in a handful.
#
#   NOISE_PARTS uploads extra parts the completion does not list, and it is what
#   makes this land. Not for the reason you might expect: those parts are skipped
#   at filer_multipart.go:431-434 and never reach the cleanup loop. What they
#   lengthen is the recursive delete itself -- doBatchDeleteFolderMetaAndData
#   lists the directory and fires a notification PER CHILD before the single
#   DeleteFolderChildren store call, so more children means a longer phase in
#   which the part entries still exist and the parent is still there. That is the
#   state the cleaner has to find. Measured: 0/20 attempts at NOISE_PARTS=0,
#   hits within a few attempts at 200.
#
# WHAT YOU SHOULD SEE
#   leftover .uploads/<uploadId>/ next to a healthy, readable object;
#   then after the cleaner + a vacuum: HEAD says 10485760, GET returns 0 bytes.
set -u

WEED=${1:?usage: $0 /path/to/weed [attempts]}
ATTEMPTS=${2:-8}
DIR=/tmp/sw-repro-$$
# -s3.port.iceberg=0 below disables the Iceberg REST catalog, which otherwise
# binds a FIXED 8181 and collides with any other weed s3 on the box.
MASTER=19333; VOLUME=18080; FILER=18888; S3=18333
BUCKET=repro
KEY=bigobject   # overwritten per attempt
NOISE_PARTS=${NOISE_PARTS:-200}      # see header: this is what makes the kill land

export AWS_ACCESS_KEY_ID=REPROKEY AWS_SECRET_ACCESS_KEY=REPROSECRET
export AWS_DEFAULT_REGION=us-east-1 AWS_EC2_METADATA_DISABLED=true
E="--endpoint-url http://127.0.0.1:$S3"

mkdir -p "$DIR"
cat > "$DIR/s3.json" <<'JSON'
{"identities":[{"name":"repro","credentials":[{"accessKey":"REPROKEY","secretKey":"REPROSECRET"}],
  "actions":["Admin","Read","Write","List","Tagging"]}]}
JSON

start_server() {
  ( cd "$DIR" && nohup "$WEED" server -dir="$DIR" -ip=127.0.0.1 \
      -master.port=$MASTER -volume.port=$VOLUME -volume.max=30 \
      -filer -filer.port=$FILER \
      -s3 -s3.port=$S3 -s3.port.iceberg=0 -s3.config="$DIR/s3.json" >> "$DIR/weed.log" 2>&1 & )
  for _ in $(seq 1 60); do
    aws $E s3api list-buckets >/dev/null 2>&1 && return 0
    sleep 1
  done
  echo "server did not come up; see $DIR/weed.log"; exit 1
}

stop_server() {
  ps -eo pid,args | grep "$DIR" | grep -v grep | awk '{print $1}' \
    | while read -r p; do kill -9 "$p" 2>/dev/null; done
  sleep 3
}

cleanup() { stop_server; }
trap cleanup EXIT

uploads_children() {
  curl -s -H 'Accept: application/json' \
    "http://127.0.0.1:$FILER/buckets/$BUCKET/.uploads/?limit=100" \
  | python3 -c 'import sys,json;d=json.load(sys.stdin);e=d.get("Entries") or [];print(len(e));[print(x["FullPath"].split("/")[-1]) for x in e]' 2>/dev/null
}

# true once THIS key's .versions pointer names a version, i.e. it is published.
# Must be scoped to the key: every object in a versioned bucket has a .versions
# entry, so an unscoped check fires on the control object immediately.
published() {
  curl -s -H 'Accept: application/json' \
    "http://127.0.0.1:$FILER/buckets/$BUCKET/?limit=500" \
  | KEY="$1" python3 -c '
import sys,json,os
want = "/" + os.environ["KEY"] + ".versions"
d=json.load(sys.stdin)
for e in (d.get("Entries") or []):
    if e["FullPath"].endswith(want):
        if "Seaweed-X-Amz-Latest-Version-File-Name" in (e.get("Extended") or {}):
            print("yes"); break
' 2>/dev/null
}


# The kill has to land between "the object is published" and "the upload
# directory is removed" -- on a loopback single-node cluster that is only tens of
# milliseconds. A shell poll loop cannot do it: each iteration spawns curl and
# python, ~80ms, wider than the window itself. This helper is one process with a
# persistent HTTP connection polling every 2ms.
cat > "$DIR/killer.py" <<'PYEOF'
import http.client, json, os, signal, sys, time

filer_port, bucket, key, pidfile = sys.argv[1], sys.argv[2], sys.argv[3], sys.argv[4]
want = "/" + key + ".versions"
conn = http.client.HTTPConnection("127.0.0.1", int(filer_port))
deadline = time.monotonic() + 60

def published():
    conn.request("GET", "/buckets/%s/?limit=500" % bucket, headers={"Accept": "application/json"})
    body = conn.getresponse().read()
    for e in (json.loads(body).get("Entries") or []):
        if e["FullPath"].endswith(want):
            return "Seaweed-X-Amz-Latest-Version-File-Name" in (e.get("Extended") or {})
    return False

while time.monotonic() < deadline:
    try:
        if published():
            for pid in open(pidfile).read().split():
                try:
                    os.kill(int(pid), signal.SIGKILL)
                except ProcessLookupError:
                    pass
            print("killed")
            sys.exit(0)
    except Exception:
        conn.close()
        conn = http.client.HTTPConnection("127.0.0.1", int(filer_port))
    time.sleep(0.002)
print("timeout")
PYEOF

echo "### workdir $DIR"
start_server
aws $E s3api create-bucket --bucket $BUCKET >/dev/null
aws $E s3api put-bucket-versioning --bucket $BUCKET \
    --versioning-configuration Status=Enabled >/dev/null
head -c 5242880 /dev/urandom > "$DIR/p1.bin"
head -c 5242880 /dev/urandom > "$DIR/p2.bin"
cat "$DIR/p1.bin" "$DIR/p2.bin" > "$DIR/whole.bin"
printf 'x' > "$DIR/tiny.bin"

# a second, untouched object: its fids are the control proving the 404 probe works
aws $E s3api put-object --bucket $BUCKET --key control --body "$DIR/p1.bin" >/dev/null
echo "### control object written"

HIT=0
for attempt in $(seq 1 "$ATTEMPTS"); do
  echo
  echo "### attempt $attempt/$ATTEMPTS"
  KEY="bigobject-$attempt"      # a fresh key each attempt; no interference
  UP=$(aws $E s3api create-multipart-upload --bucket $BUCKET --key $KEY \
        --query UploadId --output text)
  E1=$(aws $E s3api upload-part --bucket $BUCKET --key $KEY --upload-id "$UP" \
        --part-number 1 --body "$DIR/p1.bin" --query ETag --output text)
  E2=$(aws $E s3api upload-part --bucket $BUCKET --key $KEY --upload-id "$UP" \
        --part-number 2 --body "$DIR/p2.bin" --query ETag --output text)
  if [ "$NOISE_PARTS" -gt 0 ]; then
    echo "  uploading $NOISE_PARTS unlisted parts..."
    for n in $(seq 3 $((NOISE_PARTS + 2))); do
      aws $E s3api upload-part --bucket $BUCKET --key $KEY --upload-id "$UP" \
          --part-number "$n" --body "$DIR/tiny.bin" >/dev/null 2>&1
    done
  fi
  cat > "$DIR/parts.json" <<JSON
{"Parts":[{"ETag":$E1,"PartNumber":1},{"ETag":$E2,"PartNumber":2}]}
JSON

  ps -eo pid,args | grep "$DIR" | grep -v grep | awk '{print $1}' > "$DIR/pids"
  python3 "$DIR/killer.py" "$FILER" "$BUCKET" "$KEY" "$DIR/pids" > "$DIR/killer.out" 2>&1 &
  KILLER=$!

  ( aws $E s3api complete-multipart-upload --bucket $BUCKET --key $KEY \
      --upload-id "$UP" --multipart-upload file://"$DIR/parts.json" \
      > "$DIR/complete.txt" 2>&1 & )

  wait $KILLER 2>/dev/null
  KILLED=$(cat "$DIR/killer.out" 2>/dev/null)
  sleep 3
  ps -eo pid,args | grep "$DIR" | grep -v grep | awk '{print $1}' \
    | while read -r p; do kill -9 "$p" 2>/dev/null; done
  sleep 2
  echo "  killed mid-cleanup: $KILLED"
  start_server

  # The leftover must be THIS attempt's uploadId. Leftovers accumulate across
  # attempts, and an earlier attempt's directory holds fids belonging to an
  # object that was never published -- purging that proves nothing.
  #
  # It also has to still CONTAIN part entries. The gateway's own rm is recursive
  # (children first, parent last), so killing inside it leaves an EMPTY
  # .uploads/<uploadId>/, which the cleaner purges harmlessly.
  LEFT_DIR=""
  LEFT=$(curl -s -H 'Accept: application/json' \
    "http://127.0.0.1:$FILER/buckets/$BUCKET/.uploads/$UP/?limit=100" \
    | python3 -c 'import sys,json;d=json.load(sys.stdin);print(len(d.get("Entries") or []))' 2>/dev/null)
  LEFT=${LEFT:-0}
  [ "$LEFT" -ge 1 ] && LEFT_DIR=$UP
  SIZE=$(aws $E s3api head-object --bucket $BUCKET --key $KEY \
          --query ContentLength --output text 2>/dev/null)
  echo "  this upload's leftover holds ${LEFT:-0} part entries ; object ContentLength = ${SIZE:-none}"
  if [ "${LEFT:-0}" -ge 1 ] && [ "${SIZE:-0}" = "10485760" ]; then
    echo "  => BUG STATE REPRODUCED: a committed, readable object with a lingering .uploads/<uploadId>/"
    HIT=1
    HIT_KEY=$KEY
    HIT_UPLOAD=$UP
    break
  fi
  # drop this attempt's leftovers metadata-only so they cannot pollute the next
  curl -s -o /dev/null -X DELETE \
    "http://127.0.0.1:$FILER/buckets/$BUCKET/.uploads/$UP/?recursive=true&ignoreRecursiveError=true&skipChunkDeletion=true"
done

if [ "$HIT" != "1" ]; then
  echo
  echo "### did not land in the window in $ATTEMPTS attempts."
  echo "### raise the attempt count and retry; each attempt is roughly a coin flip."
  exit 2
fi
KEY=$HIT_KEY

echo
echo "### the object is healthy right now"
aws $E s3api get-object --bucket $BUCKET --key $KEY "$DIR/got.bin" >/dev/null 2>&1 \
  && cmp -s "$DIR/whole.bin" "$DIR/got.bin" && echo "  GET 200, byte-identical to source"

echo
echo "### fids (needed for the probe below)"
curl -s -H 'Accept: application/json' \
  "http://127.0.0.1:$FILER/buckets/$BUCKET/$KEY.versions/?limit=5" \
| python3 -c '
import sys,json
d=json.load(sys.stdin)
for e in (d.get("Entries") or []):
    for c in (e.get("chunks") or []): print(c["file_id"])' > "$DIR/victim.fids"
# versioning is on, so the control object lives in control.versions/ too
curl -s -H 'Accept: application/json' \
  "http://127.0.0.1:$FILER/buckets/$BUCKET/control.versions/?limit=5" \
| python3 -c '
import sys,json
d=json.load(sys.stdin)
for e in (d.get("Entries") or []):
    for c in (e.get("chunks") or []): print(c["file_id"])' > "$DIR/control.fids"
sed 's/^/  victim  /' "$DIR/victim.fids"; sed 's/^/  control /' "$DIR/control.fids"

echo
echo "### running the scheduled task"
echo "  (the leftover holds $LEFT part entries whose fids the object also references)"
printf 'lock\ns3.clean.uploads -timeAgo=0\nunlock\n' | "$WEED" shell -master=127.0.0.1:$MASTER 2>&1 | grep -E 'purge|error' | sed 's/^/  /'

# chunk deletion is queued asynchronously by the filer, so the tombstones do not
# land the instant the purge returns. Probing too early shows 200 and reads like
# nothing happened.
echo
echo "### waiting for the queued chunk deletions to land"
sleep 12
echo "### needle probe (control proves the probe itself works)"
probe() { printf '  %-8s %-14s direct=%s readDeleted=%s\n' "$2" "$1" \
  "$(curl -s -o /dev/null -w '%{http_code}' "http://127.0.0.1:$VOLUME/$1")" \
  "$(curl -s -o /dev/null -w '%{http_code}' "http://127.0.0.1:$VOLUME/$1?readDeleted=true")"; }
while read -r f; do probe "$f" victim; done < "$DIR/victim.fids"
while read -r f; do probe "$f" control; done < "$DIR/control.fids"

echo
echo "### vacuum, then read the object again"
printf 'lock\nvolume.vacuum -garbageThreshold=0.001\nunlock\n' | "$WEED" shell -master=127.0.0.1:$MASTER >/dev/null 2>&1
sleep 5
H=$(aws $E s3api head-object --bucket $BUCKET --key $KEY --query ContentLength --output text 2>/dev/null)
rm -f "$DIR/after.bin"
G=$(aws $E s3api get-object --bucket $BUCKET --key $KEY "$DIR/after.bin" 2>&1 >/dev/null)
N=$(stat -c%s "$DIR/after.bin" 2>/dev/null || echo 0)
echo "  HEAD ContentLength : $H"
echo "  GET wrote          : $N bytes"
[ -n "$G" ] && echo "  GET error          : $G"
echo
if [ "$H" = "10485760" ] && [ "$N" != "10485760" ]; then
  echo "### DATA LOSS CONFIRMED: the object advertises $H bytes and returns $N."
else
  echo "### object still intact -- the leftover was not purged, or the vacuum did not run."
fi
echo "### logs and state left in $DIR"
```

</details>

Please let me know if you have a preferred way of reporting these issues, or if what I'm doing right now is fine.

chrislusf:
## Summary

Fixes #10663 — a committed multipart object could be corrupted by `s3.clean.uploads` (on by default via the master admin worker).

A completed object's chunks **are** the part chunks — completion references the fids, it doesn't copy bytes. Completion then removed `.uploads/<uploadId>` metadata-only as a best-effort post-commit step (error logged at `V(1)` and ignored). A filer error or gateway crash in that window left the upload directory referencing the same chunks as the published object; the next cleaner run purged it with `recursive=true` and no `skipChunkDeletion`, freeing the live object's needles. `HEAD` keeps returning the size while `GET` returns 0 bytes.

Changes, one per commit:

- **Versioned completion commits in one `ObjectTransaction`** on the owner filer: `PUT` the version file, metadata-only `DELETE` of `.uploads/<id>`, `RECOMPUTE_LATEST`. The transaction has no rollback, so the order picks which partial states are reachable — `PUT` first keeps chunks referenced at all times, and removing the upload dir before the recompute means a published object never coexists with a leftover `.uploads` dir. Unused part entries are freed first since the metadata-only dir delete would otherwise leak their chunks.
- **Suspended and non-versioned completion** removes `.uploads/<id>` inside the routed object `PUT` transaction (same mutation), keeping the unrouted `mkFile` fallback on post-commit cleanup.
- **`s3.clean.uploads`** now checks whether a stale upload already completed — the object entry or any version file under `<key>.versions` carrying the upload id — and purges metadata-only (`skipChunkDeletion=true`) in that case. If the lookup fails, the upload is skipped for that run rather than risk live chunks. This also covers leftovers created before this fix or on fallback paths.

The concurrent-abort race noted in the issue is out of scope (abort locks on uploadId, completion on bucket/object).

#### Test plan

- [x] `go test ./weed/s3api/` — new tests assert the transaction carries `PUT` → metadata-only `DELETE .uploads/<id>` → `RECOMPUTE_LATEST`, and that the non-versioned `PUT` rides the removal
- [x] `go test ./weed/shell/` — `uploadCompleted` covered for object-entry, version-file, absent, and error cases
- [ ] Manual repro from the issue script (crash mid-completion + `s3.clean.uploads`) — happy to run if wanted

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11375" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when completing routed and versioned multipart uploads.
  * Prevented completion after an upload has been removed.
  * Preserved uploads when completion status cannot be determined.
  * Prevented stale-upload cleanup from deleting chunks used by completed objects.
  * Improved handling and reporting of upload-removal and unused-part cleanup errors.
  * Prevented cleanup failures from affecting completed multipart objects.
  * Corrected list pagination limit handling on 32-bit platforms.
  * Improved recovery of certain previously completed uploads under suspended versioning.


chrislusf:
Addressed the automated review findings in three follow-up commits:

- `bc35ef93b` — unused-part deletion failures now abort the completion instead of leaking chunks (the upload directory stays, so the request is retriable). Flagged by Devin Review and Greptile.
- `8d6852bab` — the commit transaction now carries an `IF_EXISTS` precondition on `.uploads/<id>`, so a delete that does not take the object lock (abort, lifecycle, `s3.clean.uploads`) racing the commit fails with `NoSuchUpload` instead of publishing an object over freed chunks. Flagged by Devin Review. The remaining micro-window inside the transaction can only be closed by routing those deleters through the same `LockKey`, which issue #10663 scopes as separate work.
- `fc1cacb9e` — on a partial transaction failure the version file is rolled back only while the upload directory survives; if the transaction already removed it, the version is kept (its chunks stay referenced) so a retry resolves it via `SeaweedFSUploadId` and reconciliation can repair the latest pointer. Flagged by Devin Review, Greptile, and CodeRabbit.

`go test ./weed/s3api ./weed/shell` passes locally.

chrislusf:
Follow-up addressing the remaining review findings and CI:

- `835e06f6d` — post-commit upload-directory removal now runs only after unused-part cleanup succeeds, so a failed entry delete keeps its metadata (and chunk references). Flagged by CodeRabbit.
- `539fe809f` — fixed the linux/386 CI failure: `filer_pb.List` computed `count > int(limit)`, and `int(math.MaxUint32)` wraps to -1 on 32-bit, dropping every streamed entry. Compared in `uint64`; identical semantics on 64-bit. This was a pre-existing bug my new test exposed.
- `a9d975aed` — `s3.clean.uploads` now derives the object location the same way `getEntryNameAndDir` does, so a trailing-slash key (`dir/`) resolves to `<bucket>/dir/dir` instead of missing and purging live chunks. Flagged by CodeRabbit.
- `5276ce2c5` — a retried `CompleteMultipartUpload` whose upload directory is gone now checks the regular path for an entry carrying the upload id and re-runs the suspended null-marker finalize, so a failed finalize no longer strands a committed-but-hidden object. Gated to suspended buckets so a newer versioned write is never demoted. Flagged by CodeRabbit.

`go test ./weed/s3api ./weed/shell` passes locally.

PR #11345, opened 2026-09-16
fix(volume): reject negative Size, recover poisoned store lock

franchb:
P0-1 from franchb/seaweedfs#10.

Negative Size from the wire (Size is i32) flowed into get_actual_size → as usize wrap → vec! panic while store RwLock held → poisoned lock bricks node until restart. Write path ts_offset also wrapped in release.

Changes:
- Add validate_wire_size (reject <=0, >1<<30 WIRE_MAX_NEEDLE_SIZE) in needle.rs with boundary tests.
- Defense-in-depth guards in read_needle_blob_unlocked, read_needle_blob_and_parse, write_needle_blob_and_index → InvalidData (keeps get_actual_size signature for deleted sizes).
- gRPC ReadNeedleBlob/WriteNeedleBlob validate first → InvalidArgument.
- store_read/store_write helpers recover poison via into_inner (these two handlers only).

Tests: test_validate_wire_size_boundaries (8 asserts), test_read_blob_negative_does_not_panic (red: capacity overflow panic), handler InvalidArgument tests. cargo test --lib 573 PASS, clippy/fmt clean.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11345" target="_blank">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=3">
    Devin Review
  </picture>
</a>

## Summary by CodeRabbit

- **Bug Fixes**
  - Negative blob sizes are rejected safely with clear validation errors.
  - Zero-size and empty blob entries are handled correctly.
  - Read requests exceeding supported message limits, including framing overhead, are rejected before processing.
  - Valid zero-size and large on-disk blob entries are no longer incorrectly rejected.

- **Tests**
  - Added coverage for negative, zero, maximum, and oversized blob sizes.
  - Added regression tests for validation and error handling.

PR #11399, opened 2026-09-19
fix(volume): validate sizes in ReadNeedleBlob and WriteNeedleBlob

hsdfat:

`volume.check.disk`, replica sync and `volume.merge` copy needles with the `ReadNeedleBlob`/`WriteNeedleBlob` RPCs. On the Go volume server, both RPCs trust the size they are given:

1. **A negative size in `ReadNeedleBlob` can kill the volume server.** The size goes through `Volume.ReadNeedleBlob` into `make()` in `needle.ReadNeedleBlob`. From -44 down on v3 (-36 on v2), the length is negative, and `make` panics with `makeslice: len out of range`. The volume gRPC server has no recovery interceptor, so the process exits. I reproduced this through `pb.NewGrpcServer`. Smaller negative sizes don't panic, but they return bytes that are not a record.
2. **`WriteNeedleBlob` accepts a negative size** if the blob header carries the same value. The blob is appended to `.dat`, and the needle is indexed with that size, which reads as deleted.
3. **`WriteNeedleBlob` never checks the blob length against the size.** The blob is appended as is.
   - If the length is not a multiple of 8, `.dat` leaves the padding grid. Every later ordinary write to the volume is then indexed at a truncated offset and reads back as `EOF`.
   - If the length is off by exactly 8, as when a v3 record is sent to a v2 volume, the grid holds. But a `.dat` scan reads the extra bytes as the next record.

#11345 added guards for 1 and 2 to the Rust volume server. Its comments say "(Go parity)", but the Go side had no such guards.

The PR has two commits, so the second one can be dropped on its own.

- **Commit 1, negative-size guards.** `Volume.ReadNeedleBlob` and `Volume.WriteNeedleBlob` now return an error when `size < 0`. These are the storage-level functions where #11345 put its Rust guards. Size 0 still passes, because delete records carry it and `volume.merge` copies them.
- **Commit 2, length check.** After the header check from #10565, `WriteNeedleBlob` now requires `len(blob) == GetActualSize(size, v.Version())`.

What the length check changes for callers:
- The in-tree callers already send exact lengths:
  - `volume.check.disk` and replica sync copy the record that `ReadNeedleBlob` returns from another replica of the same volume;
  - `volume.merge` serializes with the version it allocates the target volume with.
- The only copy it newly refuses is one between volumes of different needle versions. I tested that case on master, and it already writes a broken record:
  - a v3 record sent to a v2 volume is accepted at every size and leaves 8 bytes that a scan misreads;
  - a v2 record sent to a v3 volume fails the existing timestamp check for 7 of 8 sizes. For the other sizes it lands 8 bytes short, so it reads back as `EOF` while it is the last record, and a scan misparses the record after it.

- One path runs into that case already on master: `volume.merge` of a volume created with an older needle version. A freshly allocated target reports `Version() == 3` in memory until it is reloaded, because `maybeLoadVolumeInfo` defaults `volumeInfo.Version` to the current version. The merge then sends v2 records into a volume that believes it is v3, and on master most of them fail the timestamp check and the rest are written 8 bytes short. With this commit they all fail with the length error. The in-memory version is a separate bug, and I can send a fix for it on its own.

The Rust server doesn't check the length either. I left it alone so this PR stays Go-only, and I can mirror commit 2 there if you want it.

# How is the PR tested?

- `TestReadNeedleBlobRejectsNegativeSize` (sizes -1, -100 and `MinInt32`):
  - On master, -100 and `MinInt32` panic with `makeslice: len out of range` (the test recovers the panic), and -1 returns without an error.
  - Now all three return an error, and sizes 0 and the needle's own size still read.
- `TestWriteNeedleBlobRejectsNegativeSize` (-1, -5):
  - On master, the blob is appended and needle 2 is indexed with the negative size.
  - Now it is rejected and `.dat` is unchanged.
- `TestWriteNeedleBlobRejectsLengthMismatch` (1 byte too long, 1 byte short, 8 bytes too long):
  - On master, all three are appended. After the first two, the next ordinary write reads back as `EOF`.
  - Now all three are rejected, `.dat` is unchanged, and the next write reads back.
- `TestWriteNeedleBlobRoundTrip` passes before and after. It checks that a needle's own record and a size-0 delete record both pass the length check, and that later writes still read back.
- `TestWriteNeedleBlobRejectsSizeMismatch` (#10565) passes unchanged.
- These pass:
  - `go build ./weed/...`, `go vet ./weed/storage/... ./weed/server/`;
  - `go test ./weed/storage/... ./weed/shell/ ./weed/server/`;
  - the blob RPC integration tests in `test/volume_server/grpc`.
- `go test -race ./weed/storage/` passes, except for `TestCheckDiskSpaceProbesWithTheDiskTypeThreshold` and `TestMountEcShards_*`. Those fail the same way on master.

Not in this PR: the Rust server also caps the size at 1 GiB in the two RPC handlers. On the Go side, a huge positive size still makes `ReadNeedleBlob` allocate up to about 2 GB before the short read.

# Checks
- [ ] I will add related wiki document changes and link to this PR after merging. (No wiki change needed.)

# Checks for AI generated PRs

## Summary by CodeRabbit

- **Bug Fixes**
  - Rejects invalid negative or deleted needle sizes during read and write operations, preventing crashes and malformed storage updates.
  - Rejects blobs whose actual length does not match the declared size.
  - Rejects blobs when their embedded needle ID does not match the requested ID.
  - Prevents invalid writes from modifying data files or indexes.
  - Preserves reliable handling of valid data, including zero-size deletion records and subsequent writes.


chrislusf:
Pushed a review fix (edd3048):

- `needle.ReadNeedleBlob` now rejects deletion-marker sizes before `make()`, so direct callers (vacuum, scan paths) are covered too, not just the `Volume.ReadNeedleBlob` wrapper.
- The volume-level guards use `size.IsDeleted()` and comments were trimmed.

On the needle-id observation: a blob can legitimately be written under a different needle id than its serialized header — that clone path is exercised by the grpc integration tests. Rejecting it would change established RPC behavior, so this PR keeps the header checks to size and length. The index-vs-record divergence Greptile notes on rebuild is pre-existing and out of scope here.

Issue #11401, opened 2026-09-20
S3 multipart upload with ChecksumAlgorithm=SHA256 does not expose checksum via HeadObject


raind-cd:
## Summary

#9076 fixed flexible checksums for presigned `PutObject` uploads, but multipart uploads still do not retain or expose the SHA-256 checksum through `HeadObject` in SeaweedFS 4.47.

This appears to be a separate path: `CreateMultipartUpload` → `UploadPart` → `CompleteMultipartUpload`.

## Environment

- SeaweedFS 4.47 (commit `c5073360007d28385a33426a42ac3e4ec504c5a3`)
- `weed mini` with a local HTTP S3 endpoint
- AWS SDK for .NET `AWSSDK.S3 4.0.103.3`

## Reproduction

The following multipart upload completes successfully. The completion request uses the SDK helper so the uploaded part's ETag and checksum are both passed to `CompleteMultipartUpload`.

```csharp
var checksum = Convert.ToBase64String(SHA256.HashData(body));

var initiated = await s3.InitiateMultipartUploadAsync(
    new InitiateMultipartUploadRequest
    {
        BucketName = bucket,
        Key = key,
        ContentType = "application/octet-stream",
        ChecksumAlgorithm = ChecksumAlgorithm.SHA256,
    });

using var stream = new MemoryStream(body);
var part = await s3.UploadPartAsync(new UploadPartRequest
{
    BucketName = bucket,
    Key = key,
    UploadId = initiated.UploadId,
    PartNumber = 1,
    PartSize = body.Length,
    InputStream = stream,
    ChecksumSHA256 = checksum,
});

var complete = new CompleteMultipartUploadRequest
{
    BucketName = bucket,
    Key = key,
    UploadId = initiated.UploadId,
};
complete.AddPartETagsAndChecksums([part]);
await s3.CompleteMultipartUploadAsync(complete);

var head = await s3.GetObjectMetadataAsync(bucket, key);
Assert.Equal(checksum, head.ChecksumSHA256);
```

## Actual behavior

- The multipart upload completes successfully.
- `head.ChecksumSHA256` is `null`.
- The corresponding `x-amz-checksum-sha256` response header is absent from `HeadObject`.

## Expected behavior

When a multipart upload requests `ChecksumAlgorithm=SHA256` and supplies the part checksum, SeaweedFS should retain and return the resulting SHA-256 checksum through `HeadObject` (or clearly document that multipart flexible checksums are not supported).

[...]

PR #11403, opened 2026-09-20
s3: multipart upload checksum inheritance and validation

chrislusf:
## Summary

Fixes the multipart checksum handling gaps behind #11401, verified with the actual AWS SDK for .NET `AWSSDK.S3 4.0.103.3` repro from the issue.

**What the investigation found**

- The composite SHA-256 checksum was already computed, stored, and returned by `CompleteMultipartUpload` and by `HeadObject` when `x-amz-checksum-mode: ENABLED` is sent — matching AWS.
- `head.ChecksumSHA256` came back `null` in the repro because .NET `GetObjectMetadataAsync` never sends `x-amz-checksum-mode` (verified on the wire: `SignedHeaders=host;x-amz-content-sha256;x-amz-date`). AWS returns the checksum the same way only when the mode is enabled, so this matches AWS; the .NET caller needs `ChecksumMode = ChecksumMode.ENABLED` on `GetObjectMetadataRequest`.
- `ChecksumType=FULL_OBJECT` + SHA256 returning 400 is also AWS-conformant — AWS supports SHA-256 multipart checksums only as `COMPOSITE`.

**Real gaps this PR fixes**

- `UploadPart` sent without checksum headers now inherits the algorithm declared at `CreateMultipartUpload`, so every part keeps a checksum — the same invariant AWS maintains.
- `CompleteMultipartUpload` now parses the per-part `Checksum*` XML elements (previously ignored entirely) and validates them: `InvalidRequest` when a checksum is missing for an upload created with a checksum algorithm, `BadDigest` on mismatch with the stored part checksum, and `BadDigest` when `x-amz-checksum-type` contradicts the upload's resolved type.

#### Test plan

- [x] `TestIssue11401` — exact .NET repro flow: explicit part checksum, complete, `HeadObject` with/without checksum mode
- [x] `TestMultipartPartInheritsChecksumAlgorithm` — header-less part inherits the declared algorithm
- [x] `TestCompleteMultipartUploadValidatesPartChecksums` — missing → `InvalidRequest`, wrong → `BadDigest`
- [x] Existing `test/s3/checksum` package passes (`TestMultipartUploadReturnsObjectChecksum`, `TestMultipartCopyPreservesChecksum`)
- [x] Verified end-to-end with real `AWSSDK.S3 4.0.103.3`: `head(mode)` returns `...=-1` COMPOSITE, wrong part checksum → `BadDigest` 400

## Summary by CodeRabbit

* **Bug Fixes**
  * Multipart uploads now inherit the configured checksum algorithm when individual parts omit one.
  * Conflicting checksum algorithms on uploaded parts are rejected.
  * Multipart completion validates missing, invalid, mismatched, and inconsistent part checksums.
  * Full-object checksum uploads validate supplied checksums correctly.
  * Completed uploads return accurate composite or full-object checksums.
  * Object checksum visibility follows checksum-mode settings supplied through headers or query parameters.

* **Tests**
  * Added coverage for checksum inheritance, validation, completion, and metadata retrieval in multipart uploads.


chrislusf:
Addressed the review findings:

- Greptile "Conflicting Algorithm Accepted" — `cb7f12de0` rejects an `UploadPart` whose checksum algorithm differs from the one declared at `CreateMultipartUpload` with `InvalidRequest` (AWS behavior).
- CodeRabbit "FULL_OBJECT handled separately" — `09124cac6` requires per-part checksums only for `COMPOSITE`; for `FULL_OBJECT` the whole-object checksum is validated from the `x-amz-checksum-*` request header (`BadDigest` on mismatch).
- Greptile "Query Checksum Bypasses Validation" — `17cb45e9f` uses `lookupHeaderOrQuery` so presigned requests carrying checksum values in the query string are validated too.
- Also added `014198239`: `x-amz-checksum-mode` is accepted case-insensitively from headers or query params so presigned reads can opt into checksum headers like on AWS.

PR #11425, opened 2026-09-21
[Volume] Keep DAT and index state consistent after async batch Sync failure

### LINKED ISSUE #11400: [Volume] Keep DAT and index state consistent after async batch Sync failure
## Description

The asynchronous batch worker in `Volume.startWorker()` applies multiple write/delete requests before calling `DataBackend.Sync()` once for the whole batch.

Each request may already have modified multiple layers:

- append a data record or tombstone to `.dat`;
- update the in-memory `NeedleMap`;
- append or update the backing index, such as `.idx` or LevelDB.

If `DataBackend.Sync()` fails, the current recovery path truncates `.dat` back to the batch's starting offset and marks the requests as failed. However, it does not roll back the corresponding `NeedleMap` or backing index changes.

The failure path is explicitly called out by the existing TODO in [volume_write.go](https://github.com/seaweedfs/seaweedfs/blob/master/weed/storage/volume_write.go#L416):

```go
// todo: this may generate dirty data or cause data inconsistent, may be weed need to panic?
```

The current logic is effectively:

```text
capture .dat end offset
apply all requests
update .dat and index state
call DataBackend.Sync()

if Sync fails:
    truncate .dat back to the original end offset
    mark requests as failed
```

This can leave the Volume in a state where:

```text
.dat:
  the newly appended records have been truncated

NeedleMap:
  still points to the truncated offsets

.idx or the backing index:
  may still contain mappings or tombstones for those records
```

For example, when a batch overwrites an existing Needle, the new version is appended to `.dat` and the NeedleMap is updated to the new offset. If the subsequent batch `Sync()` fails, `.dat` is truncated, but the NeedleMap may continue pointing to the removed record. Subsequent reads can then return EOF, checksum errors, or other invalid data errors.

The inline `syncWrite()` path already has [rollbackUnflushedWrite](https://github.com/seaweedfs/seaweedfs/blob/master/weed/storage/volume_write.go#L181-L204), which restores the previous mapping after an fsync failure. The asynchronous batch path does not have equivalent rollback behavior.

## Scope

- Keep `.dat`, the in-memory `NeedleMap`, and the backing index consistent after a batch `Sync()` failure.
- Handle batches containing:
  - new writes;
  - overwrite writes;
  - deletes;
  - mixed write/delete requests.
- Decide whether the batch should:
  - fully restore its previous state; or
  - fail closed by marking the Volume unavailable/read-only when safe rollback is not possible.
- Do not allow a NeedleMap entry to point beyond the valid `.dat` tail after a failed batch.
- Add fault-injection tests for `DataBackend.Sync()` and `Truncate()` failures.
- Verify both reads in the current process and reads after reloading the Volume.
- Do not change the storage format or public API.

## Expected behavior

```text
batch Sync succeeds
=> all requests complete successfully and .dat/index state is consistent

batch Sync fails
=> requests are reported as failed
=> either the whole batch state is restored
   or the Volume stops serving further I/O
=> no in-memory or persistent index entry points to truncated .dat data
```

## Impact

This affects the asynchronous durable-write path, primarily requests using `fsync=true`.

A failed batch can leave the running Volume with `.dat` and index state referring to different versions of the data. This may cause read failures or require a Volume restart/recovery.

This issue is separate from the Filer-level asynchronous Chunk deletion behavior. It concerns the Volume storage layer after a batch has already entered the durable write path.

ssshr-66:
Fixes #11400

The asynchronous batch worker processes multiple write and delete requests before calling `DataBackend.Sync()` once for the whole batch.

Before the batch is synced, each request may already have:

- Appended a data record or tombstone to `.dat`.
- Updated the in-memory `NeedleMap`.
- Appended or updated the backing index, such as `.idx` or LevelDB.

If the batch `Sync()` fails, the current implementation only attempts to truncate `.dat` back to the batch's starting offset and marks the requests as failed.

This can leave the Volume in an inconsistent state:

- `.dat` has been truncated.
- The in-memory `NeedleMap` still points to an offset from the failed batch.
- `.idx` or LevelDB may still contain mappings from the failed batch.
- Reads in the current process or after reloading the Volume may access invalid data.

The existing TODO also notes that this failure path may generate dirty data or cause data inconsistency.

- Save the batch starting position in `.dat`.
- Save the exact byte position of `.idx` before processing the batch.
- Save the original mapping for every affected Needle.
- Save the Mapper's file, deletion, byte-count, and maximum-key statistics.
- When the batch `Sync()` fails:
  - Restore changed in-memory or LevelDB mappings.
  - Truncate `.idx` back to its exact starting position.
  - Restore the `.idx` append offset.
  - Sync the recovered index.
  - Truncate `.dat` back to the batch starting position and sync it.
  - Restore the Mapper statistics.
  - Mark every request in the failed batch as failed.
- Add coverage for new writes, overwrites, deletes, and mixed write/delete batches.
- If the rollback itself cannot be completed, mark the Volume as unavailable/read-only and reject subsequent I/O instead of continuing with an unverified state.
- This fail-closed behavior is limited to rollback failures. If the original batch `Sync()` fails but the complete rollback succeeds, the Volume remains usable.
- Preserve the existing single-write rollback behavior, non-durable write behavior, storage format, public APIs, and Filer asynchronous deletion semantics.

# How is the PR tested?

The following tests and checks pass:

- `go test ./weed/storage -count=1`
- Focused batch rollback tests with `go test -race`
- `go vet ./weed/storage/...`
- `git diff --check`

The added tests cover:

- A new Needle whose batch Sync fails.
- An overwrite whose batch Sync fails.
- A delete whose batch Sync fails.
- A mixed write/delete batch rolled back as one unit.
- Repeated operations on the same Needle ID within one batch.
- The LevelDB NeedleMapper.
- Volume reload after rollback.
- Fail-closed behavior when `Truncate()` fails.
- Rejecting subsequent reads and writes on an unavailable Volume.
- Preventing `MarkVolumeWritable` from bypassing the unavailable state.

The full storage-package race test still reports existing unrelated races in:

- `TestCheckDiskSpaceProbesWithTheDiskTypeThreshold`
- `TestLevelDbNeedleMap_Concurrency`

The focused race tests for the new batch rollback behavior pass.

# Checks

# Checks for AI generated PRs

- [x] The PR is kept as minimum as possible.

---

<a href="https://app.devin.ai/review/seaweedfs/seaweedfs/pull/11425" target="_blank"><picture><source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-devin-review-dark.svg?v=4">Devin Review</picture></a>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved recovery after batch write or sync failures by rolling back incomplete changes and preserving data consistency.
  * Volumes that encounter unrecoverable storage errors now enter an unavailable, read-only state.
  * Reads and writes fail clearly when a volume is unavailable.
  * Prevented unavailable volumes from being manually marked writable until the underlying issue is resolved.
  * Added safeguards to keep volume indexes and statistics consistent after failed operations.


ssshr-66:
@chrislusf 
Hi Chris,

Thank you for taking the time to push several commits to this PR. I really appreciate your help.

I’m sorry for adding to your workload. I understand that maintaining a project like this as an independent maintainer means handling many issues and PRs. I don’t want to be someone who opens an issue or PR and then disappears, leaving you to carry the follow-up work.

I also realize that my proposals and changes may be incomplete or based on assumptions that need revisiting. Please feel free to point out what you think should be changed or suggest a better direction. I’ll do my best to respond and update my changes within a business day or two. If personal circumstances prevent me from doing that promptly, I’ll leave a note to let you know I’ve seen your feedback and am following up. I don’t want to leave you wondering whether I’ll take care of it.

I really enjoy this project, and I have great respect for the time and care you’ve put into developing and maintaining it over the years. Thank you again for your help.

tb3088:
look good!

